PR #36, opened 2021-08-11
Messaging/remove profile calls

vikramkumar19:
This PR removes the profile calls being made to the profile service to fetch traits since profile sync is in place.
## Test Plan
- [x] Testing Completed on Stage
Screen Shot 2021-08-23 at 4 15 28 PM
Screen Shot 2021-08-23 at 4 11 41 PM
Screen Shot 2021-08-23 at 4 11 28 PM


packages/browser-destinations/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@segment/browser-destinations",
-  "version": "3.3.1",
+  "version": "3.3.2-alpha.0",
```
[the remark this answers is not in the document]
vikramkumar19: This is just to sync with the main. No bump in the version

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -85,19 +26,19 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       required: true
     },
-    to: {
+    email: {
```
omegabytes, line 29: Any specific reason we changed `to` and `toName`?

omegabytes:
Looks good to me, I don't have enough context on actions integrations to approve though :D

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
vikramkumar19: Just wanted to be consistent across the segment. To does not provide much context where as email does and also firstName

vikramkumar19:
> Looks good to me, I don't have enough context on actions integrations to approve though :D

Nah! Just wanted you to have a look at this which closes the loop. Will have someone from my team and Destination team take a look

vikramkumar19:
> > I believe we should be able to use template mapping to resolve them, I am not 100% on that yet will be doing few tests before that conclusion.
> 
> I don't think we can merge this before figuring out how to support Handlebars/Mustache syntax in `subject` and `body` fields.

I have tested this out. Have the mustache render it from profile object.

vikramkumar19:
> > > I believe we should be able to use template mapping to resolve them, I am not 100% on that yet will be doing few tests before that conclusion.
> > 
> > 
> > I don't think we can merge this before figuring out how to support Handlebars/Mustache syntax in `subject` and `body` fields.
> 
> I have tested this out. Have the mustache render it from profile object.

image
image

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -85,19 +26,19 @@ const action: ActionDefinition<Settings, Payload> = {
...
       label: 'To Email',
       description: 'The Email Address to send an email to',
       type: 'string',
       required: true,
-      default: { '@path': '$.properties.email_address' }
+      default: { '@path': '$.properties.email' }
     },
-    toName: {
+    firstName: {
```
mshwery, line 36: These are considered breaking changes for any instances already connected to the destination. If you only have dev/test instances connected you can simply delete those subscriptions and create new ones. If there are customers connected to this destination you will want to make _additive changes only_ instead of deleting the old fields.

Also FYI we wrote a new script in fab-5-engine that makes it a little bit easier to migrate specific workspaces/destinations to new fields + defaults.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/index.ts
```diff
@@ -11,34 +11,9 @@ const destination: DestinationDefinition<Settings> = {
     fields: {
       sendGridApiKey: {
         label: 'API Key',
-        type: 'string',
+        type: 'password',
         description: 'The Api Key for your sendGrid account',
         required: true
-      },
-      profileApiEnvironment: {
```
mshwery, line 18: I don't recall 100%, but we may not _delete_ old setting definitions that get removed during `push`. You may need to manually delete these stale settings via Partner Portal.

mshwery (approved):
left some comments, and keep in mind there are some technically "breaking" changes here, so beware if you have customers connected.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
vikramkumar19: Currently, these are dev/test instances. No customers connected yet. Glad to know we have a script for these changes. I will take a look at them.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/index.ts
vikramkumar19: I had previously changed the setting's they did not reflect using the push. I went ahead and deleted them manually from the partner portal. I will keep an eye out this time if the changes make it through using push!

vikramkumar19:
> left some comments, and keep in mind there are some technically "breaking" changes here, so beware if you have customers connected.

Thanks Matt! I will keep a look out going forward. However, these are test accounts

PR #41, opened 2021-08-12
GROW-26: add Twilio destination action #minor

laurenfb:
This destination action allows a customer to send an SMS using the twilio API to one of their users. It still calls the profiles API; I'd like to get this to production before removing those calls now that we have configurable sync, in the spirit of smaller shippable units.

Here's the PR made by @vadimdemedes, originally reviewed by @mshwery in fab-5-engine: https://github.com/segmentio/fab-5-engine/pull/272 - ~I have added 3 commits since then~. (now a bunch of commits)

SMS received in hushed app inbox:
File

another screenshot, showing profile traits resolved:
File (1)


vikramkumar19 (approved):
Looks good!

PR #65, opened 2021-08-17
Stratconn-831 Update GA4 with missing fields

nick-Ag:
The GA4 API is still changing and so several of our actions became out of date. This PR updates those actions to the current GA4 spec. These all had `items` objects which were missing fields, and select_promotion was also missing top level values.

Updated fields:
```
add_to_cart

begin_checkout

purchase

select_item

select_promotion

view_item
```

# Testing
Of all the updated fields only `select_promotion` since the rest of them had a few fields added to their items objects which won't require testing. To test the top-level fields changes to `select_promotion` I added test mappings to the `selectPromotion.test.ts` file. All unit tests are passing.


packages/destination-actions/src/destinations/google-analytics-4/beginCheckout/index.ts
```diff
@@ -71,11 +66,21 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          description: 'Currency of the purchase or items associated with the event, in 3-letter ISO 4217 format.'
+        },
         discount: {
           label: 'Discount',
           type: 'number',
           description: 'Monetary value of discount associated with a purchase.'
         },
+        index: {
+          label: 'Index',
+          type: 'integer',
```
lcampos, line 81: Looks like we are not consistent on the `index` definition across events with some having it as `integer` and others as `number`. Let's update them to be the same across events.

packages/destination-actions/src/destinations/google-analytics-4/addToCart/index.ts
```diff
@@ -58,6 +53,11 @@ const action: ActionDefinition<Settings, Payload> = {
           type: 'string',
           description: 'Coupon code used for a purchase.'
         },
+        currency: {
```
lcampos, line 56: Lets update the `ProductItem` type to include all these new properties https://github.com/segmentio/action-destinations/blob/702643c16a6c72f8c59ef187896a73c3c5bb2cc1/packages/destination-actions/src/destinations/google-analytics-4/ga4-types.ts#L1

packages/destination-actions/src/destinations/google-analytics-4/selectPromotion/index.ts
```diff
@@ -22,11 +22,31 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    },
     location_id: {
       label: 'Location ID',
       type: 'string',
       description: 'The ID of the location.'
     },
+    promotion_id: {
+      label: 'Promotion ID',
+      type: 'string',
+      description: '	The ID of the promotion associated with the event.'
```
lcampos, line 43: Lets remove the initial tab/spacing
```suggestion
      description: 'The ID of the promotion associated with the event.'
```

packages/destination-actions/src/destinations/google-analytics-4/selectPromotion/index.ts
```diff
@@ -101,18 +146,27 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          description: 'The location associated with the item.'
         },
         price: {
           label: 'Price',
           type: 'number',
           description: 'Price of the product being purchased, in units of the specified currency parameter.'
         },
-        currency: {
-          label: 'Currency',
+        promotion_id: {
```
lcampos, line 156: We should use a Promotion specific interface that extends ProductItem

lcampos (approved):

PR #66, opened 2021-08-17
update test to include proper time value

mshwery:
The user agent parsing code was incorrectly overwriting some fields that have special formatting for Amplitude.

I fixed how the user agent object gets merged and updated the tests to include the time value so we would caught something like this next time!

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
+++ b/packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
@@ -1,5 +1,5 @@
 import { URLSearchParams } from 'url'
-import type { ActionDefinition } from '@segment/actions-core'
+import { ActionDefinition, removeUndefined } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import { convertUTMProperties } from '../utm'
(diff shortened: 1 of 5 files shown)


dlasky (approved):

packages/destination-actions/src/destinations/amplitude/user-agent.ts
```diff
@@ -1,30 +1,26 @@
 import UaParser from '@amplitude/ua-parser-js'
-import { omit } from 'lodash'
```
mshwery, line 2: Also glad to remove this! We have an `omit` in actions-core that is faster and lighter weight. I dont think we even directly depend on `lodash` in `action-destinations`

nettofarah (approved):

packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
```diff
@@ -236,10 +236,11 @@ const action: ActionDefinition<Settings, Payload> = {
     }
 
     const identification = JSON.stringify({
-      ...rest,
+      // Conditionally parse user agent using amplitude's library
+      ...(userAgentParsing && parseUserAgentProperties(userAgent)),
+      // Make sure any top-level properties take precedence over user-agent properties
+      ...removeUndefined(rest),
```
mshwery, line 242: I used `removeUndefined` to make sure values like `{ device_model: undefined }` dont override valid values from the user agent.

dlasky (approved):

PR #67, opened 2021-08-17
introduce braze web presets

juliofarah:
image
image

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/index.ts
@@ -6,7 +6,7 @@ import logCustomEvent from './logCustomEvent'
 import logUser from './logUser'
 import logPurchase from './logPurchase'
 import debounce, { resetUserCache } from './debounce'
-import { defaultValues } from '@segment/actions-core'
+import { defaultValues, DestinationDefinition } from '@segment/actions-core'
 
 declare global {
   interface Window {
@@ -18,19 +18,19 @@ const presets: DestinationDefinition['presets'] = [
   {
     name: 'Log User',
     subscribe: 'type = "identify" or type = "group"',
-    parnetAction: 'logUser',
+    partnerAction: 'logUser',
     mapping: defaultValues(logUser.fields)
   },
   {
     name: 'Log Purchase',
     subscribe: 'type = "track"',
-    parnetAction: 'logPurchase',
+    partnerAction: 'logPurchase',
     mapping: defaultValues(logPurchase.fields)
   },
   {
     name: 'Log Custom Event',
     subscribe: 'type = "track"',
-    parnetAction: 'logCustomEvent',
+    partnerAction: 'logCustomEvent',
     mapping: defaultValues(logCustomEvent.fields)
   }
 ]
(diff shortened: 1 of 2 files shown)


packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -6,13 +6,35 @@ import logCustomEvent from './logCustomEvent'
...
+const presets: DestinationDefinition['presets'] = [
+  {
+    name: 'Log User',
+    subscribe: 'type = "identify" or type = "group"',
+    parnetAction: 'logUser',
+    mapping: defaultValues(logUser.fields)
+  },
+  {
+    name: 'Log Purchase',
+    subscribe: 'type = "track"',
```
juliofarah, line 26: It looks like `logPurchase` doesn't have a default subscription field

juliofarah: maybe not?

packages/browser-destinations/src/destinations/braze/logCustomEvent/index.ts
```diff
@@ -14,14 +14,18 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
...
     },
     eventProperties: {
       type: 'object',
       required: false,
       description: 'Hash of properties for this event.',
       label: 'eventProperties',
-      default: {}
+      default: {
+        '@path': '$.event'
+      }
```
juliofarah, line 28: are these changes accurate?

packages/browser-destinations/src/destinations/braze/logCustomEvent/index.ts
```diff
@@ -14,14 +14,18 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
...
     },
     eventProperties: {
       type: 'object',
       required: false,
       description: 'Hash of properties for this event.',
       label: 'eventProperties',
-      default: {}
+      default: {
+        '@path': '$.properties'
+      }
```
juliofarah, line 28: are these changes accurate?

mshwery: Do we need to strip out reserved properties in the perform function?

mshwery (commented):
Looks right to me. Keep in mind all presets should have mappings or values for all `required: true` fields

packages/browser-destinations/src/destinations/braze/logCustomEvent/index.ts
```diff
@@ -14,18 +15,24 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
...
     }
   },
   perform: (client, event) => {
-    client.logCustomEvent(event.payload.eventName, event.payload.eventProperties)
+    const reservedKeys = Object.keys(action.fields.eventProperties.properties ?? {})
```
mshwery, line 33: This doesn't make a ton of sense to me. Should this be: 
```suggestion
    const reservedKeys = Object.keys(action.fields ?? {})
```

mshwery: Oh I guess it honestly doesnt matter since there are only two keys here. I would remove the `omit` stuff since there's really no point for this action. I thought there were more top-level fields originally, sorry!

nettofarah (approved):

PR #71, opened 2021-08-17
[BE-67] Automatically scaffold tests on generate:action

daltonscharff:
[[BE-67](https://segment.atlassian.net/browse/BE-67)] 

When either a new server or browser action is created for a destination, a test file is also created inside that action's `__tests__` folder.

While considering what to include in the test template files, I looked at existing tests and tried to pull out imports that were common across every file.

Example server test file:
image

Example browser test file:
image


packages/cli/templates/actions/empty-browser-action/index.test.ts
```diff
@@ -0,0 +1,9 @@
+import { Analytics, Context } from '@segment/analytics-next'
+import * as jsdom from 'jsdom'
+import Destination from '../index'
```
mshwery, line 3: This will need to be: 
```suggestion
import { destination } from '../index'
```

packages/cli/src/commands/generate/action.ts
```diff
@@ -96,10 +99,15 @@ export default class GenerateAction extends Command {
...
           description: '',
-          slug
+          slug,
+          destination
         },
         flags.force
       )
+      if (!fs.existsSync(testsDirectory)) {
+        fs.mkdirSync(testsDirectory, { recursive: true })
+      }
```
mshwery, line 109: This shouldn't be needed.

mshwery: Instead, why not do `__tests__/{{slug}}.test.ts` as the file structure here? That way each action has it's own tests directory and test file. Then we don't need any custom code in the `generate:action` command like the above ^ to handle ensuring the directory exists.

packages/cli/src/commands/generate/action.ts
```diff
@@ -96,10 +99,15 @@ export default class GenerateAction extends Command {
...
+      fs.renameSync(path.join(targetDirectory, 'index.test.ts'), path.join(testsDirectory, `${slug}.test.ts`))
```
mshwery, line 110: If we need to support templatizing the file names we can add that to the templating code.

packages/cli/src/commands/generate/action.ts
daltonscharff: I'm not sure I follow completely. If I want to generate a new action on customer.io called `testAction`, the `generate:action testAction server` command would create a new action in `destination-actions/src/destinations/customerio/testAction` and would create an `index.ts` and `generate-types.ts` file in that folder. Are you saying to add a `__tests__` folder within that `testAction` folder as well? If so, I don't think we would need to template the file name as it would just be called `index.test.ts` every time, right?

mshwery: >Are you saying to add a __tests__ folder within that testAction folder as well?

Yes.

>If so, I don't think we would need to template the file name as it would just be called index.test.ts every time, right?

Up to you. I think it might be nice to have test file names matching their action. Otherwise it's hard to tell _which_ `index.test.ts` file you are in. But we could save that for a later PR!

daltonscharff: Ok. This makes sense. I think since the file being tested is called `index.ts`, then it makes sense for now that the test file to be called `index.test.ts`, even though I do agree that this can get confusing if working across actions. I will work on this now. Thanks!

daltonscharff:
I've updated the code to include a `__tests__/index.test.ts` file within the action directory that is created during the action:generate cli command. I've also created a ticket to adapt the CLI templating code to modify template strings in file names: https://segment.atlassian.net/browse/BE-105

packages/cli/templates/actions/empty-browser-action/__tests__/index.test.ts
```diff
@@ -0,0 +1,9 @@
+import { Analytics, Context } from '@segment/analytics-next'
+import * as jsdom from 'jsdom'
+import { destination } from '../../index'
+
+describe('{{destination}}', () => {
+  describe('{{name}}', () => {
```
mshwery, line 6: I think it's best practice to not have multiple of the same `describe` names, which would happen here if you have more than one action for a destination.

What about simply:
```suggestion
describe('{{destination}}.{{slug}}', () => {
```

mshwery (approved):
minor comments, but otherwise lgtm!

PR #73, opened 2021-08-17
push-browser-destination touch ups

juliofarah:
This PR just adds a little bit of logging to the `push-browser-destination` command and tweaks around environment variables.

The command no longer requires environment variables to be passed on. If run on the production workbench, the script will check for `SERVER_ENVIRONMENT`, present on the workbench, otherwise, it'll assume it's a stage build.

image

### DIFF
--- a/packages/browser-destinations/package.json
+++ b/packages/browser-destinations/package.json
@@ -17,7 +17,7 @@
     "build-ts": "yarn tsc -b tsconfig.build.json",
     "build-web": "NODE_ENV=production ASSET_ENV=production yarn webpack -c webpack.config.js",
     "build-web-stage": "NODE_ENV=production ASSET_ENV=stage yarn webpack -c webpack.config.js",
-    "deploy-prod": "yarn build-web && aws-okta exec plat-write -- aws s3 sync ./dist/web/ s3://segment-ajs-next-destinations-production/next-integrations/actions --grants read=id=$npm_config_prod_cdn_oai,id=$npm_config_prod_custom_domain_oai",
+    "deploy-prod": "yarn build-web && aws s3 sync ./dist/web/ s3://segment-ajs-next-destinations-production/next-integrations/actions --grants read=id=$npm_config_prod_cdn_oai,id=$npm_config_prod_custom_domain_oai",
     "deploy-stage": "yarn build-web-stage && aws-okta exec plat-write -- aws s3 sync ./dist/web/ s3://segment-ajs-next-destinations-stage/next-integrations/actions --grants read=id=$npm_config_stage_cdn_oai,id=$npm_config_stage_custom_domain_oai",
     "clean": "tsc -b tsconfig.build.json --clean",
     "postclean": "rm -rf dist",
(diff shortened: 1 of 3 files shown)


packages/browser-destinations/package.json
```diff
@@ -17,7 +17,7 @@
     "build-ts": "yarn tsc -b tsconfig.build.json",
     "build-web": "NODE_ENV=production ASSET_ENV=production yarn webpack -c webpack.config.js",
     "build-web-stage": "NODE_ENV=production ASSET_ENV=stage yarn webpack -c webpack.config.js",
-    "deploy-prod": "yarn build-web && aws-okta exec plat-write -- aws s3 sync ./dist/web/ s3://segment-ajs-next-destinations-production/next-integrations/actions --grants read=id=$npm_config_prod_cdn_oai,id=$npm_config_prod_custom_domain_oai",
+    "deploy-prod": "yarn build-web && aws s3 sync ./dist/web/ s3://segment-ajs-next-destinations-production/next-integrations/actions --grants read=id=$npm_config_prod_cdn_oai,id=$npm_config_prod_custom_domain_oai",
```
juliofarah, line 20: no need to use `aws-okta` given we must run this from the workbench

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -86,11 +86,11 @@ export default class PushBrowserDestinations extends Command {
 
       if (existingPlugin) {
         await updateRemotePlugin(input)
+        this.spinner.succeed(`Updated existing remote plugin for ${metadata.name}`)
```
juliofarah, line 89: image

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -107,23 +107,19 @@ export default class PushBrowserDestinations extends Command {
 
 async function build(): Promise<string> {
   execa.commandSync('lerna run build')
-  if (process.env.NODE_ENV === 'stage') {
-    return execa.commandSync('lerna run build-web-stage').stdout
+  if (process.env.SERVER_ENVIRONMENT === 'production') {
```
mshwery, line 110: Thoughts on making this a command flag instead of just an env var? You can source command flags from env vars too, so you could use either!

mshwery: Something like `--env`

packages/cli/src/config.ts
```diff
@@ -1,5 +1,5 @@
 export const ASSET_PATH =
-  process.env.NODE_ENV === 'production'
+  process.env.NODE_ENV === 'production' || process.env.SERVER_ENVIRONMENT === 'production'
```
mshwery, line 2: Wont this lead to `undefined` if you do `SERVER_ENVIRONMENT=stage`?

juliofarah: I think this is better

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -21,12 +21,17 @@ export default class PushBrowserDestinations extends Command {
   static examples = [`$ ./bin/run push-browser-destinations`]
 
   static flags = {
-    help: flags.help({ char: 'h' })
+    help: flags.help({ char: 'h' }),
+    env: flags.string({
+      char: 'e',
+      default: 'stage'
+    })
```
mshwery, line 28: You can source this from env vars too: 
```suggestion
    env: flags.string({
      char: 'e',
      default: 'stage',
      env: 'NODE_ENV' // or whatever
    })
```

mshwery: Also worth adding a description to the flag!

mshwery (approved):
Nice!

PR #74, opened 2021-08-18
remove additional properties instead of throwing

mshwery:
During a recent sev, we thought that we couldn't safely rollback changes because any destination configs with new fields would get invalidated via ajv because we were using `additionalProperties: false` (which would theoretically throw a 4xx validation error).

But it was also only partially true because ajv has `removeAdditional` set to true. That means that ajv itself will attempt to remove additional properties, but it will do this even when we don't want it to.

We want to be able to rollback safely, but also we want to protect the `perform` function from excess properties and preserve the type-safe schema builders have defined.

In this PR I've updated `additionalProperties: false` and tweaked the behavior to allow additional props for subproperties as well as adding tests for their removal and the coercion that happens.


dlasky (approved):
Thinking about future DX, do we want to at least display the removed keys or warn in the output?

packages/core/src/destination-kit/fields-to-jsonschema.ts
```diff
@@ -63,10 +66,11 @@ export function fieldsToJsonSchema(fields: Record<string, MinimalInputField> = {
     // Note: this is used for the json schema validation and type-generation,
     // but is not stored in the db. It only lives in the code.
     if (schemaType === 'object' && field.properties) {
+      const additionalProperties = true
```
mshwery, line 69: We want fields to accept additional properties. Later we can evaluate if builders should define this behavior, but for now we should make it universal.

packages/core/src/destination-kit/fields-to-jsonschema.ts
```diff
@@ -81,7 +85,7 @@ export function fieldsToJsonSchema(fields: Record<string, MinimalInputField> = {
   return {
     $schema: 'http://json-schema.org/schema#',
     type: 'object',
-    additionalProperties: false,
+    additionalProperties: overrides?.additionalProperties ?? false,
```
mshwery, line 88: The root object will always have it be false.

lcampos (approved):

krlv (approved):

PR #76, opened 2021-08-18
Fix type generation

mshwery:
This PR fixes a small type regression I introduced yesterday. We want type-safety even for object fields, but my changes to allow additional properties yields types like `[key: string]: unknown` on those object fields. So for now we must remove additional properties for all object fields until we give builders more control.

I added a test to verify the behavior, and a `todo` test for if/when we add different behavior based on builder definitions.

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/debounce/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/debounce/index.ts
@@ -42,7 +42,6 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
   description:
     'When enabled, it ensures that only events where at least one changed trait value are sent to Braze, and events with duplicate traits are not sent.',
   platform: 'web',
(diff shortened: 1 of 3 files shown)


packages/browser-destinations/src/destinations/braze/debounce/index.ts
```diff
@@ -42,7 +42,6 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
   description:
     'When enabled, it ensures that only events where at least one changed trait value are sent to Braze, and events with duplicate traits are not sent.',
   platform: 'web',
-  hidden: true,
```
mshwery, line 45: Unrelated, but a tiny change we want to make regardless.

krlv (approved):

lcampos (approved):

gpsamson (approved):
anything goes

PR #79, opened 2021-08-18
strip reserved keys from logUser

juliofarah:
Based on what we do for classic braze Identify, this PR omits reserved keys from `custom_attribuets` for the `logUser` action

see: https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/appboy/lib/index.js#L296-L319

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/logUser/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/logUser/index.ts
@@ -3,6 +3,7 @@ import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import appboy from '@braze/web-sdk'
 import dayjs from '../../../lib/dayjs'
+import { omit } from '@segment/actions-core'
 
 const known_traits = ['email', 'firstName', 'gender', 'city', 'avatar', 'lastName', 'phone']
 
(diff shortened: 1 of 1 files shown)


packages/browser-destinations/src/destinations/braze/logUser/index.ts
```diff
@@ -198,6 +199,15 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
     payload.phone !== undefined && user.setPhoneNumber(payload.phone)
     payload.push_subscribe !== undefined &&
       user.setPushNotificationSubscriptionType(payload.push_subscribe as appboy.User.NotificationSubscriptionTypes)
+
+    if (payload.custom_attributes !== undefined) {
+      const reservedKeys = Object.keys(action.fields.custom_attributes ?? {})
+      const customAttributes = omit(payload.custom_attributes, reservedKeys)
+
+      for (const [key, val] of Object.entries(customAttributes)) {
+        user.setCustomUserAttribute(key, val as string | number | boolean | Date | string[] | null)
```
mshwery, line 208: This is already happening up above?

juliofarah: not really!
when I removed `group_id` I realized `customAttributes` wasn't being set at all

mshwery: https://github.com/segmentio/action-destinations/pull/79/files#diff-8f5555b36eb6868dbb55a262131d23905136b9f45473d7a6c4489479ee73163eR180-R186

packages/browser-destinations/src/destinations/braze/logUser/index.ts
```diff
@@ -198,6 +199,15 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
     payload.phone !== undefined && user.setPhoneNumber(payload.phone)
     payload.push_subscribe !== undefined &&
       user.setPushNotificationSubscriptionType(payload.push_subscribe as appboy.User.NotificationSubscriptionTypes)
+
+    if (payload.custom_attributes !== undefined) {
+      const reservedKeys = Object.keys(action.fields.custom_attributes ?? {})
+      const customAttributes = omit(payload.custom_attributes, reservedKeys)
```
mshwery, line 205: This is kind of confusing... what are you trying to remove here? Those keys dont make sense to me.

juliofarah: I was just wondering if we should set `customAttributes` given we're not setting it anywhere

mshwery: But the keys you are removing here are the input field definition itself, like `multiple`, `type`, `label`, `description`, `required` etc. I dont think we want to use those as the reserved keys?

PR #81, opened 2021-08-18
Improve external developer experience

gpsamson:
This pull request aims to enable any external developer to clone, bootstrap, and test the `action-destinations` repo without access to private dependencies.

- Adds a CI step to ensure the external development experience is always maintained
- Updates some path mappings
- Ignores `browser-destinations` tests when the `analytics-next` dependency cannot be loaded.
- Adds `--ignore-engines` to Lerna's bootstrap command. The bootstrap command fails because the `@eslint/eslintrc` dependency targets Node v14.17 while this repo target v14.16: https://github.com/eslint/eslintrc/blob/main/package.json#L78
- Adds `--ignore-optional` as all optional dependencies are private or can be resolved locally. This is a necessary step because Yarn throws an error when an optional dependency is not cached (and it is not cached because it's private). Obviously, this goes against optional dependencies but I could not any recommended solution outside of "resolve this by upgrading to Yarn v2". We can remove this flag as private dependencies are removed or made open source. [0]

I've tested this branch using an external environment and the following set of commands succeed:
```
yarn install --ignore-optional --ignore-engines
yarn bootstrap
yarn test
```

Some of these stopgaps are likely to be removed in the future when there is more time to resolve them. I agree it's not the most elegant solution to use these flags. 🙇🏽 

_[0] Yarn's cache error:_
image


.github/workflows/ext.yml
```diff
@@ -0,0 +1,32 @@
...
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install
+        run: yarn global add lerna
+
+      - name: Bootstrap
+        run: yarn bootstrap
+
+      - name: Test
+        run: yarn test
```
gpsamson, line 29: I can't figure out why this fails for two tests, even though dozens of other similar tests are passing in the CI: https://github.com/segmentio/action-destinations/runs/3365630516?check_suite_focus=true#step:6:48

mshwery: Can you repro locally?

gpsamson: Negative. After a `lerna clean`, `rm -rf node_modules`, and `yarn cache clean` the following commands succeed:

```
yarn install --ignore-optional
yarn bootstrap
yarn test
```

package.json
```diff
@@ -20,7 +20,7 @@
     "types": "./bin/run generate:types",
     "lint": "eslint '**/*.ts' --cache",
     "subscriptions": "yarn workspace @segment/destination-subscriptions",
-    "test": "lerna run test --stream",
+    "test": "lerna run test --stream --ignore @segment/browser-destinations",
```
gpsamson, line 23: Rather than completely ignoring `browser-destinations`, we can programmatically ignore `browser-destinations` tests if the `analytics-next` dependency doesn't resolve.

See https://jestjs.io/docs/next/configuration#testpathignorepatterns-arraystring

packages/browser-destinations/jest.config.js
```diff
@@ -0,0 +1,27 @@
+module.exports = async () => {
```
gpsamson, line 1: Calling out inconsistency across packages if this is merged. Some configs will be defined in `package.json` adn others in `jest.config.js`.

As part of this change, we can consider introducing a root `jest.config.js` that is extended and switching other packages to `jest.config.js`.

packages/destination-actions/package.json
```diff
@@ -51,7 +51,8 @@
       "<rootDir>/dist/"
     ],
     "moduleNameMapper": {
-      "@segment/actions-core": "<rootDir>/../core/src/index.ts"
+      "@segment/actions-core": "<rootDir>/../core/src/index.ts",
+      "@segment/destination-subscriptions": "<rootDir>/../destination-subscriptions/src"
```
gpsamson, line 55: The consistent use of `moduleNameMapper` is a good argument for a global jest config.

.github/workflows/ext.yml
mshwery: Have you tried to remove your dist everywhere?

gpsamson: Yup, I just retried and ensured no `dist` folders existed -- still passes.

I'm noticing CI fails on the first test only. I wonder if there's a race condition between the bootstrap symlinks and tests (exploring that now)

juliofarah:
It looks good to me, but I'll the defer the +1 to @mshwery 

I'd like to suggest testing commands such as `push` and `push-browser-destination` before merging this PR as these seem to break very easily when we make structural changes on the repos!

Thanks Gabe!

packages/cli/package.json
```diff
@@ -79,6 +78,7 @@
   },
   "optionalDependencies": {
     "@segment/actions-core": "^3.4.2-alpha.0",
+    "@segment/browser-destinations": "^3.4.2-alpha.0",
```
mshwery, line 81: Weird that these are on alpha versions

mshwery (approved):

gpsamson:
Thanks for the reviews! I'll fix the conflicts (not sure why an alpha version was on stage at the time of writing) and ensure this doesn't break any commands.

packages/browser-destinations/jest.config.js
nettofarah: I think you may have misconfigured your editor. This file is using 4 spaces.

PR #82, opened 2021-08-18
add support for event-traits type

dlasky:
Adds support to the IR for event-traits and corresponding generation/parsing

### DIFF
--- a/packages/destination-subscriptions/src/types.ts
+++ b/packages/destination-subscriptions/src/types.ts
@@ -14,6 +14,7 @@ export type Condition =
 	| EventTypeCondition
 	| EventCondition
 	| EventPropertyCondition
+	| EventTraitCondition
 	| EventContextCondition
 
 export type GroupConditionOperator = 'and' | 'or'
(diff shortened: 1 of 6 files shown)


packages/destination-subscriptions/src/types.ts
```diff
@@ -65,6 +73,7 @@ export type ConditionType =
 	| 'event-type'
 	| 'event'
 	| 'event-property'
+	| 'event-trait'
 	| 'event-context'
 
 export type PropertyConditionType = 'event-property' | 'event-context'
```
mshwery, line 79: Do we need to add `event-trait` here?

mshwery (approved):
lgtm, do we need to update tests?

packages/destination-subscriptions/src/types.ts
dlasky: when we parse this back out in app there needs to be some way to differentiate at the UI level, currently thats done via this type explicitly

dlasky:
yep just wanted to get 👀 before proceeding

dlasky:
@mshwery one more look if you would, notably: d8eb2b2

mshwery (approved):

lcampos (approved):

PR #87, opened 2021-08-19
bump version of dest subs pkg

dlasky:


packages/destination-subscriptions/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@segment/destination-subscriptions",
-	"version": "3.4.0",
+	"version": "3.4.1",
```
lcampos, line 3: We should also update other package's that reference destination-subscriptions with this new version

lcampos (approved):

PR #89, opened 2021-08-19
[BE-104] Local Actions Server

jinapark202:
https://segment.atlassian.net/browse/BE-104

This PR creates a new CLI command that starts up the local actions server as described [here](https://paper.dropbox.com/doc/RFC-Local-Actions-Server--BQxmAAPnVspfctGjc2PYXac~Ag-69GsdN08nuDV9Vsalb9uJ). 

To start the local actions server, you need to run `./bin/run serve <DESTINATION_SLUG>` (e.g. `./bin/run serve google-analytics-4 -p 3001`. There are two optional flags, `port` (`-p`) and `directory` (`-d`). If not specified, it will default to port `3000` and the current directory.

An action can be invoked by sending a `POST` request with the format: `http://localhost:<PORT>/<DESTINATION_ACTION`. An example of a request body for `google-analytics-4`'s `search` action is as follows:

```
{
    "payload": {
        "client_id": 5,
        "search_term": "12"
    },
    "settings": {
        "measurementId": "1734",
        "apiSecret": "123"
    }
}
```

NOTE: I still need to test more and fix the Promise error, but feel free to provide feedback in the meantime.


packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+const app = express()
+app.disable('x-powered-by')
+app.use(express.json())
+
+const spinner: ora.Ora = ora()
+
+const port = process.env.port || 3000
+const server = http.createServer(app)
+
+app.post('/:action', async (req: express.Request, res: express.Response) => {
```
mshwery, line 21: This is dangerous – rejections/exceptions will bubble and crash the process. Express doesn't support async request handlers OOTB. You have to provide a special wrapper if you want to use async request functions (you can see examples in fab-5-engine itself).

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+  static description = `Starts a local development server to test your integration.`
+
+  static strict = false
+  static args = [
+    { name: 'destination', description: 'the destination to serve', required: true }
+  ]
+
+  static examples = [
+    `$ ./bin/run generate:serve`,
+    `$ ./bin/run generate:serve slack`
```
mshwery, line 26: A bit weary on adding extra keystrokes for the command – IMO I'd drop the `generate:` if possible

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+  ]
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    port: flags.integer({ char: 'p', default: 3000, description: 'port' }),
```
mshwery, line 31: Tip: you can pull the value from an env var too by adding `env: 'PORT'`

See https://oclif.io/docs/flags

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      child = fork(require.resolve('../lib/process-wrapper.ts'), ['../lib/server.ts'], {
+        cwd: process.cwd(),
+        env: {
+          ...process.env,
+          port: flags.port.toString(),
+          destination: args.destination,
+          directory: flags.directory
+        },
+        execArgv: ['-r', 'ts-node/register/transpile-only']
```
mshwery, line 64: We should consider passing the rest of argv (from `this.parse(Serve)`) so you can use all the normal Node / ts-node flags.

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+    const start = () => {
+      isRestarting = false
+
+      child = fork(require.resolve('../lib/process-wrapper.ts'), ['../lib/server.ts'], {
+        cwd: process.cwd(),
+        env: {
+          ...process.env,
+          port: flags.port.toString(),
+          destination: args.destination,
+          directory: flags.directory
```
mshwery, line 62: These env keys should be upcased

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      // Listen for `required` messages and watch the required file.
+      child.on('message', (message: Message) => {
+        if (!isRestarting && message.cmd === 'segment' && 'required' in message) {
+          watcher.add(message.required)
+        }
+      })
+    }
+
+    start()
```
mshwery, line 82: Shouldn't we `start()` at the very end?

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    watcher.on('change', (file) => {
+      this.log(chalk.greenBright`Restarting... ${file} has been modified`)
+      isRestarting = true
+
+      if (child) {
+        // Child is still running, restart upon exit
+        child.on('exit', start)
```
mshwery, line 90: Do we need to call `stop(child)` here?

mshwery (commented):
🙌 looks awesome :) left some questions/comments on mostly minor things!

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,132 @@
...
+        child = undefined
+      })
+
+      // Listen for `required` messages and watch the required file.
+      child.on('message', (message: Message) => {
+        if (!isRestarting && message.cmd === 'segment' && 'required' in message) {
+          watcher.add(message.required)
+        }
+      })
+    }
```
mshwery, line 80: This is a lil fancy. We could consider just blindly watching a directly and it would probably simplify a few things here, but this does have the benefit of only reloading when a file that actually was required changes (and that can include files outside the destination directory). I'm not opposed to keeping or dropping it.

jinapark202: I removed this to simplify things. Now we just look for changes inside the destination directory.

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+            })
+
+            const debug = await getExchanges(destination.responses)
+            res.json(debug)
+            spinner.succeed(chalk`${destination.name} action '${actionSlug}' completed with some code`)
+            return res.status(200).send()
+        }
+    } catch (err) {
+        spinner.fail()
+        return res.status(500).send(err.message)
```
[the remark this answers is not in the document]
jinapark202: `err.status` doesn't return anything. If I don't set the status as a `500`, it returns as a `200` so is it ok to just set it as `500`?

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,99 @@
...
+      default: './packages/destination-actions/src/destinations'
+    })
+  }
+
+  static destination: Destination
+
+  async run() {
+    const { args, flags } = this.parse(Serve)
+
+    const folderPath = path.join(process.cwd(), flags.directory, args.destination)
```
lcampos, line 34: Instead of asking the user for the slug/destination could we surface a menu with the existing destinations like we do in the push command ?

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,99 @@
...
+  static description = `Starts a local development server to test your integration.`
+
+  static strict = false
+  static args = [{ name: 'destination', description: 'the destination to serve', required: true }]
+
+  static examples = [`$ ./bin/run serve slack`, `$ PORT=3001 ./bin/run serve slack -p 3001`]
+
+  static flags = {
+    help: flags.help({ char: 'h' }),
+    port: flags.integer({ default: 3000, description: 'port', env: 'PORT' }),
```
lcampos, line 21: Lets have the default port be a constant used here and on `server.ts`

packages/cli/src/commands/serve.ts
jinapark202: I tried to pass in `argv` by adding it to `execArgv` as follows: `execArgv: ['-r', 'ts-node/register/transpile-only', ...argv]`. But it gave the following error where the directory path is wrong and I wasn't able to figure out a solution. What even do we need from `argv`? It doesn't seem like it holds any additional information when I log it
 ```
Error: Cannot find module '/Users/jinapark/dev/src/github.com/segmentio/action-destinations/amplitude'
```

File: packages/cli/src/lib/server.ts [hunk 12]
```diff
@@ -0,0 +1,105 @@
+import express from 'express'
+import http from 'http'
+import { once } from 'lodash'
+import logger from './logger'
+import path from 'path'
+import { loadDestination } from './destinations'
+import { Destination, CloudDestinationDefinition } from '@segment/actions-core'
```
lcampos, line 7: This is showing me error `Module '"@segment/actions-core"' has no exported member 'CloudDestinationDefinition'`

packages/cli/src/lib/http.ts
```diff
@@ -0,0 +1,55 @@
+import { DecoratedResponse as Response } from '@segment/actions-core'
```
lcampos, line 1: I suggest we rename this to something different than `http.ts` since we import node's native `http` module already and can get confusing

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+
+// Interrupt signal sent by Ctrl+C
+process.on('SIGINT', () => gracefulShutdown(0))
+
+server.on('error', (err: Error) => {
+  logger.error(`Server error: ${err.message}`, err)
+})
+
+server.listen(port, () => {
+  logger.info(`Listening at http://localhost:${port}`)
```
lcampos, line 104: It would be nice to post a message on how to call the destination that has been initialized for the server

```
$ ./bin/run serve google-analytics-4
Watching required files for changes .. 

INFO: Ready to process POST requests at http://localhost:3000/google-analytics-4
```

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    try {
+      const def = await loadDestination(targetDirectory) // loadDestinationDefinition
+      const destination = new Destination(def as CloudDestinationDefinition)
+
+      const actionSlug = req.params.action
+
+      const action = destination.actions[actionSlug]
+
+      if (action) {
```
lcampos, line 41: Should throw an error if the action was not defined or else we just keep showing the handling message to the user (`Handling google-analytics-4 action request`) giving them the impression the operation is still being processed.

packages/cli/src/lib/server.ts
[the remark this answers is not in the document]
mshwery: It _can_ throw http errors when you invoke an action. So the premise is that if (and only if) the error has a `status` property would we use it. Otherwise fallback to `500`.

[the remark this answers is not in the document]
mshwery: Can do something like `return res.status(err.status ?? 500).send(err.message)`

packages/cli/src/commands/serve.ts
mshwery: argv will contain whatever flags/args you pass when you run the server. The point of "forwarding" the original argv is so users can provide additional ts-node or node arguments that are accepted. For instance, if you want to attach a debugger, you can do that, but it requires arg forwarding.

mshwery: I'm guessing you are running into issues because we have a few flags here that weren't in my test branch. You'll have to manually split the parsed flags out?

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+  '/:action',
+  asyncHandler(async (req: express.Request, res: express.Response) => {
+    spinner.start(chalk`Handling ${process.env.DESTINATION} action request`)
+
+    const directory = process.env.DIRECTORY as string
+    const destinationPath = process.env.DESTINATION as string
+
+    // For now, include the slug in the path, but when we support external repos, we'll have to change this
+    const relativePath = path.join(directory, destinationPath, 'index.ts')
+    const targetDirectory = path.join(process.cwd(), relativePath)
```
mshwery, line 31: I wonder if this could be simplified to the point you give the server the "entrypoint" path to the definition instead of requiring 2 env vars to get there.

The serve command itself could keep things user-friendly, but this `server.ts` file could look at a single env var.

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,99 @@
...
+        child = undefined
+      })
+    }
+
+    watcher.on('change', (file) => {
+      this.log(chalk.greenBright`Restarting... ${file} has been modified`)
+
+      if (child) {
+        // Child is still running, restart upon exit
+        stop(child)
```
mshwery, line 66: This needs to come _after_ we call `child.on('exit', start)` iirc

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+      const def = await loadDestination(targetDirectory) // loadDestinationDefinition
+      const destination = new Destination(def as CloudDestinationDefinition)
+
+      const actionSlug = req.params.action
+
+      const action = destination.actions[actionSlug]
+
+      if (action) {
+        await action.execute({
+          data: req.body.payload,
```
lcampos, line 43: Should also validate `req.body` has all this data before sending it to the action or else the user just sees the request failed with a not that useful error `Unexpected token u in JSON at position`

packages/cli/src/lib/require-hook.ts
```diff
@@ -0,0 +1,37 @@
+export default function hook(wrapper: NodeModule, callback: (filename: string) => void) {
```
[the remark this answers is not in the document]
mshwery: This file is no longer used anywhere?

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const app = express()
+app.use(express.json())
+
+const spinner: ora.Ora = ora()
+
+const port = process.env.PORT || 3000
+const server = http.createServer(app)
+
+app.post(
```
lcampos, line 21: How is this working right now ? I didn't see the request parsing the body at all so `req.body` is always empty. Had to add `body-parser` so that I could successfully run this endpoint.

```
import bodyParser from 'body-parser'
...
app.use(bodyParser.json()); // for parsing application/json
```

Request
```
curl --request POST --url http://localhost:3000/search --header 'Content-Type: application/json' \
  --data '{
  "payload": {
    "client_id": "23455343467",
    "search_term": "$.properties.query"
  },
  "settings": {
    "apiSecret": "some-data",
    "measurementId": "G-<something>"
  }
}'
```

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      const actionSlug = req.params.action
+
+      const action = destination.actions[actionSlug]
+
+      if (action) {
+        await action.execute({
+          data: req.body.payload,
+          settings: req.body.settings,
+          mapping: req.body.payload,
```
lcampos, line 45: We should update this to `req.body.mapping` so we allow the user to also test defining custom mappings in the request.

```
{
  "payload": {
    "client_id": "23455343467",
    "search_term": "test call"
  },
  "settings": {
    "apiSecret": "some-data",
    "measurementId": "G-<something>"
  },
  "mapping": {
     "client_id": "$.context.userId",
     "search_term": "$.properties.query"
  }
}

```

packages/cli/src/lib/server.ts
lcampos: I actually just noticed we expect the action name to be part of the url path. Would it be better to change this to expect the destination name and action name as part of it `http://localhost:3000/google-analytics-4/search` instead of the current `http://localhost:3000/search` ? 
Something to consider if we ever want to support the server not exclusively supporting one destination at a time.

lcampos:
Left a couple of comments but this is looking really good !

packages/cli/src/lib/server.ts
mshwery: I'm not so sure we want builders to have to deal with mappings... it's only useful for them for testing `default` values really.

Also it wouldn't make sense to provide both `mapping` _and_ `payload`. The two are conflicting.

lcampos: I see, I do think there's value for a builder to be able to override the default mappings but that might be for more advanced scenarios. 

If we don't want to ask for mappings should we just hardcode an empty object instead of sending the payload in mappings ? Both `mapping` and `data` are required by `action.execute`.

mshwery: `mapping` can be the payload (it can be an object with fully resolved values). `data` doesn't need to be anything (unless we want to support passing in mapping-kit mappings)

packages/cli/src/lib/server.ts
jinapark202: So instead of having a user choose a destination when they run `./bin/run`, do you think we should just update the url parameters like `http://localhost:3000/google-analytics-4/search`?

mshwery: I don't know that it makes sense for devs to be working on multiple destinations simultaneously. I'd opt to keep things simpler and scoped to one destination at a time? My 2¢

packages/cli/src/lib/server.ts
jinapark202: Not sure why it didn't work for you. It's been working for me, both through Postman and with your curl command. I don't have `body-parser` but I do have `app.use(express.json())` , which I believe is the command used for Express 4.16+

packages/cli/src/lib/server.ts
lcampos: Yes, I agree to start with the simplest scenario I was just thinking about how I've been using the previous experience while developing destinations.

packages/cli/src/lib/server.ts
lcampos: @jinapark202 maybe I have some old dependencies in my local. If it works for you ignore this comment for now 👍

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,130 @@
...
+      choices: sortBy(Object.entries(manifest), '[1].definition.name').map(([metadataId, entry]) => ({
+        title: entry.definition.name,
+        value: metadataId
+      }))
+    })
+
+    const destination = manifest[metadataId].directory
+    const folderPath = path.join(process.cwd(), flags.directory, destination)
+
+    let child: ChildProcess | undefined
```
gpsamson, line 67:
```suggestion
    let child: ChildProcess | null = null
```
nit: Since we're assigning a value on L88, `null` is recommended as opposed to `undefined.

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,114 @@
...
+const port = process.env.PORT || DEFAULT_PORT
+const server = http.createServer(app)
+
+app.post(
+  '/:action',
+  asyncHandler(async (req: express.Request, res: express.Response) => {
+    spinner.start(chalk`Handling ${process.env.DESTINATION} action request`)
+
+    const destinationSlug = process.env.DESTINATION as string
+    const directory = process.env.DIRECTORY as string
```
gpsamson, line 28: nit: Move these up near other constants like the port

gpsamson (commented):
🚀  NICE WORK! Just minor nit comments on syntax and DX. This is game changing

packages/cli/src/commands/serve.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    const { metadataId } = await prompt<{ metadataId: string }>({
+      type: 'select',
+      name: 'metadataId',
+      message: 'Choose a definition:',
+      choices: sortBy(Object.entries(manifest), '[1].definition.name').map(([metadataId, entry]) => ({
+        title: entry.definition.name,
+        value: metadataId
+      }))
+    })
```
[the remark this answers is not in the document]
gpsamson: @lcampos I just saw that you recommended removing the destination arg. Curious of your thoughts on the above -- optional as opposed to required?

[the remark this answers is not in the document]
mshwery: Also btw you can use `autoPrompt` to automatically prompt for any args/flags that weren't provided when invoking the command!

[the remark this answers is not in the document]
lcampos: @gpsamson what Matt said is the experience I'm suggesting we have in this command

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,114 @@
...
+const spinner: ora.Ora = ora()
+
+const DEFAULT_PORT = 3000
+const port = process.env.PORT || DEFAULT_PORT
+const server = http.createServer(app)
+
+app.post(
+  '/:action',
+  asyncHandler(async (req: express.Request, res: express.Response) => {
+    spinner.start(chalk`Handling ${process.env.DESTINATION} action request`)
```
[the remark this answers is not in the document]
jinapark202: It actually does log, but because the response comes back super fast, it's hard to catch the message :)

packages/cli/src/lib/server.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    try {
+      const def = await loadDestination(targetDirectory)
+      const destination = new Destination(def)
+
+      const actionSlug = req.params.action
+      const action = destination.actions[actionSlug]
+
+      if (action) {
+        if (req.body.payload && req.body.settings) {
```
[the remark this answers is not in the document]
jinapark202: The reason for this check is because `action.execute` method doesn't allow `payload` and `settings` to be undefined. Are you suggesting changing the `execute` method itself?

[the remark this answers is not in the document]
gpsamson: Ahh, got it. I'm suggesting that we default to `{}` in the case that either is undefined.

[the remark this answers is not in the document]
jinapark202: Oh gotcha! Yeah we can do that

PR #90, opened 2021-08-20
Remove --no-git-tag-version from alpha script

gpsamson:
_Tags, we don't need no stinkin' tags_

This pull request adds a `--no-git-tag-version` flag to the `yarn alpha` script in order to avoid pushing git tags during releases. Documentation will be updated to with the `--no-git-tag-version` once this pull request is merged.

From [Lerna docs](https://github.com/lerna/lerna/tree/main/commands/publish#options):
> lerna publish supports all of the options provided by lerna version in addition to the following:

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "subscriptions": "yarn workspace @segment/destination-subscriptions",
     "test": "lerna run test --stream",
     "typecheck": "lerna run typecheck --stream",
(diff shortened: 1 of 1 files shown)


mshwery (approved):
~~Does this flag work for publish too? Didn't see it in their docs (other than `version`)~~

Oops missed your note!

PR #92, opened 2021-08-20
[BE-105] Render templated file names

daltonscharff:
[[BE-105](https://segment.atlassian.net/browse/BE-105)]

This update modifies the CLI so that templates with file names such as `{{slug}}.test` can be rendered using the Mustache templating engine.

For example:
file: `{{slug}}.test` + data: `{slug:"example"}` = filename: `example.test`

Considerations:
- This `renderTemplates` function copies all files within the template path to the target path before any template rendering occurs. This means that a files are renamed instead of copied with the correct name initially. The alternative would be to replace line 22: `fs.copySync` with a custom function, increasing code complexity. 
- If you pass in data with prohibited characters for a filename (this really only applies to Windows, I believe) fs-extra should throw an error, but this is untested


mshwery (approved):
Nice!

PR #94, opened 2021-08-20
Change alpha to canary release

dlasky:
This should prevent folks that are pushing multiple alphas across separate branches from stepping on each other.

Its not clear to me if the `--pre-dist-tag next` should stay or not from the lerna docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "subscriptions": "yarn workspace @segment/destination-subscriptions",
     "test": "lerna run test --stream",
     "typecheck": "lerna run typecheck --stream",
-    "alpha": "lerna publish prerelease --pre-dist-tag next --allow-branch '**' --no-git-tag-version"
+    "alpha": "lerna publish prerelease --pre-dist-tag next --canary --allow-branch '**' --no-git-tag-version"
   },
   "devDependencies": {
     "@types/jest": "^26.0.23",


mshwery (approved):

package.json
```diff
@@ -22,7 +22,7 @@
     "subscriptions": "yarn workspace @segment/destination-subscriptions",
     "test": "lerna run test --stream",
     "typecheck": "lerna run typecheck --stream",
-    "alpha": "lerna publish prerelease --pre-dist-tag next --allow-branch '**' --no-git-tag-version"
+    "alpha": "lerna publish prerelease --pre-dist-tag next --canary --allow-branch '**' --no-git-tag-version"
```
mshwery, line 25: Maybe we should use `--pre-dist-tag canary`

mshwery: That way we can reserve `next` to changes that aren't just to test in staging

lcampos, line 25: Based on the doc if you use `--canary` you don't need to use `prerelease` and `--pre-dist-tag` is exclusive to prereleases. However, there's this open issue https://github.com/lerna/lerna/issues/2060 with using canary which I think would cause use some problems.

dlasky: oof good find. This was an interesting tidbit in that thread: https://github.com/lerna/lerna/issues/2060#issuecomment-569349453

PR #95, opened 2021-08-23
Add fieldSchema as part of the action field payload

lcampos:
Update the CLI's push command to populate the new `fieldSchema` field. FYI this change will show diffs on all actions.

Tested successfully against staging using GA4 destination. This change depends on https://github.com/segmentio/control-plane/pull/1521

### DIFF
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -33,7 +33,7 @@ import {
   setSubscriptionPresets
 } from '../lib/control-plane-client'
 import { DestinationDefinition, hasOauthAuthentication } from '../lib/destinations'
-import { JSONSchema4 } from 'json-schema'
+import type { JSONSchema4 } from 'json-schema'
 
 type BaseActionInput = Omit<DestinationMetadataActionCreateInput, 'metadataId'>
 
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -33,6 +33,7 @@ import {
 } from '../lib/control-plane-client'
 import { DestinationDefinition, hasOauthAuthentication } from '../lib/destinations'
 import type { MinimalInputField } from '@segment/actions-core'
+import { JSONSchema4 } from 'json-schema'
 
 type BaseActionInput = Omit<DestinationMetadataActionCreateInput, 'metadataId'>
 
@@ -283,14 +284,12 @@ export default class Push extends Command {
   }
 }
 
-function getFieldPropertySchema(fieldKey: string, field: MinimalInputField) {
+function getFieldPropertySchema(fieldKey: string, field: MinimalInputField): JSONSchema4 {
   // Build a temporary object in which they key = field name and value = field properties
   // since that's the structure expected by fieldsToJsonSchema
   const tmpFieldObject: Record<string, MinimalInputField> = {}
   tmpFieldObject[fieldKey] = field
-  const schemaFormat = fieldsToJsonSchema(tmpFieldObject)
-
-  return schemaFormat.properties ? schemaFormat.properties[fieldKey] : ''
+  return fieldsToJsonSchema(tmpFieldObject)
 }
 
 function filterOAuth(optionList: string[]) {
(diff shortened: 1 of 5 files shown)


gpsamson (approved):
Change LGTM 👍🏽. I don't have much context, what exactly does the new `fieldSchema` property give us that others (eg; `label`, `required`, etc.) can't?

lcampos:
> Change LGTM 👍🏽. I don't have much context, what exactly does the new `fieldSchema` property give us that others (eg; `label`, `required`, etc.) can't?

@gpsamson we'll be storing the json shape for the fields and use that to render the new object field UI. I can share the SDD with you directly.

dlasky:
maybe i'm overthinking it, but dont we really only need `field.properties` in this case? since we already have the other top level stuff?

lcampos:
> maybe i'm overthinking it, but dont we really only need `field.properties` in this case? since we already have the other top level stuff?

You are correct that we only need that part for the object field UI work but Matt highlighted that storing the whole JSON for all field types would allow us to run client side checks (unrelated to object field changes and would need other changes) in the future.

yarn.lock
```diff
@@ -969,9 +969,9 @@
     source-map-support "^0.5.16"
 
 "@babel/runtime-corejs3@^7.12.1":
-  version "7.14.9"
-  resolved "https://registry.yarnpkg.com/@babel/runtime-corejs3/-/runtime-corejs3-7.14.9.tgz#fb21b1cf11650dcb8fcf4de2e6b3b8cf411da3f3"
-  integrity sha512-64RiH2ON4/y8qYtoa8rUiyam/tUVyGqRyNYhe+vCRGmjnV4bUlZvY+mwd0RrmLoCpJpdq3RsrNqKb7SJdw/4kw==
+  version "7.15.3"
```
lcampos, line 972: All these changes came from upgrading to the latest control-plane-service-client

packages/cli/src/commands/push.ts
```diff
@@ -152,7 +152,8 @@ export default class Push extends Command {
             choices: null,
             dynamic: field.dynamic ?? false,
             placeholder: field.placeholder ?? '',
-            allowNull: field.allowNull ?? false
+            allowNull: field.allowNull ?? false,
+            fieldSchema: field
```
mshwery, line 156: qq: would it be better to pass the json schema-tized version of the `field` definition? otherwise it looks like we are passing a bunch of redundant fields here that are already columns.

dlasky: thats what i was trying to get at here: https://github.com/segmentio/action-destinations/pull/95#issuecomment-904014969

lcampos: Both formats work for the object field UI, I'll change to json schema since that will allow us to use it for other things outside object field UI.

packages/cli/src/commands/push.ts
```diff
@@ -281,6 +283,16 @@ export default class Push extends Command {
...
 }
 
+function getFieldPropertySchema(fieldKey: string, field: MinimalInputField) {
+  // Build a temporary object in which they key = field name and value = field properties
+  // since that's the structure expected by fieldsToJsonSchema
+  const tmpFieldObject: Record<string, MinimalInputField> = {}
+  tmpFieldObject[fieldKey] = field
+  const schemaFormat = fieldsToJsonSchema(tmpFieldObject)
+
+  return schemaFormat.properties ? schemaFormat.properties[fieldKey] : ''
```
lcampos, line 293: Original field sample:
```
currency: {
  label: 'Currency',
  type: 'string',
  description: 'Currency of the purchase or items associated with the event, in 3-letter ISO 4217 format.'
}
```

This method returns the following structure
```
{
  title: 'Currency',
  description: 'Currency of the items associated with the event, in 3-letter ISO 4217 format.',
  type: 'string',
  format: undefined,
  default: undefined
}
```

Based on the original output from fieldsToJsonSchema
```
{
  '$schema': 'http://json-schema.org/schema#',
  type: 'object',
  additionalProperties: false,
  properties: {
    currency: {
      title: 'Currency',
      description: 'Currency of the items associated with the event, in 3-letter ISO 4217 format.',
      type: 'string',
      format: undefined,
      default: undefined
    }
  },
  required: []
}
```

@dlasky @mshwery let me know what you think about the format

mshwery: I think we may want to override additionalProperties? Or let builders decide

lcampos: Yes, I missed that.

dlasky (approved):

PR #96, opened 2021-08-23
[LIBWEB-860] resolves several issues with UTM and referrer properties on amplitude

dlasky:
- for `identify` referrer was clobbering UTM

- for `logEvent` it looks like we don't actually need to send a 2nd $identify event because the first event with no `user_properties` was actually stepping on the provided utm updates


packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
```diff
@@ -228,19 +230,23 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    const { utm_properties, referrer, userAgent, userAgentParsing, ...rest } = payload
+
+    const properties = rest as AmplitudeEvent
 
     if (Object.keys(utm_properties ?? {}).length || referrer) {
-      userProperties = { ...convertUTMProperties(payload), ...convertReferrerProperty(payload) }
+      properties.user_properties = mergeUserProperties(
+        omit(properties.user_properties ?? {}, ['utm_properties', 'referrer']),
+        convertUTMProperties(payload),
+        convertReferrerProperty(payload)
```
mshwery, line 241: I'm a little concerned passing in `payload` when we've specifically extracted properties from it above (line 233). Can we pass a more specific subset so we don't risk the same issues we had previously?

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
```diff
@@ -205,14 +213,14 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      })
-    }
+    // if (Object.keys(payload.utm_properties ?? {}).length || payload.referrer) {
+    //   const user_properties = mergeUserProperties(convertUTMProperties(payload), convertReferrerProperty(payload))
+    //   events.push({
+    //     ...properties,
+    //     event_type: '$identify',
+    //     user_properties
+    //   })
+    // }
```
mshwery, line 223: Can we remove commented out code?

mshwery (commented):
Looks like this isn't quite ready since there is quite a bit of code commented out, but generally the remaining code changes look ok!

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
dlasky: 😓

PR #97, opened 2021-08-24
[LIBWEB-739] add min_id_length support to all amplitude partner actions

dlasky:
[LIBWEB-739] 

adds support for a `min_id_length` property. This is only officially documented for the v2 api, however subsequent testing revealed that it is supported for the additional apis.

[LIBWEB-739]: https://segment.atlassian.net/browse/LIBWEB-739

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
+++ b/packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
@@ -348,7 +348,7 @@ describe('Amplitude', () => {
             "api_key",
             "undefined",
             "identification",
-            "{\\"user_id\\":\\"some-user-id\\",\\"device_id\\":\\"some-anonymous-id\\",\\"country\\":\\"United States\\",\\"city\\":\\"San Francisco\\",\\"language\\":\\"en-US\\",\\"user_properties\\":{\\"some-trait-key\\":\\"some-trait-value\\"},\\"library\\":\\"segment\\"}",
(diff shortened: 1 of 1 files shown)


packages/destination-actions/src/destinations/amplitude/groupIdentifyUser/index.ts
```diff
@@ -67,10 +67,21 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       description: 'Value of the group',
       required: true
+    },
+    min_id_length: {
+      label: 'Minimum ID Length',
+      description: 'Minimum permitted length for user_id and device_id fields',
```
mshwery, line 73: I think this description could be clearer for customers. When would I want to use this? I see in the ticket this is to override the default min length for specific circumstances. Maybe we could include that here or add a markdown link to their docs so customers can figure it out.

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
```diff
@@ -181,6 +192,13 @@ const action: ActionDefinition<Settings, Payload> = {
       properties.session_id = dayjs.utc(session_id).valueOf()
     }
 
+    if (min_id_length && min_id_length > 0) {
+      options = { min_id_length }
+    }
+
+    // console.log('PROPERTIES', properties)
+    // debugger
```
mshwery, line 200: Remove

mshwery (commented):
lgtm but need to clean up the commented code

packages/destination-actions/src/destinations/amplitude/groupIdentifyUser/index.ts
```diff
@@ -67,10 +67,22 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      allowNull: true,
+      type: 'integer'
     }
   },
   perform: async (request, { payload, settings }) => {
     const groupAssociation = { [payload.group_type]: payload.group_value }
+    const { min_id_length } = payload
+    let options
+    if (min_id_length && min_id_length > 0) {
+      options = JSON.stringify({ min_id_length })
```
mshwery, line 84: Why are we stringifying?

PR #114, opened 2021-08-30
[braze web] Remove changeUser calls from trackEvent and trackPurchase

juliofarah:
This PR removes `changeUser` calls from `trackEvent` and `trackPurchase` actions as `analytics.track()` doesn't accept a `userId` and will always send the one from the most recent identify call


juliofarah:
> Can you remove the dotfile that got added?

Oh vim :face_palm:

nielst:
I've been back and forth on this myself... changeUser needs a userId, and analytics.track() doesn't accept one, so we dont really have one that we can pass to changeUser. The BrazeSDK itself generates the BrazeID and includes on the track calls.

If they remove all other actions, the track call will still be sent to Braze with the BrazeID

PR #116, opened 2021-08-30
[braze-web] Fix trackPurchase products mapping

juliofarah:
the `trackPurchase` action wasn't being invoked because of a few issues:
- the customer passes in product properties in snake_case instead of camelCase;
- the `products` array is located inside the `purchaseProperties` object;

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts
@@ -67,8 +67,8 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
     const reservedKeys = Object.keys(action.fields.products.properties ?? {})
     const purchaseProperties = omit(payload.purchaseProperties, reservedKeys)
 
-    if (payload.purchaseProperties?.products && Array.isArray(payload.purchaseProperties?.products)) {
-      payload.purchaseProperties?.products?.forEach((product) => {
+    if (purchaseProperties?.products && Array.isArray(purchaseProperties?.products)) {
+      purchaseProperties?.products?.forEach((product) => {
         const result = client.logPurchase(
           product.product_id,
           product.price,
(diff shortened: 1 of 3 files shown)


packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts
```diff
@@ -27,7 +27,7 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
       label: 'Products',
       description: 'List of products purchased by the user',
       properties: {
-        productId: {
+        product_id: {
```
mshwery, line 30: This will be a breaking change for any existing instances.

packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts
```diff
@@ -39,7 +39,7 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
           required: true,
           description: `The price paid. Base units depend on the currency. As an example, USD should be reported as Dollars.Cents, whereas JPY should be reported as a whole number of Yen. All provided values will be rounded to two digits with toFixed(2)`
         },
-        currencyCode: {
+        currency: {
```
mshwery, line 42: Same here

mshwery (approved):

packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts
```diff
@@ -67,19 +67,21 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
...
 
-    payload.products?.forEach((product) => {
-      const result = client.logPurchase(
-        product.productId,
-        product.price,
-        product.currencyCode ?? 'USD',
-        product.quantity ?? 1,
-        purchaseProperties
-      )
+    if (payload.purchaseProperties?.products && Array.isArray(payload.purchaseProperties?.products)) {
```
mshwery, line 70: Any reason not to use `purchaseProperties` defined right above?

juliofarah: nope, not really.
gonna refactor

packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts
juliofarah: That's OK for now, as we only have @nielst 's source connected to that destination

PR #117, opened 2021-08-30
[braze web] Fix traits being passed as custom_attributes on

juliofarah:
This PR fixes traits (such as `firstName` and `lastName`) being passed as `custom_attributes` to braze

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
@@ -174,6 +174,7 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
       }
     }
 
+    // Adding `firstName` and `lastName` here as these fields are mapped using cammel_case.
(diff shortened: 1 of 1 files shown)


packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
```diff
@@ -174,9 +174,9 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
       }
     }
 
+    const reservedFields = [...Object.keys(action.fields ?? {}), 'firstName', 'lastName']
```
mshwery, line 177: What's the reasoning for these other keys? Maybe include a code comment

mshwery:
```suggestion
    const reservedFields = [...Object.keys(action.fields), 'firstName', 'lastName']
```

mshwery (approved):

PR #118, opened 2021-08-30
add nested resolution for @literal directives

mshwery:
I noticed that `@literal` directives containing an object with nested directives wasn't getting fully resolved. This fixes that behavior and adds some tests.


gpsamson (approved):

dlasky:
🚒

PR #122, opened 2021-09-01
Facebook Conversions API Destination

nick-Ag:
Adds a new Facebook Conversions API destination using the actions framework. [FB CAPI Spec](https://paper.dropbox.com/doc/Facebook-Conversions-API-Actions-Spec--BQrh7_xm1L2Nb5Iia4CIGuETAg-pIPezlMTTiTRm43JCuG6p).

This destination is the actions version of the 'Facebook Pixel Server Side' classic destination.

Facebook requires an access token to make calls to graph api and this PR is dependent on implementing builder defined secrets in the actions framework. For testing purposes I've implemented the token as a workspace setting that can be passed in. The PR won't be merged in until the builder defined secrets work is added to it.


packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    return request(`https://graph.facebook.com/v11.0/${settings.pixelId}/events`, {
+      method: 'POST',
+      json: {
+        data: [
+          {
+            event_name: 'Purchase',
+            event_time: payload,
+            action_source: payload.action_source,
+            user_data: {
+              em: payload.user_data.email,
```
nick-Ag, line 52: Facebook requires most of these user data fields to be [hashed to accept the event.](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/customer-information-parameters#normalize-and-hash) Would it make sense to validate that these are hashed? I was going to implement something like this to accomplish that https://melvingeorge.me/blog/check-if-string-is-valid-sha256-hash-javascript

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/__tests__/index.test.ts
```diff
@@ -0,0 +1,94 @@
+import nock from 'nock'
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
+import Destination from '../../index'
+import { token } from '/Users/nicholas.aguilar/Desktop/token'
+
+const testDestination = createTestIntegration(Destination)
+
+const settings = {
+  pixelId: '545342626773734',
```
nick-Ag, line 9: This is just my test pixel, not sensitive

packages/destination-actions/src/destinations/actions-facebook-conversions/index.ts
```diff
@@ -0,0 +1,34 @@
+import type { DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+import purchase from './purchase'
+import { token } from '/Users/nicholas.aguilar/Desktop/token'
```
nick-Ag, line 4: Using a staging token and pulling it in

packages/destination-actions/src/destinations/actions-facebook-conversions/constants.ts
```diff
@@ -0,0 +1,181 @@
+export const CURRENCY_ISO_CODES = [
```
mericsson, line 1: If you use `Set` instead of `Array` you will get a minor perf improvement when calling `.has` vs `.includes` (assuming integration is handling large throughput of events this will be worth it -- https://egghead.io/lessons/javascript-use-es6-sets-to-improve-javascript-performance )


```suggestion
export const CURRENCY_ISO_CODES = new Set([
```

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/__tests__/index.test.ts
```diff
@@ -0,0 +1,94 @@
...
+            '@path': '$.properties.currency'
+          },
+          value: {
+            '@path': '$.properties.value'
+          }
+          //No user data mapping
+        }
+      })
+    } catch (e) {
+      expect(e.message).toBe('Must include at least one user data property')
```
mericsson, line 91: I think this test may result in false positives as it is currently configured. You are expecting an error, right? If no error is thrown the test will still pass. You may need to update the test to have pattern:

```js
try {
  perform()
  fail('Did not expect `perform` to succeed')
} catch (err) {
  expect(err.message).toBe('Must include at least one user data property')
}
```

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    content_ids: content_ids,
+    action_source: {
+      label: 'Action Source',
+      description: 'The action source of the event',
+      type: 'string'
+    },
+    user_data: user_data
+  },
+  perform: (request, { payload, settings }) => {
+    if (!CURRENCY_ISO_CODES.includes(payload.currency)) {
```
mericsson, line 27: If you make the `Set` update https://github.com/segmentio/action-destinations/pull/122/files#r699840738

... then this changes to

```suggestion
    if (!CURRENCY_ISO_CODES.has(payload.currency)) {
```

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    if (payload.action_source === 'website' && payload.user_data.client_user_agent === undefined) {
+      throw new IntegrationError(
+        'If action source is "Website" then client_user_agent must be defined',
+        'Misconfigured required field',
+        400
+      )
+    }
+
+    return request(`https://graph.facebook.com/v11.0/${settings.pixelId}/events`, {
```
mericsson, line 43: For our classic integration, we have this `11.0` behind a flagon flag so we can roll out API version updates slowly. fwiw until we have versioning here we will not be able to do that.

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/index.ts
mericsson: It looks like we hash in the classic integration -- so I think we want to hash here as well too.

packages/destination-actions/src/destinations/index.ts
```diff
@@ -23,6 +23,7 @@ export const manifest: Record<MetadataId, ManifestEntry> = {}
  * in production and sync them into staging via `sprout`.
  */
 register('60f64ae3eaebd66d17d28e9f', './1plusx')
+register('6127d2a468778ec5f0023103', './actions-facebook-conversions')
```
mericsson, line 26: Wonder if this should be

```suggestion
register('6127d2a468778ec5f0023103', './facebook-conversions-api')
```

to match other destinations and name of integration.

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    action_source: {
+      label: 'Action Source',
+      description: 'The action source of the event',
+      type: 'string'
+    },
+    user_data: user_data
+  },
+  perform: (request, { payload, settings }) => {
+    if (!CURRENCY_ISO_CODES.includes(payload.currency)) {
+      throw new Error(`${payload.currency} is not a valid currency code.`)
```
mericsson, line 28: Curious, why `Error` here and elsewhere `IntegrationError`?

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/__tests__/index.test.ts
```diff
@@ -0,0 +1,94 @@
...
+        event,
+        settings,
+        mapping: {
+          currency: {
+            '@path': '$.properties.currency'
+          },
+          value: {
+            '@path': '$.properties.value'
+          }
+          //No user data mapping
```
mericsson, line 87:
```suggestion
          // No user data mapping included. This should cause action to fail.
```

mericsson:
@nick-Ag Thanks for adding this! Can you update PR description to have brief description of changes included in the PR as well as what is remaining before it is ready for full review / merge?

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/__tests__/index.test.ts
lcampos: We should replace this with a test value (e.g. '123456..`) since all the requests are mocked

packages/destination-actions/src/destinations/index.ts
lcampos, line 26: We don't need to use the `actions-` prefix here, let's remove it and follow what we've done in other destinations.

packages/destination-actions/src/destinations/actions-facebook-conversions/index.ts
lcampos: We can also replace this with a field in `authentication` so others can run this code or if we want to test this in an environment like stage

packages/destination-actions/src/destinations/actions-facebook-conversions/constants.ts
nick-Ag: Good call, we can also use the set for the GA4 constant.ts as well

nick-Ag: Actually, GA4 and FB CAPI are going to both use this same set. Would it make sense to pull the `constants.ts` out into /lib so they can both import from the same file. And later on if other destinations need to check currency it can be reused

packages/destination-actions/src/destinations/actions-facebook-conversions/index.ts
mshwery: Would recommend adding [`dotenv`](https://www.npmjs.com/package/dotenv) and storing these in a `.env` file.

mshwery: I'm going to add this in a separate PR. You can use it once it's landed in `main`.

packages/destination-actions/src/destinations/actions-facebook-conversions/constants.ts
mshwery: I would hesitate to introduce a cross-destination dependency. IMO we should keep destinations as isolated as possible.

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/__tests__/index.test.ts
mshwery: The jest way to do this is:
```ts
await expect(promiseReturningFn()).rejects.toThrowError(/your error/)
```

More info in the docs: 
https://jestjs.io/docs/tutorial-async

Alternately, you can tell jest how many assertions you expect in the test:
```ts
expect.assertions(1)
```

packages/destination-actions/src/destinations/index.ts
mshwery: It doesn't really matter tbh. Convention is that the directory names do not have the redundant `actions-` part that is found in most of these destinations' slugs.

packages/destination-actions/src/destinations/actions-facebook-conversions/purchase/index.ts
mshwery: Use `IntegrationError` as you can provide more information, including status codes.

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-user-data.ts
```diff
@@ -0,0 +1,248 @@
+import { InputField } from '@segment/actions-core/src/destination-kit/types'
+import { createHash } from 'crypto'
+import { US_STATE_CODES, COUNTRY_CODES } from './constants'
+import { Payload } from './addToCart/generated-types'
+
+// Implementation of Facebook user data object
+// https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/customer-information-parameters
+export const user_data_field: InputField = {
+  label: 'User Data',
```
kdaswani, line 9: I see User Data appearing in each Action like this in staging:
Screen Shot 2021-11-10 at 11 59 58 AM

Is that expected?

kdaswani (commented):
Reviewed - looks good. I added some comments and revisions. Thanks @nick-Ag!

packages/destination-actions/src/destinations/facebook-conversions-api/custom/index.ts
```diff
@@ -0,0 +1,57 @@
+import { ActionDefinition, IntegrationError } from '@segment/actions-core'
+import { action_source, event_id, event_source_url, event_time } from '../fb-capi-properties'
+import { hash_user_data, user_data_field } from '../fb-capi-user-data'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { API_VERSION } from '../constants'
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Custom',
```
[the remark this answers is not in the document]
nick-Ag: I think it would be best to leave `custom_data` as object-key so users can add fields as needed, and so we won't nave to make any changes if FB adds new object properties. (Implementing now). This change says 'As a user, I want to add a key value pair to an object, but only if the builder has not provided details on the shape of the object (child fields).' so we won't be able to take both approaches right?

packages/destination-actions/src/destinations/facebook-conversions-api/search/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    contents: contents,
+    currency: currency,
+    event_id: event_id,
+    event_source_url: event_source_url,
+    search_string: {
+      label: 'Search String',
+      description: 'Search String',
+      type: 'string',
+      default: {
+        '@path': '$.properties.query'
```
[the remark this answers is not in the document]
nick-Ag: Is there an example of what the object shape would look like? If users provide a mapping to a string inside the object it would work but without that it would fail. I don't think it's currently possible to stringify an object since actions will reject if the types don't match before we hit the `perform` method

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
```diff
@@ -0,0 +1,106 @@
...
+    '@path': '$.timestamp'
+  }
+}
+
+export const action_source: InputField = {
+  label: 'Action Source',
+  description: 'This field allows you to specify where your conversions occurred.',
+  type: 'string',
+  default: {
+    '@path': '$.properties.action_source'
```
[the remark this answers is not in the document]
nick-Ag: Segment events don't have `action_source` do they? (Didn't see them when I searched rn), I just copy pasted that default from spec, but will remove

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
```diff
@@ -0,0 +1,106 @@
...
+    '@path': '$.properties.currency'
+  }
+}
+
+export const value: InputField = {
+  label: 'Value',
+  description: 'The value of a user performing this event to the business.',
+  type: 'number',
+  default: {
+    '@path': '$.properties.revenue'
```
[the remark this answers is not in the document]
nick-Ag: I'll move the default to the individual action level instead of using coalesce. 
Going to do:
addToCart, viewContent, value: default: `$.properties.price`
Purchase event value: default: `$.properties.revanue`

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
```diff
@@ -0,0 +1,106 @@
...
+  type: 'string'
+}
+
+export const content_ids: InputField = {
+  label: 'Content IDs',
+  description: 'Product IDs associated with the event, such as SKUs (e.g. ["ABC123", "XYZ789"]).',
+  type: 'string',
+  multiple: true,
+  default: {
+    '@path': '$.properties.product_id'
```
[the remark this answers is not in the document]
nick-Ag: I think it's best to not override defaults since it might cause confusion in the future (if the override is removed it would still have a default). I'll move these defaults to the action level

packages/destination-actions/src/destinations/facebook-conversions-api/custom/index.ts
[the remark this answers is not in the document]
kdaswani: Correct! My second suggestion was to not use an object type field and instead make each object field in the FB docs a separate mapping. But, I agree we should just add a single mapping for `custom_data` and make it an `object` type, and then customers can add fields as needed without us providing any shape/predefined fields.

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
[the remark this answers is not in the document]
kdaswani: They usually don't since we don't tell customers to do so in our spec. So let's remove and not provide a default for this. Thanks!

packages/destination-actions/src/destinations/facebook-conversions-api/custom/index.ts
```diff
@@ -0,0 +1,63 @@
...
+        '@path': '$.event'
+      }
+    },
+    event_time: { ...event_time, required: true },
+    user_data: user_data_field,
+    event_id: event_id,
+    event_source_url: event_source_url,
+    custom_data: {
+      label: 'Custom Data',
+      description: 'The custom data object which can be used to pass the custom properties described at: https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/custom-data',
```
kdaswani, line 28: Thanks for doing this @nick-Ag! Small thing, but is it possible to not include the entire URL and instead hyperlink it.

```suggestion
      description: 'The custom data object which can be used to pass the custom properties. See [here](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/custom-data) for more information.',
```

packages/destination-actions/src/destinations/facebook-conversions-api/search/index.ts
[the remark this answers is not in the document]
kdaswani: Hmm I don't think we have an example in our docs but I imagine it would look like this:
`query: ['shirts', 'blue']`

I'm not sure how common that is though so let's just update the description for `search_string`. I'll add a suggestion for that.

packages/destination-actions/src/destinations/facebook-conversions-api/search/index.ts
```diff
@@ -0,0 +1,79 @@
...
+    user_data: user_data_field,
+    content_category: content_category,
+    content_ids: content_ids,
+    contents: contents,
+    currency: currency,
+    event_id: event_id,
+    event_source_url: event_source_url,
+    search_string: {
+      label: 'Search String',
+      description: 'Search String',
```
kdaswani, line 24:
```suggestion
      description: 'A search query made by a user. This must be a string.',
```

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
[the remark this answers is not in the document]
kdaswani: Out of curiosity, what does `multiple: true` do? If we put a default on the multi-product Actions of `properties.products.id` would it grab the various IDs in the payload and put them together in an array of strings? FB [expects an array of strings](https://developers.facebook.com/docs/marketing-api/conversions-api/parameters/custom-data) so want to make sure I understand this correctly :)

packages/destination-actions/src/destinations/facebook-conversions-api/purchase/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  defaultSubscription: 'type = "track" and event = "Order Completed"',
+  fields: {
+    action_source: { ...action_source, required: true },
+    currency: { ...currency, required: true },
+    event_time: { ...event_time, required: true },
+    user_data: user_data_field,
+    value: { 
+      ...value, 
+      required: true, 
+      default: { '@path': '$.properties.revanue' }
```
kdaswani, line 20:
```suggestion
      default: { '@path': '$.properties.revenue' } 
```

packages/destination-actions/src/destinations/facebook-conversions-api/initiateCheckout/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    content_category: content_category,
+    content_ids: content_ids,
+    contents: contents,
+    currency: currency,
+    event_id: event_id,
+    event_source_url: event_source_url,
+    num_items: num_items,
+    value: { 
+      ...value,
+      default: { '@path': '$.properties.revanue' }
```
kdaswani, line 25:
```suggestion
      default: { '@path': '$.properties.revenue' } 
```

packages/destination-actions/src/destinations/facebook-conversions-api/purchase/index.ts
```diff
@@ -0,0 +1,104 @@
...
+      ...value, 
+      required: true, 
+      default: { '@path': '$.properties.revenue' } 
+    },
+    content_ids: content_ids,
+    content_name: content_name,
+    content_type: content_type,
+    contents: {
+      ...contents,
+      default: [{
```
kdaswani, line 27: @nick-Ag Is there supposed to be a default mapping with `arrayPath` or something so that there's a default object in the UI? cc @dlasky 

If so, we'll need that object to be `properties.products` for Purchase and Initiate Checkout. We'll have the object be `properties` for ViewContent, Search and AddToCart

packages/destination-actions/src/destinations/facebook-conversions-api/addToCart/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    content_name: content_name,
+    content_type: content_type,
+    contents: { 
+      ...contents,
+      default: [{
+        item_price: {
+          '@path': '$.properties.price'
+        },
+        id: {
+          '@path': '$.product_id'
```
kdaswani, line 26: I'm not sure what the correct syntax is when we use `arrayPath` but I think this should be found within properties.

```suggestion
          '@path': '$.properties.product_id'
```

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-user-data.ts
kdaswani: This looks better now!

packages/destination-actions/src/destinations/facebook-conversions-api/initiateCheckout/index.ts
```diff
@@ -0,0 +1,88 @@
...
+  title: 'Initiate Checkout',
+  description: 'Send event when a user enters the checkout flow',
+  defaultSubscription: 'type = "track" and event = "Checkout Started"',
+  fields: {
+    action_source: { ...action_source, required: true },
+    event_time: { ...event_time, required: true },
+    user_data: user_data_field,
+    content_category: content_category,
+    content_ids: content_ids,
+    contents: contents,
```
kdaswani, line 18: We should add the same `contents` object fields and defaults as Purchase to this InitiateCheckout Action

packages/destination-actions/src/destinations/facebook-conversions-api/search/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  title: 'Search',
+  description: 'Send event when a user searches content or products',
+  defaultSubscription: 'type = "track" and event = "Products Searched"',
+  fields: {
+    action_source: { ...action_source, required: true },
+    event_time: { ...event_time, required: true },
+    user_data: user_data_field,
+    content_category: content_category,
+    content_ids: content_ids,
+    contents: contents,
```
kdaswani, line 18: We should add the same `contents` object fields and defaults as AddToCart to this Search Action.

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
```diff
@@ -0,0 +1,116 @@
...
+  description: 'Either product or product_group based on the content_ids or contents being passed.',
+  type: 'string'
+}
+
+export const contents: InputField = {
+  label: 'Contents',
+  description: 'An array of JSON objects that contains the quantity and the International Article Number (EAN) when applicable, or other product or content identifier(s). id and quantity are the required fields.',
+  type: 'object',
+  multiple: true,
+  properties: {
```
kdaswani, line 52: Are `id` and `quantity` required? Out of curiosity, if `contents` is not required for certain Actions but `id` and `quantity` are required in the `contents` object, what happens? Should we put requiredness for `contents` and sub-fields at the Action-level?

packages/destination-actions/src/destinations/facebook-conversions-api/addToCart/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  fields: {
+    action_source: { ...action_source, required: true },
+    event_time: { ...event_time, required: true },
+    user_data: user_data_field,
+    content_ids: { ...content_ids, default: { '@path': '$.properties.product_id' } },
+    content_name: content_name,
+    content_type: content_type,
+    contents: { 
+      ...contents,
+      default: [{
```
kdaswani, line 21: @nick-Ag Is there supposed to be a default mapping with `arrayPath` or something so that there's a default object in the UI? Asked the same thing on the Purchase Action.

If so, we'll need that object to be `properties.products` for Purchase and Initiate Checkout. We'll have the object be `properties` for ViewContent, Search and AddToCart

packages/destination-actions/src/destinations/facebook-conversions-api/viewContent/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    content_name: content_name,
+    content_type: content_type,
+    contents: { 
+      ...contents,
+      default: [{
+        item_price: {
+          '@path': '$.properties.price'
+        },
+        id: {
+          '@path': '$.product_id'
```
kdaswani, line 29: I'm not sure what the correct syntax is when we use arrayPath but I think this should be found within properties.

```suggestion
          '@path': '$.properties.product_id'
```

packages/destination-actions/src/destinations/facebook-conversions-api/viewContent/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    action_source: { ...action_source, required: true },
+    event_id: event_id,
+    event_source_url: event_source_url,
+    content_ids: { ...content_ids, default: { '@path': '$.properties.product_id' } },
+    content_category: content_category,
+    content_name: content_name,
+    content_type: content_type,
+    contents: { 
+      ...contents,
+      default: [{
```
kdaswani, line 24: Same question about including some `arrayPath` to the `properties` object

packages/destination-actions/src/destinations/facebook-conversions-api/addToCart/index.ts
dlasky: https://github.com/segmentio/action-destinations/blob/main/packages/core/src/mapping-kit/README.md has some examples

kdaswani: Thanks, Dan! Would the default be something like this for single-product Actions like Add to Cart, View Content and Search?

```
default: {
  "@arrayPath": ["$.properties", {
    "id": { "@path": "$.product_id" },
    "item_price": { "@path": "$.price" },
    "quantity": { "@path": "$.quantity" },
  }]
}
```

And this for multi-product Actions like Purchase and Checkout Initiated?

```
default: {
  "@arrayPath": ["$.properties.products", {
    "id": { "@path": "$.product_id" },
    "item_price": { "@path": "$.price" },
    "quantity": { "@path": "$.quantity" },
  }]
}
```

mshwery: fyi `@arrayPath` should only be used for array fields. if not an array:

```
default: {
  "id": { "@path": "$.product_id" },
  "item_price": { "@path": "$.price" },
  "quantity": { "@path": "$.quantity" }
}
```

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
nick-Ag: FB has some conflicting docs on this, but in practice their API only rejects if there is no id included, so I enforced that in the `perform` method.

packages/destination-actions/src/destinations/facebook-conversions-api/addToCart/index.ts
```diff
@@ -0,0 +1,118 @@
...
+      throw new IntegrationError(
+        'If action source is "Website" then client_user_agent must be defined',
+        'Misconfigured required field',
+        400
+      )
+    }
+
+    const valid_delivery_categories = ['in_store', 'curbside', 'home_delivery']
+    if (payload.contents) {
+      payload.contents.forEach((obj, index) => {
```
brennan, line 73: Do we first need to check whether `payload` is iterable?

packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/custom.test.ts
```diff
@@ -0,0 +1,154 @@
...
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
+import Destination from '../index'
+
+const testDestination = createTestIntegration(Destination)
+const settings = {
+  pixelId: '123321',
+  token: process.env.TOKEN
+}
+
+describe.only('FacebookConversionsApi', () => {
```
brennan, line 11: Remove `.only`.

packages/destination-actions/src/destinations/facebook-conversions-api/addToCart/index.ts
nick-Ag: I don't think so, if a non-array is mapped to `contents` actions will throw an error before we reach this `perform` block

packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/viewContent.test.ts
```diff
@@ -78,7 +78,7 @@ describe('FacebookConversionsApi', () => {
       expect(responses[0].status).toBe(201)
     })
 
-    it('should handle default mappings', async () => {
+    it.only('should handle default mappings', async () => {
```
brennan, line 81: Delete `only`.

packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/addToCart.test.ts
```diff
@@ -0,0 +1,274 @@
...
+const testDestination = createTestIntegration(Destination)
+
+const settings = {
+  pixelId: '123321',
+  token: process.env.TOKEN
+}
+describe('FacebookConversionsApi', () => {
+  describe('AddToCart', () => {
+    it('should handle a basic event', async () => {
+      nock(`https://graph.facebook.com/v11.0/${settings.pixelId}`).post(`/events`).reply(201, {})
```
brennan, line 14: v11.0 deprecates September 14, 2023, so I think we're good to leave this version rather than proactively bumping to v12.0.

packages/destination-actions/src/destinations/facebook-conversions-api/addToCart/index.ts
```diff
@@ -0,0 +1,123 @@
...
+              '@path': '$.price'
+            }
+          }
+        ]
+      }
+    },
+    currency: currency,
+    event_id: event_id,
+    event_source_url: event_source_url,
+    value: { ...value, default: { '@path': '$.properties.price' } }
```
brennan, line 53: Wondering if grabbing `'$.properties.price'` here will be an issue given that arrifying the event's `properties` object mutates the underlying object. Here, IIUC, `price` will be in `properties[i].price` rather than in `properties.price`. See: https://sourcegraph.segment.com/github.com/segmentio/action-destinations/-/blob/packages/core/src/arrify.ts#L13

kdaswani: I just tested this in staging with the following event: 

```
{
	"channel": "server",
	"context": {
		"ip": "0.0.0.0",
		"library": {
			"name": "unknown",
			"version": "unknown"
		},
		"traits": {
			"email": "kiara@test.com"
		}
	},
	"event": "Product Added",
	"integrations": {},
	"messageId": "api-21c3e985j2DXsgWAqxt2rs977Dn",
	"projectId": "fmPPhKvVqei9wG2C5ps4ci",
	"properties": {
		"category": "toys",
		"currency": "USD",
		"name": "Legos",
		"price": 19,
		"product_id": "507f1f77439011",
		"quantity": 1
	},
	"receivedAt": "2021-11-29T22:57:19.057Z",
	"timestamp": "2021-11-29T22:57:19.057Z",
	"type": "track",
	"userId": "123456",
	"version": 2
}
```

Looks like `properties.price` was picked up as the `value` and the test succeeds:
Screen Shot 2021-11-30 at 10 31 12 AM

@nick-Ag in staging I'm still seeing the wrong mapping in the test `contents` object but in your feature branch, it is working.

kdaswani: Result from the above test event in Nick's `bump-actions-core` feature branch:
Screen Shot 2021-11-30 at 10 34 17 AM

brennan (approved):
Given all unit tests pass, and the extent of manual testing, this PR looks good to me. I also just pushed a commit to remove `.only` from some of the Facebook CAPI unit tests.

PR #123, opened 2021-09-01
[BE-69] automatically capture request snapshots

jinapark202:
JIRA: https://segment.atlassian.net/browse/BE-69

This PR adds a CLI command to automatically generate and run snapshot tests for a specific destination, allowing developers to see how their changes affect requests.

CLI command:
`./bin/run generate:snapshots` or `./bin/run generate:snapshots amplitude`

This command generates a `snapshot.test.ts` file within the `__tests__` folder of the specified destination. This file disables external network requests, spins up a nock server, and generates test data for each action within the destination. It also creates two tests per action - one for all required fields and another for all fields (required and optional). 

The command also generates a `__snapshots__/snapshot.test.ts.snap` file within the same `__tests__` folder, which contains all of the snapshots for the tests mentioned above.


packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,72 @@
...
+beforeAll(() => {
+  // Disable external network requests
+  nock.disableNetConnect()
+  // But allow localhost connections so we can test local routes and mock servers.
+  nock.enableNetConnect('127.0.0.1')
+
+  if (!nock.isActive()) {
+    nock.activate()
+  }
+})
```
gpsamson, line 18: I think this could be removed if we pass the `--config <path-to>/package.json` when [calling `jest` programmatically](https://github.com/segmentio/action-destinations/pull/123/files#diff-1c797c14fe4b26d781a69596bca449f3a309d8b8219a538f98b759c6fcf58fcaR74). If we're sharing the [config](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/package.json#L58-L60) then the nock commands in `setup-after-env.ts` should run.

packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,72 @@
...
+        event,
+        settings: settingsData as Settings,
+        mapping: event.properties,
+        useDefaultMappings: false
+      })
+
+      const requestBody = responses[0].request.body
+      if (requestBody) {
+        expect(requestBody.toString()).toMatchSnapshot()
+      }
```
gpsamson, line 44: Would it be cleaner if we move most of this testing logic to the [`TestDestination` class](https://github.com/segmentio/action-destinations/blob/main/packages/core/src/create-test-integration.ts#L35)?

We could introduce a `snapshotAction` or `fuzzAction` as part of the class that takes care of nocking, 
generating test data, calling `testAction`, and returning responses?

gpsamson: This would greatly simplify tests. With the above, they would look something like?

```
it('snapshots search' => {
    const res = testDestination.snapshotAction('search')
    const requestBody = responses[0].request.body
      if (requestBody) {
        expect(requestBody.toString()).toMatchSnapshot()
      }
})
```

packages/destination-actions/src/destinations/twilio/index.ts
```diff
@@ -38,7 +38,7 @@ const destination: DestinationDefinition<Settings> = {
     }
   },
   actions: {
-    sendSMS
+    sendSms
```
gpsamson, line 41: @mshwery @dlasky Can you guys confirm whether or not this casing change would break things? iiuc, the current key `sendSMS` is saved in the database and is used to reference the action. Changing the action key to `sendSms` would end up breaking things.

packages/destination-actions/src/lib/test-data.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  let fields = Object.keys(action.fields)
+
+  if (flag === 'required') {
+    fields = fields.filter((name) => action.fields[name].required || name.includes('id'))
+  }
+
+  for (const name of fields) {
+    const field = action.fields[name]
```
gpsamson, line 49:
```suggestion
  if (flag === 'required') {
    fields = fields.filter(() => action.fields[name].required || name.includes('id'))
  }

  for (const [name, field] of Object.entries(action.fields)) {
   if (flag === 'required' && (!field.required || !name.includes('id'))) {
     continue
   }
```

[`Object.entries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) is a helpful tool for iterating over objects when you need both the key and value

packages/destination-actions/src/lib/test-data.ts
```diff
@@ -0,0 +1,75 @@
+import { ActionDefinition, DestinationDefinition } from '@segment/actions-core'
+
+const testData: { [key: string]: any } = {
```
gpsamson, line 3: Wonder if there's any benefit in eventually testing multiple variations of randomized test data? As mentioned in the RFC, this would require that inputs are also saved alongside snapshots so would probably be a follow up.

packages/destination-actions/src/lib/test-data.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  if (flag === 'required') {
+    fields = fields.filter((name) => action.fields[name].required || name.includes('id'))
+  }
+
+  for (const name of fields) {
+    const field = action.fields[name]
+    const { properties, type, format, multiple } = field
+
+    if (properties) {
```
gpsamson, line 52: Since `properties` implements the same interface ([`InputField`](https://github.com/segmentio/action-destinations/blob/8a565982805522fd1f26d5507e7f8e5c97771317/packages/core/src/destination-kit/types.ts#L94)) as `fields`, could we use recursion here to keep code DRY?

packages/cli/src/commands/generate/snapshots.ts
```diff
@@ -0,0 +1,87 @@
...
+        },
+        true
+      )
+      this.spinner.succeed(`Created snapshot tests for ${destinationName} destination`)
+    } catch (err) {
+      this.spinner.fail(`Snapshot test creation failed: ${chalk.red(err.message)}`)
+      this.exit()
+    }
+
+    exec(`jest --testPathPattern=${targetPath} --updateSnapshot`, (error, stdout, stderr) => {
```
gpsamson, line 74: Can consider [`execSync`](https://nodejs.org/api/child_process.html#child_process_child_process_execsync_command_options) to avoid callbacks.

packages/cli/src/commands/generate/snapshots.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (error) {
+        console.log('error', error)
+        return
+      }
+
+      if (stderr) {
+        console.log(stderr)
+        return
```
gpsamson, line 82: I'm not sure if we should return early here. _I think_ there might be a case where `stderr` and `stdout` both contain data, per [the `exec` docs](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback).

packages/cli/src/commands/generate/snapshots.ts
```diff
@@ -0,0 +1,87 @@
...
+      )
+      this.spinner.succeed(`Created snapshot tests for ${destinationName} destination`)
+    } catch (err) {
+      this.spinner.fail(`Snapshot test creation failed: ${chalk.red(err.message)}`)
+      this.exit()
+    }
+
+    exec(`jest --testPathPattern=${targetPath} --updateSnapshot`, (error, stdout, stderr) => {
+      if (error) {
+        console.log('error', error)
```
gpsamson, line 76: What do you think about treating the error as a CLI error? In other words, calling `this.spinner.fail` because there was an issue calling `exec`.

packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      const responses = await testDestination.testAction(actionSlug, {
+        event,
+        settings: settingsData as Settings,
+        mapping: event.properties,
+        useDefaultMappings: false
+      })
+
+      const requestBody = responses[0].request.body
+      if (requestBody) {
+        expect(requestBody.toString()).toMatchSnapshot()
```
gpsamson, line 68: Should we also be diffing the URL and headers?

gpsamson: I wonder if we can pretty-print snapshots so they're easily diffed? For example, it's hard to catch the difference in the first snapshot versus the second:

1.
```
- "[{\\"group_properties\\":{\\"some-trait-key\\":\\"some-trait-value\\"},\\"group_value\\":\\"some-value\\",\\"group_type\\":\\"some-type\\",\\"library\\":\\"segment\\"}]",
+ "[{\\"group_properties\\":{\\"some-trait-key\\":\\"some-trait-value\\"},\\"group_value\\":\\"some-value\\",\\"group_type\\":\\"some-type\\",\\"library\\":\\"twilio\\"}]",
```

2.
```
    "group_type": "some-type",
-  "library": "segment"
        }]
    "group_type": "some-type",
+  "library": "twilio"
        }]
```

gpsamson (commented):
Awwyiss! Nice work!

I'm wondering if these snapshot tests should also be a part of the  tests generated as part of the `generate:action` command?

packages/destination-actions/src/destinations/twilio/index.ts
dlasky: i believe it would create a 2nd action cause we don't currently delete or 'sync' anything. is that right @mshwery

packages/destination-actions/src/lib/test-data.ts
dlasky: you wouldn't be able to randomize with snaps per se unless as you said it baked first, but fuzzing would have saved me on a SEV so i'm in favor of that generally even if its only for selectable (boolean, select) values and the rest are some preset test value.

packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
gpsamson: Ouuu, snapshot serializers might help solve this! https://medium.com/@luisvieira_gmr/jest-snapshot-serializers-6a96f5c362a1

packages/destination-actions/src/lib/test-data.ts
gpsamson: Hmm. This is a bit hacky, but you could rely on snapshots to store the inputs as well. When a snapshot is first created you can save an object like:
```
{
    "input": { "fuzzed": true },
    "output": { "url": "", "headers": {}, "body": {} }
}
```

Then you can hook into the `jest-snapshot` package to read back and parse the snapshot data:
```
import JestSnapshot from 'jest-snapshot'
const data = JestSnapshot.utils.getSnapshotData('../<path-to-snap>/snapshot', 'none')
// somehow parse data (type: Record<string, string>) for the input
```

packages/cli/src/commands/generate/snapshots.ts
```diff
@@ -0,0 +1,81 @@
...
+import chalk from 'chalk'
+import path from 'path'
+import ora from 'ora'
+import globby from 'globby'
+
+export default class GenerateSnapshots extends Command {
+  private spinner: ora.Ora = ora()
+  static description = `Generates TypeScript definitions for an integration.`
+
+  static examples = [`$ ./bin/run generate:snapshots`, `$ ./bin/run generate:snapshots slack`]
```
mshwery, line 14: Look at this implementation, I wonder why have this be a separate command at all? We could include it in the main `init` and `generate:action` commands?

packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
mshwery: Pretty sure you can omit the `toString`.

packages/destination-actions/src/destinations/twilio/index.ts
mshwery: These are case-sensitive. This would technically be a breaking change (well, if you remove the previous action)

mshwery: Let's remove this change from this PR to keep things isolated.

packages/core/src/create-test-integration.ts
```diff
@@ -64,6 +66,28 @@ class TestDestination<T> extends Destination<T> {
...
     return responses
   }
+
+  async snapshotAction(destination: DestinationDefinition<T>, actionSlug: string, isRequiredOnly: boolean) {
+    const action = destination.actions[actionSlug]
+    const [eventData, settingsData] = generateTestData(destination, action, isRequiredOnly)
+
+    nock(/.*/).persist().get(/.*/).reply(200)
+    nock(/.*/).persist().post(/.*/).reply(200)
+    nock(/.*/).persist().put(/.*/).reply(200)
```
mshwery, line 76: Two things:

1. Let's move this into the generated test file instead. 
2. You need to release these nocks after the tests since you are using `persist`

packages/core/src/create-test-integration.ts
```diff
@@ -64,6 +66,28 @@ class TestDestination<T> extends Destination<T> {
 
     return responses
   }
+
+  async snapshotAction(destination: DestinationDefinition<T>, actionSlug: string, isRequiredOnly: boolean) {
+    const action = destination.actions[actionSlug]
+    const [eventData, settingsData] = generateTestData(destination, action, isRequiredOnly)
```
mshwery, line 72: Seems like all this could be moved to the generated file.

packages/core/src/test-data.ts
```diff
@@ -0,0 +1,79 @@
...
+  boolean: true,
+  datetime: '2021-01-17',
+  integer: 1,
+  number: 5,
+  string: 'test1234',
+  text: 'test text',
+  object: {
+    testObject: 'data1234'
+  }
+}
```
mshwery, line 13: Thoughts on using https://www.npmjs.com/package/chance instead of hard-coded values?

mshwery: Though this probably works fine. If you introduce `chance` you'll have to store the data fixtures somewhere so we have deterministic tests in CI (otherwise the data would change every time you run the test).

packages/core/src/test-data.ts
```diff
@@ -0,0 +1,79 @@
...
+  const settingsData: any = {}
+
+  const authentication = destination.authentication
+  if (authentication) {
+    for (const settingKey in authentication.fields) {
+      const type = authentication.fields[settingKey].type
+      settingsData[settingKey] = testData[type]
+      if (settingKey === 'endpoint') {
+        settingsData[settingKey] = 'https://www.example.com'
+      }
```
mshwery, line 50: Can we rely on `format: 'uri'` instead? I don't think this covers all the url fields as-is.

packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
jinapark202: The reason I used `toString` is because `request.body` comes back as a `Buffer` type and the output becomes unreadable.

packages/core/src/create-test-integration.ts
jinapark202: The `setup-after-env.ts` file currently removes all mocks (including persistent ones) before each unit test with the command: `nock.cleanAll()`. Should they be run after the tests instead (or at least after the last test is run)?

packages/cli/templates/actions/snapshot/__tests__/snapshot.test.ts
mshwery: Ah, yeah request bodies need to be parsed. You should be able to do `await request.text()` or `await request.json()` here.

packages/core/src/test-data.ts
jinapark202: With `Chance`, can I use fixed seeds (like the destination name) to get deterministic tests rather than storing data fixtures? Or do you see any flaws with this method? https://chancejs.com/usage/seed.html

packages/cli/src/commands/generate/snapshots.ts
jinapark202: Do you think the snapshot tests should be at the destination level or the destination action level (i.e. should we include it to the `init` command or the `generate:action` command? Or were you thinking we should include it for both commands?)? Also, do we just want to create the test file and not actually run it on command?

packages/core/src/test-data.ts
gpsamson: Neat! I think seeds should work here. You could potentially build them with `Chance(destination.name, action.name)`.

packages/cli/src/commands/generate/snapshots.ts
mshwery: Both – so on `init` it would scaffold the destination-level snapshot tests, and on `generate:action` it would be for the action.

The snapshots would of course require getting updated any time you change the schema for the destination or actions.

packages/core/src/create-test-integration.ts
mshwery: >The `setup-after-env.ts` file currently removes all mocks (including persistent ones) before each unit test

Ah missed that. 

I think we do want an "afterAll" hook that cleans up any remaining nocks.

packages/core/src/test-data.ts
mshwery: Also totally happy to start with the basics (what you already have) and we can iterate to a more sophisticated solution (only if needed).

package.json
```diff
@@ -74,5 +75,8 @@
         "isolatedModules": true
       }
     }
+  },
+  "dependencies": {
+    "chance": "^1.1.8"
```
gpsamson, line 80:
```suggestion
```

packages/destination-actions/src/lib/test-data.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (isRequiredOnly && !(field.required || name.includes('id'))) {
+      continue
+    }
+
+    const { properties } = field
+
+    if (properties) {
+      let subData: any = {}
+      const propertyFields = Object.keys(properties)
+      if (isRequiredOnly) propertyFields.filter((name) => properties[name].required)
```
gpsamson, line 78: `filter()` returns a new array and doesn't modify the array in place
```suggestion
      if (isRequiredOnly) {
          propertyFields = propertyFields.filter((name) => properties[name].required)
      }
```

package.json
gpsamson, line 80: We should have this as a `devDependencies` since it's exclusively used for testing.

mshwery:
If this is used by the CLI it needs to be a dependency not a dev dep (think
of the CLI as an installable production thing)

– Matt

gpsamson (approved):
One last request is to add docs!

PR #127, opened 2021-09-02
Add `toEmail` mapping to Twilio for sending test SMS

vadimdemedes:
This is necessary to send test SMS to a custom phone number specified in the UI by the customer


packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
```diff
@@ -74,6 +74,11 @@ const action: ActionDefinition<Settings, Payload> = {
       required: true,
       default: { '@path': '$.userId' }
     },
+    toNumber: {
+      label: 'To Number',
```
mshwery, line 78: Should this be labeled like `Test Recipient Phone Number`?

PR #130, opened 2021-09-02
Add package-lock.json to gitignore

gpsamson:
This pull request adds `package-lock.json` to the git ignore list.


.gitignore
```diff
@@ -4,3 +4,4 @@ dist
 *.log
 *.tsbuildinfo
 .eslintcache
+package-lock.json
```
mshwery, line 7: Let's also add an `.npmrc` file that has:
```
package-lock=false
```

mshwery (approved):

daltonscharff (approved):
LGTM!

.gitignore
gpsamson: pro move 👍🏽

PR #138, opened 2021-09-07
Pre-commit hook to catch secrets

nick-Ag:
To prevent committing certain secrets this PR adds a pre commit hook which uses trufflehog to check for secrets that a user might have added.

Screen Shot 2021-09-07 at 11 36 46 AM


.husky/pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# fetch changed files on this commit
+git diff --name-only --cached > .husky/.included.txt
```
nick-Ag, line 5: What would be a better way to do this?

.husky/trufflehog-rules.json
```diff
@@ -0,0 +1,42 @@
+{
```
nick-Ag, line 1: truffle uses this regex file to check for secrets

package.json
```diff
@@ -22,7 +22,8 @@
     "subscriptions": "yarn workspace @segment/destination-subscriptions",
     "test": "lerna run test --stream",
     "typecheck": "lerna run typecheck --stream",
-    "alpha": "lerna publish prerelease --pre-dist-tag next --allow-branch '**' --no-git-tag-version"
+    "alpha": "lerna publish prerelease --pre-dist-tag next --allow-branch '**' --no-git-tag-version",
+    "prepare": "husky install"
```
nick-Ag, line 26: I'm using husky to manage the precommit hook because it looks like that's what is already being used in the `.git` folder on my machine. The folder is untracked however and I have some questions about where exactly I should add the script?

.husky/pre-commit
```diff
@@ -0,0 +1,10 @@
...
+
+# call trufflehog on those files only
+trufflehog .git --rules .husky/trufflehog-rules.json --include_paths .husky/.included.txt --regex --entropy=NO #entropy checks throw too many false positives
```
lcampos, line 8: Have you looked at ways in which to kick off installing trufflehog when it's not available in the machine ?

.husky/pre-commit
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
marinhero: Also add this for safer execution of the script

`set -euo pipefail`

.husky/pre-commit
marinhero: Add checks for python, pip, and trufflehog. If they are not available install them.

.husky/trufflehog-rules.json
marinhero: Add a comment with the source of these regular expressions.

marinhero (changes requested):
Thank you for taking the time to go thru this over a Zoom call. I left the comments as a reminder of the things we talked about. Let me know when I can take another look.

nick-Ag:
Trufflehog was unable to check files which were uncommitted and so wasn't well suited as a pre-commit hook. Fortunately gitleaks has this option with the `--unstaged` flag. It also allows for better custom rule extensions with the `--additional-config` flag. This commit switches from trufflehog to gitleaks for that reason and has output that looks like:
Screen Shot 2021-09-07 at 6 52 10 PM

.husky/pre-commit
```diff
@@ -0,0 +1,16 @@
...
+fi
+
+if ! command -v gitleaks -h &> /dev/null
+then
+    brew install gitleaks
+fi
+
+BRANCH=$(git branch --show-current)
+
+gitleaks --repo-url=.git -v --additional-config=".husky/gitleaks-rules.toml" --branch=$BRANCH --unstaged || true
```
nick-Ag, line 16: Do we want the commit to be canceled if a secret is detected?

Whamo12: That would be a safe option.  If a secret is detected the commit should fail but it'll still be in staging.

package.json
gpsamson: I don't quite follow here. Are you asking if we should use the default `.husky` directory or use some other directory?

mshwery: I would add it to lint-staged.

.husky/gitleaks-rules.toml
```diff
@@ -0,0 +1,4 @@
+[[rules]]
+  description = "Facebook System User Access Token"
```
marinhero, line 2: You should include a section on the README that reminds other developers to include new regexes whenever they introduce a new token.

.husky/pre-commit
```diff
@@ -0,0 +1,21 @@
+`#!/usr/bin/env bash`
```
marinhero, line 1: Shouldn't this be 
```suggestion
#!/usr/bin/env bash
```
?

.husky/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+set -exo pipefail
+
+if ! command -v gitleaks -h &> /dev/null
+then
+    if -f /.dockerenv
+    then
+        docker pull zricethezav/gitleaks:latest
+        docker run -v ${PWD}:/. zricethezav/gitleaks:latest --path="/."
+    else
+        if ! command -v brew -v &> /dev/null
```
marinhero, line 12: Run this only on a mac
```suggestion
        if [ "$(uname)" == "Darwin" ]; then
            if ! command -v brew -v &> /dev/null
            then
                /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
            fi
        brew install gitleaks
        else
             echo "Unable to automatically install gitleaks. Install it yourself"
        fi
```

marinhero: Sorry indentation is weird!

.husky/pre-commit
marinhero: I think this is ok 👍

gpsamson:
> @daltonscharff and I have been testing this out locally for a couple weeks now and the main functionality is working as expected. Users would be able to install the husky and lint-staged dependencies with a `yarn install` and if they don't have gitleaks installed the husky pre-commit hook will install it. @gpsamson Do you think there are additional first-time setup steps this script should do?

Chatted about this out of band in Slack. In short, I think all looks fine but I would caution against the assumption that every developer will be on MacOS and have `brew` installed. This is usually the case for Segmenters (majority of devs in this repo) but we should still catch a failure and notify the user.

I am not sure if we want to fail the commit if `gitleaks` can't be installed and ran. This might be a question to pose in a Github Discussion post and share in our actions-collab Slack channel.

mericsson:
> > @daltonscharff and I have been testing this out locally for a couple weeks now and the main functionality is working as expected. Users would be able to install the husky and lint-staged dependencies with a `yarn install` and if they don't have gitleaks installed the husky pre-commit hook will install it. @gpsamson Do you think there are additional first-time setup steps this script should do?
> 
> Chatted about this out of band in Slack. I think all looks fine but I would caution against the assumption that every developer will be on MacOS and have `brew` installed. This is usually the case for Segmenters (majority of devs in this repo) but we should still catch a failure and notify the user.
> 
> I am not sure if we want to fail the commit if `gitleaks` can't be installed and ran. This might be a question to pose in a Github Discussion post and share in our actions-collab Slack channel.

Perhaps we update to only run if `OS == MacOS and brew is installed`? Then we can potentially add a follow-up to consider platform independent implementation?

nick-Ag:
> > > > @daltonscharff and I have been testing this out locally for a couple weeks now and the main functionality is working as expected. Users would be able to install the husky and lint-staged dependencies with a `yarn install` and if they don't have gitleaks installed the husky pre-commit hook will install it. @gpsamson Do you think there are additional first-time setup steps this script should do?
> > > 
> > > 
> > > Chatted about this out of band in Slack. I think all looks fine but I would caution against the assumption that every developer will be on MacOS and have `brew` installed. This is usually the case for Segmenters (majority of devs in this repo) but we should still catch a failure and notify the user.
> > > I am not sure if we want to fail the commit if `gitleaks` can't be installed and ran. This might be a question to pose in a Github Discussion post and share in our actions-collab Slack channel.
> > 
> > 
> > Perhaps we update to only run if `OS == MacOS and brew is installed`? Then we can potentially add a follow-up to consider platform independent implementation?
> 
> Ahh :nice: -- this works for me. It'll solve for the common case :)

Implemented this suggestion of running only on MacOS. I successfully tested on an Ubuntu machine, my local (Segment) machine, and my personal laptop (MacOS) to check if open-source users can install fine. This is ready to merge in after a review.

PR #142, opened 2021-09-08
Customer.io destination action updates

mike-engel:
Howdy y'all!

This updates the customer.io destination action with a few improvements:

- Adds an option to set the data center (EU/US) to use for API requests
- Allows `last_used` and `created_at` to be transformed to a UNIX timestamp or not for `createUpdateDevice` and `createUpdatePerson`, respectively
- Adds some tests for `createUpdateDevice` and `createUpdatePerson`
- Removes `triggerCampaign`, since that doesn't make much sense to do from Segment

The data center options do include unicode emoji characters, so if that's not allowed, please let me know and I'll remove those.

If there are other conventions I'm not adhering to, please let me know as well. Thanks!


packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
```diff
@@ -50,16 +50,16 @@ const action: ActionDefinition<Settings, Payload> = {
...
       name: payload.name,
       type: payload.type,
       data: payload.data
     }
-    let url = `https://track.customer.io/api/v1/customers/${payload.id}/events`
+    let url = `${settings.accountRegionEndpoint}/api/v1/customers/${payload.id}/events`
 
     if (payload.id === undefined) {
-      url = 'https://track.customer.io/api/v1/events'
+      url = `${settings.accountRegionEndpoint}/api/v1/events`
```
mshwery, line 62: I don't think we enforce/validate that this value is one of `choices` – because we aren't sure if `choices` always represents the full set of accepted values, or just a subset of them.

You may want to add an extra check here, or manually coerce the value into a valid one (you could check for EU and if it's not that, fall back to US or something).

Open to other thoughts around how we should handle this.

packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
```diff
@@ -50,16 +50,28 @@ const action: ActionDefinition<Settings, Payload> = {
       default: {
         '@path': '$.traits'
       }
+    },
+    convert_timestamp: {
+      label: 'Convert timestamps',
+      description: 'Convert `created_at` to a Unix timestamp (seconds since Epoch).',
```
mshwery, line 56: Curious: Do some customers not convert timestamps? I'm wondering what the use case is for that (excuse my unfamiliarity with the customer.io api)

mshwery (approved):
🔥 looks solid

packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
mike-engel: In this case, `choices`, as coded, is the complete set. Is it possible to choose something else in the segment UI? Or will the default not be set for users who have already configured the action?

packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
mike-engel: I would say that 99% of the time, it's beneficial for users to have date/times as UNIX timestamps in Customer.io since that's all that we parse. But, of course, there are some users who prefer the format they originally provided. This setting is pretty much an escape hatch.

If you think it does more harm than good to have it in, we can discuss it with Brendan to get more use cases or to decide to drop it from these actions

packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
mshwery: It's not possible to use the UI to input any other values than these choices.

packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
mshwery: That makes sense! Thanks for sharing the context.

packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
gpsamson: > Or will the default not be set for users who have already configured the action?

At the time of writing this, this is correct. We'll have to backfill the existing destination instance with the default account region endpoint value.

gpsamson (commented):
Hey, Mike! Thanks for opening this PR 🔥. Other than the region endpoint comment this looks great.

packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
mike-engel: Thanks @gpsamson! I added a code-based fallback for `accountRegionEndpoint` as well as the new timestamp conversion settings.

As a bit of feedback, it would be great to have a way to backfill new settings with the default value for users who already have the destination action configured 😄

mshwery: > As a bit of feedback, it would be great to have a way to backfill new settings with the default value for users who already have the destination action configured 😄

@mike-engel we do have a way to do this. we can target specific workspaces/destinations. It's not a tool available to builders yet (internal only) and we havent exposed any customer control over it (in the UI, etc), but it's possible! Let us know if you want to run us to run it for specific workspaces / destinations.

gpsamson (approved):
We can get this merged in and deployed on Thursday. We have a change freeze on our end that is blocking us from deploying any destination updates until then. Does this sound okay @mike-engel? 

I'll also go ahead and backfill existing instances with the default `accountRegionEndpoint` setting value after this is deployed.

mike-engel:
@gpsamson That sounds good to us. Since I added the util to handle an undefined `accountRegionEndpoint`, it's not as important to backfill it, but it would be appreciated if it's easy to do

gpsamson:
@mike-engel I did some manual end-to-end testing in our staging environment and caught one issue related to the timestamp changes. As you can see in the request below, `created_at` gets sent to Customer.io as `'X'` instead of a Unix timestamp when the Convert timestamps field is set to true.

This is due to `lib/dayjs.ts` not importing the [AdvanceFormat plugin](https://day.js.org/docs/en/plugin/advanced-format). Would you mind updating your fork to import this plugin? That should resolve the issue. On the bright side, regional endpoints are working as expected. Thanks!

image

mike-engel:
@gpsamson Thanks for the update! Glad to know the data center config is working as expected 😄 I just rebased this branch with the latest from `main` on this repo. If you'd like me to do something else, let me know!

gpsamson:
> @gpsamson Thanks for the update! Glad to know the data center config is working as expected 😄 I just rebased this branch with the latest from `main` on this repo. If you'd like me to do something else, let me know!

Hey Mike, thanks! Do you mind updating your branch so that the dayjs [AdvanceFormat plugin](https://day.js.org/docs/en/plugin/advanced-format) is being imported within the `lib/dayjs.ts` file. The `main` branch currently does not do it and this is causing Unix timestamp to be formatted incorrectly (as seen in the screenshot above).

mike-engel:
@gpsamson ah, sorry for the misunderstanding! I thought it was already in the `main` branch. I've added it to `dayjs.ts` 👍

gpsamson:
> @gpsamson ah, sorry for the misunderstanding! I thought it was already in the `main` branch. I've added it to `dayjs.ts` 👍

Perfect! I'm going to get this merged in as timestamps are passing correctly. Thanks

image

mike-engel:
@gpsamson I don't think it will, but just in case, I added a new commit which uses `unix` over `format('X')` for `created_at` and `last_used`

mshwery:
FWIW I ran into this before and used `.valueOf()` instead of the advanced syntax, which also produces the unix value: https://day.js.org/docs/en/display/unix-timestamp-milliseconds

PR #143, opened 2021-09-08
add internal handling of batches of events

mshwery:
This PR spikes on our ability to support _batching_ in Actions.

The basic premise:
- Centrifuge produces a batch of events (based on a batching key that's a hash of settings + auth + subscription) and sends it to our `integrations` cluster
- `integrations` will see the `batch` method and invoke it (will hook this up in a later PR)
- the `batch` method will proxy events + subscriptions by calling the destination's `.onBatch` method instead of the `.onEvent` method
- `.onBatch` will assume that the settings, auth data, and subscription(s) are the same for the entire batch of events
- each subscription will be parsed and evaluated only once for the entire batch
- each event will get transformed by the `subscription.mapping` and validated (this is optimized with a cached validator)
- the entire batch of transformed payloads will be passed to a `performBatch` method if defined on the action. Otherwise it will throw a 501

### DIFF
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -112,7 +112,8 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
 
     // Validate the resolved payload against the schema
     if (this.schema) {
-      validateSchema(payload, this.schema, `${this.destinationName}:${this.definition.title}`)
+      const schemaKey = `${this.destinationName}:${this.definition.title}`
+      validateSchema(payload, this.schema, { schemaKey })
       results.push({ output: 'Payload validated' })
     }
 
@@ -133,13 +134,22 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
   }
 
   async executeBatch(bundle: ExecuteBundle<Settings, InputData[]>): Promise<void> {
-    const payloads = transformBatch(bundle.mapping, bundle.data) as Payload[]
+    let payloads = transformBatch(bundle.mapping, bundle.data) as Payload[]
 
     // Validate the resolved payloads against the schema
     if (this.schema) {
+      const schema = this.schema
       const schemaKey = `${this.destinationName}:${this.definition.title}`
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      payloads.map((payload) => validateSchema(payload, this.schema!, schemaKey))
+      payloads = payloads.filter((payload) =>
+        validateSchema(payload, schema, {
+          schemaKey,
+          throwIfInvalid: false
+        })
+      )
+    }
+
+    if (payloads.length === 0) {
+      return
     }
 
     if (this.definition.performBatch) {
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -1,7 +1,7 @@
 import { EventEmitter } from 'events'
 import createRequestClient from '../create-request-client'
 import { JSONLikeObject, JSONObject } from '../json-object'
-import { transform } from '../mapping-kit'
+import { InputData, transform, transformBatch } from '../mapping-kit'
 import { fieldsToJsonSchema } from './fields-to-jsonschema'
 import { Response } from '../fetch'
 import type { ModifiedResponse } from '../types'
@@ -58,12 +58,22 @@ export interface ActionDefinition<Settings, Payload = any> extends BaseActionDef
 
   /** The operation to perform when this action is triggered */
   perform: RequestFn<Settings, Payload>
+
+  /** The operation to perform when this action is triggered for a batch of events */
+  performBatch?: RequestFn<Settings, Payload[]>
 }
 
 interface ExecuteDynamicFieldInput<Settings, Payload> {
   settings: Settings
   payload: Payload
   page?: string
+}
+
+interface ExecuteBundle<T = unknown, Data = unknown> {
+  data: Data
+  settings: T
+  mapping: JSONObject
+  auth: AuthTokens | undefined
 }
 
 /**
@@ -74,12 +84,12 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
   readonly definition: ActionDefinition<Settings, Payload>
   readonly destinationName: string
   readonly schema?: JSONSchema4
-  private extendRequest: RequestExtension<Settings, Payload> | undefined
+  private extendRequest: RequestExtension<Settings> | undefined
 
   constructor(
     destinationName: string,
     definition: ActionDefinition<Settings, Payload>,
-    extendRequest?: RequestExtension<Settings, Payload>
+    extendRequest?: RequestExtension<Settings>
   ) {
     super()
     this.definition = definition
@@ -92,12 +102,7 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
     }
   }
 
-  async execute(bundle: {
-    data: unknown
-    settings: Settings
-    mapping: JSONObject
-    auth: AuthTokens | undefined
-  }): Promise<Result[]> {
+  async execute(bundle: ExecuteBundle<Settings, InputData | undefined>): Promise<Result[]> {
     // TODO cleanup results... not sure it's even used
     const results: Result[] = []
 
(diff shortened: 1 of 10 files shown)


packages/core/src/retry.ts
```diff
@@ -0,0 +1,30 @@
+interface RetryOptions {
+  retries?: number
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onFailedAttempt?: (error: any, attemptCount: number) => PromiseLike<void> | void
+}
+
+export async function retry<T>(
```
mshwery, line 7: @lcampos I extracted this logic out just to simplify the handlers a bit. Originally I was going to end up duplicating this functionality, but ended up scrapping that by moving things around.

packages/core/src/mapping-kit/index.ts
```diff
@@ -132,6 +134,31 @@ export function transform(mapping: JSONLikeObject, data: unknown = {}): JSONObje
   return cleaned as JSONObject
 }
 
+/**
+ * Validates and transforms a mapping across multiple payloads
+ * @param mapping - the directives and raw values
+ * @param data - the array input data to apply to directives
+ */
+export function transformBatch(mapping: JSONLikeObject, data: Array<InputData> | undefined = []): JSONObject[] {
```
mshwery, line 142: This is optimized to transform multiple events with a given mapping.

We only need to validate the mapping once since it does not change. We also only need to removeUndefined values once since that works on arrays or objects.

packages/core/src/mapping-kit/index.ts
```diff
@@ -132,6 +134,31 @@ export function transform(mapping: JSONLikeObject, data: unknown = {}): JSONObje
...
+      const resolved = resolve(cloned, d as JSONObject)
+
+      return resolved
+      // Cast because we know there are no `undefined` values after `removeUndefined`
+    })
+  ) as JSONObject[]
+}
+
 function cloneJson<T>(obj: T): T {
   return JSON.parse(JSON.stringify(obj))
```
mshwery, line 163: For large batches we may want to consider something faster...

packages/core/src/destination-kit/types.ts
```diff
@@ -137,4 +137,4 @@ export type Directive = IfDirective | TemplateDirective | PathDirective
  * A function to configure a request client instance with options
  * that will be applied to every request made by that instance
  */
-export type RequestExtension<Settings, Payload = unknown> = (data: ExecuteInput<Settings, Payload>) => RequestOptions
+export type RequestExtension<Settings, Payload = undefined> = (data: ExecuteInput<Settings, Payload>) => RequestOptions
```
mshwery, line 140: So far, we don't expect `extendRequest` to accept or have an action's payload available. This is a bit of a hack until we streamline these types to make them easier to work with.

packages/core/src/destination-kit/action.ts
```diff
@@ -127,6 +132,42 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
...
+        rawData: bundle.data[i],
+        rawMapping: bundle.mapping,
+        settings: bundle.settings,
+        payload,
+        auth: bundle.auth
+      }
+      return this.performRequest(this.definition.perform, data)
+    })
+
+    await Promise.all(promises)
```
mshwery, line 168: This is a fallback in case `performBatch` isn't defined. Basically process the batch of events as if it wasn't a batched request.

packages/core/src/destination-kit/action.ts
```diff
@@ -144,16 +185,17 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
...
+    data: ExecuteInput<Settings, T>
   ): Promise<unknown> {
     const requestClient = this.createRequestClient(data)
     const response = await requestFn(requestClient, data)
     return this.parseResponse(response)
   }
 
-  private createRequestClient(data: ExecuteInput<Settings, Payload>): RequestClient {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private createRequestClient(data: ExecuteInput<Settings, any>): RequestClient {
```
mshwery, line 198: I had to do this to make `this.extendRequest` happy.

packages/core/src/mapping-kit/index.ts
```diff
@@ -109,16 +109,18 @@ function resolve(mapping: JSONLike, payload: JSONObject): JSONLike {
   return mapping
 }
 
+export type InputData = { [key: string]: unknown }
```
lcampos, line 112: This is so much better, thanks

packages/core/src/retry.ts
lcampos: Looks good. Since we're introducing batching, is if this is the right moment to remove retries in actions core and delegate the retry to centrifuge after we refresh tokens (when possible)  ?

lcampos (approved):

packages/core/src/retry.ts
mshwery: Definitely open to it @lcampos !

lcampos: FYI: one thing I learned this past week is that having this retry code allows `OAuth` actions to be used by event tester once the tokens have expired.

packages/core/src/retry.ts
```diff
@@ -0,0 +1,30 @@
...
+  input: (attemptCount: number) => PromiseLike<T> | T,
+  options?: RetryOptions
+): Promise<T> {
+  const retries = options?.retries ?? 2
```
gpsamson, line 11: Move to a constant so it's clear?

```suggestion
  const retries = options?.retries ?? DEFAULT_RETRY_ATTEMPTS
```

packages/core/src/destination-kit/index.ts
```diff
@@ -259,70 +271,81 @@ export class Destination<Settings = JSONObject> {
...
       }
 
-      const isSubscribed = validate(parsedSubscription, event)
-      if (!isSubscribed) {
-        state = 'skipped'
+      events = Array.isArray(events) ? events : [events]
+      const subscribedEvents = events.filter((event) => validate(parsedSubscription, event))
+
+      if (subscribedEvents.length === 1) {
+        return await this.executeAction(actionSlug, { ...input, event: subscribedEvents[0] })
```
gpsamson, line 341: There's an interesting edge case where a batch of events can end up only having one subscribed event, ultimately triggering `execute` instead of `executeBatch`.

If `execute` has different business logic than `executeBatch` you could end up with random, single events that cause different results in the downstream tool. An example of a classic destination that does this is the recent Adobe Analytics batch updates. The batch API supports newer fields while the single event API does not.

gpsamson (commented):
LGTM! Non-blocking comments

packages/core/src/destination-kit/index.ts
gpsamson: I notice you mentioned further up in this PR that we will fall back to `execute` when `executeBatch` is undefined. 

This makes sense when there is no previous `executeBatch` method because there is no risk of regression -- the issue would be upstream since we would be incorrectly sending batches to a destination that has never supported it.

However, this decision seems risky if there was previously an `executeBatch` method because the switch to `execute` could introduce a regression for the same reasons in my first comment (a batch of events is handled with different business logic).

It's a tricky edge case and maybe I'm overthinking it. Maybe it's okay to fallback to the `execute` in this new model?

tysonmote: I don't think you're over-thinking it. I agree that if the request is submitted as a batch or if the customer is requesting that we use batches, we should use batched codepath even if the batch has only one event.

mshwery: Agreed! Thanks, this was initially a stop-gap to help us test batching e2e without having a batch destination. Now that we are doing the webhook w/ batch support, this can go away.

mshwery: Sidenote: the thought of an action's `performBatch` having different fields than the regular `perform` sounds like a case where they should be separate actions entirely.

mshwery: Updated to always use the batch method!

packages/core/src/destination-kit/action.ts
mshwery: No longer doing this. We'll throw a 501 not implemented if that's the case.

gpsamson (approved):

PR #145, opened 2021-09-08
Surface HTTPError details when requests fail via serve command

lcampos:
Update `serve` command to return `HTTPError` details when a request fails.

Before
```
$ curl --request POST --url http://localhost:3000/postConversion ....
Bad Request
```

Now
```
$ curl --request POST --url http://localhost:3000/postConversion ....
{"status":"INVALID_REQUEST","error_statuses":[{"error_code":"INVALID_OAUTH_TOKEN","error_message":"The OAuth token is missing or invalid."}]}
```

### DIFF
--- a/packages/cli/src/lib/server.ts
+++ b/packages/cli/src/lib/server.ts
@@ -9,6 +9,8 @@ import asyncHandler from './async-handler'
 import getExchanges from './summarize-http'
 import ora from 'ora'
 import chalk from 'chalk'
+import { HTTPError } from '@segment/actions-core/request-client'
+import type { ModifiedResponse } from '@segment/actions-core/types'
 
 const app = express()
 app.use(express.json())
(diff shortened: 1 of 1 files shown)


packages/cli/src/lib/server.ts
```diff
@@ -67,45 +69,57 @@ server.on('error', (err: Error) => {
...
+            mapping: req.body.payload || {},
+            auth: req.body.auth || {}
+          })
+
+          const debug = await getExchanges(destination.responses)
+          spinner.succeed(chalk`${destination.name} action '${actionSlug}' completed`)
+          return res.status(200).json(debug)
+        } catch (err) {
+          spinner.fail()
+          let statusCode = err?.status ?? 500
```
lcampos, line 103: Lines 103 to 111 is the code change that adds HTTPError data.

packages/cli/src/lib/server.ts
```diff
@@ -67,45 +69,57 @@ server.on('error', (err: Error) => {
...
-
-      try {
-        const destination = new Destination(def as CloudDestinationDefinition)
-        const action = destination.actions[actionSlug]
-
-        if (!action) {
-          const msg = `${destination.name} action '${actionSlug}' is invalid or not found`
-          spinner.fail(chalk`${msg}`)
-          return res.status(400).send(msg)
+loadDestination(targetDirectory)
```
lcampos, line 72: Unexpected formatting, actual code change starts in line 103

packages/cli/src/lib/server.ts
```diff
@@ -67,45 +69,57 @@ server.on('error', (err: Error) => {
...
+          const debug = await getExchanges(destination.responses)
+          spinner.succeed(chalk`${destination.name} action '${actionSlug}' completed`)
+          return res.status(200).json(debug)
+        } catch (err) {
+          spinner.fail()
+          let statusCode = err?.status ?? 500
+          let msg = err?.message
+
+          if (err instanceof HTTPError) {
+            statusCode = err?.response?.status ?? 500
```
gpsamson, line 107: Should we be falling back to the existing value instead in case `err.status` had a value?

```suggestion
            statusCode = err?.response?.status ?? statusCode
```

gpsamson (approved):
Looks good. Left one small comment about retaining the original status code

jinapark202 (approved):
LGTM

packages/cli/src/lib/server.ts
```diff
@@ -67,45 +69,57 @@ server.on('error', (err: Error) => {
...
+
+    server.listen(port, () => {
+      logger.info(`Listening at http://localhost:${port} ->
+${Object.keys(def?.actions ?? {})
+  .map((action) => `  POST http://localhost:${port}/${action}`)
+  .join('\n')}`)
     })
-  )
-
-  server.listen(port, () => {
```
jinapark202, line 105: Minor but indentation looks a little weird here

lcampos: I checked again and this gets formatted by the pre-commit hook even after manually making changes to the format https://github.com/segmentio/action-destinations/blob/41b9caa497a3edfce2e71096dac86e9a91a77266/package.json#L49

packages/cli/src/lib/server.ts
mshwery: Seems like our prettier config is messed up if it's applying different formatting for you and me

mshwery (approved):

PR #147, opened 2021-09-08
Bundle browser destination actions with content sha to invalidate cache

juliofarah:
This PR kicks off the versioning story for browser action destinations.

image

```
mysql> select * from destination_definition_remote_plugins where destination_definition_id = '60fb01aec459242d3b6f20c1';
+---------------------------+--------------------------+-----------------------------------------------------------------------------------+------------------+
| destination_definition_id | name                     | url                                                                               | library_name     |
+---------------------------+--------------------------+-----------------------------------------------------------------------------------+------------------+
| 60fb01aec459242d3b6f20c1  | Braze Web Mode (Actions) | https://cdn.segment.build/next-integrations/actions/braze/5beb8ffa92c9ec709036.js | brazeDestination |
+---------------------------+--------------------------+-----------------------------------------------------------------------------------+------------------+
1 row in set (0.00 sec)
```

image


packages/cli/src/lib/web-bundles.ts
```diff
@@ -1,7 +1,24 @@
...
+export const DIST_DIR = 'packages/browser-destinations/dist/web/'
+
+export function webBundles(): { [destination: string]: string } {
+  const command = `ls ${DIST_DIR}`
+  const map: { [destination: string]: string } = {}
+  const destinations = execa.commandSync(command).stdout
+
+  destinations.split('\n').forEach((destination) => {
+    map[destination] = execa.commandSync(`ls ${DIST_DIR}/${destination}`).stdout.split('\n')[0]
+  })
```
mshwery, line 12: Could use globby here to fetch the directories / files.

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -85,7 +86,7 @@ export default class PushBrowserDestinations extends Command {
         // This MUST match the way webpack exports the libraryName in the umd bundle
         // TODO make this more automatic for consistency
         libraryName: `${entry.directory}Destination`,
-        url: `${path}/${entry.directory}.js`
+        url: `${path}/${entry.directory}/${webBundles()[entry.directory]}`
```
mshwery, line 89: I wonder if we should add a confirmation step that outputs all of these changes so you can choose to continue (like we do in `register` and `push`)?

mshwery (commented):
looks like a great start

packages/cli/src/lib/web-bundles.ts
juliofarah: Globby returns an array of strings with the full paths of the files, forcing me to have something like
```js
    const path: string[] = globby.sync(`${DIST_DIR}/${destination}/*.js`)
    const split = path[0].split('/')
    const file = split[split.length -1]
```
instead of 
```js
execa.commandSync(`ls ${DIST_DIR}/${destination}`).stdout.split('\n')[0
```

See:

```js
console.log({
      destination,
      globby: globby.sync(`${DIST_DIR}/${destination}/*.js`),
      execa: execa.commandSync(`ls ${DIST_DIR}/${destination}`).stdout.split('\n')
    })
```


```
{
  destination: 'amplitude-plugins',
  globby: [
    'packages/browser-destinations/dist/web//amplitude-plugins/9fac0b2912f3a24f96fe.js'
  ],
  execa: [
    '9fac0b2912f3a24f96fe.js',
    '9fac0b2912f3a24f96fe.js.gz',
    '9fac0b2912f3a24f96fe.js.map',
    '9fac0b2912f3a24f96fe.js.map.gz'
  ]
}
```

packages/cli/src/commands/push-browser-destinations.ts
juliofarah: As I don't have the plugins and really don't want to fetch them just to show the diff, I thought about showing something like this
image

Showing the diff that will be pushed to the database sounds enough to me. Any thoughts?

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -85,22 +90,60 @@ export default class PushBrowserDestinations extends Command {
...
       }
 
       // We expect that each definition produces a single Remote Plugin bundle
       // `name` + `metadataId` are guaranteed to be unique
       const existingPlugin = remotePlugins.find((p) => p.metadataId === metadata.id && p.name === metadata.name)
 
       if (existingPlugin) {
```
mshwery, line 100: Looks like you do have the existing plugins already

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -85,22 +90,60 @@ export default class PushBrowserDestinations extends Command {
...
+    if (!shouldContinue) {
+      this.exit()
+    }
+
+    try {
+      this.spinner.start('Persisting changes')
+      await Promise.all([
+        pluginsToUpdate.map((p) => updateRemotePlugin(p)),
+        pluginsToCreate.map((p) => createRemotePlugin(p))
+        // this.spinner.succeed(`Updated existing remote plugin for ${metadata.name}`)
```
mshwery, line 139: Let's remove dead code

mshwery (approved):
Looks great

packages/cli/src/commands/push-browser-destinations.ts
juliofarah: YAS

image

ps: we're only gonna show the diff if the values are actually different (this was just for the screenshot)

juliofarah: image

mshwery (approved):

mshwery:
@juliofarah did you figure out the publicPath issue where staging bundles where referencing production chunks?

juliofarah:
> @juliofarah did you figure out the publicPath issue where staging bundles where referencing production chunks?

Yes! The problem we were having before was that `push-browser-destinations` called `lerna run build`, building every package inside the monorepo with no ASSET_PATH (assuming `production` as the default value). Running `build web stage` then would produce a second bundle which was not picked up by my script.

Solution:
- build only the packages we need to avoid redundant builds of web libraries
- [remove any built files](https://github.com/segmentio/action-destinations/pull/147/files#diff-315bc8c7cefb61d11192cd5801e8c603f0a9150da995f8f470f672236257a727R21) to prevent old builds from being pushed to s3

PR #148, opened 2021-09-09
Update Enhanced Conversions to handle invalid/expired tokens

lcampos:
Update post conversion action to handle invalid/expired oauth tokens. Since the API returns a 400 http code instead of a 401, we are catching errors, checking if they belong to an invalid/expired oauth token and internally throwing a 401 error so we can trigger the oauth refresh token flow in actions core.

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/postConversion/index.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/postConversion/index.ts
@@ -227,9 +227,8 @@ const action: ActionDefinition<Settings, Payload> = {
       hashed_phone_number: [formatPhone(payload.phone_number)]
     })
 
-    let res
     try {
-      res = await request('https://www.google.com/ads/event/api/v1', {
+      return await request('https://www.google.com/ads/event/api/v1', {
         method: 'post',
         json: {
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/google-enhanced-conversions/postConversion/index.ts
```diff
@@ -216,13 +227,31 @@ const action: ActionDefinition<Settings, Payload> = {
...
+        json: {
+          pii_data: { ...pii_data, address: [address] },
+          ...conversionData
+        }
+      })
+    } catch (err) {
+      // Google returns a 400 when using invalid tokens
+      // We'll catch invalid token errors and throw an internal 401 to
+      // trigger the refresh token flow in this transaction
+      if (err instanceof HTTPError) {
```
lcampos, line 242: I prefer doing this than updating actions/core to handle specific api behaviors related to invalid authentication tokens but let me know what you think.

marinhero: Sounds reasonable to me since you'll end up leveraging the refresh token flow without any issues. 👍

marinhero (approved):

nick-Ag (approved):

PR #149, opened 2021-09-09
GRAMA-5 [personas-messaging-sendgrid] Allows journey metadata to be added to a send

lew-gordon:
We're going to want to track analytical data on sends from journeys, so we'll need to ensure that certain journey metadata is included in the webhook callback.

### DIFF
--- a/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
@@ -139,6 +139,24 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'The HTML content of the body',
       type: 'string',
       required: true
+    },
+    journeyId: {
+      label: 'Journey Id',
+      description: 'Journey id used for analytics',
+      type: 'string',
+      required: false
+    },
+    journeyStateId: {
+      label: 'Journey State Id',
+      description: 'Journey state id used for analytics',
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -139,6 +139,24 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      description: 'Journey state id used for analytics',
+      type: 'string',
+      required: false
+    },
+    audienceId: {
+      label: 'Audience Id',
+      description: 'Audience id used for analytics',
+      type: 'string',
+      required: false
     }
```
mshwery, line 160: If you want a grab bag of arbitrary data you could alternatively use an object type instead.

lew-gordon: Sounds good to me. I wasn't sure if there was a precedent for that and if the desire was to have the fields be specific.

mshwery: It's up to you and the needs of the action.

If the specific fields are arbitrary or unknown then I'd recommend an object field. If the properties are important and need to be validated individually then what you have works better.

vikramkumar19 (approved):
Looks good

lew-gordon:
Testing completed in staging

PR #151, opened 2021-09-09
[LIBWEB-877] fixes for amplitude UA parsing

dlasky:
default path did not match segment spec

added some additional non-transparent mapping to better replicate how classic handles the ua parsing


packages/destination-actions/src/destinations/amplitude/user-agent.ts
```diff
@@ -17,10 +17,12 @@ export function parseUserAgentProperties(userAgent?: string): ParsedUA {
   const device = parser.getDevice()
   // eslint-disable-next-line @typescript-eslint/no-unsafe-call
   const os = parser.getOS()
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+  const browser = parser.getBrowser()
 
   return {
-    os_name: os.name,
-    os_version: os.version,
+    os_name: browser.name ?? os.name,
```
dlasky, line 24: @nielst @spencerattick in particular this is the change to better match classic behavior

mshwery (approved):

PR #152, opened 2021-09-09
[BE-102] Automate Package Publishing

daltonscharff:
[RFC: Publish action-destinations Package Using CI Pipeline](https://paper.dropbox.com/doc/RFC-Publish-action-destinations-Package-Using-CI-Pipeline--BSEadZPFIYUeAi9sUggLCRTlAg-H0LJ3fjhkyJLVICXjGOTn)
[BE-102](https://segment.atlassian.net/browse/BE-102)

## Background
One pain point of this repository is that it requires developers to publish new packages from their local machines. I recently proposed an RFC (linked above) that outlined a plan to automate this package publishing process. The proposed solution, `auto`, did not work as seamlessly with monorepos as expected, so the solution in this PR is a bit different (and forgoes the changelog requirement altogether). The RFC contains a lot of good discussion and ideas on alternate implementations, which is why I've included it as a reference document. Since there are many different ideas, I intend this solution to be iterated on and modified in the future. My goal of this PR is to see if this is an agreeable first step in the right direction.

## Developer Experience
This PR establishes a `publish` GitHub action that can run **manually via workflow dispatch** or **automatically when a new pull request is merged**. 

[auto-publish-demo](https://github.com/segmentio/auto-publish-demo) is a repository that demonstrates this experience. Feel free experiment and merge PRs on that repository to get a better understanding. Since it is a demo, no `hello-world` or `random-number-generator` packages are published, but the [release tags](https://github.com/segmentio/auto-publish-demo/releases) are updated after the publish action is run. Running the `publish` action on `action-destinations` publishes NPM packages as well as updates release tags.

### Option 1: Manual trigger
1. In this repository, navigate to the 'Actions' tab
2. Under 'Workflows', click 'Publish'
3. In the workflow history box, click 'Run workflow'
4. Ensure the `main` branch is selected and enter the semver version you would like to publish
    - accepts any valid command for [lerna publish](https://github.com/lerna/lerna/tree/main/commands/publish#readme) (e.g., major, minor, patch)

### Option 2: Automated via PR
1. Clone the repository
2. Create a new branch
3. Modify one of the packages in the `./packages` directory
4. Push your changes
5. Create a pull request
6. Specify a label associated with the semver version you would like to publish
    - valid labels include `major`, `minor`, or `patch`
7. Merge the pull request

## Potential Issues
- All PRs that contain a label of `major`, `minor`, or `patch` will be published upon PR merge.
  - maybe only PRs with the label `publish` (or similar) should be published
- No changelog is generated
  - would be a nice addition if we began batching changes together

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,95 @@
+name: Publish
+
+on:
+  pull_request:
+    branches: [main]
+    types: [closed]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Semver version to bump'
+        required: true
+        default: 'patch'
+
+jobs:
+  get-version:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.workflow.outputs.version || steps.major.outputs.version ||  steps.minor.outputs.version ||  steps.patch.outputs.version }}
+    steps:
+      - id: patch
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'patch') }}
+        run: echo "::set-output name=version::patch"
+      - id: minor
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'minor') }}
+        run: echo "::set-output name=version::minor"
+      - id: major
+        if: ${{ contains(github.event.pull_request.labels.*.name, 'major') }}
+        run: echo "::set-output name=version::major"
+      - id: workflow
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: echo "::set-output name=version::${{ github.event.inputs.version }}"
+
+  test-build-and-publish:
+    runs-on: ubuntu-latest
+
+    needs: get-version
+    if: ${{ needs.get-version.outputs.version }}
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - uses: actions/cache@v2
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Build
+        run: NODE_ENV=production yarn build
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Test
+        run: yarn test
+
+      - name: destination-subscriptions size
+        run: |
+          if $(lerna changed | grep -q destination-subscriptions); then
+            yarn subscriptions size
+          fi
+
+      - name: Set author
+        run: |
+          git config user.email "github-actions@github.com"
+          git config user.name "github-actions"
+
+      - name: Publish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: lerna version ${{ needs.get-version.outputs.version }} --yes
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -89,8 +89,8 @@ jobs:
 
       - name: Set author
         run: |
-          git config user.email "github-actions@github.com"
-          git config user.name "github-actions"
+          git config user.email "`git log --format='%ae' HEAD^!`"
+          git config user.name "`git log --format='%an' HEAD^!`"
 
       - name: Publish
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,95 @@
...
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Build
+        run: NODE_ENV=production yarn build
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Test
+        run: yarn test
```
gpsamson, line 78: I'm not sure this is feasible since the `publish` workflow can be triggered without a `ci` workflow, but it would be nice to forgo these steps and rely on the `ci` workflow to lint and test.

https://github.com/segmentio/action-destinations/actions/workflows/ci.yml

.github/workflows/publish.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: destination-subscriptions size
+        run: |
+          if $(lerna changed | grep -q destination-subscriptions); then
+            yarn subscriptions size
+          fi
+
+      - name: Set author
+        run: |
+          git config user.email "github-actions@github.com"
+          git config user.name "github-actions"
```
gpsamson, line 89: Are we able to get assign owner of the commit rather than a generic user?

gpsamson (commented):
I'm excited about this! Nice work spiking and speccing this out 🔥

.github/workflows/publish.yml
daltonscharff: I agree. I think we could change these steps to only run if the publish action is triggered from a workflow dispatch since they'd be redundant if it's triggered from a PR merge.

gpsamson: Ahh, that's how you'd do it. Neat!

.github/workflows/publish.yml
daltonscharff: Yes I think this is possible. I'll work on making that change.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,97 @@
+name: Publish
+
+on:
+  pull_request:
```
lcampos, line 4: Can we use `path` or `path-ignore` to only trigger the action on meaningful code changes ?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Set author
+        run: |
+          git config user.email "github-actions@github.com"
+          git config user.name "github-actions"
+
+      - name: Publish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: lerna version ${{ needs.get-version.outputs.version }} --yes
```
lcampos, line 97: We'll need to add `--allow-branch` or modify the current allowBranch config if we want to use this with a branch different than `main`
https://github.com/segmentio/action-destinations/blob/7cbdf06b2cdd333b92bc3b23a1e8d2547e055ccc/lerna.json#L11

.github/workflows/publish.yml
daltonscharff: I like this. I'll add it!

.github/workflows/publish.yml
daltonscharff: This now sets these values to the name and email associated with the most recent commit in the branch.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Test
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: yarn test
+
+      - name: destination-subscriptions size
+        run: |
+          if $(lerna changed | grep -q destination-subscriptions); then
+            yarn subscriptions size
+          fi
```
gpsamson, line 88: It's probably okay to ignore this step
```suggestion
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Build
+        run: NODE_ENV=production yarn build
+
+      - name: Lint
+        if: ${{ github.event_name == 'workflow_dispatch' }}
```
gpsamson, line 77: 🔥

.github/workflows/publish.yml
gpsamson: Shouldn't this line be `lerna publish`?

I'd be down to merge this in without support for non-`main` branches, then follow up to support those. I think there's some additional complexity that may warrant more thinking and additional workflow. Is there a way to publish prerelease versions with branch names or commit SHAs? Are there additional guardrails we want in place for pull request publishes?

daltonscharff: It should be `lerna publish`. That was an oversight that I've fixed now. And I agree that maybe the non-main branch publishes could come later due to added complexity.

mshwery:
What happens when someone adds multiple version labels (major + minor + patch or some combination of 2 or more)?

mshwery:
Also, do we have any considerations for how a developer can publish a canary (that will not conflict with other canaries that have never been merged)?

daltonscharff:
> What happens when someone adds multiple version labels (major + minor + patch or some combination of 2 or more)?
https://github.com/segmentio/action-destinations/blob/2bd1339801f896598b515c1845d9999284cdd18b/.github/workflows/publish.yml#L20
Major has precedence over minor, which has precedence over patch. So applying the labels `major` and `patch` to a single PR would make for a major version bump.

daltonscharff:
> Also, do we have any considerations for how a developer can publish a canary (that will not conflict with other canaries that have never been merged)?

Maybe in order to publish a canary, the user can go into the github actions tab and select their branch. Seeing that the selected branch isn't `main`, the publish workflow can then publish it with as a canary? That functionality hasn't been built out here yet, but I wonder if this would meet requirements.

gpsamson (approved):
:shipit: +1 to supporting prerelease publishing in a follow-up PR since it will introduce a bit more complexity to the workflow.

PR #155, opened 2021-09-10
GRAMA-7 [personas-messaging-twilio] Allows journey metadata to be added to a send

lew-gordon:
We're going to want to track analytical data on sends from journeys. This will allow us to add that data via custom args. This will require additional backend setup to ensure that the action-destination is configured with the proper source base webhook.

This follows from #149. My plan is probably to wait on feedback so that this can take a similar approach.

### DIFF
--- a/packages/destination-actions/src/destinations/personas-messaging-twilio/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-twilio/index.ts
@@ -43,6 +43,13 @@ const destination: DestinationDefinition<Settings> = {
         description: 'Source ID',
         type: 'string',
         required: true
+      },
+      webhookUrl: {
+        label: 'Webhook URL',
+        description: 'Webhook URL that will receive all events for the sent message',
+        type: 'string',
+        format: 'uri',
+        required: false
       }
     },
     testAuthentication: (request) => {
--- a/packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
@@ -90,6 +90,12 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'Message to send',
       type: 'text',
       required: true
+    },
+    customArgs: {
+      label: 'Custom Arguments',
+      description: 'Additional custom arguments that will be opaquely sent back on webhook events',
+      type: 'object',
+      required: false
     }
   },
   perform: async (request, { settings, payload }) => {
@@ -113,16 +119,31 @@ const action: ActionDefinition<Settings, Payload> = {
     // and we no longer need to call the profiles API first
     const token = Buffer.from(`${settings.twilioAccountId}:${settings.twilioAuthToken}`).toString('base64')
 
+    const body = new URLSearchParams({
+      Body: Mustache.render(payload.body, { profile }),
+      From: payload.fromNumber,
+      To: phone
+    })
+
+    const webhookUrl = settings.webhookUrl
+    const customArgs = payload.customArgs
+    if (webhookUrl && customArgs) {
+      // Webhook URL parsing has a potential of failing. I think it's better that
+      // we fail out of any invocation than silently not getting analytics
+      // data if that's what we're expecting.
+      const webhookUrlWithParams = new URL(webhookUrl)
+      for (const key of Object.keys(customArgs)) {
+        webhookUrlWithParams.searchParams.append(key, String(customArgs[key]))
+      }
+      body.append('StatusCallback', webhookUrlWithParams.toString())
+    }
+
     return request(`https://api.twilio.com/2010-04-01/Accounts/${settings.twilioAccountId}/Messages.json`, {
       method: 'POST',
       headers: {
         authorization: `Basic ${token}`
       },
-      body: new URLSearchParams({
-        Body: Mustache.render(payload.body, { profile }),
-        From: payload.fromNumber,
-        To: phone
-      })
+      body
     })
   }
 }
--- a/packages/destination-actions/src/destinations/personas-messaging-twilio/__tests__/send-sms.test.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-twilio/__tests__/send-sms.test.ts
@@ -17,12 +17,19 @@ for (const environment of ['stage', 'production']) {
 
   const endpoint = `https://profiles.segment.${environment === 'production' ? 'com' : 'build'}`
 
+  beforeEach(() => {
+    nock(`${endpoint}/v1/spaces/d/collections/users/profiles/user_id:jane`).get('/traits?limit=200').reply(200, {
+      traits: {}
+    })
+  })
+
+  afterEach(() => {
+    twilio.responses = []
+    nock.cleanAll()
+  })
+
   describe(`${environment} - send SMS`, () => {
     it('should abort when there is no `phone` external ID', async () => {
-      nock(`${endpoint}/v1/spaces/d/collections/users/profiles/user_id:jane`).get('/traits?limit=200').reply(200, {
-        traits: {}
-      })
-
       nock(`${endpoint}/v1/spaces/d/collections/users/profiles/user_id:jane`)
         .get('/external_ids?limit=25')
         .reply(200, {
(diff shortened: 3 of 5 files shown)


packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
```diff
@@ -103,7 +109,7 @@ const action: ActionDefinition<Settings, Payload> = {
       traits
     }
 
-    const phone = payload.toNumber || profile.phone
+    const phone = payload.toNumber || profile.phone || profile.traits.phone
```
lew-gordon, line 112: I wasn't sure if we actually needed this. This was the only way I could get it to work in my local end to end testing in stage.

laurenfb: I ran into this too initially. I think the reason here is that maybe you've added your phone number to the user's profile as a trait rather than an external id. @vadimdemedes told me he wants to use external ids for this purpose, and the external id needs to be mapped as an identifier (lots of work, sorry). I guess we could use `profile.traits.phone` as well - sadly Vadim is out for all of next week so we can't ask him his thoughts behind this. I think his idea was that a customer could have a custom `phone` profile trait that we didn't want to collide with. 

you need an identify event like this:

```
{
    "userId": "<create a user id here>",
    "context": {
        "externalIds": [{
            "id": "+17072478800",
            "type": "phone",
            "collection": "users",
            "encoding": "none"
        }]
    },
    "integrations": {},
    "traits": {
        "phone": "+17072478800",
        "cat": "dummy"
    }
}```

there's a little more info about this in the "chicken scratch" section of our SDD at the very bottom - https://paper.dropbox.com/doc/SDD-Destination-Actions-for-Personas-Messaging-SendGrid-and-TwilioWIP--BSKR1EAJJJIwEXxS295ZAyrzAg-sLGQ3j2Hvh9qLX7nXCjJE

Comment (lew-gordon) [reply packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts:112]: Oh, interesting. Thank you for the context!

File: packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts [hunk 2]
```diff
@@ -113,16 +119,31 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    const body = new URLSearchParams({
+      Body: Mustache.render(payload.body, { profile }),
+      From: payload.fromNumber,
+      To: phone
+    })
+
+    const webhookUrl = settings.webhookUrl
+    const customArgs = payload.customArgs
+    if (webhookUrl && customArgs) {
+      // Webhook URL parsing has a potential of failing. I think it's better that
```
laurenfb, line 131: I think your comment makes sense - what is the experience like for the customer when the webhook fails, I wonder?

lew-gordon: Yeah, I wasn't sure. I was thinking that since this is really an internal server error and analytics are a large reason for these messaging orchestrations we could alert on this. I wasn't sure if there was a clear way to indicate that "analytics is degraded" outside of an explicit failure.

laurenfb: Is it possible to try with a funky webhook and see what happens? Maybe a PM would want to weigh in on the experience, but I don't know. Not a blocking comment, just a wondering.

lew-gordon: I think this is a great idea. I'll definitely test this out before merging.

packages/destination-actions/src/destinations/personas-messaging-twilio/index.ts
```diff
@@ -43,6 +43,12 @@ const destination: DestinationDefinition<Settings> = {
         description: 'Source ID',
         type: 'string',
         required: true
+      },
+      webhookUrl: {
+        label: 'Webhook URL',
+        description: 'Webhook URL that will receive all events for the sent message',
+        type: 'string',
```
mshwery, line 50: You can add 
```suggestion
        type: 'string',
        format: 'uri',
```

If this is expected to be a valid uri.

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
```diff
@@ -90,6 +90,12 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'Message to send',
       type: 'text',
       required: true
+    },
+    customArgs: {
+      label: 'Custom Args',
+      description: 'Additional custom args that will be opaquely sent back on webhook events',
```
mshwery, line 96: Might be worth spelling out "args" or using a different word that's more approachable for customers – soon action fields will be programmatically generating docs in our catalog docsite.

```suggestion
      description: 'Additional custom arguments that will be opaquely sent back on webhook events',
```

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
mshwery: As a note (non-blocking): we should avoid hard-coded mappings wherever possible for actions. This is one of the main pain points that actions solves – customers are able to configure how properties from an event are transformed into the payload that gets sent to the partner destination. By hard-coding the behavior, that means the customer can no longer transparently see what value will be sent to the end tool, or where it's coming from.

I think in this case you have to use the fallbacks as you have it, because we don't have the profile data at the time the action's mappings are resolved. If we did, you could expect a customer to control the mapping transparently in their config with mapping-kit directives or literal values:
```json
{
  "toNumber": {
    "@if": {
      "exists": { "@path": "$.profile.phone" },
      "then": { "@path": "$.profile.phone" },
      "else": { "@path": "$.profile.traits.phone" }
    }
  }
}
```

The `profile` / `profile.traits` would have to be properties of the event for this mapping configuration to work.

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
lew-gordon: Should I spell out the label, too? I wasn't sure where that was used.

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
lew-gordon: Hey @laurenfb what do you think about this? ☝️  I figured not implementing this for this PR to keep the scope of the change down, but I think it'd be a great change for a future PR.

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
mshwery: TBH I'm not sure how this is rendered by Personas UI – but in the main catalog and action editor it looks like this: 
Screen Shot 2021-09-10 at 3 06 01 PM

This screenshot is of the Slack action-destination: 
https://github.com/segmentio/action-destinations/blob/5bd7ff8b79e903ecc8af2ec2b2b12f21f23bbbd2/packages/destination-actions/src/destinations/slack/postToChannel/index.ts#L8-L40

So you can see how the label + description are displayed to users.

lew-gordon: Gotcha. I think based off of that and this view from testing the SendGrid action it makes more sense to have the label be readable.

image

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
vikramkumar19: I do like this approach :)

laurenfb: agreed, I like @mshwery's approach (thanks Matt!). I will make a ticket for it.

lew-gordon:
Testing completed in staging.

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
lew-gordon: Yeah just a normal sending error for the action. Looks like it just won't send and ends up in the standard error metrics section for the action-destination.

lew-gordon:
@laurenfb when you get a moment could I get another quick review, please?

packages/destination-actions/src/destinations/personas-messaging-twilio/sendSms/index.ts
laurenfb: filed https://segment.atlassian.net/browse/GROW-336

laurenfb (approved):
I can't quite tell what changes were added since the force push, but nothing stands out. 👍

PR #158, opened 2021-09-13
Add webhook action

tysonmote:


packages/destination-actions/src/destinations/webhook/postJSON/index.ts
```diff
@@ -0,0 +1,31 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Post JSON',
```
mshwery, line 6: WDYT of `Send to Webhook` or `POST Webhook`?

packages/destination-actions/src/destinations/webhook/postJSON/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  description: 'POST JSON to a custom URL',
```
mshwery, line 7: Why limit to json? Make that configurable with the action fields.

What if we add `payload_type: 'form' | 'json'` and `body` could be `data` (so it's a bit more agnostic to content-type).

packages/destination-actions/src/destinations/webhook/postJSON/index.ts
```diff
@@ -0,0 +1,31 @@
...
+      description: 'URL to deliver JSON-encoded message to.',
+      type: 'string',
+      required: true,
+      format: 'uri'
+    },
+    body: {
+      label: 'Body',
+      description: 'Message to deliver to webhook URL (JSON-encoded).',
+      type: 'object',
+      default: { '@path': '$' }
```
mshwery, line 20: You know I never tried this. Not sure if it works (we aren't using jsonpath because it was so big and provided way more features than we wanted to debug/support).

packages/destination-actions/src/destinations/webhook/postJSON/index.ts
mshwery: Could even consider pulling in the request method as a field – unless we want them to be separate actions (unclear if they would have different schemas. I can imagine a `GET` different than a `POST|PUT|PATCH`.

mshwery (approved):
let's merge this and we can iterate – and i can add a `performBatch` to my branch

packages/destination-actions/src/destinations/webhook/postJSON/index.ts
tysonmote: Ah, yeah -- forgot about that. It looks like this doesn't work. I'm pushing a fix shortly.

packages/destination-actions/src/destinations/webhook/postJSON/index.ts
tysonmote: Yeah, that makes a lot of sense. Pushing a commit in a bit.

packages/core/src/get.ts
```diff
@@ -5,17 +5,18 @@
 export function get<T = unknown, Default = undefined>(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   obj: any,
-  path: string | string[],
-  defValue?: Default
```
tysonmote, line 9: Just a heads up -- this argument was unused so I removed it.

packages/core/src/get.ts
```diff
@@ -5,17 +5,18 @@
...
+  path: string | string[]
 ): T | Default | undefined {
-  // If path is not defined or it has false value
-  if (!path) return defValue
```
mshwery, line 12: Were we not using this argument ever?

packages/destination-actions/src/destinations/webhook/index.ts
```diff
@@ -0,0 +1,15 @@
+import type { DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+
+import send from './send'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Webhook',
+  slug: 'webhook',
```
mshwery, line 8:
```suggestion
  slug: 'actions-webhook',
```

packages/destination-actions/src/destinations/webhook/send/index.ts
```diff
@@ -0,0 +1,52 @@
...
+      description: 'URL to deliver data to.',
+      type: 'string',
+      required: true,
+      format: 'uri'
+    },
+    method: {
+      label: 'Method',
+      description: 'HTTP method to use.',
+      type: 'string',
+      default: 'POST',
```
mshwery, line 34:
```suggestion
      choices: [
        { label: 'POST', value: 'POST' },
        { label: 'PUT', value: 'PUT' },
        { label: 'PATCH', value: 'PATCH' },
        { label: 'DELETE', value: 'DELETE' },
        { label: 'GET', value: 'GET' }
      ],
      default: 'POST',
```

mshwery: Maybe we should support a shorthand for `['Same Label <> Value']` strings.

mshwery (approved):

packages/core/src/get.ts
tysonmote: Nope! I tried to comment this but GitHub lost it...

PR #159, opened 2021-09-13
Remove actions-core from optional deps

gpsamson:
Removes `@segment/actions-core` as an _optional_ dependency of other packages. The external workflow doesn't break with this change: https://github.com/segmentio/action-destinations/runs/3592019198


mshwery (approved):
Do we need to `yarn install` after this change? Will it impact the lockfile at all?

daltonscharff (approved):
Looks good to me!

gpsamson:
> Do we need to `yarn install` after this change? Will it impact the lockfile at all?

I ran `yarn install` a few times and the lockfile didn't change. CI would have failed since we use `--frozen-lockfile`

PR #160, opened 2021-09-13
make github cancel jobs that take too long

mshwery:
Prevent this insanity:

Screen Shot 2021-09-13 at 4 31 37 PM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,4 +103,4 @@ jobs:
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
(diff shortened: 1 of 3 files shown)


gpsamson (approved):
LGTM. Looks like Snyk is flagging a vulnerability though?

daltonscharff (approved):

mshwery:
> LGTM. Looks like Snyk is flagging a vulnerability though?
> 
> See: https://github.com/segmentio/action-destinations/pull/160/checks?check_run_id=3592276056#step:9:61

Yeah – annoyingly it's a package that we cannot upgrade because the consumer dependencies may not be prepared for esmodules (the patch is only available as pure esm)

PR #161, opened 2021-09-13
Load cps-client at runtime

gpsamson:
Oclif suffers from a bug where all commands are imported (but not invoked) anytime the CLI is called, regardless of which command it is (see https://github.com/oclif/oclif/issues/606). This issue causes confusing errors to be logged when the builder doesn't have `@segment/control-plane-service-client` installed:

```
(node:19492) [MODULE_NOT_FOUND] Error Plugin: @segment/actions-cli: Cannot find module '@segment/control-plane-service-client'
Require stack:
- /Users/builder/action-destinations/packages/cli/src/lib/control-plane-service.ts
- /Users/builder/action-destinations/packages/cli/src/lib/control-plane-client.ts
- /Users/builder/action-destinations/packages/cli/src/commands/push-browser-destinations.ts
- /Users/builder/action-destinations/node_modules/@oclif/config/lib/plugin.js
- /Users/builder/action-destinations/node_modules/@oclif/config/lib/config.js
- /Users/builder/action-destinations/node_modules/@oclif/config/lib/index.js
- /Users/builder/action-destinations/node_modules/@oclif/command/lib/command.js
- /Users/builder/action-destinations/node_modules/@oclif/command/lib/index.js
- /Users/builder/action-destinations/packages/cli/bin/run
module: @oclif/config@1.17.0
task: toCached
plugin: @segment/actions-cli
root: /Users/builder/action-destinations/packages/cli
See more details with DEBUG=*
(Use `node --trace-warnings ...` to show where the warning was created)
```

This pull request updates the CLI so the `@segment/control-plane-service-client` dependency is loaded when commands are run/invoked so `MODULE_NOT_FOUND` are no longer logged.

When a command is invoked and the CPS client can't be loaded, this warning will be printed:
```
❯ ./bin/run push
 ›   Warning: This command is only available to Segmenters. If you are a Segment internal builder please run `yarn
 ›   install` and ensure the control-plane-service-client is installed.
```

### DIFF
--- a/packages/cli/src/commands/push-browser-destinations.ts
+++ b/packages/cli/src/commands/push-browser-destinations.ts
@@ -7,6 +7,7 @@ import ora from 'ora'
 import { assetPath } from '../config'
 import type { RemotePlugin } from '../lib/control-plane-service'
 import { prompt } from '../lib/prompt'
+import loadCPS from '../lib/control-plane-service'
 import {
   getDestinationMetadatas,
   getRemotePluginByDestinationIds,
@@ -34,6 +35,18 @@ export default class PushBrowserDestinations extends Command {
   static args = []
 
   async run() {
+    try {
+      await loadCPS()
+    } catch (err) {
+      if (err.code === 'MODULE_NOT_FOUND') {
+        this.warn(
+          'This command is only available to Segmenters. If you are a Segment internal builder please run `yarn install` and ensure the control-plane-service-client is installed.'
+        )
+      } else {
+        this.error(err.message)
+      }
+      this.exit()
+    }
     const { flags } = this.parse(PushBrowserDestinations)
     const { destinationIds } = await prompt<{ destinationIds: string[] }>({
       type: 'multiselect',
--- a/packages/cli/src/commands/push-browser-destinations.ts
+++ b/packages/cli/src/commands/push-browser-destinations.ts
@@ -36,7 +36,7 @@ export default class PushBrowserDestinations extends Command {
 
   async run() {
     try {
-      await loadCPS()
+      loadCPS()
     } catch (err) {
       if (err.code === 'MODULE_NOT_FOUND') {
         this.warn(
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -81,7 +81,7 @@ export default class Push extends Command {
 
   async run() {
     try {
-      await loadCPS()
+      loadCPS()
     } catch (err) {
       if (err.code === 'MODULE_NOT_FOUND') {
         this.warn(
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -11,6 +11,7 @@ import chalk from 'chalk'
 import { uniq, pick, omit, sortBy, mergeWith } from 'lodash'
 import { diffString } from 'json-diff'
 import ora from 'ora'
+import loadCPS from '../lib/control-plane-service'
 import type {
   ClientRequestError,
   DestinationMetadata,
@@ -79,6 +80,19 @@ export default class Push extends Command {
   static args = []
 
   async run() {
+    try {
+      await loadCPS()
+    } catch (err) {
+      if (err.code === 'MODULE_NOT_FOUND') {
+        this.warn(
+          'This command is only available to Segmenters. If you are a Segment internal builder please run `yarn install` and ensure the control-plane-service-client is installed.'
+        )
+      } else {
+        this.error(err.message)
+      }
+      this.exit()
+    }
+
     const { flags } = this.parse(Push)
 
     const { metadataIds } = await prompt<{ metadataIds: string[] }>({
--- a/packages/cli/src/lib/control-plane-client.ts
+++ b/packages/cli/src/lib/control-plane-client.ts
@@ -1,5 +1,5 @@
-import {
-  controlPlaneService,
+import loadCPS from '../lib/control-plane-service'
+import type {
   DestinationMetadata,
   DestinationMetadataAction,
   DestinationMetadataActionCreateInput,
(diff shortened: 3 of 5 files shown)


packages/cli/src/lib/control-plane-service.ts
```diff
@@ -1,16 +1,26 @@
...
-  timeout: 10000,
-  headers: {
-    // All calls from this script are system-to-system and shouldn't require authz checks
-    // TODO remove this once we support auth tokens
-    'skip-authz': '1'
+let client: Client
+
+export default async function loadCPS(): Client {
+  if (!client) {
+    const module = await import('@segment/control-plane-service-client')
```
mshwery, line 10: Instead of dynamic import, can we use `require`?
```suggestion
export default function createClient(): Client {
  if (!client) {
    const module = require('@segment/control-plane-service-client')
```

mshwery: You may need an eslint ignore for the `require` but there should be some prior art somewhere:
https://github.com/segmentio/action-destinations/blob/df68122c34cfcac6f8cabb28cd6ea64f3b57bbf5/packages/cli/src/lib/destinations.ts#L21

packages/cli/src/lib/control-plane-client.ts
```diff
@@ -160,6 +168,7 @@ export async function updateRemotePlugin(plugin: RemotePlugin): Promise<RemotePl
 }
 
 export async function createRemotePlugin(plugin: RemotePlugin): Promise<RemotePlugin> {
+  const controlPlaneService = await loadCPS()
```
mshwery, line 171: You actually shouldn't need this `loadCPS` wrapper function if you do `require('../path/to/module')` since it will only be required in that codepath.

packages/cli/src/lib/control-plane-client.ts
```diff
@@ -64,6 +67,7 @@ export async function updateDestinationMetadata(
 }
 
 export async function setSubscriptionPresets(metadataId: string, presets: DestinationSubscriptionPresetInput[]) {
+  const controlPlaneService = await loadCPS()
   const { data, error } = await controlPlaneService.setDestinationSubscriptionPresets(NOOP_CONTEXT, {
```
mshwery, line 71: Seems that the client isn't being typed properly...

packages/cli/src/lib/control-plane-client.ts
gpsamson: Ahh, you're right! 🙇🏽

packages/cli/src/commands/push.ts
```diff
@@ -79,6 +80,19 @@ export default class Push extends Command {
...
+      if (err.code === 'MODULE_NOT_FOUND') {
+        this.warn(
+          'This command is only available to Segmenters. If you are a Segment internal builder please run `yarn install` and ensure the control-plane-service-client is installed.'
+        )
+      } else {
+        this.error(err.message)
+      }
+      this.exit()
+    }
+
```
gpsamson, line 95: I purposely left this block even though helper functions in `lib/control-plane-client` call it to in order to give immediate feedback to the builder. Rather than failing after they fill out the prompts.

packages/cli/src/lib/control-plane-service.ts
```diff
@@ -1,16 +1,29 @@
...
+      name: 'control-plane-service',
+      url: 'http://control-plane-service.segment.local',
+      userAgent: 'Segment (actions cli)',
+      timeout: 10000,
+      headers: {
+        // All calls from this script are system-to-system and shouldn't require authz checks
+        // TODO remove this once we support auth tokens
+        'skip-authz': '1'
+      }
+    })
```
gpsamson, line 25: Couldn't quite resolve this type issue so I did what any experienced engineer would do and ignored the line. :)

From what I've read, it's due to the require statement happening at runtime but I couldn't find any helpful suggestions. Anyone have any thoughts on how to resolve?

packages/cli/src/commands/push-browser-destinations.ts
```diff
@@ -34,6 +35,18 @@ export default class PushBrowserDestinations extends Command {
   static args = []
 
   async run() {
+    try {
+      loadCPS()
```
mshwery, line 39: Do we need to use the module somewhere?

packages/cli/src/lib/control-plane-service.ts
mshwery: I think it can't find the types because it's a `require`.

You could try to do `/** @type {import('@segment/control-plane-service-client')} */` right above it. Might need a `typeof` in there...

gpsamson: I tried with and without `typeof` I'm still getting type errors. I would also assume that the `import type { Client }`(see [here](https://github.com/segmentio/action-destinations/pull/161/files#diff-926602a4b37e13925620923036bdea084c5df7a7c92c772e2749fb7b0744fda9R2)) would help things but I guess not?

packages/cli/src/commands/push-browser-destinations.ts
gpsamson: This block aims to give immediate feedback to the builder. If we remove this block, an external builder could fill out prompts before being notified that the command won't work for them.

mshwery:
@gpsamson what are your thoughts around making the internal parts of the CLI a separate oclif plugin? That way we could make this process smoother – people could do `yarn install` directly, and avoid obscure optional dep errors that may cause confusion. I think it'll be a little bit of time before we are ready to release all of the internal commands.

gpsamson:
> @gpsamson what are your thoughts around making the internal parts of the CLI a separate oclif plugin? That way we could make this process smoother – people could do `yarn install` directly, and avoid obscure optional dep errors that may cause confusion. I think it'll be a little bit of time before we are ready to release all of the internal commands.
> 
> (If you think this PR solves it in a more direct/simpler way, let's keep going with it)

@mshwery I agree with that approach! That (imo) is the most correct way to mitigate these issues.

gpsamson:
Closing in favor of a separate internal command plugin. Tracking in Jira (see backlog)

PR #162, opened 2021-09-13
Add setVars to FS interface

danieljackins:
This PR extends the FS interface to include setVars to avoid TS/linter errors.

### DIFF
--- a/packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
+++ b/packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
@@ -59,7 +59,7 @@ const action: BrowserActionDefinition<Settings, typeof FullStory, Payload> = {
   },
   perform: (client, event) => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    let newTraits: any
+    let newTraits: any = {}
 
     if (event.payload.traits) {
       newTraits = Object.entries(event.payload.traits).reduce(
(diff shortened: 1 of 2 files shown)


packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
```diff
@@ -58,7 +58,9 @@ const action: BrowserActionDefinition<Settings, typeof FullStory, Payload> = {
     }
   },
   perform: (client, event) => {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
     let newTraits: any = {}
```
mshwery, line 62: Can this be typed? `Record<string, unknown>` ?

packages/browser-destinations/src/destinations/fullstory/viewedPage/index.ts
```diff
@@ -2,6 +2,14 @@ import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import * as FullStory from '@fullstory/browser'
+declare global {
+  interface Window {
+    // setVars is not available on the FS client yet.
+    FS: {
```
mshwery, line 8: Are we able to get declaration merging on the FS namespace itself, without having to add it to the `global.Window`?

mshwery (approved):
looks good. left a couple comments!

packages/browser-destinations/src/destinations/fullstory/viewedPage/index.ts
danieljackins: What would that look like? `setVars` doesn't exist on the client, or do you mean something else?

mshwery: I mean by using declaration merging (like i mentioned in your original PR) on the FS types, instead of the window object.

PR #163, opened 2021-09-13
add choices to interface for InputField and enforce as enum

mshwery:
We have choices exposed for global/auth settings, but forgot to expose it for action fields. This PR fixes that!

We also will need to hook it up in Control Plane:
https://github.com/segmentio/control-plane/blob/250b8c149fcce3a3caf2e9142debd2a3a05a4528/src/resources/destination-metadata/actions/action.ts#L29-L44

And in the UI:
https://github.com/segmentio/app/blob/c00d71b09c716ad98676918d1b7493d7bcf6baf0/client/containers/DestinationActions/components/RichInput/RichInput.tsx#L283-L293

### DIFF
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -142,6 +142,18 @@ export default class Push extends Command {
             )
           }
 
+          let choices: null | { label: string; value: string } = null
+          if (Array.isArray(field.choices) && field.choices.length > 0) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+            choices = field.choices.map((choice: string | { label: string; value: string }) => {
+              if (typeof choice === 'string') {
+                return { label: choice, value: choice }
+              }
+
+              return choice
+            })
+          }
+
           return {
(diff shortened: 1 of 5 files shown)


tysonmote (approved):

packages/cli/src/commands/push.ts
```diff
@@ -142,6 +142,18 @@ export default class Push extends Command {
...
+          if (Array.isArray(field.choices) && field.choices.length > 0) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+            choices = field.choices.map((choice: string | { label: string; value: string }) => {
+              if (typeof choice === 'string') {
+                return { label: choice, value: choice }
+              }
+
+              return choice
+            })
+          }
```
mshwery, line 155: This is to ensure it's always an array of items like `{ label, value }` even if you've specified the choices using the shorthand `string[]` style.

lcampos (approved):

danielstjules (approved):

juliofarah (approved):

PR #165, opened 2021-09-14
Fix local development

danieljackins:
Local dev was broken after versioning was introduced, as it was looking for something like fullstory.js rather than [hash].js. This PR adds a quick fix (live reloading still won't work properly with this fix)

### DIFF
--- a/packages/browser-destinations/package.json
+++ b/packages/browser-destinations/package.json
@@ -24,7 +24,7 @@
     "prepublishOnly": "yarn build",
     "test": "jest",
     "typecheck": "tsc --noEmit",
-    "dev": "concurrently \"webpack serve\" \"webpack -c webpack.config.js --watch\""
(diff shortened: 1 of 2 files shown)


nettofarah (approved):

packages/browser-destinations/package.json
```diff
@@ -24,7 +24,7 @@
     "prepublishOnly": "yarn build",
     "test": "jest",
     "typecheck": "tsc --noEmit",
-    "dev": "concurrently \"webpack serve\" \"webpack -c webpack.config.js --watch\""
+    "dev": "yarn postclean && yarn build && concurrently \"webpack serve\" \"webpack -c webpack.config.js --watch\""
```
mshwery, line 27: Why not switch local dev mode's file name in webpack? That probably would make it work like it did before during local, and then the production mode would have been versioned?

Not opposed to this solution either  – was just curious as what made you land on this direction instead?

PR #167, opened 2021-09-14
Add fullstory production id

danieljackins:
I had initially registered separate stage and production versions of fullstory instead of registering prod and then syncing with sprout. This changes it to the proper production ID

### DIFF
--- a/packages/browser-destinations/src/destinations/index.ts
+++ b/packages/browser-destinations/src/destinations/index.ts
@@ -30,3 +30,4 @@ register('5f7dd6d21ad74f3842b1fc47', './amplitude-plugins')
 register('60fb01aec459242d3b6f20c1', './braze')
 register('60f9d0d048950c356be2e4da', './braze-cloud-plugins')
 register('6138f272036e7c6650387fa4', './fullstory')
+register('6141153ee7500f15d3838703', './fullstory') // fullstory-prod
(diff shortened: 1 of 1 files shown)


packages/browser-destinations/src/destinations/index.ts
```diff
@@ -30,3 +30,4 @@ register('5f7dd6d21ad74f3842b1fc47', './amplitude-plugins')
 register('60fb01aec459242d3b6f20c1', './braze')
 register('60f9d0d048950c356be2e4da', './braze-cloud-plugins')
 register('6138f272036e7c6650387fa4', './fullstory')
+register('6141153ee7500f15d3838703', './fullstory') // fullstory-prod
```
juliofarah, line 33: what's the difference between these two?

danieljackins: the other is for stage (i couldnt push there without adding this either)

mshwery: This is going to cause problems. IDs need to match across environments.

packages/browser-destinations/src/destinations/index.ts
```diff
@@ -29,4 +29,4 @@ function register(id: MetadataId, destinationPath: string) {
 register('5f7dd6d21ad74f3842b1fc47', './amplitude-plugins')
 register('60fb01aec459242d3b6f20c1', './braze')
 register('60f9d0d048950c356be2e4da', './braze-cloud-plugins')
-register('6138f272036e7c6650387fa4', './fullstory')
+register('6141153ee7500f15d3838703', './fullstory')
```
mshwery, line 32: Swweeeet. You'll likely need to sync destinations (if you havent already) from prod → stage with sprout. You can follow the commands in the [Actions Interface doc](https://paper.dropbox.com/doc/Destination-Actions-Interface--BS2PEtlubTnMxHOLNFmgn6hVAg-BS1RKGj9VKTcjqLrCSuBV#:uid=595103768835854214449522&h2=Prereq-2.-Sync-production-dest).

mshwery (approved):

PR #168, opened 2021-09-15
Test mapping support for array of object

lcampos:


packages/core/src/mapping-kit/__tests__/index.test.ts
```diff
@@ -215,3 +215,43 @@ describe('remove undefined values in objects', () => {
...
+      }
+    ]
+    // @ts-ignore
+    expect(transform(mapping, { foo: [{ manchu: 'bar' }, { fighters: 'dave' }] })).toEqual([
+      { one: 'bar', two: 'dave' }
+    ])
+  })
+
+  test('multiple objects in array field with field specific mappings', () => {
+    const mapping = { '@path': '$.foo.baz..num' }
```
lcampos, line 246: This scenario is a gap based on the syntax support in the original mapping-kit. I think we do want to add support for it now that we'll have the individual key-value pair UI in place.

@mshwery @dlasky any objection with me adding this to https://github.com/segmentio/action-destinations/blob/main/packages/core/src/get.ts ?

dlasky: no objection, do we want to support * as well from the json path spec?

one super minor nit, when i read this i thought that `num` was part of the proposed spec, so maybe use a different name JUST in this test (price maybe?)

lcampos: Specific to supporting wildcards `*`, if what we want is to navigate arrays I would propose we keep following GJSON syntax and use the `#` character since the `*` one has different usage https://github.com/tidwall/gjson/blob/master/SYNTAX.md#path-structure

And I will definitely change the test names I have here 👍

tysonmote: Following the gjson syntax would be ideal, yeah. One of these days we'll need to either get a Go/JS compatibility test suite going. Or compile the Go version to WASM! 😅

mericsson:
@lcampos should we close this?

PR #176, opened 2021-09-17
Fix default values for tracking page calls

danieljackins:


juliofarah (approved):

packages/browser-destinations/src/destinations/fullstory/index.ts
```diff
@@ -23,7 +23,7 @@ export const destination: BrowserDestinationDefinition<Settings, typeof FullStor
       name: 'Track Page Events',
       subscribe: 'type = "page"',
       partnerAction: 'trackEvent',
```
mshwery, line 25: Should this be `viewedPage`?

PR #180, opened 2021-09-17
Added new field send to gate sending requests to Twilio and Sendgrid

vikramkumar19:
This PR is adds a send field to messaging DA to gate sending only test feature until pilot


packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -146,8 +146,18 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'object',
       required: false
     },
+    send: {
+      label: 'Send Message',
+      description: 'Send Message can have true/false based on to send message or not',
+      type: 'boolean',
+      required: false,
+      default: { '@template': false }
```
mshwery, line 154:
```suggestion
      default: false
```

Keep in mind `default` are only applied in the UI upon creating a new action subscription. If you are using a separate UI or if you are programmatically creating the customer subscriptions + mappings, these defaults aren't automatically applied anywhere. They are more of an "initial value" than anything tbh.

They are not used at runtime.

vikramkumar19: Thank for the info. I do have these changes done programmatically to have the values set. Currently do not have any customers using these actions yet. However, Will like to know if we have any documentation to introduce incremental changes to the field while customers are subscribed. You did mention we could use a script for changes like these.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -146,8 +146,18 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'object',
       required: false
     },
+    send: {
+      label: 'Send Message',
+      description: 'Send Message can have true/false based on to send message or not',
```
mshwery, line 151: non-blocking suggestion
```suggestion
      label: 'Send Message?',
      description: 'Whether or not the message should actually get sent.',
```

mshwery (approved):

mshwery:
> LGTM. I'm guessing we can't place this behind a feature flag though?

Correct. This is because the actions repo is public and accepts public/external contributions. We don't have any internal feature flagging capabilities exposed (yet).

PR #181, opened 2021-09-20
[cli] Don't push null choices

gpsamson:
This pull request avoids pushing a null value if there is no choices defined for an action's field. CPS throws an error if choices is defined as anything but an array.

From Matt: 
image


gpsamson:
Closing in favor of a control-plane fix to allow null values.

PR #188, opened 2021-09-22
Batch field

mshwery:
This PR introduces a new action field for actions that support batching. We'll automatically inject a boolean for customers to control whether or not to use batching when it's supported.

### DIFF
--- a/packages/cli/src/commands/push.ts
+++ b/packages/cli/src/commands/push.ts
@@ -176,7 +176,7 @@ export default class Push extends Command {
             fieldKey: 'enable_batching',
             type: 'boolean',
             label: 'Enable Batching?',
-            description: 'When enabled, Segment will try to batch events.',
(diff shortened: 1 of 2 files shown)


tysonmote (approved):

gpsamson (approved):

packages/cli/src/commands/push.ts
```diff
@@ -170,6 +170,21 @@ export default class Push extends Command {
           }
         })
 
+        // Automatically include a field for customers to control batching behavior, when supported
+        if (typeof action.performBatch === 'function') {
+          fields.push({
+            fieldKey: 'enable_batching',
```
lcampos, line 176: Not immediately needed but we should add some checks in case someone decides to create their own `enable_batching` field which would clash with this one.

lcampos (approved):

PR #194, opened 2021-09-24
[Amplitude] Remove @if directive from session_id reading

juliofarah:
As we're ready migrating users from the old session_id format (`integrations.Amplitude.session_id` => `integrations.Actions Amplitude.session_Id`), this PR removes the usage of the `@if` directive used during the migration.


mshwery (approved):
nooice

PR #195, opened 2021-09-24
[Customer.io] Fix use of timestamps in bad test

gpsamson:
This pull request fixes a test where the `last_used` request property was being compared against the wrong value. Prior to this change, the test assumed the `last_used` field was set to default to `$.context.lastUsed` and not `$.timestamp`.

`last_used` field definition:
```
    last_used: {
      label: 'Last Used',
      description: 'Timestamp for when the device was last used. Default is current date and time.',
      type: 'string',
      default: {
        '@path': '$.timestamp'
      }
    },
```


gpsamson:
@mike-engel Would you mind reviewing this as well? It shouldn't cause conflict with https://github.com/segmentio/action-destinations/pull/187.

mshwery (approved):
lgtm though you may be able to get rid of the mock Segment event entirely if you can pass in all the required `mappings` as literals instead of mapping-kit directives. You could remove the `useDefaultMappings` in that case.

PR #199, opened 2021-09-28
Add `fromDomain` setting to Sendgrid

vadimdemedes:
Needed to store `fromDomain` email setting value for https://segment.atlassian.net/browse/GROW-264.

### DIFF
--- a/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
@@ -90,8 +90,7 @@ const action: ActionDefinition<Settings, Payload> = {
     fromDomain: {
       label: 'From Domain',
       description: 'Verified domain in Sendgrid',
-      type: 'string',
-      required: true
+      type: 'string'
     },
     fromEmail: {
       label: 'From Email',
--- a/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
@@ -87,6 +87,12 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'Email to send to when testing',
       type: 'string'
     },
+    fromDomain: {
+      label: 'From Domain',
+      description: 'Verified domain in Sendgrid',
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -87,6 +87,12 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'Email to send to when testing',
       type: 'string'
     },
+    fromDomain: {
+      label: 'From Domain',
+      description: 'Verified domain in Sendgrid',
+      type: 'string',
+      required: true
```
vikramkumar19, line 94: Do we want to make this required? Since this will not be used to make a call to Sendgrid.

vikramkumar19 (approved):
LGTM! Left a comment on required field.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
vadimdemedes: Yeah, you are right, will mark it as not required.

vadimdemedes: Actually, we might want to follow what we did with `previewText`. It's also not used when sending an API call, but it's required for UI to properly work with email audience actions.

vikramkumar19: Gotcha! Just a note this will be a break fix and existing Destination Actions need to be run a migration script

vadimdemedes: I wouldn't bother with that, since we haven't launched and all of them are used for testing.

vadimdemedes: Let's actually mark it as not required for now so that we don't break the UI for anyone who might be working with it, and after I roll out UI changes for this feature I can make it required again.

vikramkumar19: sounds good!

PR #200, opened 2021-09-28
[LIBWEB-916] add userId and name to FQL IR

mshwery:
This enhances the destination-subscriptions intermediate representation (the thing we translate between UI <> FQL for displaying and parsing) to include a couple new top-level attributes: `name` and `userId`. This will allow users to construct subscriptions like `'type = "page" and name = "Homepage"'` (page calls with a given page name) or `'type = "track" and userId != null'` (only identified users)

### DIFF
--- a/packages/destination-subscriptions/src/__tests__/fql.test.ts
+++ b/packages/destination-subscriptions/src/__tests__/fql.test.ts
@@ -94,6 +94,33 @@ test('event = "Product Added"', () => {
 	})
 })
 
+test('userId != null', () => {
+	testFql('userId != null', {
+		type: 'group',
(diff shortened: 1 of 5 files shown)


gpsamson (approved):

gpsamson:
We could have supported this without any changes if we switched to `fql-wasm`, correct?

mshwery:
> We could have supported this without any changes if we switched to `fql-wasm`, correct?

Not entirely. We still need the IR to be updated to support the UI components. `fql-wasm` is a potential future though, but would require us to build components that construct raw fql strings instead of the IR.

PR #202, opened 2021-09-28
Hardcode datetime for snapshot test data

jinapark202:
This PR hardcodes datetime for snapshot test data to make it deterministic.

### DIFF
--- a/packages/destination-actions/src/lib/test-data.ts
+++ b/packages/destination-actions/src/lib/test-data.ts
@@ -11,7 +11,7 @@ function setTestData(seedName: string, type: string, fieldName?: string, format?
       val = chance.bool()
       break
     case 'datetime':
-      val = chance.date()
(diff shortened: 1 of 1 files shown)


gpsamson (approved):

packages/destination-actions/src/lib/test-data.ts
```diff
@@ -11,7 +11,7 @@ function setTestData(seedName: string, type: string, fieldName?: string, format?
       val = chance.bool()
       break
     case 'datetime':
-      val = chance.date()
+      val = new Date(2021, 1, 1)
```
mshwery, line 14: I think we want this to be a ISO string or a unix number, too.

gpsamson: Oof, miss on my end. Most actions will do something like `new Date(payload.field)` of `day.utc(payload.field)`

PR #207, opened 2021-09-29
Customerio action followups

gpsamson:
This is a duplicate of https://github.com/segmentio/action-destinations/pull/187 which was reverted while we followed up on the best way to structure the device-related actions.

The Customer.io team will open another follow-up PR to test device-related changes but want to get these follow-ups merged and deployed beforehand.

PR #209, opened 2021-10-01
[Customer.io] Destination action followups

mike-engel:
> This is a self-owned PR alternate of #207 

Howdy y'all!

After #142 was merged and deployed, our QA and product teams had several things to tweak and add.

- Copy changes throughout
- Update scope of `createUpdateDevice` to include `Application Opened` event
- Add `deleteDevice` action
- Add `timestamp` and `convertTimestamp` to the `trackEvent` action
- Update how `Account Region` is displayed, so that the picker and the set value are the same to the user


packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
```diff
@@ -35,32 +47,53 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    },
     data: {
-      label: 'Data',
-      description: 'Custom data to include with the event.',
+      label: 'Event Attributes',
+      description: 'Optional data to include with the event.',
       type: 'object',
       default: {
         '@path': '$.properties'
       }
```
mshwery, line 70: If it's optional, do we want to include it by default?

packages/destination-actions/src/destinations/customerio/trackApplicationInstalledOrOpened/index.ts
```diff
@@ -5,13 +5,13 @@ import type { Payload } from './generated-types'
 import { trackApiEndpoint } from '../utils'
 
 const action: ActionDefinition<Settings, Payload> = {
-  title: 'Create or Update Device',
-  description: "Update a person's device in Customer.io or create it if it doesn't exist.",
+  title: 'Track Application Installed or Application Opened Event',
+  description: `Track an "Application Installed" or "Application Installed" event to create or update a person's device.`,
```
mshwery, line 9: >Track an "Application Installed" or "Application Installed" event

Copypasta?

packages/destination-actions/src/destinations/customerio/index.ts
```diff
@@ -66,16 +65,28 @@ const destination: DestinationDefinition<Settings> = {
...
-      partnerAction: 'createUpdateDevice',
-      mapping: defaultValues(createUpdateDevice.fields)
+      name: 'Track Application Installed or Application Opened Event',
+      subscribe: 'event = "Application Installed" or event = "Application Opened"',
+      partnerAction: 'trackApplicationInstalledOrOpened',
+      mapping: defaultValues(trackApplicationInstalledOrOpened.fields)
+    },
+    {
+      name: 'Track Application Uninstalled Event',
+      subscribe: 'type = "track" and event = "Application Uninstalled"',
```
mshwery, line 75: iirc you don't even need the `type = "track"` part because you wouldn't get an event with that key for other types of Segment events, but you can leave it in if you want :)

mshwery (approved):
Seems legit! Nice changes. Left a couple non-blocking comments. :)

PR #214, opened 2021-10-02
Tiktok conversions action destination

coreyching:
Opening a PR to add an actions-destination with TikTok Events API.

[TikTok Events API Documentation](https://ads.tiktok.com/marketing_api/docs?id=1701890979375106
)

The Events API enable customers to power TikTok solutions like dynamic product ads, custom targeting, campaign optimization and attribution.

The events API is a single endpoint that accepts several different supported web events (see documentation for more details).

As discussed with @mshwery - The TikTok Web Events API has the same requirements across events and is a single endpoint. Thus, **I built one single action called reportWebEvent** to power different events. To map Segment related events to Tiktok, I built 5 presets that take care of mapping names from Segment Spec named events. However, customers can still create their own mappings if desired. 

PII data that is sent in the context object has been hashed using sha-256.

There are additional events in Tiktok that do not map to a specific Segment Event. If customers would like to add those, they are able to create their own subscription that can specify the event name they'd like.


packages/browser-destinations/src/destinations/fullstory/viewedPage/index.ts
```diff
@@ -44,8 +44,8 @@ const action: BrowserActionDefinition<Settings, typeof FullStory, Payload> = {
   perform: (_, event) => {
     if (event.payload.pageName) {
       window.FS.setVars('page', { pageName: event.payload.pageName, ...event.payload.properties })
-    } else {
-      window.FS.setVars('page', event.payload.properties || {})
+    } else if (event.payload.properties) {
```
mshwery, line 47: Not sure why there are extraneous changes in your PR. You may want to merge latest `main` (or rebase on top of it).

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { formatEmail, formatPhone, formatUserId } from './formatter'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Report Web Event',
```
mshwery, line 7: Love that you are using TikTok's own phrasing/terminology from their API docs!

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  description: '',
```
mshwery, line 8: Definitely add a description when you get a chance :)

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  fields: {
+    event: {
+      label: 'Event Name',
+      type: 'string',
+      description: 'Conversion event name. Please refer to Tiktok "Web Event" section for accepted event names.',
```
mshwery, line 13: Tip: you can use markdown links in the `description` so users can navigate directly to the right place

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
...
+      }
+    },
+    timestamp: {
+      label: 'Event Timestamp',
+      type: 'string',
+      description: 'Timestamp that the event took place. Timestamp with ISO 8601 format.'
+    },
+    type: {
+      label: 'Event Type',
+      description: 'Override event type. Ex. "page".',
```
mshwery, line 33: Is this limited to a subset of allowed values?

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
...
+          then: { '@path': '$.userId' },
+          else: { '@path': '$.anonymousId' }
+        }
+      }
+    }
+  },
+  perform: (request, { payload, settings }) => {
+    const eventId = payload.event_id
+      ? payload.event_id.toString() + '_' + (Math.random() + 1).toString(36).substring(7)
+      : ''
```
mshwery, line 89: This does take away a little control from customers, because they can't guarantee that providing a deterministic `event_id` value is indeed deterministic (because we are appending values to it).

But I understand why you did this. @segmentio/build-experience-team we should consider providing another unique identifier – something like the `subscription.id` + `event.messageId` would be stable.

mshwery: To make it available for users as an event variable would require defining a "special" variable that isn't found in the Segment event, but is guaranteed to exist at runtime.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
...
+      hashedPhoneNumber: formatPhone(payload.phone_number)
+    }
+
+    // Request to tiktok Events Web API
+    return request('https://business-api.tiktok.com/open_api/v1.2/pixel/track/', {
+      method: 'post',
+      headers: {
+        'Access-Token': settings.accessToken,
+        'content-type': 'application/json'
+      },
```
mshwery, line 103: If it makes sense, you can apply this `Access-Token` header for all requests made by this `request` client by defining an `extendRequest` function at the top level of the destination definition: 
https://paper.dropbox.com/doc/Destination-Actions-Interface--BTpQVSaMjnbLjoki9NK7One7Ag-BS1RKGj9VKTcjqLrCSuBV#:uid=730817748397331546138957&h2=HTTP-Requests

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -0,0 +1,121 @@
...
+      hashedEmail: formatEmail(payload.email),
+      hashedPhoneNumber: formatPhone(payload.phone_number)
+    }
+
+    // Request to tiktok Events Web API
+    return request('https://business-api.tiktok.com/open_api/v1.2/pixel/track/', {
+      method: 'post',
+      headers: {
+        'Access-Token': settings.accessToken,
+        'content-type': 'application/json'
```
mshwery, line 102: When you use the `json` shorthand (as you did below) you don't need to explicitly set the content-type header as it is done for you.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/formatter.ts
```diff
@@ -0,0 +1,40 @@
+import { createHash } from 'crypto'
```
mshwery, line 1: @segmentio/build-experience-team this may be something we should provide OOTB for destination authors.

I'm a bit wary of authors relying on direct `import` statements, because the more packages that are imported, the less control we have over the type of environment actions can run in.

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -0,0 +1,119 @@
...
+      // If you do not have a way to validate the authentication fields safely,
+      // you can remove the `testAuthentication` function, though discouraged.
+      return request('https://business-api.tiktok.com/open_api/v1.2/oauth2/advertiser/get/', {
+        json: {
+          access_token: settings.accessToken,
+          app_id: settings.appId,
+          secret: settings.secretKey
+        }
+      })
+    }
```
mshwery, line 111: Niiiiiice

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  },
+  {
+    name: 'Order Completed',
+    subscribe: 'event = "Order Completed',
+    partnerAction: 'reportWebEvent',
+    mapping: {
+      ...defaultValues(reportWebEvent.fields),
+      event: {
+        '@template': 'PlaceOrder'
+      }
```
mshwery, line 61: btw, if you are using a literal value (no event variables, no curlies) you can skip the `@template` directive altogether!
```suggestion
      event: 'PlaceOrder'
```

mshwery (approved):
Looks great! I left a few minor comments, mostly suggestions you can take 'em or leave 'em.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
coreyching: ah good to know thanks. i'll remove here since it is implicit. thanks!

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
coreyching: Ooh that makes sense. thanks will update!

packages/browser-destinations/src/destinations/fullstory/viewedPage/index.ts
coreyching: updated!

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
coreyching: Done.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
coreyching: added hyperlink

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
coreyching: fixed description. override was not accurate.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
coreyching: Ah I see. Thanks for flagging for future. Can update if the new variable is provided.

dlasky: is this just to have the character counts match?

PR #216, opened 2021-10-04
add mixpanel action destinations

Alex-Kim-mp:


packages/destination-actions/src/destinations/mixpanel/index.ts
```diff
@@ -0,0 +1,94 @@
...
+        required: false
+      }
+    },
+    testAuthentication: (request, { settings }) => {
+      return request(`https://mixpanel.com/app/api/validate-project-token/`, {
+        method: 'post',
+        body: JSON.stringify({
+          api_secret: settings.apiSecret,
+          project_token: settings.projectToken
+        })
```
mshwery, line 81: Here you can skip the `JSON.stringify` if you use the shorthand `json` property instead of `body`. This has the added benefit of also automatically including the correct content-type header:
```suggestion
        json: {
          api_secret: settings.apiSecret,
          project_token: settings.projectToken
        }
```

packages/destination-actions/src/destinations/mixpanel/alias/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      description: 'New alias to merged with the previous id. Each alias can only map to one previous id',
+      default: {
+        '@if': {
+          exists: { '@path': '$.userId' },
+          then: { '@path': '$.userId' },
+          else: { '@path': '$.anonymousId' }
+        }
+      }
+    },
+    previous_id: {
```
mshwery, line 22: Should this field be required? If not, you may get requests that have no `payload.previousId` and based on the `perform` function below those requests will go all the way to mixpanel.

packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
```diff
@@ -0,0 +1,55 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Group Identify User',
+  description: 'Updates or adds properties to a group profile. The profile is created if it does not exist.',
+  defaultSubscription: 'type = "group"',
+  fields: {
```
mshwery, line 9: I noticed there are no required fields here, is that intentional/ok?

packages/destination-actions/src/destinations/mixpanel/trackEvent/index.ts
```diff
@@ -0,0 +1,369 @@
...
+        utm_medium: utm.utm_medium,
+        utm_source: utm.utm_source,
+        utm_term: utm.utm_term,
+        ...payload.event_properties
+      }
+    }
+
+    if (!settings.apiSecret) {
+      throw new Error('Missing api secret')
+    }
```
mshwery, line 358: 2 things: 
1. why not do this at the very top?
2. You should use the `IntegrationError` which lets you provide a `code` and `status` number used in the http response. You can import this from `@segment/actions-core`

mshwery (approved):
Looks good! 

I left a few non-blocking comments that I would recommend attending to in a follow-up PR. I noticed quite a few of the action field definitions don't have required fields and wanted to make sure that was intentional!

We'll get this first rev shipped privately so you can test it e2e :)

PR #218, opened 2021-10-04
invert url logic so anon goes to correct url always

mshwery:
If `userId` is getting sent as `null` (as it can via ajs) our flexible coercion/validation rules turn `null` into `''` for string fields. Unfortunately that means that the logic here doesn't work 100% of the time because we are checking explicitly that `payload.id === undefined` not `''`.

I've inverted the logic that constructs the anonymous vs identified endpoints (and request `body`) so the logic is a bit more resilient. Now, only when a valid, _non-empty_ `id` exists do we attempt to send to the identified endpoint. All empty values (`null`, `undefined`, `''`) will instead send to the anonymous url.

cc @mike-engel

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
+++ b/packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
@@ -90,9 +90,12 @@ const action: ActionDefinition<Settings, Payload> = {
       data: payload.data,
       timestamp
     }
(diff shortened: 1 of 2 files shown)


mericsson (approved):

daltonscharff (approved):

mike-engel (commented):
Thanks for looking into this @mshwery! This looks great to me.

PR #225, opened 2021-10-05
Doc updates

markzegarelli:
First cut of public facing docs for creating an action-based destination.

Readme is largely untouched, but links to creation (new) and testing (existing) guides.

Also added a documentation template and instructions for submitting docs.

### DIFF
--- a/packages/docs/create.md
+++ b/packages/docs/create.md
@@ -181,7 +181,7 @@ For each action or authentication scheme, you define a collection of inputs and 
 
 The Segment CLI introspects field definitions when you run `./bin/run generate:types` to generate their TypeScript declarations. This ensures the `perform` function is strongly-typed.
 
-Define fields with a JSON scheme. If your editor or IDE provides good Intellisense and autocompletion, you should see the allowed properties.
+Define fields following the field schema If your editor or IDE provides good Intellisense and autocompletion, you should see the allowed properties.
 
 As mentioned above, the `perform` function contains the code that defines what the action does.
 


README.md
```diff
@@ -1,8 +1,15 @@
 # Action Destinations
 
-Action Destinations are a way to build streaming destinations on Segment.
+Action Destinations are a way to build streaming destinations on Segment. To begin, follow the instructions in Get Started below.
 
-## Local Development
+Fore more detailed instruction, see the following READMEs:
+
+- [Create a Destination Action](./packages/docs/create.md)
+- [Troubleshooting](./packages/docs/testing.md)
```
mshwery, line 8: Can we also move `/docs` out of `/packages` and at the top level of the repo?

README.md
```diff
@@ -22,7 +29,7 @@ yarn bootstrap
 yarn test
 ```
 
-## Actions CLI
+### Actions CLI
```
mshwery, line 32: I think this was intended to be a top-level heading (not hierarchically under "Get Started")

packages/docs/doc-template.md
```diff
@@ -0,0 +1,76 @@
+---
+# The end name should be similar to `Slack (Actions) Destination`
+title: <destination_name>
+hide-boilerplate: true
+hide-dossier: true
```
mshwery, line 5: Are we wanting to expose this as a PR template when creating a new destination?

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+The security of customers and partners is a top priority at Segment. Before you begin building, review the [Acceptable Use Policy](https://segment.com/legal/acceptable-use-policy/), and keep in mind:
+
+- Follow a secure software-development lifecycle, which enables you to both create code that is safe for Segment customers and their end users, and maintain and raise the security of that code over time.
+- If you or your code comes into contact with Segment customer or end-user data for any reason, protect it with commercially reasonable methods throughout the data lifecycle, including creating, handling, transporting, and destruction.
+- If you suspect a security event, incident, or breach related to this project, contact [Segment Security](mailto:security@segment.com) for assistance with your investigation and communications.
+- Practice modern and common-sense security for any scenario that is not explicitly stated.
+
+### Configure your development environment
+
+Fork the `segmentio/action-destinations` repository, connect to NPM and Yarn, and ensure a compatible version of Node is installed.
```
mshwery, line 20: Might be worth commenting that internal Segmenters can simply clone the `segmentio/action-destinations` repo itself (no fork required for internal folks)?

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+
+Run the test suite to ensure the environment is properly configured.
+
+```sh
+git clone https://github.com/segmentio/action-destinations.git
```
mshwery, line 25: This should be replaced with their fork

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+cd action-destinations
+npm login
+yarn login
+# Requires node 14.17, optionally: nvm use 14.17
+yarn --ignore-engines --ignore-optional
```
mshwery, line 30: I think we should remove `--ignore-engines`... Why is this currently here // cc @gpsamson ?

mshwery: Also, internal Segmenters shouldn't `--ignore-optional` or the CLI won't fully work for them.

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Create a destination
+
+Once you've configured your environment, you're ready to begin building your first destination. All commands, unless noted otherwise, should run from the root of the project folder. For example, `./action-destinations`
+
+> Run `./bin/run --help` at any time to see a list of available commands.
+
+### Scaffold the new destination
+
+To begin, run `./bin/run init` to scaffold the project's directory structure, and create a minimal implementation of the new destination. The initialization sets the following information:
```
mshwery, line 43: Might be worth calling out that `init` **does not register or deploy the integration.**

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+
+- Integration name
+- Integration slug
```
mshwery, line 46: May be worth describing what the slug is.

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+yarn --ignore-engines --ignore-optional
+yarn bootstrap
+yarn test
+```
+
+## Create a destination
+
+Once you've configured your environment, you're ready to begin building your first destination. All commands, unless noted otherwise, should run from the root of the project folder. For example, `./action-destinations`
+
+> Run `./bin/run --help` at any time to see a list of available commands.
```
mshwery, line 39: I think we auto-generate CLI docs too: https://github.com/segmentio/action-destinations/blob/main/packages/cli/README.md

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Scaffold the new destination
+
+To begin, run `./bin/run init` to scaffold the project's directory structure, and create a minimal implementation of the new destination. The initialization sets the following information:
+
+- Integration name
+- Integration slug
+- Authentication template (choose one of Custom Auth, Browser Destination, Basic Auth, OAuth2 Auth, or Minimal)
```
mshwery, line 47: Can we list "Browser Destination" as `(experimental)`

packages/docs/create.md
```diff
@@ -0,0 +1,224 @@
...
+
+After you've created the scaffold for the action, add logic that defines what the action does. Here, you'll define the fields that the action expects to receive, and write the code that performs the action.
+
+#### Action fields
+
+For each action or authentication scheme, you define a collection of inputs and `fields`. Input fields define what the user sees in the Action Editor within the Segment App. In an action, these fields accept input from the incoming Segment event.
+
+The Segment CLI introspects field definitions when you run `./bin/run generate:types` to generate their TypeScript declarations. This ensures the `perform` function is strongly-typed.
+
+Define fields with a JSON scheme. If your editor or IDE provides good Intellisense and autocompletion, you should see the allowed properties.
```
mshwery, line 184: It's not really JSON Schema exactly.

mshwery: It follows our field schema in the db, though there are some jsonschema-compatible parts

mshwery (approved):
Looks amazing. I left some non-blocking comments, but would definitely support getting this in early and following up with separate PRs.

I think we want to outline the entire interface like we do in our paper doc.

packages/docs/doc-template.md
markzegarelli: I don't think a PR template is the way to go here. We need them to actually create the markdown file, not describe the changes in the PR body. There are likely easier ways to do this though and I'm open to suggestions.

packages/docs/create.md
markzegarelli: I suggest we have a separate internal doc for internal users. The doc we have now will get messy if we try address both audiences in a single doc.

packages/docs/create.md
markzegarelli: Can users see the field schema? Or should we just remove anything about schema?

packages/docs/create.md
mshwery: Yeah, but some readers may not know the term "slug" – maybe describe that this is a url-and-human-friendly stable identifier for the destination, along with an example.

PR #228, opened 2021-10-05
validate settings before running actions

mshwery:
I realized that we only validate the settings during `testAuthentication`, but really we should validate them before we invoke actions, too. Otherwise we could end up with runtime errors or invalid settings in the action's `perform`.


dlasky (approved):

dlasky:
good catch!

lcampos (approved):

PR #231, opened 2021-10-07
Add Friendbuy trackCustomer action and profile browser plugin

dhull:


packages/browser-destinations/src/destinations/friendbuy-plugins/index.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Switch from unknown to the partner SDK client types
+export const destination: BrowserDestinationDefinition<Settings, unknown> = {
+  name: 'Friendbuy Plugins',
+  slug: 'friendbuy-plugins',
+  mode: 'device',
+
+  initialize: async ({ _settings, _analytics }, _deps) => {
+    return {}
+  },
```
dhull, line 15: I modeled this after `packages/browser-destinations/src/destinations/amplitude-plugins/index.ts`. I don't know if it should be doing any more.

packages/browser-destinations/src/destinations/friendbuy-plugins/profile/index.ts
```diff
@@ -0,0 +1,37 @@
...
+      }
+      const trackerJson = JSON.parse(friendbuyLocalStorageJson).tracker
+      if (!trackerJson) {
+        return
+      }
+      const profile = JSON.parse(trackerJson).tracker
+      if (!profile) {
+        return
+      }
+      context.updateEvent('integrations.Actions Friendbuy.profile', profile)
```
dhull, line 30: Is this the correct place to store my profile in the updated event? (This is copied from `segmentio-action-destinations/packages/browser-destinations/src/destinations/amplitude-plugins/sessionId/index.ts`.)

packages/browser-destinations/src/destinations/friendbuy-plugins/profile/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import type { Payload } from './generated-types'
+
+export const friendbuyLocalStorageKey = 'persist:friendbuy-msdk-06192019-root'
+
+// Change from unknown to the partner SDK types
+const action: BrowserActionDefinition<Settings, unknown, Payload> = {
+  title: 'Profile',
+  description: 'Attaches the Friendbuy profile to browser events',
+  platform: 'web',
+  defaultSubscription: 'type = "track" or type = "identify"',
```
dhull, line 12: What is the syntax for `defaultSubscription`?

packages/destination-actions/src/destinations/friendbuy/trackCustomer/__tests__/index.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    const name = 'John Doe'
+    const email = 'john.doe@example.com'
+    // const profile = 'JWT profile'
+    const event = createTestEvent({
+      userId,
+      traits: {
+        name,
+        email
+      },
+      // integrations: { 'Actions Friendbuy': { profile } },
```
dhull, line 32: Based on the `@path` `$.integrations.Actions Friendbuy.profile` it seems that my profile from my browser plugin should go here, but TypeScript says that `integrations.[integration]` should be a boolean. (This is from the `Integrations` type.)

This path is modeled after the `session_id` path in `packages/destination-actions/src/destinations/amplitude/event-schema.ts`, and does seem to match the browser-plugin code.

packages/destination-actions/src/destinations/friendbuy/index.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Friendbuy',
+  slug: 'friendbuy',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'custom',
+    fields: {
+      merchantId: {
```
dhull, line 22: I'm confused about whether I should be using camelCase or snake_case for these field names. I think I saw something that said that we should use whatever is standard for our language and it would be converted.

packages/destination-actions/src/destinations/friendbuy/trackCustomer/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      encodeURIComponent(
+        JSON.stringify({
+          customer: { id: data.payload.customerId, email: data.payload.email, name: data.payload.name }
+        })
+      )
+    )
+    const metadata = base64Encode(
+      JSON.stringify({
+        url: get(data, ['rawData', 'context', 'page', 'url']),
+        title: get(data, ['rawData', 'context', 'page', 'title'])
```
dhull, line 55: Is `rawData` correct here to get the context? I was going off the event generated by `createTestEvent`. This does not match the TypeScript `ExecuteInput` interface definition.

packages/browser-destinations/src/destinations/friendbuy-plugins/profile/index.ts
mshwery: Subscriptions use [FQL](https://segment.com/docs/config-api/fql/) but one thing to be mindful of here is that we currently don't support _grouping_ in this case.

Also, `defaultSubscription` is totally optional. It will autopopulate the subscription trigger when customers select this action.

packages/destination-actions/src/destinations/friendbuy/index.ts
mshwery: We don't convert anything automagically. It doesn't _really_ matter which casing you use, as long as the keys are unique. I would recommend that you are consistent with whichever choice you make!

packages/destination-actions/src/destinations/friendbuy/index.ts
```diff
@@ -0,0 +1,41 @@
...
+        label: 'Merchant ID',
+        description: 'Your Friendbuy Merchant ID.',
+        type: 'string',
+        required: true
+      }
+    },
+    testAuthentication: (_request, { settings }) => {
+      if (!isUuid(settings.merchantId)) {
```
mshwery, line 30: Looks like `uuid` is a `format` we may be able to support!

>"uuid": New in draft 2019-09 A Universally Unique Identifier as defined by RFC 4122. Example: 3e4666bf-d5e5-4aa7-b8ce-cefe41c7568a

If that's the case, you could add `format: 'uuid'` to the `merchantId` field and it would automatically get validated before invoking the `testAuthentication` method.

packages/destination-actions/src/destinations/friendbuy/trackCustomer/__tests__/index.test.ts
mshwery: Ah, the types are a bit wrong it seems. Technically you can put non boolean values in the integration object too. We can see about raising this issue in the analytics repo where those types are defined.

mshwery: Something to keep in mind (like in Amplitude's case), the enrichment only occurs when the customer has installed this destination on a browser source. In any other scenario, you may not have a value for `profile`.

packages/destination-actions/src/destinations/friendbuy/index.ts
```diff
@@ -0,0 +1,41 @@
...
+const uuidRegex = /^[0-9a-f]{8}-(?:[0-9a-f]{4}-){3}[0-9a-f]{12}$/
+
+function isUuid(s: string) {
+  return uuidRegex.test(s)
+}
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Friendbuy',
+  slug: 'friendbuy',
+  mode: 'cloud',
```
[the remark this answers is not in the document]
dhull: I answered this question in the Slack channel.

PR #232, opened 2021-10-07
CLI Improvements

gpsamson:
This pull request improves various aspects of the CLI:

- Internal commands are moved into a separate package/plugin called `cli-internal` and conditionally loaded by the `cli` package if the `control-plane-client` dependency is available. Tests for this package is skipped.
- Fixes and re-enables CLI tests


packages/cli/src/hooks/init.ts
```diff
@@ -0,0 +1,35 @@
...
+      type: 'user',
+      name: '@segment/cli-internal',
+      root: path
+    }
+  ])
+}
+
+async function isCpsSupported() {
+  try {
+    await import('@segment/control-plane-service-client')
```
mshwery, line 28: Can't you `require.resolve` rather than actually importing the module?

packages/cli/src/hooks/init.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const cpsSupport = await isCpsSupported()
+  if (!cpsSupport) {
+    return
+  }
+
+  const path = require.resolve('@segment/cli-internal')
+
+  // @ts-ignore loadPlugins is not recognized as a valid method
+  await config.loadPlugins(path, 'user', [
```
mshwery, line 17: Can we fix the linter warning in github?

packages/cli/package.json
```diff
@@ -80,8 +80,7 @@
     "tslib": "^2"
   },
   "optionalDependencies": {
-    "@segment/browser-destinations": "^3.11.1",
-    "@segment/control-plane-service-client": "github:segmentio/control-plane-service-js-client.git#master"
+    "@segment/browser-destinations": "^3.11.1"
```
mshwery, line 83: Does this still need to be optional given it's in the same repo?

packages/cli-internal/src/lib/prompt.ts
```diff
@@ -0,0 +1,44 @@
...
+  questions: prompts.PromptObject | Array<prompts.PromptObject>,
+  options: prompts.Options = {}
+): Promise<Answers> {
+  return prompts(questions, { onCancel, ...options }) as Promise<Answers>
+}
+
+/**
+ * Given a set of parsed flags and questions, only prompt for answers that are missing
+ */
+export async function autoPrompt<F extends flags.Output>(
```
mshwery, line 20: Isn't this used for other public commands?

PR #233, opened 2021-10-08
cli: validate command

mshwery:
This adds the beginnings of a `validate` command. We can do other local validations here to prevent common mistakes or ensure the definition is valid before a change gets merged (and later `push`ed into Segment which may encounter an error).

Here's the current output:
```sh
➜  action-destinations git:(main) ✗ ./bin/run validate
✔ Validating presets for 1plusX
✔ Validating presets for Actions Amplitude
✔ Validating presets for Braze Cloud Mode
✔ Validating presets for Customer.io
✔ Validating presets for Google Analytics 4
✔ Validating presets for Actions Google Enhanced Conversions
✖ Validating presets for Mixpanel (Actions):
  The preset "Page Calls" references a field "event_type" that the "trackEvent" action does not define.
  The preset "Screen Calls" references a field "event_type" that the "trackEvent" action does not define.
✔ Validating presets for Actions Personas Messaging Sendgrid
✔ Validating presets for Personas Messaging Twilio (Actions)
✔ Validating presets for Pipedrive
✔ Validating presets for Slack
✔ Validating presets for Twilio
✔ Validating presets for Webhook
✔ Validating presets for Braze Web Mode
✔ Validating presets for Fullstory (Actions)
    Error: One or more validation errors were found.
```

If we want to get nicer diffing/error output we could consider using a testing framework and run it programmatically via `validate`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,9 @@ jobs:
       - name: Lint
         run: yarn lint
 
+      - name: Validate
+        run: yarn validate
+
       - name: Test
         run: yarn test
 
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "bootstrap": "lerna bootstrap",
     "build": "./bin/run generate:types && lerna run build --stream --ignore @segment/actions-cli",
     "types": "./bin/run generate:types",
+    "validate": "./bin/run validate",
     "lint": "eslint '**/*.ts' --cache",
     "subscriptions": "yarn workspace @segment/destination-subscriptions",
     "test": "lerna run test --stream",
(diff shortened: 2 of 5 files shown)


jinapark202 (approved):
👏

mshwery:
> My first thought was oh we should also have an option to specify a destination, but that was quickly followed by: what if we also add that for `push` as well or really any of the destination specific commands (maybe using .env) so builders don't have to constantly pick their own destination.

100%

I had outlined somewhere that we should allow the command to also work with the current working directory as the destination entrypoint. So you could cd into your destination and run all commands from there and it would be scoped to that destination (tests, generate:types, generate:action, register, even init).

PR #242, opened 2021-10-11
[Customer.io] More minor copy tweaks

mike-engel:
This fixes a few minor things that we didn't catch when testing before:

- Make the use of periods on field descriptions consistent
- Fix a pluralization typo in the `createUpdatePerson` description
- Rename the old `trackApplicationInstalledOrOpened` test file to `createUpdateDevice`


mike-engel:
@mshwery whoops! Too early in the morning and I missed that. I also committed the generated types updates. I had expected the git hooks to do that, but I guess it only updates them?

PR #244, opened 2021-10-11
GROW-407 - block demo workspace email domains from execution

laurenfb:
Our demo-generating tool uses the email domains `gmailx`, `hotmailx`, `aolx`, and `yahoox`. Even though some of these are "real" domains, we don't want to risk sending email to them and compromising our Sendgrid reputation.

Tested locally and unit tested.

https://segment.atlassian.net/browse/GROW-407

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Refer [here](./docs/testing.md) for more information about testing your destinat
 
 ## Debugging
 
-Pass the node flag `--inspect` when you run the local server, and then you can attach a debugger from your IDE.
+Pass the Node flag `--inspect` when you run the local server, and then you can attach a debugger from your IDE. The `serve` command will pass any extra args/flags to the underlying Node process. 
 
 ### Configuring
 


README.md
```diff
@@ -76,6 +76,10 @@ If a CLI command fails to work properly, run the command with `DEBUG=*` at the b
 
 Refer [here](./docs/testing.md) for more information about testing your destination actions.
 
+## Debugging
```
vikramkumar19, line 79: Thanks for updating the docs 👍

vikramkumar19 (approved):
LGTM!

README.md
```diff
@@ -76,6 +76,10 @@ If a CLI command fails to work properly, run the command with `DEBUG=*` at the b
...
+
+Pass the node flag `--inspect` when you run the local server, and then you can attach a debugger from your IDE.
```
mshwery, line 81: Thanks for this! Maybe we can make it more obvious that any flag will work (not just `--inspect`)

```suggestion
Pass the Node flag `--inspect` when you run the local server, and then you can attach a debugger from your IDE. The `serve` command will pass any extra args/flags to the underlying Node process. 
```

mshwery (approved):

PR #250, opened 2021-10-12
Friendbuy browser destination

dhull:


packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
```diff
@@ -0,0 +1,149 @@
...
+      // - `revenue` is the sum of the costs of the items being ordered.
+      // - `subtotal` is `revenue` minus any discount.
+      // - `total` is `subtotal` plus tax and shipping.
+      label: 'Amount Source',
+      description: 'Source of purchase amount to send to Friendbuy.',
+      type: 'string',
+      required: true,
+      choices: amountSourceChoices as unknown as string[],
+      default: 'total'
+    },
```
dhull, line 37: My intention is to allow the merchant to configure which of the order fields (`revenue`, `subtotal`, or `total`) that they would like to use as the value of the track purchase call. Is this the best way to do this?

packages/browser-destinations/src/destinations/friendbuy/trackCustomer/index.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+function getName(payload: Payload): string | undefined {
+  // prettier-ignore
+  return (
+    payload.name                          ? payload.name :
+    payload.firstName && payload.lastName ? `${payload.firstName} ${payload.lastName}`
+    :                                       undefined
+  )
+}
```
mshwery, line 74: Something like this you would likely want customers to have control over. Rather than opaquely mapping it under the hood, we would encourage you to use a single field that represents `name` (which you already have).

If you truly need to handle fallbacks on behalf of users (taking customization control away from them), then you can do something like this imperatively in code, or you could keep customers in control by using mapping-kit's `@if` directive to achieve it.

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
mshwery: If you need to limit it to a predefined list of static choices, this is a great way to go ☝️ 

However, if the possible choices are unknown (or customer-specific) it's best to provide a default and then let them  fill it in however they prefer if they need to customize where it's coming from.

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  perform: (friendbuyAPI, data) => {
+    // console.log('trackPurchase.perform', JSON.stringify(data, null, 2))
+    const amount =
+      data.payload[data.payload.amountSource as AmountSource] ||
+      data.payload.total ||
+      data.payload.subtotal ||
+      data.payload.revenue
+    if (amount === undefined) {
+      return
+    }
```
mshwery, line 127: Why have 4 fields that all _might_ contain the value? Can you have a single field that the customer can use to say where it should be coming from in their events?

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    const products =
+      data.payload.products && data.payload.products.length > 0
```
mshwery, line 130: It looks like you've required the `products` field. Should that be optional?

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
dhull: I think I might have a basic misunderstanding what Segment does. My assumption was that if I'm a merchant using e-commerce platform X, then I could configure the Segment Source for X, which would then automatically generate and send the various Segment e-commerce Spec events, and that the Friendbuy Destination code would have the opportunity to be triggered by those events to transform and send our trackPurchase events to our back end.

I guess what you're saying, though, is that I can just have a single `amount` field, and maybe defined a default path of `$.properties.total`, but that merchants can override that if they want?

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
dhull: Yes, thanks, it should be optional.

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
mshwery: >which would then automatically generate and send the various Segment e-commerce Spec events

There might be some sources that do this, but that's an implementation detail. 

With Segment, customers will connect sources (could be a website, an app, a server) to destinations. When the source is a website, we'll automatically load any connected destination that has a JavaScript SDK + functionality defined on Segment's platform (like the destination you are building right now). 

As events are sent through the Segment's Analytics.js client, the client will fan out those events to destinations, which can respond to those events by enriching them with additional data, or can trigger behaviors in their own SDKs/APIs. 

In the case you described our mutual customer has Segment running on their website/app. When they connect the Friendbuy destination, Segment will automatically load the Friendbuy destination + initialize the Friendbuy SDK (as you are defining in your code here). That customer will get a configuration that works out of the box, based on the sane defaults you are defining here, but ultimately they are in control of:
- which events will trigger Friendbuy actions
- how event data will get transformed into the payload that Friendbuy's actions expect

If they have not been using the e-commerce spec exactly, or if they have some other constraints that cause some data to be defined differently than the defaults suggest, they can freely change every field in the UI to pull values from any part of a Segment event. Here's a sample from the UI of what customers can configure:
Screen Shot 2021-10-12 at 9 08 30 PM

This functionality gives a lot of freedom and transparency to how Segment and destinations map values from the event.

juliofarah:
@dhull if you need to test your changes manually, you should be able to do so following the steps documented on https://github.com/segmentio/action-destinations/blob/main/packages/browser-destinations/README.md#local-testing .

This will spin a dev server where you can select the destination you want to test, then you can tweak the subscriptions as you see fit and load the destination. Once loaded, you can fire events by clicking the `track` / `identify` buttons, or even use chrome's dev tools to fire events via command line (eg `windows.analytics.track('my-event')`)

packages/browser-destinations/src/destinations/friendbuy/trackPurchase/index.ts
dhull: I've converted this to use mapping-kit.

```typescript
  amount: {
    label: 'Amount Source',
    description: 'Purchase amount to be considered when evaluating reward rules.',
    type: 'number',
    required: true,
    default: {
      '@if': {
        exists: { '@path': '$.properties.total' },
        then: { '@path': '$.properties.total' },
        else: {
          '@if': {
            exists: { '@path': '$.properties.subtotal' },
            then: { '@path': '$.properties.subtotal' },
            else: { '@path': '$.properties.revenue' }
          }
        }
      }
    }
  },

```

mshwery: Ah, one note – we don't currently support any nesting in mapping-kit directives in the UI so this will probably break things.

I would recommend keeping defaults simple. Choose the default value that tends to make the most sense for customers. Where are they putting the amount? If they need to pull it from somewhere else, they can reconfigure the field themselves. 

cc @dlasky

dlasky: seems like an argument for splitting coalesce into a variadic method

packages/browser-destinations/src/destinations/friendbuy/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    name: 'Track Sign Up',
+    subscribe: trackSignUpDefaultSubscription,
+    partnerAction: 'trackSignUp',
+    mapping: defaultValues(trackSignUpFields)
+  }
+]
+
+export const destination: BrowserDestinationDefinition<Settings, FriendbuyAPI> = {
+  name: 'Friendbuy',
+  slug: 'friendbuy',
```
juliofarah, line 46: We already have a destination called `Friendbuy` so we're gonna need to change this.
We've been following this pattern:

```suggestion
  name: 'Friendbuy (Actions)',
  slug: 'actions-friendbuy',
```

How does it sound?

packages/browser-destinations/src/destinations/friendbuy/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    const friendbuyBaseHost = window.friendbuyBaseHost ?? 'fbot.me'
+
+    friendbuyAPI.merchantId = settings.merchantId
+    friendbuyAPI.push(['merchant', settings.merchantId])
+    await Promise.all([
+      dependencies.loadScript(`https://static.${friendbuyBaseHost}/friendbuy.js`),
+      dependencies.loadScript(`https://campaign.${friendbuyBaseHost}/${settings.merchantId}/campaigns.js`)
+    ])
+
+    return friendbuyAPI
```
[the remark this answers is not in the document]
dhull: They don't. (They do overwrite the friendbuyAPI's `push` method. Actually, I believe the `await` isn't necessary so I'll remove that.)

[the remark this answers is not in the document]
mshwery: Ok! So long as this `initialize` method returns the initialized friendbuy SDK you should be fine.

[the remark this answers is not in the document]
dhull: I'm assuming that the return value from `initialize` is an opaque value that is passed as the first argument to the `BrowserActionDefinition`'s `perform` function and that there aren't any other requirements on it.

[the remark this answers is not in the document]
mshwery: Correct! It can be whatever you want, but usually it's the SDK/client that you want to interact with.

dhull:
I've updated the PR in response to the review @mshwery's review.

mshwery (approved):

dhull:
@juliofarah Thanks, I will!

PR #252, opened 2021-10-12
change userId to distinctId for track

Alex-Kim-mp:


packages/destination-actions/src/destinations/mixpanel/trackEvent/index.ts
```diff
@@ -340,7 +343,7 @@ const action: ActionDefinition<Settings, Payload> = {
         $screen_density: payload.screen_density,
         $source: 'segment',
         $wifi_enabled: payload.wifi,
-        id: payload.user_id, // this is just to maintain backwards compatibility with the classic segment integration
+        id: payload.distinct_id, // this is just to maintain backwards compatibility with the classic segment integration
```
mshwery, line 346: I'm curious why we need this backwards compatibility? Is it something on the Mixpanel side?

PR #272, opened 2021-10-15
GDPR Deletion support for actions

dlasky:


packages/cli/src/lib/server.ts
```diff
@@ -71,6 +75,39 @@ server.on('error', (err: Error) => {
...
+
+    if (supportsDelete) {
+      app.post(
+        '/delete',
+        asyncHandler(async (req: express.Request, res: express.Response) => {
+          spinner.start(chalk`Handling ${process.env.DESTINATION}#delete request`)
+
+          try {
+            if (destination.onDelete) {
+              await destination.onDelete(req.body.payload ?? {}, req.body.settings ?? {})
```
dlasky, line 89: @jinapark202 @mshwery what do y'all think about the signature here? it feels a little weird to me tbh at least how its mapped to serve

packages/core/src/destination-kit/types.ts
```diff
@@ -150,3 +153,21 @@ export type Directive = IfDirective | TemplateDirective | PathDirective
...
+}
+
+/**
+ * A function to perform a deletion request for GDPR or PII related data
+ *
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Deletion<Settings, Payload, Return = any> = (
+  request: RequestClient,
+  data: ExecuteInput<Settings, Payload>
```
mshwery, line 172: Should this be:
```suggestion
  data: ExecuteInput<Settings, DeletionPayload>
```

packages/core/src/create-request-client.ts
```diff
@@ -4,6 +4,10 @@ import prepareResponse from './middleware/after-response/prepare-response'
 import createInstance, { AllRequestOptions, RequestOptions } from './request-client'
 import type { ModifiedResponse } from './types'
 
+export interface ResponseError extends Error {
+  status?: number
+}
```
mshwery, line 9: You might not need this type. The fetch client (if set to throw on non-2xx responses) will throw a `HTTPError` which is exported there. Can we use that instead?

packages/core/src/destination-kit/index.ts
```diff
@@ -194,6 +197,10 @@ export class Destination<Settings = JSONObject> {
     this.authentication = destination.authentication
     this.responses = []
 
+    if (this.definition.onDelete) {
+      this.onDelete = this._onDelete
```
mshwery, line 201: Will we need to `.bind(this)`?

packages/core/src/destination-kit/index.ts
```diff
@@ -360,14 +368,15 @@ export class Destination<Settings = JSONObject> {
...
+    } catch (err) {
+      const error = err as unknown as JSONObject
       results = [{ error }]
 
       if (error.name === 'AggregateAjvError' || error.name === 'ValidationError') {
         error.status = 400
       }
 
-      throw error
+      throw err
```
mshwery, line 379: For safety, shouldn't we `throw error`? I realize you are modifying the same object, but it might be clearer.

packages/core/src/destination-kit/index.ts
```diff
@@ -360,14 +368,15 @@ export class Destination<Settings = JSONObject> {
         // there should only be 1 item in the subscribedEvents array
         return await this.executeAction(actionSlug, { ...input, event: subscribedEvents[0] })
       }
-    } catch (error) {
+    } catch (err) {
+      const error = err as unknown as JSONObject
```
mshwery, line 372: Shouldn't this be cast to an error type? Or use `instanceof` checks to narrow the type to an error?

packages/core/src/destination-kit/index.ts
```diff
@@ -397,6 +406,32 @@ export class Destination<Settings = JSONObject> {
     return this.onSubscriptions(events, settings, options)
   }
 
+  /** Pass a single deletion event to the destination for execution
+   * note that this method is conditionally added if the destination supports it
+   */
+  public async _onDelete(event: SegmentEvent, settings: JSONObject): Promise<Result> {
```
mshwery, line 412: Should this be `private` given that we only expose it as a public method as `onDelete`?

packages/core/src/destination-kit/index.ts
```diff
@@ -397,6 +405,32 @@ export class Destination<Settings = JSONObject> {
...
+
+    const data: ExecuteInput<Settings, DeletionPayload> = { payload, settings: destinationSettings, auth }
+    const context: ExecuteInput<Settings, undefined> = { settings: destinationSettings, payload: undefined, auth }
+
+    this.validateSettings(destinationSettings)
+
+    const options = this.extendRequest?.(context) ?? {}
+    const requestClient = createRequestClient(options)
+
+    //TODO: retry logic
```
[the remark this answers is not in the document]
mshwery: Not sure I follow. Can you or @lcampos elaborate?

packages/cli/src/lib/server.ts
```diff
@@ -99,25 +135,29 @@ loadDestination(targetDirectory)
...
 
+    const deleteRoute = supportsDelete ? `POST http://localhost:${port}/delete` : ''
+
     server.listen(port, () => {
       logger.info(`Listening at http://localhost:${port} -> 
 ${Object.keys(def?.actions ?? {})
   .map((action) => `  POST http://localhost:${port}/${action}`)
-  .join('\n')}`)
+  .join('\n')}
+  ${deleteRoute}`)
```
mshwery, line 160: Curious what this looks like when you run serve. Got a screenshot/video?

packages/cli/src/lib/server.ts
```diff
@@ -71,6 +75,39 @@ server.on('error', (err: Error) => {
...
+            const error = err as ResponseError
+            let statusCode = error?.status ?? 500
+            let msg = error?.message
+
+            if (err instanceof HTTPError) {
+              statusCode = err?.response?.status ?? statusCode
+              msg = ((err.response as ModifiedResponse).data as string) ?? (err.response as ModifiedResponse).content
+            }
+
+            return res.status(statusCode).send(msg + '<br/>' + error.stack?.split('\n').join('<br/>'))
```
mshwery, line 106: Can we send json? or something a bit nicer than plain text?

mshwery:
```suggestion
            return res.status(statusCode).json({ error })
```

packages/cli/src/lib/server.ts
```diff
@@ -71,6 +75,39 @@ server.on('error', (err: Error) => {
...
+            spinner.succeed(chalk`${destination.name} delete completed`)
+            return res.status(200).json(debug)
+          } catch (err) {
+            spinner.fail()
+            const error = err as ResponseError
+            let statusCode = error?.status ?? 500
+            let msg = error?.message
+
+            if (err instanceof HTTPError) {
+              statusCode = err?.response?.status ?? statusCode
```
mshwery, line 102: If it's an HTTPError, isn't it guaranteed to exist?
```suggestion
              statusCode = err.response?.status ?? statusCode
```

mshwery (approved):
✨ 

left a few non-blocking comments/questions!

packages/core/src/destination-kit/index.ts
dlasky: I did that originally, apparently error doesnt overlap with JSONObject at all

packages/cli/src/lib/server.ts
dlasky: Screen Shot 2021-10-15 at 11 42 49 AM

dlasky: thinking we should maybe expose the other methods as well here just for completeness and maybe make 1 section for 'actions' and one for the 'helper' methods?

packages/core/src/destination-kit/index.ts
dlasky: we can, but `_onDelete` doesn't reference the class instances at all for anything so it shouldn't matter

packages/core/src/create-request-client.ts
dlasky: will check, was just trying to clean up all the red i saw from error types not being handled

packages/core/src/destination-kit/index.ts
mshwery: I'm confused why it would need to.

packages/cli/src/lib/server.ts
mshwery: 100%

`testAuthentication` should be testable, as well as dynamic fields.

mshwery: Can happen in another PR!

packages/cli/src/lib/server.ts
mshwery: up to you!

packages/core/src/destination-kit/index.ts
dlasky: i was just trying to suppress all the TS errors we had around... errors here.

packages/cli/src/lib/server.ts
```diff
@@ -71,6 +75,39 @@ server.on('error', (err: Error) => {
...
 loadDestination(targetDirectory)
   .then((def) => {
+    const destination = new Destination(def as CloudDestinationDefinition)
+    const supportsDelete = destination.onDelete
+
+    if (supportsDelete) {
+      app.post(
+        '/delete',
+        asyncHandler(async (req: express.Request, res: express.Response) => {
+          spinner.start(chalk`Handling ${process.env.DESTINATION}#delete request`)
```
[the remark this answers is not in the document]
dlasky: its copied from @jinapark202's code below, and it seems to work in practice so i suspect it does get overridden somewhere

PR #273, opened 2021-10-15
fix `yarn typecheck`

mshwery:
`yarn typecheck` has been broken since we introduced references/composite tsconfigs. Now it works!

### DIFF
--- a/packages/cli/src/lib/server.ts
+++ b/packages/cli/src/lib/server.ts
@@ -4,13 +4,16 @@ import { once } from 'lodash'
 import logger from './logger'
 import path from 'path'
 import { loadDestination } from './destinations'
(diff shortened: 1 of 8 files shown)


packages/cli/src/lib/server.ts
```diff
@@ -4,13 +4,16 @@ import { once } from 'lodash'
...
+  DestinationDefinition as CloudDestinationDefinition,
+  HTTPError,
+  ModifiedResponse
+} from '@segment/actions-core'
 import asyncHandler from './async-handler'
 import getExchanges from './summarize-http'
 import ora from 'ora'
 import chalk from 'chalk'
-import { HTTPError } from '@segment/actions-core/request-client'
-import type { ModifiedResponse } from '@segment/actions-core/types'
```
mshwery, line 13: Core doesn't export these paths.

PR #274, opened 2021-10-18
Upgrade jest

mshwery:
This PR upgrades to the latest version of Jest, which touts faster test initialization and environment improvements (uses the less expensive "node" env by default, used to be the expensive "jsdom").

This is a no-op PR in that it does not impact any production code.


mshwery:
> is the intercom delete intentional?

Yep:

> I also removed the Intercom destination because it was causing some weird test failures / hanging, and it's not a destination we've actually shipped anywhere.

PR #281, opened 2021-10-19
Remove empty values from fields when not matching their data types

mshwery:
This PR introduces some logic to _remove empty values like `null` `undefined` or `''`_ when the field's type doesn't support them. This is because Ajv coerces values to match the allowed data type of a field, but in some cases cannot coerce values in a sensible way. 

We need to decide if it makes more sense to reject messages because of uncoercible empty values, or to strip them out. Optional fields would pass when they get stripped out, while required fields would obviously fail.

## Testing

### DIFF
--- a/packages/core/src/remove-empty-values.ts
+++ b/packages/core/src/remove-empty-values.ts
@@ -2,7 +2,7 @@ import type { JSONSchema4 } from 'json-schema'
 import { isArray, isObject } from './real-type-of'
 import { arrify } from './arrify'
 
-export function removeEmptyValues(obj: unknown, schema: JSONSchema4) {
+export function removeEmptyValues(obj: unknown, schema: JSONSchema4 = {}): unknown {
   if (isArray(obj)) {
     return obj.filter((item) => removeEmptyValues(item, schema) !== undefined)
   }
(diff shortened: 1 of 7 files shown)


packages/core/src/remove-empty-values.ts
```diff
@@ -0,0 +1,40 @@
+import type { JSONSchema4 } from 'json-schema'
```
dlasky, line 1: i know its redundant with the other tests but I would feel more confident modifying this in the future if it had tests explicitly for this method. non blocking

packages/core/src/remove-empty-values.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    return newObj
+  }
+
+  const schemaType = arrify(schema.type)
+
+  // Only remove `null` when the schema doesn't allow 'null'
+  // because it may not be coercible to a valid value in Ajv
+  if (obj === 'null' && !schemaType.includes('null')) {
```
gpsamson, line 29: Everything else looks good but the `'null'` string is throwing me off. Is this intentional? I know we have a test for null values specifically but maybe it's stripped by a different code path?

PR #283, opened 2021-10-19
use native globals, specify mainFields

mshwery:
This PR focuses on optimizing our webpack bundles.

| How it started | How it's going |
|---|---|
|Screen Shot 2021-10-19 at 4 05 17 PM|Screen Shot 2021-10-19 at 4 05 36 PM|

## Testing

Tested locally with the webpack bundle analyzer

### DIFF
--- a/packages/browser-destinations/package.json
+++ b/packages/browser-destinations/package.json
@@ -73,6 +73,9 @@
     "transformIgnorePatterns": [
       "/node_modules/(?!@segment/analytics-next).+\\.js$"
     ],
+    "setupFilesAfterEnv": [
+      "<rootDir>/test/setup-after-env.ts"
+    ],
     "forceExit": true
   }
 }
--- a/packages/browser-destinations/package.json
+++ b/packages/browser-destinations/package.json
@@ -24,7 +24,7 @@
     "prepublishOnly": "yarn build",
     "test": "jest",
     "typecheck": "tsc -p tsconfig.build.json --noEmit",
-    "dev": "NODE_ENV=development concurrently \"webpack serve\" \"webpack -c webpack.config.js --watch\""
+    "dev": "NODE_ENV=development concurrently \"webpack serve --open\" \"webpack -c webpack.config.js --watch\""
   },
   "dependencies": {
     "@braze/web-sdk": "^3.3.0",
--- a/packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
+++ b/packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
@@ -1,11 +1,11 @@
 import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import * as FullStory from '@fullstory/browser'
-import { camelCase } from 'lodash'
+import type { FS } from '../types'
+import camelCase from 'lodash/camelCase'
 
 // Change from unknown to the partner SDK types
-const action: BrowserActionDefinition<Settings, typeof FullStory, Payload> = {
+const action: BrowserActionDefinition<Settings, FS, Payload> = {
   title: 'Identify User',
   description: 'Sets user identity variables',
   platform: 'web',
@@ -57,7 +57,7 @@ const action: BrowserActionDefinition<Settings, typeof FullStory, Payload> = {
       }
     }
   },
-  perform: (client, event) => {
+  perform: (FS, event) => {
     let newTraits: Record<string, unknown> = {}
 
     if (event.payload.traits) {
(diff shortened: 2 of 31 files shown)


packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
```diff
@@ -1,11 +1,11 @@
 import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import * as FullStory from '@fullstory/browser'
-import { camelCase } from 'lodash'
+import type { FS } from '../types'
+import camelCase from 'lodash/camelCase'
```
mshwery, line 5: Load only the lodash module we need.

packages/browser-destinations/src/destinations/fullstory/index.ts
```diff
@@ -50,7 +56,7 @@ export const destination: BrowserDestinationDefinition<Settings, typeof FullStor
     initScript({ debug: settings.debug, org: settings.orgId })
     await dependencies.loadScript('https://edge.fullstory.com/s/fs.js')
     await dependencies.resolveWhen(() => Object.prototype.hasOwnProperty.call(window, 'FS'), 100)
-    return FullStory
+    return window.FS
```
mshwery, line 59: Use the same client everywhere.

packages/browser-destinations/src/destinations/fullstory/types.ts
```diff
@@ -0,0 +1,6 @@
+import * as FullStory from '@fullstory/browser'
+
+export type FS = typeof FullStory & {
+  // setVars is not available on the FS client yet.
+  setVars: (eventName: string, eventProperties: object) => {}
+}
```
mshwery, line 6: Trying to export just the types, instead of the entire thing (because it is not used anywhere).

packages/browser-destinations/webpack.config.js
```diff
@@ -76,6 +76,7 @@ module.exports = {
       path.resolve(__dirname, 'node_modules'),
       'node_modules'
     ],
+    mainFields: ['exports', 'module', 'browser', 'main'],
```
mshwery, line 79: This fixes the way `@segment/fql-ts` gets resolved and shaves off ~10kb gzipped from the core bundle.

Previously webpack would prefer `browser` over `module` which we don't really want in this case.

mshwery: Note: this works because libs like `cross-fetch` only have a `main` and `browser` specified, but no `module`

packages/core/package.json
```diff
@@ -12,13 +12,11 @@
   "exports": {
     ".": {
       "require": "./dist/cjs/index.js",
-      "import": "./dist/esm/index.js",
-      "default": "./dist/cjs/index.js"
```
mshwery, line 16: This fixes the way webpack was resolving the core runtime (previously used cjs)

packages/core/src/__tests__/request-client.test.ts
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/no-empty-function */
 import AbortController from 'abort-controller'
 import createTestServer from 'create-test-server'
-import { URLSearchParams } from 'url'
```
mshwery, line 4: Prefer the native module which is widely available:
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#browser_compatibility

We may need to polyfill this for i.e.? cc @nettofarah @pooyaj

packages/core/src/request-client.ts
```diff
@@ -1,6 +1,5 @@
 import AbortController from 'abort-controller'
 import { CustomError } from 'ts-custom-error'
-import { URL, URLSearchParams } from 'url'
```
mshwery, line 3: Prefer the native globals: 
https://developer.mozilla.org/en-US/docs/Web/API/URL#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#browser_compatibility

packages/core/tsconfig.build.json
```diff
@@ -2,6 +2,7 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "composite": true,
+    "module": "esnext",
```
mshwery, line 5: The tsc builds still output cjs + esm versions. But this will get the webpack `projectReferences` thing working better.

juliofarah (approved):

packages/core/src/__tests__/request-client.test.ts
nettofarah: I doubt actions will work on IE tbh. I think we can try this and see if someone complains

packages/browser-destinations/src/destinations/fullstory/identifyUser/index.ts
mshwery: cc @danieljackins

packages/browser-destinations/src/destinations/fullstory/index.ts
mshwery: cc @danieljackins

packages/browser-destinations/package.json
```diff
@@ -24,7 +24,7 @@
     "prepublishOnly": "yarn build",
     "test": "jest",
     "typecheck": "tsc -p tsconfig.build.json --noEmit",
-    "dev": "NODE_ENV=development concurrently \"webpack serve\" \"webpack -c webpack.config.js --watch\""
+    "dev": "NODE_ENV=development concurrently \"webpack serve --open\" \"webpack -c webpack.config.js --watch\""
```
mshwery, line 27: `--open` automatically opens the browser to the demo page.

PR #284, opened 2021-10-20
match fql behavior in IR validation

mshwery:
This PR (mostly automated indentation changes) makes the destination-subscriptions IR match the validation behavior of FQL more closely. Instead of conditionally check traits vs properties, we should _always use what the user specified_. You can use the IR to build conditions against properties, context, and traits explicitly.

## Testing

### DIFF
--- a/packages/core/src/__tests__/destination-kit.test.ts
+++ b/packages/core/src/__tests__/destination-kit.test.ts
@@ -174,9 +174,9 @@ describe('destination kit', () => {
     test('should succeed when property filtering is specified', async () => {
       const destinationTest = new Destination(destinationCustomAuth)
       const testEvent: SegmentEvent = {
-        properties: { field_one: 'test input' },
+        properties: { a: 'foo', field_one: 'test input' },
         traits: {
-          a: 'foo'
+          b: 'foo'
         },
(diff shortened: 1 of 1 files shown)


packages/destination-subscriptions/src/validate.ts
```diff
@@ -1,135 +1,105 @@
...
-	}
-
-	if (condition.type === 'userId') {
-		return validateValue(data.userId, condition.operator, condition.value)
-	}
-
-	if (condition.type === 'event-property') {
-		const properties = ['identify', 'group'].includes(data.type)
-			? data.traits
-			: data.properties
```
mshwery, line 62: This was removed.

packages/destination-subscriptions/src/validate.ts
```diff
@@ -1,135 +1,105 @@
...
-			get(properties, condition.name),
-			condition.operator,
-			condition.value
-		)
-	}
-
-	if (condition.type === 'event-trait') {
-		const properties = ['identify', 'group'].includes(data.type)
-			? data.traits
-			: data.properties
```
mshwery, line 74: This was removed.

PR #305, opened 2021-10-29
Reduce Canary Package Creation

daltonscharff:
This change separates the publish steps for canary and production packages while also reducing the number of canary packages created. **The process for deploying production packages with Github Actions has not changed. However, the process for deploying canary packages has.**

Previously, canary packages were created as soon as a code change within a PR occurred. This lead to a lot of failures, since our NPM token has not yet been updated, which also lead to a lot of emails. Additionally, this can cause a lot of unnecessary packages, as not every user will be interested in utilizing the canary packages.

Now, the process for creating a canary package is as simple as commenting `/publish-canary` on your PR. When the Github Action registers your comment, you will receive a reply saying "Publishing canary package..." followed by a message informing you that the canary package was created.

image

## Testing

Testing was conducted in the [auto-publish-demo](https://github.com/segmentio/auto-publish-demo) repository.


daltonscharff:
I would have liked to have the second automated comment return the response from Lerna, but it's a multiline string, which is hard to set as a variable and use in a different step.

.github/workflows/publish.yml
```diff
@@ -100,17 +89,4 @@ jobs:
           npm whoami
 
       - name: Publish production package
-        if: ${{ github.event_name == 'workflow_dispatch' || (github.event.action == 'closed' && github.event.pull_request.merged == true) }}
```
gpsamson, line 103: Shouldn't this around to still gate production packages?

mshwery: I don't think this `if` is needed now because these are now the only 2 conditions that trigger this workflow.

.github/workflows/publish.yml
```diff
@@ -100,17 +89,4 @@ jobs:
...
         run: lerna publish ${{ needs.get-version.outputs.version }} --conventional-commits --no-changelog --conventional-graduate --no-verify-access --yes
-
-      - name: Replace forward slashes in branch name
-        if: ${{ github.event.action == 'opened' || github.event.action == 'synchronize' }}
-        env:
-          github_ref: ${{github.head_ref}}
-        run: |
-          branch="${github_ref////-}"
-          echo "BRANCH=$branch" >> $GITHUB_ENV
```
mshwery, line 112: We _might_ still need this if you are allowed to manually trigger the workflow on any branch.

.github/workflows/publishCanary.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Notify end
+        uses: actions/github-script@v4
+        with:
+          script: |
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'Canary package publishing complete. See the actions tab for more details.',
+            });
```
mshwery, line 92: This is awesome!

2 questions
1. can we get the final output of lerna (the published packages) maybe with grep or something and include those in the PR comment? 
2. even more amazing would be auto-linking to npm

.github/workflows/publish.yml
daltonscharff: The if statement at line 47 now protects us against PRs that are closed but not merged. https://github.com/segmentio/action-destinations/blob/0f15af921d03a2d19a7b12447c3e6533ada208d1/.github/workflows/publish.yml#L47

.github/workflows/publish.yml
daltonscharff: Good catch. I think a good solution may be to modify line 47 to ensure that `github.event_name == 'workflow_dispatch' && branch == master`

.github/workflows/publishCanary.yml
daltonscharff: I would like to see both of those improvements as well, however I was struggling on Friday to persist the Lerna result across workflow steps. I can take another look at this today.

.github/workflows/publish.yml
daltonscharff: This has been fixed.

.github/workflows/publishCanary.yml
mshwery: We can punt on that part until later. I think PAPI metadata is probably a bigger deal // @mericsson

mshwery (approved):

daltonscharff:
The final output of Lerna is now displayed in the comment confirming the packages have been published.
image

PR #311, opened 2021-11-02
add auth field demarcation for UI

dlasky:
tags auth fields as such

## Testing

### DIFF
--- a/packages/cli-internal/src/commands/push.ts
+++ b/packages/cli-internal/src/commands/push.ts
@@ -372,11 +372,11 @@ export function getOptions(
       validators.push(['required', `The ${fieldKey} property is required.`])
     }
 
-    // Everything in `authentication.fields` should be private. Otherwise, public is fine
-    const isPrivateSetting = typeof authFields === 'object' && fieldKey in authFields
-
     // If the field exists in auth fields.
     const isAuth = typeof authFields === 'object' && fieldKey in authFields
+
+    // Everything in `authentication.fields` should be private. Otherwise, public is fine
+    const isPrivateSetting = isAuth
 
     let type: DestinationMetadataOption['type'] = schema.type
     if (Array.isArray(schema.choices)) {
--- a/packages/cli-internal/src/commands/push.ts
+++ b/packages/cli-internal/src/commands/push.ts
@@ -375,6 +375,9 @@ export function getOptions(
     // Everything in `authentication.fields` should be private. Otherwise, public is fine
     const isPrivateSetting = typeof authFields === 'object' && fieldKey in authFields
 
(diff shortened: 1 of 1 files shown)


packages/cli-internal/src/commands/push.ts
```diff
@@ -399,9 +402,15 @@ export function getOptions(
 
     // Preserve existing properties unless specified by the action field definition
     const existing = existingOptions[fieldKey]
+    const tags = existing.tags ?? []
```
dlasky, line 405: not sure if this will be a problem for existing empties...

packages/cli-internal/src/commands/push.ts
```diff
@@ -375,6 +375,9 @@ export function getOptions(
     // Everything in `authentication.fields` should be private. Otherwise, public is fine
     const isPrivateSetting = typeof authFields === 'object' && fieldKey in authFields
 
+    // If the field exists in auth fields.
+    const isAuth = typeof authFields === 'object' && fieldKey in authFields
```
mshwery, line 379: Since this is identical to `isPrivateSetting` maybe we just rename `isPrivateSetting` → `isAuthSetting` and use that value for both the tags + `private` property?

packages/cli-internal/src/commands/push.ts
```diff
@@ -399,9 +402,15 @@ export function getOptions(
 
     // Preserve existing properties unless specified by the action field definition
     const existing = existingOptions[fieldKey]
+    const tags = existing.tags ?? []
+
+    if (isAuth && !tags.includes('authentication:test')) {
+      tags.push('authentication:test') //valid values here can eventually be test, no-test if needed
+    }
```
mshwery, line 409: I'm curious why we wouldn't use `category:authentication`? We want these fields grouped together, right?

mshwery (approved):
Couple minor things but otherwise LGTM!

packages/cli-internal/src/commands/push.ts
dlasky: yeah i was just thinking if that ever diverged in the future, hence duplicating it, but i could certainly just assign the vars together.

packages/cli-internal/src/commands/push.ts
dlasky: I don't believe so my understanding of the grouping logic (which admittedly could be flawed) is:

1 'virtual' group containing all required items (regardless of auth status)
any actual groups via categories
1 'virtual' group containing all non-required items

dlasky: I was also thinking this gives us the ability to make testing optional should that arise for some scenario while still considering it an auth field

PR #312, opened 2021-11-02
[braze web] Add default value to sdkVersion; prevent redundant initializations

juliofarah:
We found out that the library was not loading because `sdkVersion` is not being populated by the Segment App. Also, the braze team has reported that braze web is being initialized twice.

This PR checks if the `sdkVersion` field is present on braze (web) configuration and, if not, sets `3.3` as the default value and removes braze's snippet in favor of the existing `loadScript` implementation.

## Testing

Before:
image

After:
image

image


danieljackins (approved):
:shipit:

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -251,12 +250,11 @@ export const destination: BrowserDestinationDefinition<Settings, typeof appboy>
   initialize: async ({ settings }, dependencies) => {
     try {
       const { endpoint, api_key, sdkVersion, automaticallyDisplayMessages, ...expectedConfig } = settings
-
-      initScript(sdkVersion)
+      const version = sdkVersion ?? '3.3'
```
dlasky, line 253: shouldn't this always be set in settings? wonder if there is an upstream issue

dlasky (approved):

packages/browser-destinations/src/destinations/braze/index.ts
juliofarah: I would assume there's an issue either on the app or whatever other upstream service that populates these fields. For some reason, `sdkVersion` was not populated for a couple of sources I tested (including @nielst 's)

mshwery: If it was added later (after the destination instances had been created) that could be the reason the default wasn't provided.

At any rate, for fields that are required and can use a fallback value, this is definitely a smart choice!

PR #313, opened 2021-11-02
Don't Persists Credentials on Publish

daltonscharff:
Adds "persist-credentials" so that the actions secret takes precedence.


gpsamson (approved):
rad

PR #316, opened 2021-11-04
make user_alias props optional and validate manually

mshwery:
This PR takes a more nuanced approach for Braze's `user_alias` object. We need to ensure that messages are not rejected when this field is invalid (but present). Instead, we want to manually validate and extract the proper shape of this object before sending it to Braze.

## Testing


dlasky (approved):
lgtm, maybe want a few tests for the util methods and handling?

PR #317, opened 2021-11-04
Use NPM Publish Tokens

daltonscharff:
Previously, the NPM token being used to publish packages was a read-only token, causing a 403 error.
image

I've added a secret NPM_PUBLISH_TOKEN that should now be able to publish packages successfully.


mshwery (approved):

gpsamson (approved):
LFG

PR #319, opened 2021-11-05
removing requirements to add in secret key and app id

coreyching:

Removed the requirement for a customer to sign up for TikTok Developer App and no longer is a required field for authentication. 

This simplifies the workflow so customers can just provide access token and pixel id. 
## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/tiktok-conversions/index.ts
+++ b/packages/destination-actions/src/destinations/tiktok-conversions/index.ts
@@ -71,18 +71,6 @@ const destination: DestinationDefinition<Settings> = {
         type: 'string',
         required: true
       },
-      secretKey: {
-        label: 'Secret Key',
-        description: 'TikTok App Secret Key. You can find this key in the "Basic Information" tab of your TikTok app.',
-        type: 'string',
-        required: true
-      },
-      appId: {
-        label: 'App Id',
-        description: 'TikTok App Id. You can find this key in the "Basic Information" tab of your TikTok app.',
-        type: 'string',
-        required: true
-      },
       pixel_code: {
         label: 'Pixel Code',
         type: 'string',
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -92,13 +80,13 @@ const destination: DestinationDefinition<Settings> = {
...
+      // Send a blank event to events API.
+      return request('https://business-api.tiktok.com/open_api/v1.2/pixel/track/', {
         json: {
-          access_token: settings.accessToken,
-          app_id: parseInt(settings.appId, 10),
-          secret: settings.secretKey
+          pixel_code: settings.pixel_code,
+          event: "Test Event",
+          timestamp: '',
+          context: {}
```
gpsamson, line 89: Just noting the hardcoded values!

gpsamson (approved):

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
mshwery: Will this request actually fail if the pixel_code is invalid?

coreyching: @gpsamson thanks - I used hard-coded values since it's just to confirm authentication. Are you suggesting I create a const for the test values? 

@mshwery yes there needs to be a valid pixel_code for the request to complete.

gpsamson: Ahh, I understand now. It's fine as is. I would consider renaming the event to `Test Authentication Event` so it's clear where that event is being produced. No strong opinions though! :shipit:

PR #323, opened 2021-11-08
Modify GitHub Token Secret

daltonscharff:
The previous github token secret did not have the correct permissions to write. Using the standard Github token that set by default should fix this issue.

## Testing

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -61,7 +61,7 @@ jobs:
           cache: 'yarn'
 
       - name: Use Github Personal Access Token
-        run: git config --global url."https://${{ secrets.GITHUB_TOKEN }}@github.com/".insteadOf ssh://git@github.com/
+        run: git config --global url."https://${{ secrets.NEW_GH_TOKEN }}@github.com/".insteadOf ssh://git@github.com/
 
       - name: Install Dependencies
         run: yarn install --frozen-lockfile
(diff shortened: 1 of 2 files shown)


gpsamson (commented):
Are you able to only use `GITHUB_TOKEN` for the publishing step? It may not be compatible with the dependency install because it won't have access to private repositories like the `NEW_GH_TOKEN` secret does.

mshwery (approved):

daltonscharff:
> Are you able to only use `GITHUB_TOKEN` for the publishing step? It may not be compatible with the dependency install because it won't have access to private repositories like the `NEW_GH_TOKEN` secret does.

I've made the change to update only the publishing step to use that new token. According to this doc, it should work: https://github.com/lerna/lerna/tree/main/commands/version#--create-release-type

.github/workflows/publishCanary.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           cache: 'yarn'
 
       - name: Use Github Personal Access Token
-        run: git config --global url."https://${{ secrets.GITHUB_TOKEN }}@github.com/".insteadOf ssh://git@github.com/
+        run: git config --global url."https://${{ secrets.GH_PAT }}@github.com/".insteadOf ssh://git@github.com/
```
gpsamson: Should we be using `NEW_GH_TOKEN` for consistency?

gpsamson (approved):

.github/workflows/publishCanary.yml
daltonscharff: Yeah that makes sense

.github/workflows/publish.yml
```diff
@@ -94,4 +94,6 @@ jobs:
           npm whoami
 
       - name: Publish production package
+        env: 
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
mshwery, line 98: I think you aren't setting this in the right place.

daltonscharff: I think, according to [this doc](https://github.com/lerna/lerna/tree/main/commands/version#--create-release-type), adding a `GH_TOKEN` environment variable tells Lerna to use it as an auth token.

PR #324, opened 2021-11-08
[cli] Set default regions to US only

gpsamson:
This pull request sets the default supported regions to `us-west-2` when pushing destination definitions. Action destinations do not support multi-region deployments at the time of writing this.

Note that it will **always** overwrite the current value, so if a change in Partner Portal is made it will be overwritten.

## Testing

### DIFF
--- a/packages/cli-internal/src/commands/push.ts
+++ b/packages/cli-internal/src/commands/push.ts
@@ -275,7 +275,8 @@ export default class Push extends Command {
             advancedOptions: [], // make sure this gets cleared out since we don't use advancedOptions in Actions
             basicOptions,
             options,
-            platforms
+            platforms,
+            supportedRegions: ['us-west-2'] // always default to US until regional action destinations are supported
           }),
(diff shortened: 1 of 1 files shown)


packages/cli-internal/src/commands/push.ts
```diff
@@ -275,7 +275,8 @@ export default class Push extends Command {
             advancedOptions: [], // make sure this gets cleared out since we don't use advancedOptions in Actions
             basicOptions,
             options,
-            platforms
+            platforms,
+            supportedRegions: ['us-west-2'] // always default to US until regional action destinations are supported
```
mshwery, line 279: Should this be handled in CPS instead?

mshwery (approved):

packages/cli-internal/src/commands/push.ts
gpsamson: I would say yes if supported regions of actions were to be us-only in the long term, but since multi-region support is slated for Q1 I think this is a safe (short-term) compromise.

PR #325, opened 2021-11-08
use GITHUB_TOKEN

mshwery:
This updates the github token we are pulling in.

## Testing
NA

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -61,7 +61,7 @@ jobs:
           cache: 'yarn'
 
       - name: Use Github Personal Access Token
-        run: git config --global url."https://${{ secrets.NEW_GH_TOKEN }}@github.com/".insteadOf ssh://git@github.com/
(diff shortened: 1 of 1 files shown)


dlasky (approved):

.github/workflows/publish.yml
```diff
@@ -61,7 +61,7 @@ jobs:
           cache: 'yarn'
 
       - name: Use Github Personal Access Token
-        run: git config --global url."https://${{ secrets.NEW_GH_TOKEN }}@github.com/".insteadOf ssh://git@github.com/
+        run: git config --global url."https://${{ secrets.GITHUB_TOKEN  }}@github.com/".insteadOf ssh://git@github.com/
```
gpsamson, line 64: I _think_ this will fail because `GITHUB_TOKEN` isn't scoped to access private repos like `control-plane-js-client`

mshwery: but... `NEW_GH_TOKEN` doesnt have read/write permissions to the repo either 🤔

PR #329, opened 2021-11-10
add @arrayPath directive to resolve arrays and transform item shapes

mshwery:
This PR introduces a new directive with the same ol' jsonpath-like syntax for not only looking up an array path, but also remapping the keys with directives/raw values of their own.

You can do stuff like this:

Input:
```json
{
  "properties": {
    "products": [{ "productId": 1 }, { "productId": 2 }]
  }
}
```

Simple mapping (not really any different that `@path`):
```json
{
  "@arrayPath": ["$.properties.products"]
}
```

Result:
```json
[
  {
    "productId": 1
  },
  {
    "productId": 2
  }
]
```

Mappings with item shape:
```json
{
  "@arrayPath": ["$.properties.products", {
    "some_other_key": { "@path": "$.productId" }
  }]
}
```

Result:
```json
[
  {
    "some_other_key": 1
  },
  {
    "some_other_key": 2
  }
]
```

## Testing

Tested locally and via CI unit tests.

### DIFF
--- a/packages/core/src/mapping-kit/README.md
+++ b/packages/core/src/mapping-kit/README.md
@@ -62,6 +62,7 @@ Output:
   - [@path](#path)
   - [@template](#template)
   - [@literal](#literal)
+  - [@arrayPath](#array-path)
 
 <!-- tocstop -->
 
@@ -443,4 +444,57 @@ n/a
 
 Mappings:
 { "@literal": true } => true
+```
+
+### @arrayPath
+
+The @arrayPath directive resolves an array value at a given path (much like @path), but allows you to specify the shape of each item in the resulting array. You can use directives for each key in the given shape, relative to the root array.
+
+Input:
+```json
+{
+  "properties": {
+    "products": [{ "productId": 1 }, { "productId": 2 }]
+  }
+}
+```
+
+Mapping:
+```json
+{
+  "@arrayPath": ["$.properties.products"]
+}
+```
+
+Result:
+```json
+[
+  {
+    "productId": 1
+  },
+  {
+    "productId": 2
+  }
+]
+```
+
+Mappings with item shape:
+```json
+{
+  "@arrayPath": ["$.properties.products", {
+    "some_other_key": { "@path": "$.productId" }
+  }]
+}
+```
+
+Result:
+```json
+[
+  {
+    "some_other_key": 1
+  },
+  {
+    "some_other_key": 2
+  }
+]
 ```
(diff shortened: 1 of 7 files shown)


packages/core/src/mapping-kit/index.ts
```diff
@@ -83,7 +103,6 @@ registerDirective('@literal', (value, payload) => {
 
 /**
  * Resolves a mapping value/object by applying the input payload based on directives
- * *WARNING* This function mutates `mapping` when an object
```
mshwery, line 86: fixed this so we dont have to clone all over the place

packages/core/src/mapping-kit/index.ts
```diff
@@ -68,6 +68,26 @@ registerDirective('@if', (opts, payload) => {
...
+  const [path, itemShape] = data as [string, undefined | JSONObject]
+  if (typeof path !== 'string') {
+    throw new Error(`@arrayPath expected args[0] to be string, got ${realTypeOf(path)}`)
+  }
+
+  const root = get(payload, path.replace('$.', '')) as JSONLike
+
+  // If a shape has been provided, resolve each item in the array with this shape
+  if (isArray(root) && realTypeOf(itemShape) === 'object' && Object.keys(itemShape as JSONObject).length > 0) {
+    return root.map((item) => resolve(itemShape, item as JSONObject))
```
mshwery, line 85: Note: this is treating `item` as a the root object. That means you won't be able to use paths that reference anything _above/outside_ the resolved array item.

mshwery: This is both intentional, and a limitation.

The most likely use case is that customers want to map values from each object.

dlasky: IMO its a good limitation, anything beyond that and we start stepping firmly into functions territory

packages/core/src/mapping-kit/README.md
```diff
@@ -444,3 +445,56 @@ n/a
...
+```json
+[
+  {
+    "some_other_key": 1
+  },
+  {
+    "some_other_key": 2
+  }
+]
+```
```
dlasky, line 500: 🔥

dlasky (approved):

packages/core/src/mapping-kit/index.ts
lcampos: Agreed, these type of limits makes it easy to maintain and easier to reason

lcampos (approved):

PR #340, opened 2021-11-15
Fix test imports

gpsamson:
- Fixes an issue where tests would fail without a `yarn build`
- Removes the build step when testing the set of setup commands documented for external devs

## Testing


packages/destination-actions/tsconfig.json
```diff
@@ -3,10 +3,20 @@
   "compilerOptions": {
     "baseUrl": ".",
     "paths": {
-      "@segment/actions-core": ["../core/src"],
-      "@segment/destination-subscriptions": ["../destination-subscriptions/src"]
+      "@segment/ajv-human-errors": [
```
gpsamson, line 6: need to fix formatting

.github/workflows/ext.yml
```diff
@@ -31,8 +31,8 @@ jobs:
...
         run: yarn bootstrap --no-ci
 
-      - name: Build
-        run: NODE_ENV=production yarn build
-
       - name: Test
         run: yarn test
+
+      - name: Build
+        run: yarn build
```
mshwery, line 38: Should this be `NODE_ENV=production`? or is that not needed?

packages/core/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
+    "baseUrl": "./",
```
mshwery, line 4: Is this needed?

mshwery: I've found baseUrl and rootDir changes to introduce somewhat unpredictable breakages, so I'm generally hesitant.

mshwery (approved):

packages/core/tsconfig.json
gpsamson: I think `baseUrl` is needed if `paths` is used..

.github/workflows/ext.yml
gpsamson: I removed it from the external workflow to better emulate the dev workflow. I don't think devs will specify production when running locally.

PR #344, opened 2021-11-17
Increase actions-cli test timeouts

gpsamson:
I've noticed that the `actions-cli` tests have been shaky in the main branch due to constant timeouts -- the first test pushes close to the current 5s timeout.

This pull request adds `isolatedModules` to the cli's jest config (see [SOF](https://stackoverflow.com/questions/45087018/jest-simple-tests-are-slow)). It also removes a `console.warn` call which seems to cause some unnecessary and incorrect error logs in the test output.

As a follow up, we should consider moving away from ts-jest.

Example test run which completes in ~4.7s:
image

## Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,6 @@
   },
   "jest": {
     "preset": "ts-jest",
-    "testTimeout": 10000,
     "globals": {
       "ts-jest": {
         "isolatedModules": true
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
   },
   "jest": {
     "preset": "ts-jest",
+    "testTimeout": 10000,
     "globals": {
       "ts-jest": {
         "isolatedModules": true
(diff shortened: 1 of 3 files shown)


packages/cli/src/hooks/init.ts
```diff
@@ -3,7 +3,7 @@ import { dirname } from 'path'
 
 const hook: Hook<'init'> = async function ({ config }) {
   if (process.env.NODE_ENV === 'test') {
-    console.warn('skipping cli-internal plugin because of tests')
+    // console.warn('skipping cli-internal plugin because of tests')
```
jinapark202, line 6: Should we remove this line altogether?

jinapark202 (approved):

package.json
```diff
@@ -74,6 +74,7 @@
   },
   "jest": {
     "preset": "ts-jest",
+    "testTimeout": 10000,
```
mshwery, line 77: womp womp. do we know why any tests are taking that long? are some making external requests?

mshwery (approved):

package.json
gpsamson: Pretty lame, I know. I couldn't pin point why it's so slow in tests. It's just TS compilation + some filesystem calls.

PR #350, opened 2021-11-19
Add metronome destination

cozmo:

This PR adds [Metronome](https://getmetronome.com) as a destination using the new `action-destinations` framework. For context - Metronome is a billing platform that allows companies to transform their product's raw event stream (such as `api_request.serviced` and `data.uploaded` events) into bills for their customers. This context may be useful when reviewing some of our defaults (for example, we default to `$.context.groupId` instead of `$.userId`, since our customers are usually billing organizations not end users).

Documentation for how to send events to Metronome is located [here](https://docs.getmetronome.com/getting-usage-data-into-metronome/overview), and the api endpoint specific documentation for our ingest endpoint can be found [here](https://docs.getmetronome.com/api#operation/ingest).

## Testing

Biggest thing is - I have no idea how this is going to look or be used in the Segment UI, so I'm kinda crossing my fingers that this will cause a reasonable UX. Once it's deployed we'll try to run through it with our Segment account so we can take screenshots for our documentation. Once we do that we may make another PR with anything we want to change? Not sure if there's a better way to do that?

### DIFF
--- a/packages/destination-actions/src/destinations/metronome/index.ts
+++ b/packages/destination-actions/src/destinations/metronome/index.ts
@@ -6,6 +6,7 @@ import sendEvent from './sendEvent'
 const destination: DestinationDefinition<Settings> = {
   name: 'Metronome',
   mode: 'cloud',
+  slug: 'metronome-actions',
 
   authentication: {
     scheme: 'custom',
--- a/packages/destination-actions/src/destinations/metronome/index.ts
+++ b/packages/destination-actions/src/destinations/metronome/index.ts
@@ -1,0 +1,52 @@
+import { defaultValues, DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+
+import sendEvent from './sendEvent'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Metronome',
+  slug: 'metronome',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'custom',
+    fields: {
+      apiToken: {
+        type: 'string',
+        label: 'API Token',
+        description: 'Your Metronome API Token',
+        required: true
+      }
+    },
+    testAuthentication: async (request) => {
+      const response = await request('https://api.getmetronome.com/v1/ingest', {
+        method: 'post',
+        body: JSON.stringify([]),
+        throwHttpErrors: false
+      })
+      // An empty set of events will return a 400 response whereas a bad token will return a 403.
+      if (response.status === 400) {
+        return true
+      }
+      throw new Error('Invalid API Token')
+    }
+  },
+  extendRequest: ({ settings }) => {
+    return {
+      headers: { Authorization: `Bearer ${settings.apiToken}` }
+    }
+  },
+  actions: {
+    sendEvent
+  },
+  presets: [
+    {
+      name: 'Send track events to Metronome',
+      subscribe: 'type = "track"',
+      partnerAction: 'sendEvent',
+      mapping: defaultValues(sendEvent.fields)
+    },
+  ]
+}
+
+export default destination
--- a/packages/destination-actions/src/destinations/metronome/index.ts
+++ b/packages/destination-actions/src/destinations/metronome/index.ts
@@ -5,7 +5,6 @@ import sendEvent from './sendEvent'
 
 const destination: DestinationDefinition<Settings> = {
   name: 'Metronome',
-  slug: 'metronome',
   mode: 'cloud',
 
   authentication: {
--- a/packages/destination-actions/src/destinations/metronome/sendEvent/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/metronome/sendEvent/__tests__/index.test.ts
@@ -1,0 +1,150 @@
+import nock from 'nock'
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
+import Destination from '../../index'
+
+const testDestination = createTestIntegration(Destination)
+
+describe('Metronome.sendEvent', () => {
+  describe("process", () => {
+    it('should send an event', async () => {
+      const event = createTestEvent();
+
+      nock('https://api.getmetronome.com').post('/v1/ingest').reply(200)
+
+      const responses = await testDestination.testAction('sendEvent', {
+        event,
+        useDefaultMappings: true,
+        settings: {
+          apiToken: "mock-api-token",
+        },
+        mapping: {
+          event_type: { '@path': "$.event" },
+          properties: { '@path': "$.context" },
+        },
+      });
+
+      expect(responses.length).toBe(1)
+      expect(responses[0].status).toBe(200)
+
+      // TODO: Feels like there should be a lighter weight to do this check
+      expect(responses[0].options.headers).toMatchInlineSnapshot(`
+        Headers {
+          Symbol(map): Object {
+            "authorization": Array [
+              "Bearer mock-api-token",
+            ],
+            "user-agent": Array [
+              "Segment (Actions)",
+            ],
+          },
+        }
+      `)
+
+      expect(responses[0].options.json).toEqual([
+        {
+          event_type: event.event,
+          customer_id: event?.context?.groupId ?? event.userId ?? event.anonymousId,
+          properties: event.context,
+          transaction_id: event.messageId,
+          timestamp: new Date(event.timestamp ?? "").toISOString(),
+        }
+      ])
+    })
+
+
+    it('should convert a non rfc3999 timestamp', async () => {
+      const event = createTestEvent({
+        timestamp: "2021-01-01"
+      });
+
+      nock('https://api.getmetronome.com').post('/v1/ingest').reply(200)
+
+      const responses = await testDestination.testAction('sendEvent', {
+        event,
+        useDefaultMappings: true,
+        settings: {
+          apiToken: "mock-api-token",
+        },
+        mapping: {
+          event_type: { '@path': "$.event" },
+          properties: { '@path': "$.context" },
+        },
+      });
+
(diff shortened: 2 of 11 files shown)


packages/destination-actions/src/destinations/metronome/sendEvent/__tests__/index.test.ts
```diff
@@ -0,0 +1,150 @@
...
+        mapping: {
+          event_type: { '@path': "$.event" },
+          properties: { '@path': "$.context" },
+        },
+      });
+
+      expect(responses.length).toBe(1)
+      expect(responses[0].status).toBe(200)
+
+      // TODO: Feels like there should be a lighter weight to do this check
```
gpsamson, line 29: Some destination builders have convert the headers to a JSON object before comparing: https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/customerio/__tests__/createUpdateDevice.test.ts#L42-L45.

You can also consider plucking the `Authorization` header and checking for `mock-api-token` to simplify the test.

gpsamson (commented):
INCREDIBLE! Thank you for the thorough testing, comments, and reference links 🙇🏽 

The only comment that needs addressing before this destination is deployed for testing is the one regarding default field path of the Send Event action.

packages/destination-actions/src/destinations/metronome/sendEvent/index.ts
```diff
@@ -0,0 +1,95 @@
...
+      description: 'The timestamp at which this event occurred.',
+      required: true,
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    event_type: {
+      type: 'string',
+      label: 'event_type',
+      description: 'The Metronome event_type.',
```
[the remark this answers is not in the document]
gpsamson: This may be applicable to other descriptions. I notice markdown is used throughout the Metronome docs.

packages/destination-actions/src/destinations/metronome/sendEvent/index.ts
```diff
@@ -0,0 +1,95 @@
...
+          else: {
+            '@if': {
+              exists: { '@path': '$.userId' },
+              then: { '@path': '$.userId' },
+              else: { '@path': '$.anonymousId' }
+            }
+          }
+        }
+      }
+    },
```
gpsamson, line 55: I'm 90% certain nested if/else isn't supported but left me confirm with the team. You may have to decide on a single backup value.

My recommendation is that you use `userId` and `anonymousId` as the primary and back up values since both are a part of the [Segment Spec](https://segment.com/docs/connections/spec/common/). Customer's events will almost always have one of those two fields set, whereas only a subset customers will be adding `groupId` to the context object.

cozmo: If that's the case can we update the types? They explicitly make it seem supported right now https://github.com/segmentio/action-destinations/blob/main/packages/core/src/destination-kit/types.ts#L133-L141

mshwery: I would discourage this in general. You want your defaults to be as simple as possible. 

The UI will not be able to support this type of nesting even if the framework does. 

We can consider tightening the types to prevent it!

cozmo: I feel like to that end the default should maybe just be `.groupId`? I don't think it's likely our customers will want to send individual end user IDs to our platform, and if the goal is a simple default then I think `.groupId` makes the most sense. Customers can/should be encouraged to change the settings for this field to map to whatever information they want here, since groupId may not always exist? 

Does that seem reasonable?

gpsamson: Your explanation makes sense. I agree that just having `$.context.groupId` as the default is the best route. If `$.context.groupId` doesn't exist in their events then request to Metronome will start to fail with a 400 status code until customers tailor their mapping.

cozmo: Makes sense. Switched to use `$.context.groupId` in b01f3ed. I don't totally understand when I should be using `$.groupId` (at the root) vs `$.context.groupId`, but as per this discussion stuck with `$.context.groupId`.

packages/destination-actions/src/destinations/metronome/sendEvent/__tests__/index.test.ts
cozmo: Unfortunately you'll notice that those calls are all on _response_ headers, not on _request_ headers. Response headers have been typed correctly [here](https://github.com/segmentio/action-destinations/blob/main/packages/core/src/types.ts#L7-L10) as `Headers & { toJSON: () => Record<string, string> }`. This allows the caller to easily get an object back, of which they can introspect in any number of ways. 

Unfortunately the same is not true of the _request_ headers, which are typed [here](https://github.com/segmentio/action-destinations/blob/main/packages/core/src/request-client.ts#L13) (incorrectly) as `Record<string, string>`. I think the error stems from the fact that you _can_/_should_ pass a raw object _into_ request, but under the hood it'll be converted into an instance of [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers), which is actually fairly incompatible with `Record<string, string>` type. This means you cannot call things like `.entries()`, `.toJSON()`, or `.get()`, which severely limits the introspection you can do in the tests. As such we're stuck either using this verbose snapshot API _or_ casting to `unknown` to override the incorrect types (`response?.options?.headers as unknown as Headers`). I have a fairly strong aversion to type overriding, so we're stuck using the verbose snapshots. 

I'm biased and surely am missing context but from the outside I'd say the ideal solution is to type this field using the correct types 😄. Until then the snapshot seems fine. 

I updated the comment to explain this in more depth.

packages/destination-actions/src/destinations/metronome/index.ts
```diff
@@ -0,0 +1,52 @@
+import { defaultValues, DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+
+import sendEvent from './sendEvent'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Metronome',
+  slug: 'metronome',
```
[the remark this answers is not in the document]
mshwery: Isn't this slug preferred when registering the destination?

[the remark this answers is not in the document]
cozmo: I don't know how any of this stuff works internally to Segment obviously but from where I'm sitting I'd much prefer our (new) destination to have the clean `metronome` slug, vs letting our old one (which we're killing off after releasing this) from having it. Not sure if they share the same namespace or not?

packages/destination-actions/src/destinations/metronome/sendEvent/index.ts
```diff
@@ -0,0 +1,95 @@
...
+      description: 'The Metronome transaction ID uniquely identifies an event to ensure Metronome only processes each event once.',
+      required: true,
+      default: {
+        '@path': '$.messageId'
+      }
+    },
+    customer_id: {
+      type: 'string',
+      label: 'customer_id',
+      description: 'The Metronome customer ID or ingest alias this event should be associated with.',
```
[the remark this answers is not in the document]
cozmo: Good to know, will leave that up to our docs writer(s). I imagine if we have docs about setting up Segment specifically we maybe would link to them, but would want to see how this appears in the UI first.

packages/destination-actions/src/destinations/metronome/sendEvent/__tests__/index.test.ts
```diff
@@ -26,7 +31,11 @@ describe('Metronome.sendEvent', () => {
       expect(responses.length).toBe(1)
       expect(responses[0].status).toBe(200)
 
-      // TODO: Feels like there should be a lighter weight to do this check
+      // The `options.headers` property is incorrectly typed as Record<string, string>,
+      // when in practice it is an instance of Headers. Until this is fixed,
+      // we'll use a snapshot to check the headers. Ideally this check could just be
+      // `expect(responses[0].options.headers.get("authorization")).toBe("Bearer mock-api-token")`
```
gpsamson, line 37: Thanks for pointing this out! We have some type updates to follow up with on our end. In the meantime, I agree with using snapshots over type overrides.

gpsamson (approved):

packages/destination-actions/src/destinations/metronome/sendEvent/__tests__/index.test.ts
```diff
@@ -26,7 +31,11 @@ describe('Metronome.sendEvent', () => {
...
+      console.log(responses[0].options.headers instanceof Headers)
```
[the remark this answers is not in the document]
gpsamson:
```suggestion
```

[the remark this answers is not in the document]
cozmo: Oh 🤦‍♂️

packages/destination-actions/src/destinations/metronome/index.ts
```diff
@@ -0,0 +1,51 @@
+import { defaultValues, DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+
+import sendEvent from './sendEvent'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Metronome',
+  mode: 'cloud',
+
```
gpsamson, line 9: Matt corrected me, we should have a slug here. 
```suggestion
slug: 'actions-metronome',
```

cozmo: Added

PR #351, opened 2021-11-19
default group key

Alex-Kim-mp:
Per https://mixpanel.slack.com/archives/C02D9DFHFC2/p1637281794002300?thread_ts=1637275909.001900&cid=C02D9DFHFC2, invalid presets for quick setup actions have been changed to throw an error now. So a solution was to provide a default value, which is now `company`.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
@@ -10,15 +10,14 @@ const action: ActionDefinition<Settings, Payload> = {
     group_key: {
       label: 'Group Key',
       type: 'string',
-      description: 'The group key',
-      required: true,
-      default: 'company'
+      description: 'The group key. If this is not set, it is defaulted to "name".',
+      default: 'name'
     },
     traits: {
       label: 'Group Properties',
       type: 'object',
-      description: 'Properties to set on the group profile',
-      required: true,
+      description:
+        'Properties to set on the group profile. Make sure to have a field that corresponds to the above group key.',
       default: {
         '@path': '$.traits'
       }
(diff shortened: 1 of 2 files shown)


jinapark202:
Can you also fix `Error` to use `IntegrationError` with this PR?

mshwery (approved):
The change looks fine to me but I'm not familiar if this is a sane default for most customers using the presets.

Another option would be to omit the default you added here, and instead remove this action from the presets.

Alex-Kim-mp:
What does removing this action from the presets do? Would users not be able to make this call anymore?

mshwery:
> What does removing this action from the presets do? Would users not be able to make this call anymore?

Removing the preset means it wont automatically be subscribed when users connect a new instance of the destination. Presets are automatically applied, and _must_ have all required fields provided, i.e. it must work out-of-the-box without user intervention. 

If it's not a preset, customers can still manually subscribe to the action and configure it themselves.

Alex-Kim-mp:
Okay, I think we'd prefer to have the preset available then? Maybe we can use `name` instead since that's standardized in the spec. We also have an error check since that key must exist in the traits object

packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
```diff
@@ -10,24 +10,26 @@ const action: ActionDefinition<Settings, Payload> = {
     group_key: {
       label: 'Group Key',
       type: 'string',
-      description: 'The group key',
-      required: true
+      description: 'The group key. If this is not set, it is defaulted to "name".',
+      required: true,
```
jinapark202, line 14: Do you want it to be `required` here? Based on your description and the change you made to the `generated-types` file, it seems like you want `group_key` to be optional.

jinapark202 (approved):

PR #354, opened 2021-11-23
Update GA4 Items array defaults

nick-Ag:
This PR updates the Items array defaults in GA4, and refactors the codebase to pull the items array out into a common shared file. Unit tests are also updated as well.

[Link to Spec](https://paper.dropbox.com/doc/GA4-Cloud-mode-Actions--BWb9JcLSQOglLG7swjdNeQk5Ag-TazwZKvI6E9yBIwajtkwz)
## Testing


nick-Ag:
After merge this one will require we update and deploy `fab-5-ops` with the newly published npm package since `generated-types` files were updated. It will also require we run `./bin/run push` on production workbench

packages/destination-actions/src/destinations/google-analytics-4/addPaymentInfo/index.ts
```diff
@@ -3,150 +3,83 @@ import { CURRENCY_ISO_CODES } from '../constants'
...
-          label: 'Index',
-          type: 'number',
-          description: 'The index/position of the item in a list.'
-        },
-        item_brand: {
-          label: 'Brand',
-          type: 'string',
-          description: 'Brand associated with the product.'
+        ...items.properties,
+        creative_name: {
```
kdaswani, line 34: @nick-Ag why did we add `creative_name`, `creative_slot`, `promotion_name` and `promotion_id` to the `items` array for Add Payment Info? I don't think these are fields in the `items` array for that event as per: https://developers.google.com/analytics/devguides/collection/ga4/reference/events#add_payment_info

kdaswani (commented):
Looks good! Added a comment. And then similar to FB CAPI, let's not merge until we figure out the `arrayPath` issue that is affecting single-product events:

- Add to Cart
- Add to Wishlist
- Remove from Cart
- Select Item
- Select Promotion
- View Item
- View Promotion

nick-Ag:
Going to wait on testing of FB CAPI to be done, then deploy this to stage for testing, then merge in. Today or early tomorrow

kdaswani:
@nick-Ag I think we still need to update the `arrayPath` and subfield defaults for these actions:

- Add to Cart
- Add to Wishlist
- Remove from Cart
- Select Item
- Select Promotion
- View Item
- View Promotion

`arrayPath` should be `properties` similar to what we did with the single-product events in FB CAPI. Thanks!

packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
```diff
@@ -0,0 +1,206 @@
...
+      label: 'Brand',
+      type: 'string',
+      description: 'Brand associated with the product.'
+    },
+    item_category: {
+      label: 'Category',
+      type: 'string',
+      description: 'Product category.'
+    },
+    item_category2: {
```
brennan, line 155: Do we still need `item_category2` et. seq. defined here? It looks like these have been removed from the actions.

packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts
```diff
@@ -11,182 +22,57 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      }
-    },
-    tax: {
-      label: 'Tax',
-      type: 'number',
-      description: 'Total tax associated with the transaction.',
-      default: {
-        '@path': '$.properties.tax'
+        '@arrayPath': [
+          '$.properties.products',
```
brennan, line 36: Definitely not a blocker, but it looks as if these defaults defaults are shared across many files. Should we abstract them into a single file?

packages/destination-actions/src/destinations/google-analytics-4/refund/index.ts
```diff
@@ -9,159 +10,56 @@ const action: ActionDefinition<Settings, Payload> = {
...
-    },
-    shipping: {
-      label: 'Shipping',
-      type: 'number',
-      description: 'Shipping cost associated with a transaction.'
-    },
+    client_id: { ...client_id, required: true },
+    currency: { ...currency },
+    transaction_id: { ...transaction_id, required: true },
+    value: { ...value },
```
brennan, line 16: Does value need a default here?

packages/destination-actions/src/destinations/google-analytics-4/addPaymentInfo/index.ts
```diff
@@ -3,150 +3,60 @@ import { CURRENCY_ISO_CODES } from '../constants'
...
-      label: 'Payment Type',
-      type: 'string',
-      description: 'The chosen method of payment.',
-      default: {
-        '@path': '$.properties.payment_method'
-      }
-    },
+    client_id: { ...client_id, required: true },
+    currency: { ...currency },
+    value: { ...value, type: 'string' },
```
brennan, line 15: Why should the type of `value` here be 'string' rather than its default, 'number'?

packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts
nick-Ag: Inside `ga4-properties` we could create two `items` InputFields which are identical but have the different defaults. So like `items_single_item` & `items_multi_item`. I lean towards that option, any thoughts?

brennan (dismissed):
Generally looks really good to me! I'll be happy to approve once stage testing is completed.

Otherwise, a few things:

(1) Let's make sure that all generated types are up-to-date.
(2) Generally, I'm a bit confused about how default paths are being handled. Sometimes we assign a default path in the action, and sometimes we don't. In particular, I'm seeing inconsistencies with `value` and `currency` (perhaps others, too).

packages/destination-actions/src/destinations/google-analytics-4/addToCart/index.ts
```diff
@@ -3,141 +3,54 @@ import type { Settings } from '../generated-types'
...
     },
-    value: {
-      label: 'Value',
-      type: 'number',
-      description: 'The monetary value of the event, in units of the specified currency parameter.',
-      default: {
-        '@path': '$.properties.value'
-      }
-    }
+    value: { ...value, default: { '@path': '$.properties.value' } }
```
brennan, line 53: There is no default path for `value` in `ga4-properties`, so we're proactively setting a default path here. In other actions, we don't set a default value. Should we be consistent?

packages/destination-actions/src/destinations/google-analytics-4/beginCheckout/index.ts
```diff
@@ -9,148 +10,53 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      label: 'Currency',
-      type: 'string',
-      description: 'Currency of the purchase or items associated with the event, in 3-letter ISO 4217 format.',
-      default: {
-        '@path': '$.properties.currency'
-      }
-    },
+    client_id: { ...client_id, required: true },
+    coupon: { ...coupon, default: { '@path': '$.properties.coupon' } },
+    currency: { ...currency, default: { '@path': '$.properties.currency' } },
```
brennan, line 15: Here, we set a default path for `currency`. There is no default path for `currency` in `ga4-properties`. In other actions, we don't set a default path for `currency`. Should we be consistent?

packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
```diff
@@ -0,0 +1,206 @@
...
+    '@path': '$.properties.payment_method'
+  }
+}
+
+export const items: InputField = {
+  label: 'Products',
+  description: 'The list of products purchased.',
+  type: 'object',
+  multiple: true,
+  //required: true, needed for: addPaymentInfo, addToWishlist
```
brennan, line 108: Remove commented-out line? I see we specify `required: true` in the actions themselves where the items array is indeed required.

packages/destination-actions/src/destinations/google-analytics-4/removeFromCart/index.ts
```diff
@@ -9,131 +10,45 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      label: 'Currency',
-      type: 'string',
-      description: 'Currency of the purchase or items associated with the event, in 3-letter ISO 4217 format.'
-    },
-    value: {
-      label: 'Value',
-      type: 'number',
-      description: 'The monetary value of the event.'
-    },
+    client_id: { ...client_id, required: true },
```
brennan, line 13: If `client_id` is always required, should we just specify that in the `ga4-properties` file?

packages/destination-actions/src/destinations/google-analytics-4/selectPromotion/index.ts
```diff
@@ -9,165 +10,66 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      label: 'Promotion ID',
-      type: 'string',
-      description: 'The ID of the promotion associated with the event.'
-    },
-    promotion_name: {
-      label: 'Promotion Name',
-      type: 'string',
-      description: 'The name of the promotion associated with the event.'
-    },
+    promotion_id: { ...promotion_id },
```
brennan, line 21: Curious why no defaults for `promotion_id` and `promotion_name`? I see a default defined in the `viewPromotion` action. But further, wondering why we don't just define default paths for promotion_id and promotion_name in the `ga4-properties` file?

packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts
brennan: We could just abstract [this part](https://github.com/segmentio/action-destinations/blob/50f5a66751cdaba2fffc7bd87be3d5a647f908ca/packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts#L37-L68), right? That's the same for all ecommerce events - those with a products array, as well as single-product events.

packages/destination-actions/src/destinations/google-analytics-4/addPaymentInfo/index.ts
nick-Ag: This is a weird one, the type of value on this action was 'string' so I just copied that over. But the GA4 spec shows that we should be sending a number: [link](https://developers.google.com/analytics/devguides/collection/ga4/reference/events#add_payment_info). But when I actually test it the event goes through without error. I'll investigate a bit more why this field is string for only this action

Screen Shot 2021-12-01 at 12 00 23 PM
Screen Shot 2021-12-01 at 12 00 31 PM

packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
kdaswani: Yes, we still need this as builder-defined for now until customers can add their own custom subfields. 
I'm pretty sure that even though we don't set a default in each action for `item_category2` etc. they still appear because they are defined within `items`. I see that in staging:
Screen Shot 2021-12-01 at 12 16 16 PM

packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts
kdaswani: ^ Yes, that's correct because the subfield defaults are all the same. The only difference is that single-product events have an arrayPath of `properties` and multi-product events have an arrayPath of `properties.products`. Happy to have us abstract that so it's not repetitive in each action if you'd like!

packages/destination-actions/src/destinations/google-analytics-4/beginCheckout/index.ts
kdaswani: good point - let's move this default to `ga4-properties` and always default to `$.properties.currency`

packages/destination-actions/src/destinations/google-analytics-4/addToCart/index.ts
kdaswani: good point - let's move this default to `ga4-properties` and always default to `$.properties.value`

packages/destination-actions/src/destinations/google-analytics-4/selectPromotion/index.ts
kdaswani: We could do this instead. `promotion_id` and `promotion_name` are only used in viewPromotion and selectPromotion so it makes sense to just keep defaults consistent for both and move the defaults to `ga4-properties`. Can we do this?

packages/destination-actions/src/destinations/google-analytics-4/refund/index.ts
kdaswani: Let's put a default for `value` in `ga4-properties` to be consistent across the board

packages/destination-actions/src/destinations/google-analytics-4/selectPromotion/index.ts
nick-Ag: We can do that yes, but `promotion_id` and `promotion_name` are reused inside the `items` array, so setting a global default would copy the `$.properties.name/promotion_id` value at both the top level and as a property shared between every array item in `items`. We would get something like

```
payload: {
  promotion_name: 'abc',
  items: [ {promotion_name: 'abc'}, ... ]}
}
```

packages/destination-actions/src/destinations/google-analytics-4/beginCheckout/index.ts
nick-Ag: I think there are Segment events (like Cart Viewed) that don't have a currency property? So for the View Cart action for example having the global default wouldn't ever map to anything?

kdaswani: That's fine! I know some customers add properties that aren't in our defined spec. So we can put it in there was a default and if the customer doesn't have `properties.currency` in the event, we would just ignore or they could map something else.

packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
```diff
@@ -0,0 +1,287 @@
...
+  type: 'string',
+  description: 'Currency of the items associated with the event, in 3-letter ISO 4217 format.'
+}
+
+export const value: InputField = {
+  label: 'Value',
+  type: 'number',
+  description: 'The monetary value of the event.',
+  default: {
+    '@path': '$.properties.value'
```
kdaswani, line 88: thank you!

packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts
```diff
@@ -11,182 +22,20 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      type: 'number',
-      description: 'The monetary value of the event, in units of the specified currency parameter.',
-      default: {
-        '@path': '$.properties.total'
-      }
-    }
+    transaction_id: { ...transaction_id, required: true },
+    shipping: { ...shipping },
+    tax: { ...tax },
+    value: { ...value, default: { '@path': '$.properties.total' } }
```
kdaswani, line 38: To confirm, this will override `properties.value` with `properties.total` as the default right? For the Purchase action? That is what we want for Purchase.

packages/destination-actions/src/destinations/google-analytics-4/refund/index.ts
kdaswani: @nick-Ag can we do the same thing here that we do for Purchase? Override `properties.value` with `properties.total` as the default? That is what we want for Purchase and Refund based on our spec.

packages/destination-actions/src/destinations/google-analytics-4/selectPromotion/index.ts
kdaswani: Ah good point, didn't realize that those two were also in the `items` array. Let's not do that then. Especially because I think this might break the `items` array since it would try to look in `properties.properties.promotion_name` or `properties.properties.products.promotion_name` right? 

Let's just leave this as is, but add the same defaults to `selectPromotion` as we have for `viewPromotion`.

packages/destination-actions/src/destinations/google-analytics-4/addPaymentInfo/index.ts
brennan: Yeah, it seems that if `number` also works, we should use `number` instead of string. Passing a string may be causing side effects downstream in Google that we can't see/are unaware of.

packages/destination-actions/src/destinations/google-analytics-4/purchase/index.ts
nick-Ag: Yes that's correct

kdaswani: Great - that's perfect

packages/destination-actions/src/destinations/google-analytics-4/customEvent/index.ts
```diff
@@ -1,25 +1,14 @@
...
-          exists: { '@path': '$.userId' },
-          then: { '@path': '$.userId' },
-          else: { '@path': '$.anonymousId' }
-        }
-      }
-    },
+    clientId: { ...client_id },
     name: {
       label: 'Event Name',
       description: 'The unique name of the custom event created in GA4.',
```
kdaswani, line 14: Can we add 
```default: { '@path': '$.event' }```

nick-Ag:
> Generally looks really good to me! I'll be happy to approve once stage testing is completed.
> 
> Otherwise, a few things:
> 
> (1) Let's make sure that all generated types are up-to-date. (2) Generally, I'm a bit confused about how default paths are being handled. Sometimes we assign a default path in the action, and sometimes we don't. In particular, I'm seeing inconsistencies with `value` and `currency` (perhaps others, too).

Confirmed all `generated-types` are included, and all good in staging. This is ready to go

marinhero (approved):

PR #355, opened 2021-11-24
[mixpanel] some functionality changes and fields added

Alex-Kim-mp:
- Overall: many field names and descriptions changed
- Alias: cleaned up
- Group: group Id added (explanation in comments)
- Identify: was not working properly before, added anonymousId to be tied to the identified Id, as well as an additional API call that actually identifies the user
- Track: groupId added that adds additional value to tying track calls to groups

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
@@ -7,38 +7,49 @@ const action: ActionDefinition<Settings, Payload> = {
   description: 'Updates or adds properties to a group profile. The profile is created if it does not exist.',
   defaultSubscription: 'type = "group"',
   fields: {
+    group_id: {
+      label: 'Group ID',
+      type: 'string',
+      description: 'The unique identifier of the group.',
+      default: {
+        '@path': '$.groupId'
+      }
+    },
     group_key: {
       label: 'Group Key',
       type: 'string',
-      description: 'The group key. If this is not set, it is defaulted to "name".',
-      default: 'name'
+      description:
+        'The group key you specified in Mixpanel under Project settings. If this is not specified, this is defaulted to "$group_id".'
     },
     traits: {
       label: 'Group Properties',
       type: 'object',
-      description:
-        'Properties to set on the group profile. Make sure to have a field that corresponds to the above group key.',
+      description: 'The properties to set on the group profile.',
       default: {
         '@path': '$.traits'
       }
     }
   },
   perform: (request, { payload, settings }) => {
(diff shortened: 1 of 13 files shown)


packages/destination-actions/src/destinations/mixpanel/identifyUser/index.ts
```diff
@@ -26,17 +35,42 @@ const action: ActionDefinition<Settings, Payload> = {
...
   },
 
-  perform: (request, { payload, settings }) => {
-    const data = {
-      $token: settings.projectToken,
-      $distinct_id: payload.user_id,
-      $set: payload.traits
+  perform: async (request, { payload, settings }) => {
+    if (!settings.projectToken) {
+      throw new IntegrationError('Missing project token')
```
mshwery, line 40: I'd recommend passing a status code as the 2nd arg (for validation errors you should probably use `400` or `422`).

mshwery (approved):
lgtm! left a couple minor comments

jinapark202 (approved):
LGTM. I would just make the small change that Matt suggested and then we could deploy this change after the code freeze tomorrow

packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
```diff
@@ -7,38 +7,49 @@ const action: ActionDefinition<Settings, Payload> = {
...
     }
-    if (!payload.traits[group_key]) {
-      throw new IntegrationError('Group traits does not include proper group key', 'Misconfigured required field')
-    }
+    /*
+     * Getting the group id from payload.traits[group_key] is primarily for backwards compatibility with the “classic” Segment integration.
+     * In that integration, you specify the id as a trait like {[group_key]: group_id} and the group_id from the Segment payload was ignored.
+     * Now, we simplify this by using the group_id from the segment payload directly.
+     */
+    const group_id = payload.traits[group_key] || payload.group_id
```
jinapark202, line 43: Do you want to require `payload.group_id`? Because if both `payload.traits[group_key]` and `payload.group_id` are null, `group_id` will also be null (unless that's what you want?)

Alex-Kim-mp: we're expecting the scenario to be that the user doesn't change it and maintains it to be the default mapping, so perhaps we could generate a random ID just as a fallback

jbwyme: I think we can safely mark `payload.group_id` as required right? On the Segment side, there will always be a `$.groupId` right?

PR #356, opened 2021-11-24
[BE-246] Amplitude Actions EU Endpoint

daltonscharff:
[\[BE-246\]](https://segment.atlassian.net/browse/BE-246)

## Summary
This change allows users to specify whether to use the `north_america` or `europe` regional endpoint for an Amplitude destination. 

In order to support different mappings between NA and EU endpoints, I created a mapping for each Amplitude endpoint used in amplitude-actions along with a `getEndpointByRegion` function. This function creates a safer way to access the endpoint map, as it defaults the region to `north_america` if the region is undefined or otherwise invalid.

## Testing

[BE-246]: https://segment.atlassian.net/browse/BE-246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/destination-actions/src/destinations/amplitude/index.ts
```diff
@@ -68,27 +69,60 @@ const destination: DestinationDefinition<Settings> = {
           'Amplitude project secret key. You can find this key in the "General" tab of your Amplitude project.',
         type: 'string',
         required: true
+      },
+      endpoint: {
+        label: 'Endpoint Region',
+        description: 'Choose the region where you would like to send your data.',
```
gpsamson, line 75:
```suggestion
        description: 'The region to send your data.',
```

packages/destination-actions/src/destinations/amplitude/index.ts
```diff
@@ -68,27 +69,60 @@ const destination: DestinationDefinition<Settings> = {
...
+          {
+            label: 'North America',
+            value: 'north_america'
+          },
+          {
+            label: 'Europe',
+            value: 'europe'
+          }
+        ],
+        default: 'https://api.amplitude.com'
```
gpsamson, line 88: Should this default be `north_america`?

packages/destination-actions/src/destinations/amplitude/create-endpoint.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Generates a region-specific Amplitude endpoint
+ *
+ * @param path path to api endpoint (e.g., '/usermap')
+ * @param region specified region ('north_america' | 'europe')
+ * @param options change protocol or subdomains of endpoint
+ * @returns api endpoint string
+ */
+export default function createEndpoint(
```
gpsamson, line 27: This function overcomplicates endpoints quite a bit compared to a static map. I would vote for / recommend a maps:

```
export const endpoints = {
        north_america: 'https://api2.amplitude.com',
        eu: 'https://api.eu.amplitude.com'
}
```

With this, it's clear what endpoint will be used per region and makes it easier for othe developers to implement more actions without needing to parse through a `createEndpoint` function. For example:
```
request(`${endpoints[settings.region]}/2/usersearch`)
```

Note the static `https` protocol since we should always be using secure connections.

packages/destination-actions/src/destinations/amplitude/index.ts
daltonscharff: Yes. Good catch.

packages/destination-actions/src/destinations/amplitude/create-endpoint.ts
daltonscharff: Great points. The mapping makes everything more readable/searchable. I've updated solution as recommended.

packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
```diff
@@ -477,7 +573,7 @@ describe('Amplitude', () => {
         }
       })
 
-      nock('https://api.amplitude.com').post('/identify').reply(200, {})
+      nock('https://api2.amplitude.com').post('/identify').reply(200, {})
```
daltonscharff, line 576: This change comes from the Amplitude documentation: https://developers.amplitude.com/docs/identify-api

packages/destination-actions/src/destinations/amplitude/regional-endpoints.ts
```diff
@@ -0,0 +1,46 @@
+export const endpoints = {
```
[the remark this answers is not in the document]
gpsamson: I would also consider setting up a snapshot test for this map because there are so many endpoint variations.

packages/destination-actions/src/destinations/amplitude/index.ts
```diff
@@ -68,19 +69,36 @@ const destination: DestinationDefinition<Settings> = {
...
+          }
+        ],
+        default: 'north_america'
       }
     },
     testAuthentication: (request, { settings }) => {
       // Note: Amplitude has some apis that use basic auth (like this one)
       // and others that use custom auth in the request body
-      return request('https://amplitude.com/api/2/usersearch?user=testUser@example.com', {
+      return request(getEndpointByRegion('usersearch', settings.endpoint) + '?user=testUser@example.com', {
```
gpsamson, line 94: To make this a bit more readable:

```suggestion
      const endpoint = getEndpointByRegion('usersearch', settings.endpoint) 
      return request(`${endpoint}?user=testUser@example.com`, {
```

gpsamson (approved):

packages/destination-actions/src/destinations/amplitude/regional-endpoints.ts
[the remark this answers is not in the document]
mshwery: I'm not opposed to this, but agree it would be nice to colocate endpoints.

might be nice to have a central file that:
- defines the region endpoint type/interface + a getter function for grabbing the region (or fallback) from settings
- each action can import the type + getter function
- the action can use the result of the getter (a region key) to grab the right url, colocated in the action.

That's the only other suggestion I would have, but again, this is also probably fine as-is.

packages/destination-actions/src/destinations/amplitude/regional-endpoints.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Retrieves Amplitude API endpoints for a given region. If the region
+ * provided does not exist, the region defaults to 'north_america'.
+ *
+ * @param endpoint name of the API endpoint
+ * @param region data residency region
+ * @returns regional API endpoint
+ */
+export function getEndpointByRegion(endpoint: keyof typeof endpoints, region?: string): string {
```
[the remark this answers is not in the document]
daltonscharff: Using `keyof typeof endpoints` as the type for `endpoint` made it so I could only pass in one of the defined keys of `endpoints`, which I found to be beneficial. If endpoint type is a string, the getEndpointByRegion function could return undefined. I would rather the compiler warn me of errors as I type than wait until they appear in tests.

Also, using `region?: Region` in the parameters might be problematic as it expects the value from `settings.endpoint` which is a string. This means every time you'd like to call the function, you would first need to cast `settings.endpoint` as type Region. I feel that since `settings.endpoint` has the opportunity to be something other than `north_america | europe`, it may be best for it to just accept a string.

PR #358, opened 2021-11-25
Pipedrive Destination Actions: removed default fields

makks129:
## Description

`default` field in `*_match_field` payload fields is causing a bug in the logic side.
We want users to be able to select a value for `*_match_field` optionally (that's why we use `required: false`), however `default` field was setting its value that couldn't be removed in the UI.

## Changes

Removed `default` field in `*_match_field` payload fields.

### DIFF
--- a/packages/destination-actions/src/destinations/pipedrive/createUpdateActivity/index.ts
+++ b/packages/destination-actions/src/destinations/pipedrive/createUpdateActivity/index.ts
@@ -23,7 +23,6 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       required: false,
       dynamic: true,
-      default: 'id'
     },
     person_match_value: {
       label: 'Person match value',
@@ -41,7 +40,6 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       required: false,
       dynamic: true,
-      default: 'id'
     },
     organization_match_value: {
       label: 'Organization match value',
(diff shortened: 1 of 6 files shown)


mshwery:
@makks129 can you link me to a destination where it wasn't allowing you to clear the `id` default? You should be able to clear any input.

makks129:
> @makks129 can you link me to a destination where it wasn't allowing you to clear the `id` default? You should be able to clear any input.

Yes sure, if you [go to](https://app.segment.com/pipedrive1/destinations/actions-pipedrive/sources/webapp_test_colorbox/instances/60ad09b7608aeae66e969bf6/actions/9a6TQYNzb9XMrGM4XNcw5o) `Create or update an Activity` and look at the `Person match field`, you will see `id` there, which cannot be removed (I guess because it's a `default` value).

mshwery:
@makks129 it's maybe slightly hidden (could use a better ux / more affordance!) but you can clear the value even for dynamic fields like this one. You shouldn't need to change the default. If anything we should improve the ux for customers instead // cc @char-mainewang @nielst @mericsson 

image

makks129:
> @makks129 it's maybe slightly hidden (could use a better ux / more affordance!) but you can clear the value even for dynamic fields like this one. You shouldn't need to change the default. If anything we should improve the ux for customers instead // cc @char-mainewang @nielst @mericsson
> 
> image

Right, good point.
But right now that's actually not what we want. We don't want to have default value here, because it replaces the value from settings (as the description of this field says: "If present, used instead of field in settings..."). So we should remove default values.

mshwery:
@makks129 ok! that makes more sense. Thanks for the explanation.

makks129:
> @makks129 ok! that makes more sense. Thanks for the explanation.

Can we get a review please? :)

mshwery (approved):
cc @segmentio/build-experience-team

PR #359, opened 2021-11-29
permit singular object roots for arrayPath

mshwery:
This will allow arrayPath's root object to be a singular object as well, which supports some new integrations.

## Testing


dlasky (approved):
looks ok, i think we should probably update docs too because as a dev that would confuse me slightly initially

PR #361, opened 2021-11-30
[Customer.io] Use created_at trait over property

mike-engel:
We've gotten a few support tickets in at Customer.io with customers confused as to why `created_at` is constantly being updated, even though it's the same person being identified. The problem is that the `created_at` property seems to be the creation time of the event, not of the identified user.

This changes the `createUpdatePerson` action to use `traits.created_at`, which is set by the user. If the request goes out without a `created_at`, customer.io uses the current date and time if it's a new person. Otherwise, it's ignored.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
+++ b/packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
@@ -38,10 +38,10 @@ const action: ActionDefinition<Settings, Payload> = {
     },
     created_at: {
       label: 'Created At',
-      description: 'A timestamp of when the person was created. Default is current date and time.',
+      description: 'A timestamp of when the person was created. Default is the created_at trait.',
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
```diff
@@ -38,10 +38,10 @@ const action: ActionDefinition<Settings, Payload> = {
     },
     created_at: {
       label: 'Created At',
-      description: 'A timestamp of when the person was created. Default is current date and time.',
+      description: 'A timestamp of when the person was created. Default is the created_at trait.',
```
mshwery, line 41: I would consider omitting language about the default. This makes it seem like if unset, `created_at` will be used as fallback, which is not true.

mshwery (approved):

PR #367, opened 2021-12-01
[fb capi] external id update

nick-Ag:
Updating the defaults for the `user_data.externalId` field. This update applies across all FB CAPI actions.
Removed the `properties` from `properties.anonymousId` and `properties.userId`

## Testing

Pushed to stage and successfully tested the mapping. 
Screen Shot 2021-12-01 at 9 29 11 AM


mshwery:
I don't think users can get back to this state via the UI by themselves. If we want coalesce support I recommend you ensure it's enabled in the UI too!

nick-Ag:
> I don't think users can get back to this state via the UI by themselves. If we want coalesce support I recommend you ensure it's enabled in the UI too!

@dlasky Is coalesce for object properties a known issue for build experience?

dlasky:
@nick-Ag wasn't on my radar. I think we need to think long and hard about allowing that as it will potentially cause performance problems as our volume grows

nick-Ag:
> @nick-Ag wasn't on my radar. I think we need to think long and hard about allowing that as it will potentially cause performance problems as our volume grows

Should we avoid using `coalesce` for object properties? Users would have to reconfigure the action to get that `coalesce` back as Matt mentioned. That workaround seems clunky

dlasky:
yeah its clunky, but also as soon as we start nesting directives we've opened up a whole can of worms, like technically it 'works' as far as mapping kit can process it correctly, but we do have to think about the impact we might have on the platform from doing so and carefully balance that against the user need.

as far as the UI experience, its probably relatively straightforward to add that to object, but there may be some trickyness with the state for the individual keys.

dlasky (approved):

PR #368, opened 2021-12-01
Update sessionID plugin

danieljackins:
This PR updates amplitude session ID plugin to not add the session ID to the integration object when All: false

## Testing

Session ID not populated when all: false
image
Session ID populated when all != false
image
Session ID populated when All: false and Actions Amplitude: true
image

### DIFF
--- a/packages/browser-destinations/src/destinations/amplitude-plugins/sessionId/index.ts
+++ b/packages/browser-destinations/src/destinations/amplitude-plugins/sessionId/index.ts
@@ -57,7 +57,10 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
 
     ls.setItem('analytics_session_id.last_access', newSession.toString())
 
-    context.updateEvent('integrations.Actions Amplitude.session_id', id)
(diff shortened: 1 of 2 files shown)


pooyaj (approved):

packages/browser-destinations/src/destinations/amplitude-plugins/sessionId/index.ts
```diff
@@ -57,7 +57,10 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
 
     ls.setItem('analytics_session_id.last_access', newSession.toString())
 
-    context.updateEvent('integrations.Actions Amplitude.session_id', id)
+    if (context.event.integrations?.All !== false || context.event.integrations['Actions Amplitude']) {
```
mshwery, line 60: I thought this was a valid case where we would want to update the session id?

```
{
  All: false,
  'Actions Amplitude': true
}
```

danieljackins: Yeah that is valid, it should satisfy the second condition here

PR #371, opened 2021-12-02
Use auto publicpath

danieljackins:
This PR changes the webpack publicPath to auto instead of being hardcoded to the segment CDN, which was breaking for customers using custom domains.

## Testing

Actions bundle is loaded from custom cdn instead of segment cdn: 
image

### DIFF
--- a/packages/browser-destinations/webpack.config.js
+++ b/packages/browser-destinations/webpack.config.js
@@ -7,12 +7,6 @@ const webpack = require('webpack')
 
 const files = globby.sync('./src/destinations/*/index.ts')
 const isProd = process.env.NODE_ENV === 'production'
-const assetPath =
-  process.env.ASSET_ENV === 'production'
-    ? 'https://cdn.segment.com/next-integrations/actions/'
-    : process.env.ASSET_ENV === 'stage'
-    ? 'https://cdn.segment.build/next-integrations/actions/'
-    : undefined
 
 const entries = files.reduce((acc, current) => {
   const [_dot, _src, _destinations, destination, ..._rest] = current.split('/')
@@ -48,7 +42,7 @@ module.exports = {
   output: {
     filename: process.env.NODE_ENV === 'development' ? '[name].js' : '[name]/[contenthash].js',
     path: path.resolve(__dirname, 'dist/web'),
-    publicPath: assetPath,
+    publicPath: 'auto', // Needed for customers using custom CDNs with analytics.js
     library: '[name]Destination',
     libraryTarget: 'umd',
     libraryExport: 'default'


packages/browser-destinations/webpack.config.js
```diff
@@ -48,7 +42,7 @@ module.exports = {
   output: {
     filename: process.env.NODE_ENV === 'development' ? '[name].js' : '[name]/[contenthash].js',
     path: path.resolve(__dirname, 'dist/web'),
-    publicPath: assetPath,
+    publicPath: 'auto', // Needed for customers using custom CDNs with analytics.js
```
mshwery, line 45: Can you share how this changes the url in the bundles themselves?

danieljackins: the url is determined at runtime, here is the webpack code - 
```
/* webpack/runtime/publicPath */
/******/ 	(() => {
/******/ 		var scriptUrl;
/******/ 		if (__webpack_require__.g.importScripts) scriptUrl = __webpack_require__.g.location + "";
/******/ 		var document = __webpack_require__.g.document;
/******/ 		if (!scriptUrl && document) {
/******/ 			if (document.currentScript)
/******/ 				scriptUrl = document.currentScript.src
/******/ 			if (!scriptUrl) {
/******/ 				var scripts = document.getElementsByTagName("script");
/******/ 				if(scripts.length) scriptUrl = scripts[scripts.length - 1].src
/******/ 			}
/******/ 		}
/******/ 		// When supporting browsers where an automatic publicPath is not supported you must specify an output.publicPath manually via configuration
/******/ 		// or pass an empty string ("") and set the __webpack_public_path__ variable from your code to use your own logic.
/******/ 		if (!scriptUrl) throw new Error("Automatic publicPath is not supported in this browser");
/******/ 		scriptUrl = scriptUrl.replace(/#.*$/, "").replace(/\?.*$/, "").replace(/\/[^\/]+$/, "/");
/******/ 		__webpack_require__.p = scriptUrl + "../";
/******/ 	})();
```

```
var url = __webpack_require__.p + __webpack_require__.u(chunkId);
```

mshwery: so they become relative? hard to tell at a glance here.

mshwery: how does this impact the remote_plugin url?

danieljackins: the way the url is set in destination_definition_remote_plugins isn't affected by this change (will still always be cdn.segment.build or cdn.segment.com, analytics-next handles the custom domains for that portion)

pooyaj (approved):

chrisradek (approved):

PR #382, opened 2021-12-07
Pipedrive Destination Actions: custom fields

makks129:
**Context**
Pipedrive offers saving data for various entities (persons, organizations, deals, leads) into, so called, custom fields.
We want our clients to be able to map values from their Segment events into Pipedrive custom fields.

**Implementation**
Because there can be a varying amount of custom fields, based on client's case, and Segment doesn't yet support adding additional action fields dynamically, we came up with this workaround:
We provide an action field called `Custom fields` which accepts JSON mapping as its value.
JSON mapping will look like this:
```
{
  custom_field_key1: value,
  custom_field_key2: value
}
````
and will allow clients to map arbitrary amount of custom fields.
In the destination action code we attach those custom fields to an entity that we want to create/update.
(This implementation was discussed with @mericsson)

## Testing

### CODE REVIEW

File: packages/destination-actions/src/destinations/pipedrive/createUpdateDeal/index.ts [hunk 1]
```diff
@@ -138,7 +136,14 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'If the deal is created, use this timestamp as the creation timestamp. Format: YYY-MM-DD HH:MM:SS',
       type: 'datetime',
       required: false
-    }
+    },
+
+    custom_fields: {
+      label: 'Custom fields',
+      description: 'New values for custom fields in JSON format.',
+      type: 'string',
```
mshwery, line 144: What are some example values here?

mshwery: I'm mostly curious how you see users entering this value. I don't think there is any json string in the Segment event that they would map to, unless they are including that in event properties themselves?

makks129: The idea is to do something like this:
image
As far as we understood it's going to work
Wdyt?

mshwery: Asking users to enter a json string isn't a great ux, imo. I would switch this to an object type field and if you can define the properties it expects, even better!

dlasky: we recently added support for user defined object keys as well in the mapping UI which is probably what you want here

mshwery: This is what it looks like for users as Dan mentioned: 
Screen Shot 2021-12-08 at 8 16 09 AM

makks129: That's fantastic guys, that's what we need.
I'll make an update.

makks129: Done

mshwery (approved):
lgtm

PR #386, opened 2021-12-08
[Customer.io] Convert all potential timestamps in payloads/traits

mike-engel:
We've had at least one customer notice that the old customer.io action in Segment.io would convert _all_ timestamps from a string to a unix timestamp, and that the new destination action only converts specific known dates.

We've decided as a team that it's more useful to try and convert everything, so this PR extends the `convert_timestamp` flag to try and convert all possible dates to unix timestamps. This is accomplished with a recursive function that will walk the entire tree of the data object being parsed.

**Note**: It looks like `gitleaks` recently made quite a few breaking changes, and no longer works with the existing configs. I've skipped that verification step with this work. The change is unrelated and should probably be managed by y'all instead 😄 

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
+++ b/packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
@@ -2,7 +2,7 @@ import dayjs from '../../../lib/dayjs'
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import { trackApiEndpoint } from '../utils'
+import { convertAttributeTimestamps, trackApiEndpoint } from '../utils'
 
 const action: ActionDefinition<Settings, Payload> = {
   title: 'Create or Update Person',
@@ -55,7 +55,7 @@ const action: ActionDefinition<Settings, Payload> = {
     },
     convert_timestamp: {
       label: 'Convert Timestamps',
-      description: 'Convert `created_at` to a Unix timestamp (seconds since Epoch).',
+      description: 'Convert dates to Unix timestamps (seconds since Epoch).',
       type: 'boolean',
       default: true
     }
@@ -63,14 +63,22 @@ const action: ActionDefinition<Settings, Payload> = {
 
   perform: (request, { settings, payload }) => {
     let createdAt: string | number | undefined = payload.created_at
-    const body: Record<string, unknown> = {
-      ...payload.custom_attributes,
-      email: payload.email,
-      anonymous_id: payload.anonymous_id
+    let customAttributes = payload.custom_attributes
+
+    if (payload.convert_timestamp !== false) {
+      if (createdAt) {
+        createdAt = dayjs.utc(createdAt).unix()
+      }
+
+      if (customAttributes) {
+        customAttributes = convertAttributeTimestamps(customAttributes)
+      }
     }
 
-    if (createdAt && payload.convert_timestamp !== false) {
-      createdAt = dayjs.utc(createdAt).unix()
+    const body: Record<string, unknown> = {
+      ...customAttributes,
+      email: payload.email,
+      anonymous_id: payload.anonymous_id
     }
 
     if (createdAt) {
--- a/packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
+++ b/packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
@@ -1,7 +1,7 @@
 import dayjs from '../../../lib/dayjs'
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
-import { trackApiEndpoint } from '../utils'
+import { convertAttributeTimestamps, trackApiEndpoint } from '../utils'
 import type { Payload } from './generated-types'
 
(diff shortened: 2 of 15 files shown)


packages/destination-actions/src/destinations/customerio/utils.ts
```diff
@@ -10,3 +13,37 @@ export enum AccountRegion {
...
+
+// Recursively walk through an object and try to convert any strings into dates
+export const convertAttributeTimestamps = (payload: Record<string, unknown>): Record<string, unknown> => {
+  const clone: Record<string, unknown> = {}
+  const keys = Object.keys(payload)
+
+  keys.forEach((key) => {
+    const value = payload[key]
+
+    if (typeof value === 'string') {
```
mshwery, line 29: Do we care about unix numbers?

packages/destination-actions/src/destinations/customerio/utils.ts
```diff
@@ -10,3 +13,37 @@ export enum AccountRegion {
...
+      const maybeDate = dayjs.utc(value)
+
+      if (maybeDate.isValid()) {
+        clone[key] = maybeDate.unix()
+
+        return
+      }
```
mshwery, line 36: Do we need any try/catching around the date parsing?

mshwery (approved):
lgtm. cc @segmentio/build-experience-team

packages/destination-actions/src/destinations/customerio/utils.ts
mike-engel: I don’t think so. I added some plain strings to the payloads in the tests and everything seemed to be ok. I think dayjs doesn’t throw, and instead keeps track of whether it’s valid or not

packages/destination-actions/src/destinations/customerio/utils.ts
mike-engel: Not here, since we’re converting date strings into Unix timestamps. If it’s already a number, then that’s great!

packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
```diff
@@ -55,24 +55,32 @@ const action: ActionDefinition<Settings, Payload> = {
...
       type: 'boolean',
       default: true
     }
   },
 
   perform: (request, { settings, payload }) => {
     let createdAt: string | number | undefined = payload.created_at
+    let customAttributes = payload.custom_attributes
+
+    if (payload.convert_timestamp !== false) {
```
jinapark202, line 68:
```suggestion
    if (payload.convert_timestamp) {
```

packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
```diff
@@ -63,22 +63,29 @@ const action: ActionDefinition<Settings, Payload> = {
...
     }
   },
 
   perform: (request, { settings, payload }) => {
     let timestamp: string | number | undefined = payload.timestamp
+    let data = payload.data
 
-    if (timestamp && payload.convert_timestamp !== false) {
-      timestamp = dayjs.utc(timestamp).unix()
+    if (payload.convert_timestamp !== false) {
```
jinapark202, line 76:
```suggestion
    if (payload.convert_timestamp) {
```

packages/destination-actions/src/destinations/customerio/trackPageView/index.ts
```diff
@@ -63,22 +63,29 @@ const action: ActionDefinition<Settings, Payload> = {
...
       default: true
     }
   },
   perform: (request, { settings, payload }) => {
     let timestamp: string | number | undefined = payload.timestamp
+    let data = payload.data
 
-    if (timestamp && payload.convert_timestamp !== false) {
-      timestamp = dayjs.utc(timestamp).unix()
+    if (payload.convert_timestamp !== false) {
```
jinapark202, line 75:
```suggestion
    if (payload.convert_timestamp) {
```

packages/destination-actions/src/destinations/customerio/trackScreenView/index.ts
```diff
@@ -63,22 +63,29 @@ const action: ActionDefinition<Settings, Payload> = {
...
       default: true
     }
   },
   perform: (request, { settings, payload }) => {
     let timestamp: string | number | undefined = payload.timestamp
+    let data = payload.data
 
-    if (timestamp && payload.convert_timestamp !== false) {
-      timestamp = dayjs.utc(timestamp).unix()
+    if (payload.convert_timestamp !== false) {
```
jinapark202, line 75:
```suggestion
    if (payload.convert_timestamp) {
```

packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
mike-engel: @jinapark202 this (and the others you pointed out) are backwards compatibility fixes. `convert_timestamp` was added after the initial release, so it could be `undefined` for early users of this action. The default should be `true`, so the only time we *don't* want to convert timestamps is if the value is `false`. If it's `true` or `undefined`, we should

jinapark202 (approved):

PR #387, opened 2021-12-10
Salesforce (Actions): Lead Action

nick-Ag:
This PR implements the Salesforce (Actions) Lead action. It is the first action implemented for the destination and will serve as prior art for the following actions.

The action supports the following CRUD operations:

- [x] Create
- [x] Update
- [x] Upsert

## Testing

Tested locally by running `./bin/run serve` and passing in an `auth` object with shape:
```
auth: {
 clientId
 clientSecret
 accessToken
 refreshToken
}
```
Screen Shot 2021-12-10 at 3 49 30 PM

Tested in stage via deploying to stage. The Oauth flow is working, and all operations (create, update, upsert) are working correctly when tested.
Screen Shot 2021-12-07 at 10 15 26 AM


packages/destination-actions/src/destinations/salesforce/index.ts
```diff
@@ -0,0 +1,61 @@
...
+      const res = await request<RefreshTokenResponse>('https://login.salesforce.com/services/oauth2/token', {
+        method: 'POST',
+        body: new URLSearchParams({
+          refresh_token: auth.refreshToken,
+          client_id: auth.clientId,
+          client_secret: auth.clientSecret,
+          grant_type: 'refresh_token'
+        })
+      })
+      console.log('res', res)
```
mericsson, line 43:
```suggestion
```

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -0,0 +1,98 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const API_VERSION = 'v53.0'
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Lead',
+  description: 'Lead action',
+  fields: {
+    crud: {
```
mericsson, line 10:
```suggestion
    operation: {
```

Wonder if this field name is more appropriate given:
1. CRUD isn't what we are doing since we may not support delete and we support Upsert -- so would be CRUU :)
2. May not be obvious to users

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      label: 'CRUD Operation',
+      description: 'CRUD',
+      type: 'string',
+      required: true,
+      choices: [
+        { label: 'Create', value: 'create' },
+        { label: 'Update', value: 'update' },
+        { label: 'Upsert', value: 'upsert' }
```
mericsson, line 18: I think better to just not support these yet rather than have placeholders. Would rather have thin slices of fully implemented so it's clear as we start testing it what is expected to work.

mericsson, line 18:
```suggestion
        { label: 'Create', value: 'create' }
```

I think better to just not support these yet rather than have placeholders. Would rather have thin slices of fully implemented so it's clear as we start testing it what is expected to work.

packages/destination-actions/src/destinations/salesforce/index.ts
```diff
@@ -0,0 +1,61 @@
+import type { DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+import { setInstanceUrl } from './salesforce-operations'
+import lead from './lead'
+
+interface RefreshTokenResponse {
```
brennan, line 6: Curious if we'll need all these properties? Even though Salesforce's oauth endpoint returns all these values, I don't think we'll need any of them, except for perhaps `instance_url`.

We may consider just using the [existing`RefreshAccessTokenResult`](https://sourcegraph.segment.com/github.com/segmentio/action-destinations@21cf91ae4c15edda3ec0d850b761c48eed61b44d/-/blob/packages/core/src/destination-kit/index.ts?L95) interface, extended with the `instance_url` property.

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      label: 'Street',
+      description: 'Street',
+      type: 'string'
+    },
+    state: {
+      label: 'State',
+      description: 'State',
+      type: 'string'
+    },
+    first_name: {
```
brennan, line 68: Nit - should `first_name` and `last_name` be consecutive in the object?

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -0,0 +1,98 @@
...
+      label: 'First Name',
+      description: 'First Name',
+      type: 'string'
+    }
+  },
+  perform: (request, { settings, payload }) => {
+    if (payload.crud === 'create') {
+      return request(`${settings.instanceUrl}/services/data/${API_VERSION}/sobjects/Lead`, {
```
brennan, line 76: Confirming that the future `settings.instanceUrl` will simple be `auth.instanceUrl`, correct?

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -0,0 +1,98 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const API_VERSION = 'v53.0'
```
brennan, line 5: Curious how often we should expect to update the API version, if that information is known?

packages/destination-actions/src/destinations/salesforce/lead/index.ts
nick-Ag: That's correct, just using `settings` for now to send events

packages/destination-actions/src/destinations/salesforce/sf-types.ts
```diff
@@ -0,0 +1,3 @@
+import type { Payload as LeadPayload } from './lead/generated-types'
+
+export type GenericPayload = Partial<LeadPayload>
```
nick-Ag, line 3: Will be able to do `Partial<LeadPayload | AccountPayload | etc >` when adding more actions

packages/destination-actions/src/destinations/index.ts
```diff
@@ -36,6 +36,7 @@ register('61a8032ea5f157ee37a720be', './metronome')
 register('6101bf0e15772f7e12407fa9', './personas-messaging-sendgrid')
 register('6116a41e2e8fc680d8daf821', './personas-messaging-twilio')
 register('5f7dd8191ad74f868ab1fc48', './pipedrive')
+register('61957755c4d820be968457de', './salesforce')
```
brennan, line 39: Verified `id` in stage partner portal.

packages/destination-actions/src/destinations/salesforce/generated-types.ts
```diff
@@ -0,0 +1,8 @@
+// Generated file. DO NOT MODIFY IT BY HAND.
+
+export interface Settings {
+  /**
+   * Base Url of the request. For local testing only. refreshAccessToken will eventually be able to fetch this value via Oauth
```
brennan, line 5: Probably want to update the description of `instanceUrl` and regenerate the `Settings` type.

packages/destination-actions/src/destinations/salesforce/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  slug: 'actions-salesforce',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'oauth2',
+    fields: {
+      instanceUrl: {
+        label: 'Instance URL',
+        description:
+          'Base Url of the request. For local testing only. refreshAccessToken will eventually be able to fetch this value via Oauth',
```
brennan, line 19: Update `instanceUrl` description.

packages/destination-actions/src/destinations/salesforce/index.ts
```diff
@@ -0,0 +1,54 @@
+import type { DestinationDefinition, RefreshAccessTokenResult } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+import lead from './lead'
+interface SalesforceRefreshAccessTokenResult extends RefreshAccessTokenResult {
```
brennan, line 4: I don't think we need this interface since we don't update `instanceUrl` on token refresh, only when a customer authorizes or re-authorizes via the `app` UI.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,138 @@
...
+        Country: payload.country,
+        PostalCode: payload.postal_code,
+        City: payload.city,
+        Email: payload.email
+      }
+    })
+  }
+
+  updateRecord = async (payload: GenericPayload, sobject: string) => {
+    if (payload.traits === undefined || Object.keys(payload.traits).length === 0) {
```
brennan, line 43: Should we use `!payload` to account for instances where `payload` could be null, an empty string, or other falsy values, or are we pretty sure that it would always be either `undefined` or `{}` here?

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,138 @@
...
+      throw new IntegrationError('Undefined Traits when using update operation', 'Undefined Traits', 400)
+    }
+
+    if (Object.keys(payload.traits).includes('Id') && payload.traits['Id']) {
+      return await this.baseUpdate(payload.traits['Id'] as string, sobject, payload)
+    }
+
+    const [recordId, err] = await this.lookupTraits(payload.traits, sobject)
+
+    if (err !== undefined) {
```
brennan, line 53:
```suggestion
    if (err) {
```

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  upsertRecord = async (payload: GenericPayload, sobject: string) => {
+    if (payload.traits === undefined || Object.keys(payload.traits).length === 0) {
+      throw new IntegrationError('Undefined Traits when using upsert operation', 'Undefined Traits', 400)
+    }
+
+    const [recordId, err] = await this.lookupTraits(payload.traits, sobject)
+
+    if (err === undefined) {
```
brennan, line 67:
```suggestion
    if (err) {
      if (err.status === 404) {
        return await this.createRecord(payload, sobject)
      }
      throw err
    }
  return await this.baseUpdate(recordId, sobject, payload)
}
```

brennan (changes requested):
Generally looks great! Happy to 👍 once the comments are addressed.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
nick-Ag: Agreed, I also added a unit test for cases where it is null or empty

brennan (approved):

PR #389, opened 2021-12-14
Friendbuy cloud destination

dhull:
Add a Friendbuy Cloud Destination to complement the Friendbuy Browser Destination so that merchants can send events to Friendbuy from their data centers.

The testing is not complete but I am submitting this PR now so you can look at how I shared some code between my browser and cloud destinations. The shared code is in `packages/browser-destinations/src/destinations/friendbuy/shared`, and is accessed from the cloud destination by a symlink that I committed to git at `packages/destination-actions/src/destinations/friendbuy/shared`. (I initially tried using relative paths from one tree to the other in my import statements but that was not allowed by the build command.)

JIRA https://segment.atlassian.net/browse/BE-452

## Testing


dlasky:
@dhull thanks for the early preview! 

Symlink definitely doesn't seem ideal, did you try setting up a typescript path to resolve the build issue? I'm playing around with that a little myself, will comment if i can get it going.

dhull:
> Symlink definitely doesn't seem ideal, did you try setting up a typescript path to resolve the build issue? I'm playing around with that a little myself, will comment if i can get it going.

Some background for other reviewers: Friendbuy has both a device/web destination and a cloud destination and I would like to share some code between them. Currently I have a directory `packages/browser-destinations/src/destinations/friendbuy/shared` with the shared code and I have a symlink from `packages/destination-actions/src/destinations/friendbuy/shared` to that directory so that the cloud destination can import the shared code. There are several possible ways that I can think of to allow this code sharing:
* The symlink approach that I described above.
* Add a Friendbuy-specific path to `packages/destination-actions/tsconfig.json` (or `tsconfig.build.json`?) to point to the shared directory `packages/browser-destinations/src/destinations/friendbuy/shared`.
* Add a generic path to `./tsconfig.json` to point to a generic shared directory where all the developers would be able to put shared code and move my shared code to that directory.
* Give up on the idea and just copy the shared code to the individual destinations.

I'd like some guidance on how I should proceed.

dlasky:
Approach 2 is what I was exploring. I think we shouldn't add a shared dependency as that will get messy fast.

dhull:
I've added a commit (https://github.com/segmentio/action-destinations/pull/389/commits/3c1ac0dcb1b63a7ffd1d98e93bbd1b143d2c0158) that uses a tsconfig.json path to allow importing the shared code into the cloud destination. `yarn build` and `yarn test` both run successfully. The config changes are:

Add the `paths` and `references` to `tsconfig.build.json`. I added the `references` entry on a separate line but a precommit hook reformatted it to a single line. I tried adding the `paths` entry only to `tsconfig.json` but `yarn build` failed without it also in `tsconfig.build.json`.

```diff
modified   packages/destination-actions/tsconfig.build.json
@@ -3,11 +3,12 @@
   "compilerOptions": {
     "composite": true,
     "rootDir": "src",
-    "outDir": "dist"
+    "outDir": "dist",
+    "paths": {
+      "@friendbuy/shared/*": ["../browser-destinations/src/destinations/friendbuy/shared/*"]
+    }
   },
   "exclude": ["**/__tests__/**/*.ts"],
   "include": ["src"],
-  "references": [
-    { "path": "../core/tsconfig.build.json" }
-  ]
+  "references": [{ "path": "../core/tsconfig.build.json" }, { "path": "../browser-destinations/tsconfig.build.json" }]
 }
```

Added a `moduleNameMapper` entry to `package.json` so that the jest tests would work.

```diff
modified   packages/destination-actions/package.json
@@ -54,7 +54,8 @@
     "moduleNameMapper": {
       "@segment/ajv-human-errors": "<rootDir>/../ajv-human-errors/src",
       "@segment/actions-core": "<rootDir>/../core/src",
-      "@segment/destination-subscriptions": "<rootDir>/../destination-subscriptions/src"
+      "@segment/destination-subscriptions": "<rootDir>/../destination-subscriptions/src",
+      "@friendbuy/shared(.*)": "<rootDir>/../browser-destinations/src/destinations/friendbuy/shared$1"
     },
     "setupFilesAfterEnv": [
       "<rootDir>/test/setup-after-env.ts"
```

`yarn test` cannot find the imports without the `paths` also updated in `tsconfig.json`.

```diff
modified   packages/destination-actions/tsconfig.json
@@ -5,7 +5,8 @@
     "paths": {
       "@segment/ajv-human-errors": ["../ajv-human-errors/src"],
       "@segment/actions-core": ["../core/src"],
-      "@segment/destination-subscriptions": ["../destination-subscriptions/src"]
+      "@segment/destination-subscriptions": ["../destination-subscriptions/src"],
+      "@friendbuy/shared/*": ["../browser-destinations/src/destinations/friendbuy/shared/*"]
     }
   },
   "exclude": [],
```

dlasky:
@dhull conferred with team yesterday and got consensus that your latest approach is what we'd like (nice work!) Is this still draft for the rest of it or would you like a proper review?

dhull:
> @dhull conferred with team yesterday and got consensus that your latest approach is what we'd like (nice work!) Is this still draft for the rest of it or would you like a proper review?

Thanks! I have currently added Friendbuy-specific paths to the tsconfig.json files. Would you prefer for me see if I can use a generic path (`@segment/browser-destinations`) so it wouldn't be Friendbuy-specific?

Other than that, I just pushed one minor change and the code is ready for a full review.

packages/destination-actions/src/destinations/friendbuy/cloudUtil.ts
```diff
@@ -0,0 +1,54 @@
+import { JSONObject, RequestClient, RequestOptions } from '@segment/actions-core'
+
+import { Settings } from './generated-types'
+
+export const friendbuyBaseHost = 'fbot-sandbox.me'
+export const mapiUrl = `https://mapi.${friendbuyBaseHost}`
+
+export async function createRequestParams(
```
benhorowitz, line 8: Would this be more clearly named as `createPostRequest` or similar? It's doing more than creating parameters.

dhull: I named it that because it's returning the value for the second argument to the `request` call (which I guess is actually named "options" instead of "params") but I'm not stuck on the name.

packages/destination-actions/src/destinations/friendbuy/trackSignUp/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Track Sign Up',
+  description: 'Record when a customer signs up for a service.',
+  fields: Object.assign({}, trackSignUpFields, contextFields),
+
+  perform: async (request, { settings, payload }) => {
+    const friendbuyPayload = mapEvent(trackSignUpMapi, payload as unknown as AnalyticsPayload)
+
+    if (!friendbuyPayload) {
```
benhorowitz, line 47: Would it make sense to centralize this logic into `mapEvent` and throw an `IntegrationError` instead?

dhull: Done.

benhorowitz (approved):

PR #395, opened 2021-12-22
Cordial Destinations Actions

anatolii-cordial:
Cordial Destinations Actions implementation

## Testing


packages/destination-actions/src/destinations/cordial/createContactactivity/__tests__/index.test.ts
```diff
@@ -0,0 +1,31 @@
+import nock from 'nock'
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
+import Destination from '../../index'
+
+const testDestination = createTestIntegration(Destination)
+
+describe('Cordial.createContactactivity', () => {
+  it('should work with default mappings', async () => {
+    nock(/.*/).post('/v2/contactactivities').reply(200, {})
```
dlasky, line 9: see previous

packages/destination-actions/src/destinations/cordial/index.ts
```diff
@@ -0,0 +1,50 @@
...
+    fields: {
+      apiKey: {
+        label: 'API Key',
+        description: 'Your Cordial API Key',
+        type: 'string',
+        required: true
+      },
+      endpoint: {
+        label: 'Endpoint',
+        description: "Cordial API endpoint. Leave default, unless you've been provided with another one",
```
dlasky, line 26: is there a list of these? or are they individualized?

packages/destination-actions/src/destinations/cordial/user-identifier.ts
```diff
@@ -0,0 +1,6 @@
+export function getUserIdentifier(identifyByKey: string, identifyByValue: string): {} {
```
dlasky, line 1: should we type this?

dlasky (commented):
a couple of nitpicks--given that we only restore mocks `afterAll` it might make more sense to be more cautious with some of the very generous regexes being used in nock throughout. If possible I would consider specifying a few test urls to address your flexible endpoint scenarios and just include those in your default regex for testing.

packages/destination-actions/src/destinations/cordial/user-identifier.ts
espectrio: provided interface, thanks

packages/destination-actions/src/destinations/cordial/index.ts
espectrio: I've added choices, however, would it be possible to enter custom URI in such case? We might need to enter dynamic development endpoints there.

packages/destination-actions/src/destinations/cordial/addContactToList/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,106 @@
...
+const actionSlug = 'addContactToList'
+const destinationSlug = 'Cordial'
+const seedName = `${destinationSlug}#${actionSlug}`
+
+describe(`Testing snapshot for ${destinationSlug}'s ${actionSlug} destination action:`, () => {
+  it('default fields, group name', async () => {
+    const action = destination.actions[actionSlug]
+    const [eventData, settingsData] = generateTestData(seedName, destination, action, false)
+
+    nock(/.*/)
```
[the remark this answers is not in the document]
espectrio: generateTestData would return random URI for settings here, should we have our own generator for snapshot tests?

packages/destination-actions/src/destinations/cordial/addContactToList/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    } catch (err) {
+      expect(rawBody).toMatchSnapshot()
+    }
+  })
+
+  it('default fields, group id only', async () => {
+    const action = destination.actions[actionSlug]
+    const [eventData, settingsData] = generateTestData(seedName, destination, action, false)
+
+    nock(/.*/)
```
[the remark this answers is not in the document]
espectrio: generateTestData would return random URI for settings here, should we have our own generator for snapshot tests?

packages/destination-actions/src/destinations/cordial/addContactToList/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      .post(/\/.*\/contacts/)
+      .reply(200, {})
+    nock(/.*/)
```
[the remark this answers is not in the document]
espectrio: generateTestData would return random URI for settings here, should we have our own generator for snapshot tests?

espectrio:
@dlasky Thanks for review! I've pushed fixes, please take a look.

packages/destination-actions/src/destinations/cordial/index.ts
dlasky: no, so if that's the more important use case feel free to use the text input, wasn't a strong suggestion just wondering if there were some defined options that might make it easier for users configuring this in UI. (ie if the use case was more like EU/US endpoints which i've seen in a few destinations)

packages/destination-actions/src/destinations/cordial/addContactToList/__tests__/snapshot.test.ts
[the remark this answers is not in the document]
dlasky: good call out, I don't think we've run into this scenario yet since the previous destinations that used url as a configured param existed prior to the generator. Let me discuss with the team and get back to you on this one!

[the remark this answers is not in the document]
dlasky: if you want to get unblocked on this in the short term what we can do is restore the wildcard regex, and add some nock restores `afterEach` in the `describe` just in case

anatolii-cordial:
@dlasky nock.cleanAll() was added. Looks like all your suggestions are fixed now. Please take a look and give us know if something else needed to fix.

dlasky (approved):

PR #397, opened 2022-01-05
GA4 - Updated currency checking

nick-Ag:
Was seeing errors in production related to the `viewCart` action. The error:
```
TypeError: Cannot read property 'currency' of undefined
    at perform (/usr/src/integrations/node_modules/@segment/action-destinations/dist/destinations/google-analytics-4/viewCart/index.js:26:81)
```
[Sentry Link](https://sentry.io/organizations/segment/issues/2865802208/?project=1110287&query=is%3Aunresolved&statsPeriod=14d)
The error is triggered when a user passes in a payload such as:
```
	"payload": {
		"client_id": "123",
		"items": []
	}
```
Since `items[0]` is undefined we can't access `items[0].currency`

This PR updates the relevant if statement across all actions that use it.

## Testing

Deployed to staging and copied the relevant event from the customer to use as a test event. The event now correctly triggers a detailed error message rather than a vague runtime error.
Screen Shot 2022-01-05 at 12 28 05 PM

### DIFF
--- a/packages/destination-actions/src/destinations/google-analytics-4/addToWishlist/index.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/addToWishlist/index.ts
@@ -34,7 +34,7 @@ const action: ActionDefinition<Settings, Payload> = {
      * If set at the event level, item-level currency is ignored. If event-level currency is not set then
      * currency from the first item in items is used.
      */
-    if (payload.currency === undefined && payload.items && payload.items[0].currency === undefined) {
+    if (payload.currency === undefined && (!payload.items || !payload.items[0] || !payload.items[0].currency)) {
       throw new IntegrationError(
         'One of item-level currency or top-level currency is required.',
         'Misconfigured required field',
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/google-analytics-4/addToWishlist/index.ts
```diff
@@ -34,7 +34,7 @@ const action: ActionDefinition<Settings, Payload> = {
      * If set at the event level, item-level currency is ignored. If event-level currency is not set then
      * currency from the first item in items is used.
      */
-    if (payload.currency === undefined && payload.items && payload.items[0].currency === undefined) {
+    if (payload.currency === undefined && (!payload.items || !payload.items[0] || !payload.items[0].currency)) {
```
brennan, line 37: Nit - but since we reuse this logic four places, should we abstract into a single helper method? Otherwise, looks fine to me.

brennan (approved):

PR #410, opened 2022-01-10
Update Operation for Salesforce

nick-Ag:
Implementation of the update operation with three different lookup criteria: `external_id`, `trait`, and `record_id`. 

## Testing

Tested by running local server.
Update operation works in combination with `external_id`, `trait`, and `record_id`. Records are correctly being updated downstream in Salesforce.
Screen Shot 2022-01-11 at 2 30 37 PM
Screen Shot 2022-01-11 at 2 30 44 PM


packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -84,6 +122,62 @@ const action: ActionDefinition<Settings, Payload> = {
...
+        }
+
+        const SOQLQuery = buildQuery(payload.trait_field, payload.trait_value, 'Lead')
+
+        const res = await request<ExternalIDLookupResponse>(
+          `${settings.instanceUrl}/services/data/${API_VERSION}/query/${SOQLQuery}`,
+          { method: 'get' }
+        )
+
+        recordId = res.data.records[0].Id
```
brennan, line 157: Are SOQLQuery queries determinstic in Salesforce (e.g. is the same result returned at index 0 every time)? If not, this could lead us to update a different record each time we see the same external id, if it's tied to more than one record in Salesforce. I would probably recommend throwing an error if Salesforce returns more than one record.

Also, not sure if `Unique ID` is a field we can or should support? This could give customers more options for their lookup key.

packages/destination-actions/src/destinations/salesforce/index.ts
```diff
@@ -24,19 +20,16 @@ const destination: DestinationDefinition<Settings> = {
...
-          body: new URLSearchParams({
-            refresh_token: auth.refreshToken,
-            client_id: auth.clientId,
-            client_secret: auth.clientSecret,
-            grant_type: 'refresh_token'
-          })
-        }
-      )
-      setInstanceUrl(res.data.instance_url)
+      const res = await request<RefreshAccessTokenResult>('https://login.salesforce.com/services/oauth2/token', {
```
brennan, line 23: If possible, try to isolate changes into separate PRs. :) This looks as if it's related to oauth support rather than lookup criteria.

packages/destination-actions/src/destinations/salesforce/lead/index.ts
nick-Ag: Agree about the error, but I do wonder if there's a use case for updating all records returned by the lookup

nick-Ag: For `Unique ID` I think the user would be able to do `trait_field: Id` `trait_value: 123`. Is that what you meant?

brennan: I was thinking of Salesforce `Unique ID`, but my bad - this is a setting rather than its own field: https://help.salesforce.com/s/articleView?id=000325076&type=1

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  constructor(instanceUrl: string, request: RequestClient) {
+    this.instanceUrl = instanceUrl
+    this.request = request
+  }
+
+  createRecord = async (payload: any, sobject: string) => {
+    return this.request(`${this.instanceUrl}/services/data/${API_VERSION}/sobjects/${sobject}`, {
+      method: 'post',
+      json: {
```
nick-Ag, line 26: The shape of this json is specific to `Lead` right now but will change in the future

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,119 @@
...
+        Street: payload.street,
+        Country: payload.country,
+        PostalCode: payload.postal_code,
+        City: payload.city,
+        Email: payload.email
+      }
+    })
+  }
+
+  private buildQuery = (trait_field: string, trait_value: string, sobject: string) => {
```
nick-Ag, line 73: This is good for now but when I add multiple trait support copying the [classic integration code](https://github.com/segmentio/integrations/blob/master/integrations/salesforce/lib/v1/salesforce/sfutils.js#L16) would be a good idea

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    return res.data.records[0].Id
+  }
+
+  private lookupExternalId = async (field: string, value: string, sobject: string): Promise<string> => {
+    const res = await this.request<LookupResponseData>(
+      `${this.instanceUrl}/services/data/${API_VERSION}/sobjects/${sobject}/${field}/${value}`,
+      { method: 'get' }
+    )
+
```
nick-Ag, line 112: No need for checking the size of this response & throwing an error since Salesforce will return those statuses and the framework will automatically error out

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -1,35 +1,38 @@
-import type { ActionDefinition } from '@segment/actions-core'
+import { ActionDefinition, IntegrationError } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
+import {
```
marinhero, line 4: Would it make sense to turn this into a class since you are importing all the functions in the file?

packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts
```diff
@@ -0,0 +1,98 @@
...
+            email: 'sponge@seamail.com'
+          }
+        },
+        'Lead'
+      )
+    })
+
+    it('should lookup based on multiple traits', async () => {
+      nock(`${settings.instanceUrl}/services/data/${API_VERSION}/query`)
+        .get(`/?q=SELECT Id FROM Lead WHERE email = 'sponge@seamail.com' OR company = 'Krusty Krab'`)
```
nick-Ag, line 40: This is what we are expecting the outgoing request to look like given the traits input for `sf.updateRecord` below.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,118 @@
+import { IntegrationError, RequestClient } from '@segment/actions-core'
+
+const API_VERSION = 'v53.0'
```
brennan, line 3: I wonder if we should define this as a constant somewhere? We use `API_VERSION` in three different places.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,118 @@
...
+
+interface Records {
+  Id?: string
+}
```
brennan, line 7: Nit: Add space after line 7.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (Object.keys(payload.traits).includes('Id') && payload.traits['Id']) {
+      return await this.baseUpdate(payload.traits['Id'] as string, sobject, payload)
+    }
+
+    const recordId = await this.lookupTraits(payload.traits, sobject)
+    return await this.baseUpdate(recordId, sobject, payload)
+  }
+
+  private baseUpdate = async (recordId: string, sobject: string, payload: any) => {
+    return this.request(`${this.instanceUrl}/services/data/${API_VERSION}/sobjects/${sobject}/${recordId}`, {
```
marinhero, line 59: What would happen if on creating a lead I pass a trait, for example "State" and on update the payload doesn't have it but this wrapper is sending it as `undefined`. Would the trait be updated? What behavior are we expecting to see or does Salesforce provide?

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    if (!res || !res.data || res.data.totalSize === undefined) {
+      throw new IntegrationError('Response missing expected fields', 'Bad Response', 400)
+    }
+
+    if (res.data.totalSize === 0) {
+      throw new IntegrationError('No record found with given traits', 'Record Not Found', 404)
+    }
+
+    if (res.data.totalSize > 1) {
```
marinhero, line 109: I had to do some studying to understand the usage of this error code. Good choice Nick!

packages/destination-actions/src/destinations/salesforce/sf-properties.ts
```diff
@@ -0,0 +1,37 @@
+import { InputField } from '@segment/actions-core/src/destination-kit/types'
+import { IntegrationError } from '@segment/actions-core'
+
+export const operation: InputField = {
+  label: 'Operation',
+  description: 'Operation',
+  type: 'string',
+  required: true,
+  choices: [
+    { label: 'Create', value: 'create' },
```
marinhero, line 10: Are you thinking of adding unit tests for the success and error of each choice or would these ops being tested in the other files is enough for you?

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
nick-Ag: So if `payload.state` is undefined? In that case the request object would filter out undefined keys so that the json being sent wouldn't have a `State` key,value. So no `State` trait would be changed on Salesforce

packages/destination-actions/src/destinations/salesforce/sf-properties.ts
nick-Ag: I think testing them in the actions test files is good because it'll allow us to test the default mappings when those are added in the future (while also testing each choice)

PR #413, opened 2022-01-11
[Customer.io] Check if timestamps are valid before converting

mike-engel:
There's currently a bug where if an invalid date is passed in for `created_at` and other special dates, it will end up being converted to `NaN` by `dayjs`, and will be removed from the payload before being sent to customer.io. This adds a check to make sure the timestamp is valid—if not, it maintains the original value rather than converting to `NaN`.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/utils.ts
+++ b/packages/destination-actions/src/destinations/customerio/utils.ts
@@ -32,7 +32,11 @@ const isIsoDate = (value: string): boolean => {
 }
 
 export const convertValidTimestamp = <Value = unknown>(value: Value): Value | number => {
-  if (typeof value !== 'string') {
+  // Timestamps may be on a `string` field, so check if the string is only
+  // numbers. If it is, ignore it since it's probably already a unix timestamp.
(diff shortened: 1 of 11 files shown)


packages/destination-actions/src/destinations/customerio/utils.ts
```diff
@@ -31,6 +31,24 @@ const isIsoDate = (value: string): boolean => {
   return typeof value === 'string' && matcher.test(value) && !isNaN(Date.parse(value))
 }
 
+export const convertValidTimestamp = <Value = unknown>(value: Value): Value | number => {
+  // Timestamps may be on a `string` field, so check if the string is only
+  // numbers. If it is, ignore it since it's probably already a unix timestamp.
+  // DayJS doesn't parse unix timestamps correctly outside of the `.unix()`
+  // initializer.
+  if (typeof value !== 'string' || /^\d+$/.test(value)) {
```
dlasky, line 39: non blocking--would it make sense to use the regex trick here and then pass only passing values through to dayjs.unix(value) ?

dlasky (approved):

jinapark202:
This PR should be deployed now.

PR #424, opened 2022-01-21
TikTok Conversions clean up

kdaswani:
Draft PR for https://segment.atlassian.net/browse/STRATCONN-1118

Outstanding items:
- Need to add subfields to `contents` object
- Need to confirm we don't need extra randomization on messageId (event_id)
- Local and stage testing
- Engineer review

## Testing
- [X] Updated [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality

### DIFF
--- a/packages/destination-actions/src/destinations/tiktok-conversions/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/tiktok-conversions/__tests__/index.test.ts
@@ -8,7 +8,7 @@ const timestamp = '2021-09-2T15:21:15.449Z'
 
 describe('Tiktok Conversions', () => {
   describe('reportWebEvent', () => {
-    it('should send a successful event to reportWebEvent', async () => {
+    it('should send a successful InitiateCheckout event to reportWebEvent', async () => {
       const settings: Settings = {
         accessToken: 'test',
         pixel_code: 'test'
@@ -21,7 +21,20 @@ describe('Tiktok Conversions', () => {
         type: 'track',
         properties: {
           email: 'coreytest1231@gmail.com',
-          phone: '+1555-555-5555'
+          phone: '+1555-555-5555',
+          ttclid: '12345',
+          currency: 'USD',
+          value: 100,
+          query: 'shoes'
+        },
+        context: {
+          page: {
+            url: 'https://segment.com/',
+            referrer: 'https://google.com/'
+          },
+          userAgent:
+            'Mozilla/5.0 (iPhone; CPU iPhone OS 12_1_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/16D57',
+          ip: '0.0.0.0'
         },
         userId: 'testId123'
       })
@@ -30,11 +43,41 @@ describe('Tiktok Conversions', () => {
       const responses = await testDestination.testAction('reportWebEvent', {
         event,
         settings,
-        useDefaultMappings: true
+        useDefaultMappings: true,
+        mapping: {
+          event: 'InitiateCheckout'
+        }
       })
-
       expect(responses.length).toBe(1)
       expect(responses[0].status).toBe(200)
+      expect(responses[0].options.json).toMatchObject({
+        pixel_code: 'test',
+        event: 'InitiateCheckout',
+        event_id: 'corey123',
+        timestamp: '2021-09-2T15:21:15.449Z',
+        context: {
+          user: {
+            external_id: '481f202262e9c5ccc48d24e60798fadaa5f6ff1f8369f7ab927c04c3aa682a7f',
+            phone_number: '910a625c4ba147b544e6bd2f267e130ae14c591b6ba9c25cb8573322dedbebd0',
+            email: 'eb9869a32b532840dd6aa714f7a872d21d6f650fc5aa933d9feefc64708969c7'
+          },
+          ad: {
+            callback: '12345'
+          },
+          page: {
+            url: 'https://segment.com/',
+            referrer: 'https://google.com/'
+          },
+          ip: '0.0.0.0',
+          user_agent:
+            'Mozilla/5.0 (iPhone; CPU iPhone OS 12_1_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/16D57'
+        },
+        properties: {
+          currency: 'USD',
+          value: 100,
+          query: 'shoes'
+        }
+      })
     })
   })
 })
--- a/packages/destination-actions/src/destinations/tiktok-conversions/index.ts
+++ b/packages/destination-actions/src/destinations/tiktok-conversions/index.ts
@@ -7,15 +7,6 @@ import reportWebEvent from './reportWebEvent'
 /** used in the quick setup */
 const presets: DestinationDefinition['presets'] = [
   {
-    name: 'Initiate Checkout',
-    subscribe: 'event = "Checkout Started"',
-    partnerAction: 'reportWebEvent',
-    mapping: {
-      ...defaultValues(reportWebEvent.fields),
-      event: 'IntiateCheckout'
-    }
-  },
-  {
     name: 'View Content',
     subscribe: 'type="page"',
     partnerAction: 'reportWebEvent',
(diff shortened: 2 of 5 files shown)


packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -96,15 +84,89 @@ const action: ActionDefinition<Settings, Payload> = {
...
-    const eventId = payload.event_id
+    //our messageIds are always random per event so I do not think we need to add a random string to the end. TikTok only recommends this if you are using a sessionId, externalId, etc. Commenting out for now.
+    /*const eventId = payload.event_id
       ? payload.event_id.toString() + '_' + (Math.random() + 1).toString(36).substring(7)
       : ''
+    */
 
     const userData = {
-      hashedExternalId: formatUserId(payload.external_id),
+      hashedExternalId: formatUserId(payload.external_id || ''),
```
kdaswani, line 169: Not sure if this is the right thing to do. I removed `required: true` on external_id since it is not required (only recommended) by TikTok. But doing so was throwing this error:
Screen Shot 2022-01-21 at 1 16 09 PM

So I added a fallback to an empty string similar to email.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -96,15 +84,89 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      type: 'string',
+      description: 'The text string that was searched for.',
+      default: {
+        '@path': '$.properties.query'
+      }
     }
   },
   perform: (request, { payload, settings }) => {
-    const eventId = payload.event_id
+    //our messageIds are always random per event so I do not think we need to add a random string to the end. TikTok only recommends this if you are using a sessionId, externalId, etc. Commenting out for now.
```
kdaswani, line 162: I don't think we need this additional randomization on messageId since it is unique for each event/message.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
```diff
@@ -96,15 +84,89 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      description: 'User agent from the user’s device.',
+      default: {
+        '@path': '$.context.userAgent'
+      }
+    },
+    //need to add builder-defined subfields to contents object
+    contents: {
+      label: 'Contents',
+      type: 'object',
+      multiple: true,
```
kdaswani, line 124: Need to add builder-defined subfields. I am also curious if we can use coalesce in an object field type. Use case: depending on the event, these fields could either live in the `properties.products` array or just in the `properties` object.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
kdaswani: I've also asked the TikTok team to confirm we don't need the extra `_[random numbers]`.

kdaswani: I asked our TikTok contacts the following:
> For "event_id" is an underscore required? Our events already have a "messageId" which is unique per event so ideally we can map this directly to "event_id" without needing to add more random characters. Here's an example of what a messageId looks like: ajs-cb039a2598a1548ac001ade35b3a2c12. Would this work?

They confirmed:
> Underscore is not required. The solution you proposed would work!

The extra randomization on messageId is unnecessary so we should proceed with removing it. I have it commented out for now, but we can remove entirely too.

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
nick-Ag: Resolved by returning `undefined` from the hashing methods if those keys don't exist on the payload. (Rather than the hashed empty string)

nick-Ag:
@kdaswani Stage testing this was throwing some errors today, will try again tomorrow!

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/index.ts
nick-Ag: Screen Shot 2022-02-08 at 12 09 27 PM
Hey @kdaswani I get this response after testing an event with the randomization removed

nick-Ag: Is this expected?

kdaswani: Hmm weird! I asked TikTok this:

> For "event_id" is an underscore required? Our events already have a "messageId" which is unique per event so ideally we can map this directly to "event_id" without needing to add more random characters. Here's an example of what a messageId looks like: ajs-cb039a2598a1548ac001ade35b3a2c12. Would this work?

And they said:

> underscore is not required. The solution you proposed would work!

Seems like that's not correct though. What happens when you add an `_123` to the end of the messageId being sent?

kdaswani: I wonder if we need randomization or if we can just add `_seg` or something. I was running into issues with the unit test passing because I couldn't predict the randomization in the expected output. But maybe there's another way to do that and keep randomization 😄  @nick-Ag

nick-Ag: It looks like TikTok accepts with 200 the event_id without `_<random>` _but_ they send back that message in the screenshot which could be confusing for customers. So I added the `_seg` after every event_id

kdaswani: Sounds good, thank you Nick!

packages/destination-actions/src/destinations/tiktok-conversions/__tests__/index.test.ts
```diff
@@ -5,23 +5,169 @@ import { Settings } from '../generated-types'
 
 const testDestination = createTestIntegration(Definition)
 const timestamp = '2021-09-2T15:21:15.449Z'
+const settings: Settings = {
+  accessToken: 'test',
```
brennan, line 9: Curious why one is snake_case and one is camelCase?

nick-Ag: Good catch, changes to snake case

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -4,51 +4,108 @@ import type { Settings } from './generated-types'
...
+    '@arrayPath': [
+      '$.properties',
+      {
+        ...inner_contents
+      }
+    ]
+  }
+}
+
+const multi_product_contents = {
```
brennan, line 34: Variable name `multiProductArrayProperties`?

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -4,51 +4,108 @@ import type { Settings } from './generated-types'
...
-      event: 'IntiateCheckout'
-    }
-  },
   {
     name: 'View Content',
     subscribe: 'type="page"',
     partnerAction: 'reportWebEvent',
     mapping: {
-      ...defaultValues(reportWebEvent.fields),
+      ...single_product_contents,
```
brennan, line 53: Variable name `singleProductContents`.

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -4,51 +4,108 @@ import type { Settings } from './generated-types'
...
     }
   },
   {
-    name: 'Products Searched',
+    name: 'Search',
     subscribe: 'event = "Products Searched"',
     partnerAction: 'reportWebEvent',
     mapping: {
-      ...defaultValues(reportWebEvent.fields),
+      ...single_product_contents,
```
brennan, line 62: `singleProductContents`

packages/destination-actions/src/destinations/tiktok-conversions/index.ts
```diff
@@ -4,51 +4,108 @@ import type { Settings } from './generated-types'
...
       event: 'Search'
     }
   },
+  {
+    name: 'Add to Wishlist',
+    subscribe: 'event = "Product Added to Wishlist"',
+    partnerAction: 'reportWebEvent',
+    mapping: {
+      ...single_product_contents,
```
brennan, line 71: `singleProductContents`

packages/destination-actions/src/destinations/tiktok-conversions/reportWebEvent/formatter.ts
```diff
@@ -3,26 +3,32 @@ import { createHash } from 'crypto'
...
 }
 
 /**
  *
  * @param userId
  * @returns Leading/Trailing spaces are trimmed and then userId is hashed.
  */
-export function formatUserId(userId: string): string {
-  return hashAndEncode(userId.toLowerCase().trim())
+export function formatUserId(userId: string | undefined): string | undefined {
```
brennan, line 18: For this method and `formatPhone`, it looks as if the existing behavior is to send `undefined` values downstream to TikTok. Should we scrub the keys from the JSON payload as well if the value is falsy?

e.g. rather than send `{ hashedPhone: undefined }`, we would send `{}`.

nick-Ag: Good call, I tested it out locally and it looks like actions filters out undefined keys so we're good 👍

PR #427, opened 2022-01-25
[Customer.io] Parse ISO 8601 time offsets without a `:`

mike-engel:
We got another bug report/feature request from one of our customers who was using ISO8601 date strings in a way that we initially hadn't supported. They were using time zone offsets without a `:`, which is supported per [wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC):

> The UTC offset is appended to the time in the same way that 'Z' was above, in the form ±[hh]:[mm], ±[hh][mm], or ±[hh].

I also tried to add support for just `±[hh]`, but `dayjs`/`Date` doesn't support that format, so I've left it out for now.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/__tests__/createUpdatePerson.test.ts
+++ b/packages/destination-actions/src/destinations/customerio/__tests__/createUpdatePerson.test.ts
@@ -85,7 +85,10 @@ describe('CustomerIO', () => {
         date11: '2018-03-04T12:08:56 PDT',
         date12: '2018-03-04T12:08:56.235 PDT',
         date13: '15/MAR/18',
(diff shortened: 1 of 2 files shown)


dlasky (approved):

billmix:
@benhorowitz -- With this new change, we're seeing inconsistent timestamp conversions coming in to our customer's account, but are matched to the regex in the PR. Is there a reason tied to the deployment that only some timestamps meeting the conditions would be converted and not others? CC: @mike-engel

mike-engel:
@dlasky @benhorowitz can y’all let me know when this is deployed to production? I think the issue from @billmix is due to it not being out quite yet

PR #435, opened 2022-01-31
enable subscription tracking for mail sends with personas

pmcanseco:
The Journeys Email Builder by default populates some unsubscribe fields. The unsubscribe link doesn't work at the moment. It's configured to `[unsubscribe]` but the emails being sent don't enable subscription tracking. This PR aims to do that, making it so the unsubscribe URL will actually work.

image

## Testing

- [x] ~Added~ Updated [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality

### DIFF
--- a/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
@@ -318,7 +318,13 @@ const action: ActionDefinition<Settings, Payload> = {
             type: 'text/html',
             value: Mustache.render(bodyHtml, { profile })
           }
-        ]
+        ],
+        tracking_settings: {
+          subscription_tracking: {
+            enable: true,
+            substitution_tag: "[unsubscribe]"
+          }
+        }
       }
     })
   }
--- a/packages/destination-actions/src/destinations/personas-messaging-sendgrid/__tests__/send-email.test.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-sendgrid/__tests__/send-email.test.ts
@@ -120,7 +120,13 @@ for (const environment of ['stage', 'production']) {
             type: 'text/html',
             value: `Hi ${userData.firstName}, Welcome to segment`
           }
-        ]
+        ],
+        tracking_settings: {
+          subscription_tracking: {
+            enable: true,
+            substitution_tag: '[unsubscribe]'
+          }
+        }
       }
 
       const sendGridRequest = nock('https://api.sendgrid.com')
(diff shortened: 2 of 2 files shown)


laurenfb (approved):
🚢  it

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -318,7 +318,13 @@ const action: ActionDefinition<Settings, Payload> = {
             type: 'text/html',
             value: Mustache.render(bodyHtml, { profile })
           }
-        ]
+        ],
+        tracking_settings: {
+          subscription_tracking: {
```
lew-gordon, line 323: Should this be an action-destinations setting instead?

lew-gordon (changes requested):

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
pmcanseco: 🤷 @laurenfb @vikramkumar19 thoughts?

pmcanseco: My gut feel is that this should be hard-coded to true, since our product will only be sending marketing mail and this should only ever be disabled for transactional emails (shipping notifications/password reset emails)

laurenfb: Pablo's argument makes sense to me. I'm fine this way unless @vikramkumar19 disagrees

lew-gordon: That's true, but shouldn't we allow users of this destination-action to customize the tag at the very least? I think in the ideal of others outside of use for Twilio Engage I feel having this as a setting seems reasonable. Though maybe if the destination-actions team is OK with not having this as a setting it's something we could come back to later.

pmcanseco: This is an excellent point. The idea of this PR was just to fix the unsubscribe link in the default template. I would defer to the destinations or messaging teams if they need this to be customizable in the future. There's currently no `unsub footer` block in the editor experience so it may be a bit premature to know if custom unsub tags would be needed.

lew-gordon: Agreed! I won't be a blocker if you want to merge this as is.

lew-gordon (approved):

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
vikramkumar19: I have the same concern as @lew-gordon . Not a blocker for now.
If we plan this to be part of settings that could be done also at later stage. However, I thing to note is we would recommend the actions destination is abstracted to the customers. It relies on us how we would update the field or the settings.

vikramkumar19 (approved):
🚢 it

PR #443, opened 2022-02-02
Salesforce: Custom Object action

nick-Ag:
Adds the Custom Object action to Salesforce. This action allows users to create, update, and upsert on custom defined objects in Salesforce.

Also adds the `custom_fields` property to both the Custom Object action (where it is required) and the Lead action. This property allows users to send custom and standard attributes to Salesforce.

## Testing

Deployed successfully to stage environment.

Correctly working:
* Create Custom Object
* Update Custom Object
* Upsert Custom Object
* Create Lead record with custom fields

Screen Shot 2022-02-02 at 3 21 26 PM


packages/destination-actions/src/destinations/salesforce/__tests__/lead.test.ts
```diff
@@ -119,6 +119,64 @@ describe('Salesforce', () => {
...
+            '@path': '$.traits.email'
+          },
+          company: {
+            '@path': '$.traits.company'
+          },
+          last_name: {
+            '@path': '$.traits.last_name'
+          },
+          custom_fields: {
+            A: '1',
```
brennan, line 149: To verify, can customers set up their mapping such that a key is mapped to a directive, like `A: '$.traits.A'`. Should we add a test like that as well?

brennan: Nvmd, I see a test validating this above.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -74,22 +67,30 @@ export default class Salesforce {
...
-        Email: payload.email
-      }
+      json: JSON_data
     })
   }
 
+  private buildJSONData = (payload: GenericPayload, sobject: string) => {
+    let baseShape
+    if (payload.sobject) {
+      //sobject is a unique key for the custom object action
```
brennan, line 81: This comment and the variable names make the code a bit confusing to read. In this case, `sobject` may be a unique key for a custom object name, but it's also a variable, the value of which could be a standard object, like 'lead' (and other standard objects in the future).

I would recommend that the custom object key be something more specific, like `customObject`.

packages/destination-actions/src/destinations/salesforce/sf-object-to-shape.ts
```diff
@@ -0,0 +1,27 @@
...
+  const shapeFunction = object_to_shape.get(sobject.toLowerCase())
+
+  if (!shapeFunction) {
+    throw new Error(`Undefined Object Shape: ${sobject}`)
+  }
+
+  return shapeFunction(payload)
+}
+
+const LeadShape = (payload: GenericPayload) => {
```
brennan, line 13: Nit, I think these two const declarations should come above the export for readability.

packages/destination-actions/src/destinations/salesforce/__tests__/custom.test.ts
```diff
@@ -0,0 +1,62 @@
...
+import { API_VERSION } from '../sf-operations'
+
+const testDestination = createTestIntegration(Destination)
+
+const settings = {
+  instanceUrl: 'https://test.com'
+}
+
+describe('Salesforce', () => {
+  describe('Custom', () => {
```
marinhero, line 13: What do you think about being more specific in this description?
```suggestion
  describe('`Custom Object Tests`', () => {
```

packages/destination-actions/src/destinations/salesforce/__tests__/custom.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      })
+
+      expect(responses.length).toBe(1)
+      expect(responses[0].status).toBe(201)
+
+      expect(responses[0].request.headers).toMatchInlineSnapshot(`
+        Headers {
+          Symbol(map): Object {
+            "authorization": Array [
+              "Bearer undefined",
```
marinhero, line 45: How can this be a successful test if the authentication token is `undefined`?

packages/destination-actions/src/destinations/salesforce/custom/index.ts
```diff
@@ -0,0 +1,45 @@
...
+        'The name of the Salesforce object that records will be added or updated within. The object must be predefined in your Salesforce account. Values should end with "__c".',
+      type: 'string',
+      required: true
+    },
+    custom_fields: { ...custom_fields, required: true }
+  },
+  perform: async (request, { settings, payload }) => {
+    const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
+
+    if (!payload.sobject.endsWith('__c')) {
```
marinhero, line 25: The description for this object says `Values should end with "__c` but here we are handling the case where they don't. Shouldn't this be a UI validation instead?

packages/destination-actions/src/destinations/salesforce/index.ts
```diff
@@ -2,6 +2,8 @@ import type { DestinationDefinition } from '@segment/actions-core'
 import type { Settings } from './generated-types'
 import lead from './lead'
 
+import custom from './custom'
```
marinhero, line 5: I think we should rename this file to be more specific as well as this variable. What do you think?

packages/destination-actions/src/destinations/salesforce/sf-object-to-shape.ts
```diff
@@ -0,0 +1,27 @@
...
+  return {
```
marinhero, line 14: You might want to create a new type for `LeadShape` given that it has a defined and expected structure.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -74,22 +67,30 @@ export default class Salesforce {
...
-        PostalCode: payload.postal_code,
-        City: payload.city,
-        Email: payload.email
-      }
+      json: JSON_data
     })
   }
 
+  private buildJSONData = (payload: GenericPayload, sobject: string) => {
+    let baseShape
```
marinhero, line 79: Perhaps you can trim this decision tree:

```
let baseShape = {}

if (!payload.sobject) {
  baseShape = mapObjectToShape(payload, sobject)
}

if (payload.custom_fields) {
     baseShape = { ...baseShape, ...payload.custom_fields }
}

return baseShape
```

Would this make sense?

marinhero (commented):
I left a few comments and questions. Let me know if you have any questions. Good work!

packages/destination-actions/src/destinations/salesforce/sf-object-to-shape.ts
```diff
@@ -0,0 +1,27 @@
+import { GenericPayload } from './sf-types'
+
+export const mapObjectToShape = (payload: GenericPayload, sobject: string) => {
+  const shapeFunction = object_to_shape.get(sobject.toLowerCase())
```
mericsson, line 4:
```suggestion
  const shapeFunction = objectToShape.get(sobject.toLowerCase())
```

Segment style is typically camelCase

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -23,19 +24,11 @@ export default class Salesforce {
   }
 
   createRecord = async (payload: GenericPayload, sobject: string) => {
+    const JSON_data = this.buildJSONData(payload, sobject)
```
mericsson, line 27:
```suggestion
    const json = this.buildJSONData(payload, sobject)
```

At Segment, unusual to have snake case in variable names

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -74,22 +67,30 @@ export default class Salesforce {
...
+    let baseShape
+    if (payload.sobject) {
+      //sobject is a unique key for the custom object action
+      baseShape = {}
+    } else {
+      baseShape = mapObjectToShape(payload, sobject)
+    }
+
+    if (payload.custom_fields) {
+      baseShape = { ...baseShape, ...payload.custom_fields }
```
mericsson, line 88:
```suggestion
      // custom field mappings take priority over base shape mappings.
      baseShape = { ...baseShape, ...payload.custom_fields }
```

Just a  note, this ordering implies that any `custom_fields` mappings will take priority over any `baseShape` mappings. That may  be  expected but wanted to  call that out. Might be  worth a comment.

packages/destination-actions/src/destinations/salesforce/sf-properties.ts
```diff
@@ -24,6 +24,16 @@ export const traits: InputField = {
   type: 'object'
 }
 
+export const custom_fields: InputField = {
```
mericsson, line 27:
```suggestion
export const customFields: InputField = {
```

prefer camel case for consistency

packages/destination-actions/src/destinations/salesforce/custom/index.ts
```diff
@@ -0,0 +1,45 @@
...
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Custom Object',
+  description: 'Custom Object',
+  fields: {
+    operation: operation,
+    traits: traits,
+    sobject: {
+      label: 'Salesforce Object',
+      description:
+        'The name of the Salesforce object that records will be added or updated within. The object must be predefined in your Salesforce account. Values should end with "__c".',
```
mericsson, line 16:
```suggestion
        'The API name of the Salesforce object that records will be added or updated within. The object must be predefined in your Salesforce account. Values should end with "__c".',
```

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
nick-Ag: Great improvement, thanks!

packages/destination-actions/src/destinations/salesforce/index.ts
nick-Ag: Agreed

packages/destination-actions/src/destinations/salesforce/custom/index.ts
nick-Ag: We don't have a way to do action level validations in the UI right now. I'm going to remove this block tho since we should let users interact with Standard objects which we don't create an action for via this action

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
nick-Ag: Agreed

packages/destination-actions/src/destinations/salesforce/sf-object-to-shape.ts
nick-Ag: Can you expand more on this? Like having a `const LeadShape = (payload: GenericPayload): LeadShapeType =>` and then having a `GenericShape` for `mapObjectToShape` once I add more shapes?

packages/destination-actions/src/destinations/salesforce/__tests__/customObject.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      })
+
+      expect(responses.length).toBe(1)
+      expect(responses[0].status).toBe(201)
+
+      expect(responses[0].request.headers).toMatchInlineSnapshot(`
+        Headers {
+          Symbol(map): Object {
+            "authorization": Array [
+              "Bearer abc123",
```
marinhero, line 50: 👏  nice!

packages/destination-actions/src/destinations/salesforce/custom/index.ts
marinhero: Got it. Good call.

packages/destination-actions/src/destinations/salesforce/sf-object-to-shape.ts
marinhero: I'll DM you

marinhero (approved):
:shipit:

brennan (approved):

PR #453, opened 2022-02-08
GA4 - Add userID and params parameters to every action

nick-Ag:
This PR adds `userID` and `params` parameters to every GA4 action and sends those parameters to Google.

## Testing
Added unit tests for `userID` and `params`.
Testing was successful in stage, events were sent including the custom params.
Screen Shot 2022-02-07 at 4 43 17 PM
Screen Shot 2022-02-07 at 4 44 05 PM
Screen Shot 2022-02-07 at 4 48 41 PM

### DIFF
--- a/packages/destination-actions/src/destinations/google-analytics-4/customEvent/index.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/customEvent/index.ts
@@ -1,7 +1,7 @@
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import { client_id } from '../ga4-properties'
+import { client_id, user_id } from '../ga4-properties'
 
 const normalizeEventName = (name: string, lowercase: boolean | undefined): string => {
   name = name.trim()
@@ -19,6 +19,7 @@ const action: ActionDefinition<Settings, Payload> = {
   defaultSubscription: 'type = "track"',
   fields: {
     clientId: { ...client_id },
+    user_id: { ...user_id },
     name: {
       label: 'Event Name',
       description:
@@ -50,6 +51,7 @@ const action: ActionDefinition<Settings, Payload> = {
       method: 'POST',
       json: {
         client_id: payload.clientId,
+        user_id: payload.user_id,
         events: [
           {
             name: event_name,
--- a/packages/destination-actions/src/destinations/google-analytics-4/customEvent/index.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/customEvent/index.ts
@@ -1,7 +1,7 @@
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import { client_id, user_id } from '../ga4-properties'
+import { params, client_id, user_id } from '../ga4-properties'
 
 const normalizeEventName = (name: string, lowercase: boolean | undefined): string => {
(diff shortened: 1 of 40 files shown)


kdaswani:
This looks awesome! What a fast turnaround @nick-Ag - thank you!

marinhero (approved):
The changes look good to me, it's hard to manually test them all but I'll lean on the fact that the test suite is passing. :shipit:

packages/destination-actions/src/destinations/google-analytics-4/customEvent/index.ts
```diff
@@ -36,20 +37,15 @@ const action: ActionDefinition<Settings, Payload> = {
...
       default: false
     },
-    params: {
-      label: 'Event Parameters',
-      description: 'The event parameters to send to Google',
-      type: 'object',
-      additionalProperties: true,
-      default: { '@path': '$.properties' }
-    }
+    params: { ...params, default: { '@path': '$.properties' } }
```
brennan, line 40: I think we should remove defaulting params to all properties in an event. Customers should not be passing any PII to Google. Defaulting to automatically passing all properties puts Segment at risk of sending PII without our customers realizing it.

Curious your thoughts, @mericsson @kdaswani .

dlasky: If you grew the repo for $.properties there is quite a bit of prior art here.

kdaswani: @brennan That's a good point! Generally, we don't expect PII in the `properties` (typically it would be in `traits`) but I agree it's still a concern. Looks like we only have this default for the Custom Event action and it's always been this way. If we remove the default, what happens to customers who have already set up a Custom Event action? It stays as-is but future ones set up have no default?

dlasky: there are a ton of existing destinations doing this exact thing: ga4 (in other places), amplitude, facebook, metronome, clevertap, etc...

kdaswani: Thank you for looking into that, @dlasky! Given Google is more cautious about PII: https://support.google.com/analytics/answer/6366371?hl=en#zippy=%2Cin-this-article let's remove the `properties` default here. Not urgent but probably better to be safe. I would just like to confirm backends behavior to customers who have already configured the Custom Event action - no impact to them, right? @nick-Ag

dlasky: yeah no impact to existing unless you did a migration

kdaswani: cool yeah we don't need to change existing so no migration necessary! this could just apply moving forward. thank you all!

nick-Ag: Yea that's correct @kdaswani

nick-Ag: Will make a PR to remove that 👍

kdaswani: thank you 😅

nick-Ag: https://github.com/segmentio/action-destinations/pull/463

PR #455, opened 2022-02-08
[browser-destinations] add support for IE 11

chrisradek:
This PR adds IE 11 support to the browser destinations.

## Background

The browser destinations don't support IE 11 prior to this change. This is due to the browser destinations and some dependencies targeting ES2015+ whereas IE 11 only supports ES5 syntax. `analytics-next` will add polyfills to the browser if needed, but can't polyfill language syntax like `class` and `=>` arrow functions.

## How does this change achieve IE 11 support?

This PR updates the build steps for the browser destinations. There are now 2 webpack build steps: 
- The 1st webpack pass creates a bundle per destination similar to what's currently done.
- The 2nd webpack pass polyfills Regex unicode support and transpiles the bundle from the 1st step to ES5 using the TypeScript compiler.

### Why are 2 webpack passes needed?

2 webpack builds greatly reduces the size of the final bundle compared to 1 build. With a single build using TypeScript to transpile every file in a bundle to ES5 tree-shaking was much less efficient. I could create a separate bundle of shared logic between destinations, but there was a significant increase in total bundle size compared to the current state.

With 2 passes, webpack is able to perform tree-shaking optimally just like it does today in the 1st pass, then in the 2nd pass it takes each destination's bundle and transpiles it. The output bundles are still larger than current state but when `gzipped` the difference is minor.

### Why is TypeScript used to down-level to ES5 instead of babel?

`regeneratorRuntime`. Babel requires `regeneratorRuntime` to be available for any generators/async-await functions that are transpiled. As a library, we don't want to add this to the global scope. There is a ponyfill that is meant for library use however it is incompatible with webpack 5.

### Any downsides?

The main downside is that sourcemaps no longer point to the original typescript files - instead they point to the webpack bundle from the 1st pass. This is due to a limitation in our tooling to merge source maps. I came across some tools that attempt to merge source maps across build steps but haven't tried some of the options yet.

### How big are the bundles?
All sizes are in bytes and are an approximation from when this work began.
| Plugin                 |   Current |       IE 11 | two-pass |
| ---------------------- | --------: | ----------: | -------: |
| amplitude              |    41,559 |      13,374 |   51,414 |
| amplitude (gz)         |    12,269 |       4,443 |   13,182 |
| braze                  |   239,580 |      61,620 |  283,812 |
| braze (gz)             |    69,909 |      17,022 |   76,477 |
| braze-cloud            |    41,419 |      13,234 |   51,274 |
| braze-cloud (gz)       |    12,220 |       4,386 |   13,134 |
| friendbuy              |   239,833 |      61,993 |  283,495 |
| friendbuy (gz)         |    68,240 |      15,307 |   74,799 |
| fullstory              |   233,110 |      55,019 |  276,571 |
| fullstory (gz)         |    68,711 |      15,802 |   75,195 |
| sprig-web              |   223,169 |      45,246 |  266,659 |
| sprig-web (gz)         |    64,763 |      11,928 |   71,275 |
| stackadapt             |   221,277 |      43,367 |  264,767 |
| stackadapt (gz)        |    64,338 |      11,546 |   70,856 |
| vendor                 |         0 |     257,029 |        0 |
| vendor (gz)            |         0 |      67,998 |        0 |
| ----------------       | --------- | ----------- | -------- |
| Min (braze-cloud (gz)) |    12,220 |      72,384 |   13,134 |
| Max (braze (gz))       |    69,909 |      85,020 |   76,477 |

## Testing

I have tested loading the plugins in Firefox/Chrome/Edge/Internet Explorer 11. I've also ran Braze through the dev site (`yarn dev`) in each browser.

### DIFF
--- a/packages/browser-destinations/package.json
+++ b/packages/browser-destinations/package.json
@@ -16,8 +16,12 @@
     "analyze": "NODE_ENV=production webpack --profile --json > stats.json && webpack-bundle-analyzer --port 4200 stats.json",
     "build": "yarn clean && yarn build-ts && yarn build-web",
     "build-ts": "yarn tsc -b tsconfig.build.json",
-    "build-web": "NODE_ENV=production ASSET_ENV=production yarn webpack -c webpack.config.js",
-    "build-web-stage": "NODE_ENV=production ASSET_ENV=stage yarn webpack -c webpack.config.js",
+    "build-web-es2015": "NODE_ENV=production ASSET_ENV=production yarn webpack -c webpack.config.js",
+    "build-web-es5": "NODE_ENV=production ASSET_ENV=production yarn webpack -c webpack.config.es5.js",
+    "build-web": "yarn build-web-es2015 && yarn build-web-es5",
+    "build-web-es2015-stage": "NODE_ENV=production ASSET_ENV=stage yarn webpack -c webpack.config.js",
+    "build-web-es5-stage": "NODE_ENV=production ASSET_ENV=stage yarn webpack -c webpack.config.es5.js",
+    "build-web-stage": "yarn build-web-es2015-stage && yarn build-web-es5-stage",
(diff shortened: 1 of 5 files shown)


danieljackins (approved):
LGTM if all the testing went well! Bundle size increases aren't too much 🎉

PR #458, opened 2022-02-08
Removed tiktok token rule, it was false-positiving too much

nick-Ag:
Removes TikTok access token rule from our Gitleaks rules. TikTok tokens are random 40 char long strings and so the rule was throwing false positives on various hashed values we have in unit tests. 

## Testing


marinhero (approved):
Thanks for the nice description! :shipit:

PR #466, opened 2022-02-09
Change test authKey value to make it obvious it's a dummy value.

dhull:
The Friendbuy tests define a dummy `authKey` value to use in the unit tests that was flagged as a secret. This PR changes its value to make it more obvious it's not a real secret. It still has the format of a UUID so that it will pass validation.

## Testing

This is just a value that is used in the unit tests. I verified that the unit tests still pass but no other testing is necessary.


gurdasnijor (approved):
LGTM!

PR #472, opened 2022-02-14
Close: Add Lead External ID and Lead Custom Fields

froxCZ:
Add Lead External ID and Lead Custom Fields to Close's CreateUpdateContactAndLead action.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/close/index.ts
+++ b/packages/destination-actions/src/destinations/close/index.ts
@@ -17,6 +17,20 @@ const destination: DestinationDefinition<Settings> = {
         type: 'string',
         required: true
       },
+      lead_custom_field_id_for_company_id: {
+        label: 'Lead Custom Field ID for Company ID',
+        description:
+          "Enter the ID of a Lead Custom Field that'll be " +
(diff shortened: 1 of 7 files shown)


dlasky:
hey @froxCZ i'm taking a look!

packages/destination-actions/src/destinations/close/index.ts
```diff
@@ -17,6 +17,20 @@ const destination: DestinationDefinition<Settings> = {
         type: 'string',
         required: true
       },
+      lead_custom_field_id_for_company_id: {
```
dlasky, line 20: is there ever a case where a user might want this to be a different value depending on the segment event type or other filters? (just wondering if this should be in the action fields)

froxCZ: Very unlikely, but I will double check with the team. Nevertheless you can go ahead and merge as it is now. If we decide to move it to the Action level we will do so before public release if that's ok.

dlasky: no worries, like i said not a blocker. thank you!

PR #482, opened 2022-02-17
Builders can specify the default UI view for object fields 

nick-Ag:
This PR adds `defaultObjectUI` as an `InputField` key, allowing builders to specify the default UI view of object types.
Builders can specify four values: `keyvalue`, `object`, `keyvalue:only`, and `object:only`

`keyvalue` will default to this UI representation of object fields, allowing users to specify individual keys and their mappings. `keyvalue:only` will disable the `Edit Object` button, locking users to the key value editor.
image

`object` is the default value and will show users the object editor in the UI. `object:only` will disable the `Edit Key/Value` button, locking users to the object editor.
This is the current default, which would be confusing to customers. They would always have to click ‘Edit Key/Value’  to get to where they want to.

Mini-SDD: https://paper.dropbox.com/doc/Mini-SDD-Key-Value-Defaults-for-object-fields-in-actions--BdFvJo1UWGm5tFZCXf3AXqwIAg-Qstuf6EEm8AI5QuzExtIi

## Testing

Tested successfully in staging by running `./bin/run push` and having `defaultObjectUI` values being written to stage DB.


packages/cli-internal/src/commands/push.ts
```diff
@@ -163,6 +163,7 @@ export default class Push extends Command {
             dynamic: field.dynamic ?? false,
             placeholder: field.placeholder ?? '',
             allowNull: field.allowNull ?? false,
+            defaultObjectUI: field.defaultObjectUI,
```
dlasky, line 166: should we enforce a default here?

packages/core/src/destination-kit/types.ts
```diff
@@ -114,20 +114,31 @@ export interface InputField {
...
+    | 'ipv6' // IP address v6.
+    | 'regex' // tests whether a string is a valid regular expression by passing it to RegExp constructor.
+    | 'uuid' // Universally Unique IDentifier according to RFC4122.
+    | 'password' // hint to the UI to hide/obfuscate input strings
+    | 'text' // longer strings
+
+  /**
+   * Determines the UI representation of the object field. Only applies to object types.
+   * Key Value Editor: Users can specify individual object keys and their mappings, ideal for custom objects.
+   * Object Editor: Users can map an entire object from their payload to the object field in a single mapping.
```
dlasky, line 135:
```suggestion
   * Object Reference: Users can specify only another object in the segment event to use as the value for this key in the payload
```

dlasky (approved):
1 nit, lgtm! sorry for the delay!

packages/cli-internal/src/commands/push.ts
nick-Ag: The DB defaults to NULL when no value is provided, would that be sufficient?

nick-Ag: Tested w/ `defaultObjectUI: field.defaultObjectUI ?? null`, no difference between existing so I'll merge in

PR #485, opened 2022-02-18
process external ID subscription payload for "personas-messaging-sendgrid"

larrycustodio:
Updates the response body interface received from `fetchProfileExternalIds` on "personas-messaging-sendgrid" action, and processes the `isSubscribed` property from the email external ID.

Unit Tests:
image

Local testing:
with email external ID's  `isSubscribed` set to = `subscribed`
image

with email external ID's  `isSubscribed` set to = `true`
image

with email external ID's `isSubscribed` set to `unsubscribed`:
image

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
@@ -10,11 +10,11 @@ const insertEmailPreviewText = (html: string, previewText: string): string => {
 
   // See https://www.litmus.com/blog/the-little-known-preview-text-hack-you-may-want-to-use-in-every-email/
   $('body').prepend(`
-    <div style="display: none; max-height: 0px; overflow: hidden;">
+    <div style='display: none; max-height: 0px; overflow: hidden;'>
       ${htmlEscape(previewText)}
     </div>
 
-    <div style="display: none; max-height: 0px; overflow: hidden;">
+    <div style='display: none; max-height: 0px; overflow: hidden;'>
       ${'&nbsp;&zwnj;'.repeat(13)}&nbsp;
     </div>
   `)
@@ -47,32 +47,6 @@ const fetchProfileTraits = async (
 
   const body = await response.json()
   return body.traits
-}
-
-const fetchProfileExternalIds = async (
-  request: RequestFn,
-  settings: Settings,
-  profileId: string
-): Promise<Record<string, string>> => {
-  const endpoint = getProfileApiEndpoint(settings.profileApiEnvironment)
-  const response = await request(
-    `${endpoint}/v1/spaces/${settings.spaceId}/collections/users/profiles/user_id:${profileId}/external_ids?limit=25`,
-    {
-      headers: {
-        authorization: `Basic ${Buffer.from(settings.profileApiAccessToken + ':').toString('base64')}`,
-        'content-type': 'application/json'
-      }
-    }
-  )
-
-  const body = await response.json()
-  const externalIds: Record<string, string> = {}
-
-  for (const externalId of body.data) {
-    externalIds[externalId.type] = externalId.id
-  }
-
-  return externalIds
 }
 
 const isRestrictedDomain = (email: string): boolean => {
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -238,100 +241,117 @@ const action: ActionDefinition<Settings, Payload> = {
...
       throw new IntegrationError(
-        'Emails with gmailx.com, yahoox.com, aolx.com, and hotmailx.com domains are blocked.',
-        'Invalid input',
+        `The subscribed state found in the response is ${String(emailProfile?.isSubscribed)}`,
+        'Invalid isSubscribed value',
         400
       )
     }
+    if (['unsubscribed', 'did not subscribe'].includes(emailProfile?.isSubscribed)) {
+      return
```
lew-gordon, line 253: I wonder if we should log something here. I'm honestly just brainstorming ways on how we'd debug this.

dlasky: Could throw integration error -- I'd have to confirm if any part of that is visible in event-delivery

lew-gordon: The other hard part about this is that the customer may or may not know these destinations actually exist.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -238,100 +241,117 @@ const action: ActionDefinition<Settings, Payload> = {
...
+              substitution_tag: '[unsubscribe]'
+            }
           }
         }
-      }
-    })
+      })
+    } else {
+      throw new IntegrationError(
+        `Failed to process an the subscribe state in the response: "${emailProfile.isSubscribed}"`,
```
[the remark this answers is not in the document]
lew-gordon: Yep I think that makes sense. Rereading what I original put I think I missed a thought in there. I believe initially this could have the potential to break things if all the enrichment isn't setup properly, which is currently the case. I believe we only have it set for a couple of workspaces. I think it'd be fine to take this approach, though.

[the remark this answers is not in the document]
larrycustodio: So I guess this would be a breaking change, since we've originally envisioned this change to also support old actions that would _not_ have `isSubscribed` in the payload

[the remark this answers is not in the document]
lew-gordon: Right and I think that's OK. I mean it should just be as easy as enabling this for a workspace.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -223,115 +197,170 @@ const action: ActionDefinition<Settings, Payload> = {
...
+        },
+        type: {
+          label: 'type',
+          description: 'The external ID contact type.',
+          type: 'string',
+        },
+        isSubscribed: {
+          label: 'ID',
+          description: 'The subscription status for the identity.',
+          type: 'string',
```
lew-gordon, line 219: Is this what you were referring to where the boolean is sent as a string?

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -223,115 +197,170 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          description: 'A unique identifier for the collection.',
+          type: 'string',
+        },
+        type: {
+          label: 'type',
+          description: 'The external ID contact type.',
+          type: 'string',
+        },
+        isSubscribed: {
+          label: 'ID',
```
lew-gordon, line 217: Should the label be "Is subscribed"?

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -223,115 +197,170 @@ const action: ActionDefinition<Settings, Payload> = {
...
-        'Emails with gmailx.com, yahoox.com, aolx.com, and hotmailx.com domains are blocked.',
-        'Invalid input',
+        `The subscribed state found in the response is ${String(emailProfile?.isSubscribed) || 'null'}`,
+        'Invalid isSubscribed value',
         400
       )
     }
+    if (['unsubscribed', 'did not subscribed', 'false'].includes(emailProfile?.isSubscribed)) {
+      return
+    } else if (['subscribed', 'true', 'false'].includes(emailProfile?.isSubscribed)) {
```
lew-gordon, line 263: I think this should just be `true` / `false`, right?

lew-gordon (commented):
Looks very close, maybe I just need clarification on a few things.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
larrycustodio: yep

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -223,115 +197,170 @@ const action: ActionDefinition<Settings, Payload> = {
...
+              '@path': '$.id'
+            },
+            type: {
+              '@path': '$.type'
+            },
+            isSubscribed: {
+              '@path': '$.isSubscribed'
+            }
+          }
+        ]
```
larrycustodio, line 236: @lew-gordon did you mean to say in the old reply, that we were supposed to target `external_ids[0].is_subscribed`?

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
larrycustodio: Updated!

larrycustodio: So this was I thought was worth thinking about -- the mapping kit seems to convert the mapped fields into the defined type. Since I had `isSubscribed` set to a typed `string` since we have enums, the boolean and null values get converted to a string literal `"true" | "false" | "null"` in the callback payload, respectively.

lew-gordon: Maybe suboptimal for now to work with, but it seems we may have to treat the booleans and null from the string values. If the logic gets too tricky maybe we can switch to enum values in the payload upstream.

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
larrycustodio: Source: https://segment.atlassian.net/browse/GRAMA-310?focusedCommentId=261046

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
larrycustodio: entirely up to us on how we want to label our payload keys. the mapping that @VenkataNeeharikaKondipati shared seems to target `external_ids[<TARGET_INDEX>].isSubscribed`:
https://segment.atlassian.net/browse/GRAMA-310?focusedCommentId=261046

packages/destination-actions/src/destinations/personas-messaging-sendgrid/sendEmail/index.ts
lew-gordon: I think this is correct. I was mentioning about what I think is a human readable label.

larrycustodio: how's `subscriptionStatus` sound?

lew-gordon: Sounds good to me!

lew-gordon (approved):
Looks good to me! :shipit: ✅

dlasky (approved):

PR #489, opened 2022-02-22
Fix braze destination debouncing

zikaari:
Braze updateUserProfile debouncing was not working because of
faulty reference in the debounce middleware.

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/debounce/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/debounce/index.ts
@@ -53,8 +53,8 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
     // Only send the event to Braze if a trait has changed
     // Target all possible Braze integration names
     const shouldSend = shouldSendToBraze(event)
-    ctx.updateEvent('integrations.Braze Web Mode (Actions)', shouldSend)
-    ctx.updateEvent('integrations.Braze Cloud Mode (Actions)', shouldSend)
+    ctx.updateEvent('integrations.Braze Web Mode (Actions) updateUserProfile', shouldSend)
+    ctx.updateEvent('integrations.Braze Cloud Mode (Actions) updateUserProfile', shouldSend)
     ctx.updateEvent('integrations.Appboy', shouldSend)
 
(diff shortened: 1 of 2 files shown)


packages/browser-destinations/src/destinations/braze/debounce/index.ts
```diff
@@ -53,8 +53,8 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
     // Only send the event to Braze if a trait has changed
     // Target all possible Braze integration names
     const shouldSend = shouldSendToBraze(event)
-    ctx.updateEvent('integrations.Braze Web Mode (Actions)', shouldSend)
-    ctx.updateEvent('integrations.Braze Cloud Mode (Actions)', shouldSend)
+    ctx.updateEvent('integrations.Braze Web Mode (Actions) updateUserProfile', shouldSend)
+    ctx.updateEvent('integrations.Braze Cloud Mode (Actions) updateUserProfile', shouldSend)
```
chrisradek, line 57: Just double checking - was this actually broken in the cloud mode destination as well? I don't know if they have logic to disable the entire integration or just the partnerAction, genuinely curious.

This might be a question for the cloud-mode actions team members: do we need to add `createAlias`/`identifyUser` for the cloud mode as well or do those not make sense to debounce?

zikaari:
> debouncing was occurring without this action enabled

This was fixed by this change [here](https://github.com/segmentio/analytics-next/pull/368/files#diff-86750b3fc44d461eb42175dc89c4163276d98346393d207bbb0dfd88a6874430R20-R22)

packages/browser-destinations/src/destinations/braze/debounce/index.ts
zikaari: I can't comment on the cloud mode situation. We should involve someone who has deeper knowledge on this subject.

dlasky: cc @nielst 

I'm not sure, depends on the intention of the debounce

mericsson:
Closing due to age. Please re-open if needed.

PR #502, opened 2022-03-03
[STRATCONN-1195] Address QA comments for Adobe Target Web 

marinhero:
This PR addresses the issues that were discovered in our first round of QA.

- [x] triggerView: The default event trigger (subscription) should be event type = page
- [x] upsertProfile: This should not have separate fields for anonymousId and userId
- [x] trackEvent: There should be two separate fields for "Event Name" and "Event Type"

## Testing

### DIFF
--- a/packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
+++ b/packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
@@ -7,23 +7,23 @@ const TARGET_EVENT_TYPE = 'click'
 
 const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   title: 'Track Event',
-  description: 'Track an event',
+  description: 'Send user actions, such as clicks and conversions, to Adobe Target.',
   platform: 'web',
   defaultSubscription: 'type = "track"',
   fields: {
-    event_name: {
+    type: {
+      label: 'Event Type',
+      description: 'The event type. Please ensure the type entered here is registered and available.',
+      type: 'string',
+      default: TARGET_EVENT_TYPE
+    },
+    eventName: {
       label: 'Event Name',
       description: 'This will be sent to Adobe Target as an event parameter called "event_name".',
       type: 'string',
       default: {
         '@path': '$.event'
       }
-    },
-    type: {
-      label: 'Event Type',
-      description: 'The event type. Please ensure the type entered here is registered and available.',
-      type: 'string',
-      default: TARGET_EVENT_TYPE
     },
     properties: {
       label: 'Event Parameters',
@@ -36,16 +36,17 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   },
   perform: (Adobe, event) => {
     // Adobe Target only takes certain event types as valid parameters. We are defaulting to "click".
+    const payload = event.payload
     const event_params = {
-      ...event.payload.properties,
-      event_name: event.payload.event_name
+      ...payload.properties,
+      event_name: payload.eventName
     }
 
     const params = {
       mbox: event.settings.mbox_name,
       preventDefault: true,
       params: event_params,
-      type: TARGET_EVENT_TYPE
+      type: payload.type
     }
 
     Adobe.target.trackEvent(params)
--- a/packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
+++ b/packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
@@ -2,7 +2,7 @@ import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import { Adobe } from '../types'
-import { setMbox3rdPartyId } from '../utils'
+import { setMbox3rdPartyId, serializeProperties } from '../utils'
 
 // Adobe Target only takes certain event types as valid parameters. We are defaulting to "click".
 const TARGET_EVENT_TYPE = 'click'
@@ -53,26 +53,8 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
     const payload = event.payload
     setMbox3rdPartyId(payload.userId)
 
-    // Track does not accept arrays as valid properties, therefore we are stringifying them.
-    const serialize_properties = (props: { [key: string]: unknown } | undefined) => {
-      if (props === undefined) {
-        return {}
-      }
-
-      const serialized: { [key: string]: unknown } = {}
-
-      for (const key in props) {
-        serialized[key] = props[key]
-        if (Array.isArray(props[key])) {
-          serialized[key] = JSON.stringify(props[key])
-        }
-      }
-
-      return serialized
-    }
-
     const event_params = {
-      ...serialize_properties(payload.properties),
+      ...serializeProperties(payload.properties),
       event_name: payload.eventName
     }
 
--- a/packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
+++ b/packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
@@ -3,6 +3,7 @@ import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import { Adobe } from '../types'
 
+// Adobe Target only takes certain event types as valid parameters. We are defaulting to "click".
 const TARGET_EVENT_TYPE = 'click'
 
 const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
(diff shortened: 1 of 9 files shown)


packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -28,10 +36,9 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
...
   perform: (Adobe, event) => {
     // Adobe Target only takes certain event types as valid parameters. We are defaulting to "click".
-    const TARGET_EVENT_TYPE = 'click'
     const event_params = {
       ...event.payload.properties,
-      event_name: event.payload.type
+      event_name: event.payload.event_name
     }
 
     const params = {
```
kdaswani, line 44: Should `type: TARGET_EVENT_TYPE` actually be `type: payload.type`?

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -28,10 +36,9 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   },
   perform: (Adobe, event) => {
     // Adobe Target only takes certain event types as valid parameters. We are defaulting to "click".
-    const TARGET_EVENT_TYPE = 'click'
     const event_params = {
       ...event.payload.properties,
```
kdaswani, line 40: Do we need the `event.` preface? From other Actions destinations, I thought this would just be `payload.properties` but could be different for web destinations?

```suggestion
      ...payload.properties,
```

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -28,10 +36,9 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
...
-      event_name: event.payload.type
+      event_name: event.payload.event_name
```
kdaswani, line 41: Do we need the `event.` preface? From other Actions destinations, I thought this would just be `payload.event_name` but could be different for web destinations?

```suggestion
      event_name: payload.event_name
```

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -3,20 +3,28 @@ import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import { Adobe } from '../types'
 
+const TARGET_EVENT_TYPE = 'click'
+
 const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   title: 'Track Event',
   description: 'Track an event',
```
kdaswani, line 10:
```suggestion
  description: 'Send user actions, such as clicks and conversions, to Adobe Target.',
```

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -3,20 +3,28 @@ import type { Settings } from '../generated-types'
...
       label: 'Event Name',
-      description: 'Event type or name.',
+      description: 'This will be sent to Adobe Target as an event parameter called "event_name".',
       type: 'string',
       default: {
-        '@path': '$.type'
+        '@path': '$.event'
       }
     },
+    type: {
```
kdaswani, line 22: Can we move this field above Event Name?

packages/browser-destinations/src/destinations/adobe-target/triggerView/index.ts
```diff
@@ -6,7 +6,8 @@ import { setPageParams } from '../utils'
 
 const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   title: 'Trigger View',
-  description: 'Record a view',
+  defaultSubscription: 'type = "page"',
+  description: 'Send page-level data to Adobe Target.',
   platform: 'web',
   fields: {
     viewName: {
```
kdaswani, line 13: The description here has a typo in the word `the`. Need to remove the "w" at the end :)

kdaswani: Also we use snake case for the track event fields above but camel case for these fields, i.e. viewName. Should we standardize? Looks like other actions destinations use snake cake so maybe we should update all the fields in this action to also be snake cased?

```suggestion
    view_name: {
```

packages/browser-destinations/src/destinations/adobe-target/upsertProfile/index.ts
```diff
@@ -10,11 +10,11 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   platform: 'web',
   defaultSubscription: 'type = "identify"',
   fields: {
-    userId: {
+    mbox3rdpartyid: {
```
kdaswani, line 13:
```suggestion
    mbox_third_party_id: {
```
or
```suggestion
    user_id: {
```

As long as the display label is Mbox 3rd Party ID, the key can be either of the above! We use `user_id` in Adobe Target Cloud cc @rvadera12

packages/browser-destinations/src/destinations/adobe-target/upsertProfile/index.ts
```diff
@@ -23,15 +23,6 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
...
-    anonymousId: {
-      type: 'string',
-      required: true,
-      description: 'Anonymous identifier for the user',
-      label: 'Anonymous ID',
-      default: {
-        '@path': '$.anonymousId'
-      }
-    },
     traits: {
```
kdaswani, line 26: Can we change the `label` to be `Profile Attributes` instead so it matches the language we use in Adobe Target Cloud?

kdaswani (commented):
This looks good, thanks for making those changes! Add a few comments and changes. 

I am also curious if this chunk in the `perform` of Upsert Profile is necessary because we are already using a coalesce default, which should handle this fallback logic. Maybe we can remove this?

```
setMbox3rdPartyId(event.payload.anonymousId)
    if (event.payload.userId) {
      setMbox3rdPartyId(event.payload.userId)
    }
```

Would be great to get others on the team to review as well 😄

packages/browser-destinations/src/destinations/adobe-target/upsertProfile/index.ts
rvadera12: I used user_id instead of userId since I noticed most of the action destination fields use snake case.

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
marinhero: I think they extract it from the `event` object into a new variable. I'll do the same. Good 👁️ !

packages/browser-destinations/src/destinations/adobe-target/triggerView/index.ts
marinhero: Good catch! Looks like the convention is to have camel case. I'll switch the variables to that format instead.

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -24,21 +34,53 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
...
+          then: { '@path': '$.userId' },
+          else: { '@path': '$.anonymousId' }
+        }
+      }
     }
   },
   perform: (Adobe, event) => {
-    // Adobe Target only takes certain event types as valid parameters. We are defaulting to "click".
-    const TARGET_EVENT_TYPE = 'click'
+    const payload = event.payload
```
kdaswani, line 53: I don't think we even need to define payload again because isn't it defined in the object above? I may be getting confused but looking at cloud-mode, we don't define `payload` in the perform but are still able to target the fields set above: https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/adobe-target/updateProfile/index.ts#L37.

kdaswani: oh nevermind I think I see what we're doing! i.e. `perform: (Adobe, event)`

ignore me!

packages/browser-destinations/src/destinations/adobe-target/trackEvent/index.ts
```diff
@@ -24,21 +34,53 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
...
+    setMbox3rdPartyId(payload.userId)
+
+    // Track does not accept arrays as valid properties, therefore we are stringifying them.
+    const serialize_properties = (props: { [key: string]: unknown } | undefined) => {
```
mericsson, line 57: suggest pulling this inner function out into it's own function and camelcasing. And then adding unit tests to cover it specifically.

Is this something we are using elsewhere? if so, maybe should be a shared util function.

kdaswani (commented):
LGTM! Only remaining thing to figure out is why event params are not appearing in Target when they are in the outgoing network request 😭

marinhero:
Reviewers, we are waiting on Adobe to answer our questions regarding the event parameters. This code is still good for review. Any required changes will be addressed independently. Thanks!

rvadera12 (approved):

PR #517, opened 2022-03-11
Actions tester Server

dlasky:
Adds support for websocket server + manifest route needed by the actions tester UI

## Testing

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -6,6 +6,8 @@ import chokidar from 'chokidar'
 import ora from 'ora'
 import path from 'path'
 import globby from 'globby'
+import { WebSocketServer } from 'ws'
+import open from 'open'
 
 export default class Serve extends Command {
   private spinner: ora.Ora = ora()
@@ -77,6 +79,16 @@ export default class Serve extends Command {
       cwd: process.cwd()
     })
 
+    const DEFAULT_PORT = 3000
+    const port = parseInt(process.env.PORT ?? '', 10) || DEFAULT_PORT
+    const wss = new WebSocketServer({ port: port + 1 })
+
+    wss.on('connection', function connection(ws) {
+      watcher.on('change', () => {
+        ws.send('change')
+      })
+    })
+
     const start = () => {
       child = fork(require.resolve('../lib/server.ts'), {
         cwd: process.cwd(),
@@ -124,6 +136,9 @@ export default class Serve extends Command {
 
     watcher.once('ready', () => {
       this.log(chalk.greenBright`Watching required files for changes .. `)
+      //todo replace this with prod url
+      // open(`http://localhost:8000/partner-portal/actions-tester`)
+      void open('https://app.segment.build/partner-portal/actions-tester')
     })
 
     start()
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -136,9 +136,8 @@ export default class Serve extends Command {
 
     watcher.once('ready', () => {
       this.log(chalk.greenBright`Watching required files for changes .. `)
-      //todo replace this with prod url
-      // open(`http://localhost:8000/partner-portal/actions-tester`)
-      void open('https://app.segment.build/partner-portal/actions-tester')
+      this.log(chalk.greenBright`Visit https://app.segment.com/dev_center/actions-tester to preview your integration.`)
+      void open('https://app.segment.com/dev_center/actions-tester')
     })
 
     start()
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -7,7 +7,7 @@ import ora from 'ora'
 import path from 'path'
 import globby from 'globby'
 import { WebSocketServer } from 'ws'
-import open from 'open'
+// import open from 'open'
 
 export default class Serve extends Command {
   private spinner: ora.Ora = ora()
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/serve.ts
```diff
@@ -124,6 +136,8 @@ export default class Serve extends Command {
 
     watcher.once('ready', () => {
       this.log(chalk.greenBright`Watching required files for changes .. `)
+      this.log(chalk.greenBright`Visit https://app.segment.com/dev_center/actions-tester to preview your integration.`)
+      void open('https://app.segment.com/dev_center/actions-tester')
```
mericsson, line 140: Consider this format for consistency

```suggestion
      this.log(chalk.greenBright`Visit https://app.segment.com/dev-center/actions-tester to preview your integration.`)
      void open('https://app.segment.com/dev-center/actions-tester')
```

packages/cli/src/commands/serve.ts
```diff
@@ -124,6 +136,8 @@ export default class Serve extends Command {
 
     watcher.once('ready', () => {
       this.log(chalk.greenBright`Watching required files for changes .. `)
+      this.log(chalk.greenBright`Visit https://app.segment.com/dev-center/actions-tester to preview your integration.`)
+      void open('https://app.segment.com/dev-center/actions-tester')
```
nick-Ag, line 140: Will this link be opened every time someone runs `./bin/run serve`?

dlasky: yes, similar to starting `app`

packages/cli/src/lib/server.ts
```diff
@@ -13,15 +14,53 @@ import {
...
+const marshalError = (err: Error): ErrorOutput => {
+  const error = err as ResponseError
+  let statusCode = error?.status ?? 500
+  let msg = error?.message
+  let fields: { [key: string]: string } | undefined
+
+  if (error.name === 'AggregateAjvError') {
+    fields = {}
+    const ajvErr = error as AggregateAjvError
+    for (const fieldError of ajvErr) {
```
gurdasnijor, line 39: One of the downsides of working with ajv in a ts codebase :/ 

I'd much prefer something like https://github.com/jquense/yup#getting-started (but changing our json validation lib is a bit out of scope for this i'd imagine :) )

packages/cli/src/commands/serve.ts
```diff
@@ -77,6 +79,16 @@ export default class Serve extends Command {
       cwd: process.cwd()
     })
 
+    const DEFAULT_PORT = 3000
+    const port = parseInt(process.env.PORT ?? '', 10) || DEFAULT_PORT
+    const wss = new WebSocketServer({ port: port + 1 })
```
nick-Ag, line 84: Will the existing server at port 3000 and the new web socket always be running together? I wonder if we should allow running one or the other exclusively? (Not sure if there's a noticeable performance diff)

packages/cli/src/lib/server.ts
dlasky: that was just a refactor of some existing stuff--im not quite sure what it was being used for as I couldn't get it to reproduce, but I wanted to bring a consistent error handler on this side so the client can parse it uniformly

packages/cli/src/commands/serve.ts
dlasky: they are currently both required to function, the websocket server is used only for telling the client to update.

packages/cli/src/lib/server.ts
gurdasnijor: 👍🏽

packages/cli/src/lib/server.ts
```diff
@@ -13,15 +14,53 @@ import {
...
+  }
+
+  return { statusCode, message: msg, stack: err.stack?.split('\n'), fields, requestError: true }
+}
+
 const app = express()
 app.use(express.json())
+app.use(
+  cors({
+    origin: ['https://app.segment.build', 'https://app.segment.com', 'http://localhost:8000']
```
gurdasnijor, line 58: Should we get our eu subdomains in here too?

dlasky: Good question, its technically not user data so i'm not sure it matters for GDPR? but i suppose if a destination author's workspace is EU they would be at that domain?

packages/cli/src/lib/server.ts
```diff
@@ -103,6 +142,13 @@ function setupRoutes(def: DestinationDefinition | null): void {
 
   const router = express.Router()
 
+  router.get(
+    '/manifest',
+    asyncHandler(async (_, res: express.Response) => {
+      res.json(destination.definition)
```
gurdasnijor, line 148: I've been out of express-land for quite some time - is it not necessary to call `next()` to yield control to the next middleware in the stack anymore?

dlasky: i think thats only for middleware, routes just get req, res cause they are effectively the 'last' step in the chain

gurdasnijor (approved):
Still bringing up a local env to do some functional testing of this + https://github.com/segmentio/app/pull/12997 - going to give the 👍🏽 for now to unblock getting this released and will share results/feedback after - awesome work!  👏🏽

packages/cli/src/lib/server.ts
gurdasnijor: Ahhhhh missed that - thanks!

packages/cli/src/commands/serve.ts
dlasky: i guess it does make sense to comment this for now until the UI portion is accessible to all users

PR #527, opened 2022-03-18
Update fqlExpression to be able to handle new == operator for booleans

LLuque-twilio:

This PR address [JIRA-617](https://segment.atlassian.net/browse/BE-617)

In it, we are attempting to include a new possible operator `==` as well as update the fqlExpression that is generated when the given case operator is `==`. In app, this will allow us to uniquely identify the two mapping operators `is (string)` and `is (boolean)`. Currently, we only have `is` which treats the input as a string in every instance, even when the user actually wants it as a boolean.

Desired effect in app:
Screen Shot 2022-03-17 at 5 28 23 PM

Without this change, all values sent to fqlExpression will be returned as strings instead of their desired effect as a boolean.

This PR goes hand in hand with [this]() PR in app (currently not ready)

## Testing

### DIFF
--- a/packages/destination-subscriptions/src/generate-fql.ts
+++ b/packages/destination-subscriptions/src/generate-fql.ts
@@ -1,146 +1,138 @@
 import {
-	Subscription,
-	GroupCondition,
-	EventTypeCondition,
-	EventCondition,
-	EventPropertyCondition,
-	EventTraitCondition,
-	EventContextCondition,
-	EventUserIdCondition,
-	EventNameCondition,
-	Operator,
-	ErrorCondition
+  Subscription,
+  GroupCondition,
+  EventTypeCondition,
+  EventCondition,
+  EventPropertyCondition,
+  EventTraitCondition,
+  EventContextCondition,
+  EventUserIdCondition,
+  EventNameCondition,
+  Operator,
+  ErrorCondition
 } from './types'
 
-const stringifyValue = (
-	value: string | boolean | number | undefined
-): string => {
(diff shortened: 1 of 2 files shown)


gurdasnijor:
Code LGTM on first pass - would love to see a test around equality checking on booleans
 
Check out a few of these suites around equality checking for other data types for inspo:
https://github.com/segmentio/action-destinations/blob/main/packages/destination-subscriptions/src/__tests__/fql.test.ts#L685

packages/destination-subscriptions/src/generate-fql.ts
```diff
@@ -1,146 +1,138 @@
...
+    case 'starts_with':
+      return `match(${name}, "${String(value)}*")`
+    case 'not_starts_with':
+      return `!match(${name}, "${String(value)}*")`
+    case 'ends_with':
+      return `match(${name}, "*${String(value)}")`
+    case 'not_ends_with':
+      return `!match(${name}, "*${String(value)}")`
+    case '==':
+      return `${name} = ${Boolean(value)}`
```
gurdasnijor, line 42: Would we need the not equals `!=` operator here too?  (i'm not super familiar with what we expose in the UI just yet, so not sure if this is a requirement)

gurdasnijor:
A few potential scenarios we'd want to test:

- Null handling [Boolean(undefined | null | falsey-thing-here)] *should* always work in most modern JS engines, but this library can/probably will be run in tons of different environments that may not work the way we expect [here's the spec for reference https://262.ecma-international.org/5.1/#sec-4.3.15 ]\
-

PR #533, opened 2022-03-21
only call browser initialize once per destination

mshwery:
Hello!

I noticed when testing the Koala browser destination that the `initialize` function gets called multiple times – once for every subscription, which is a bit counterintuitive. I believe the intention of this function is to be called once no matter how many subscriptions are active (since it does things like loading a script from a cdn, etc).

After this change, initialize will only be called once per destination, and subsequent `load` invocations will yield the same promise. After the first one finishes, the `client` and `hasInitialized` will both be set.

cc @pooyaj @dlasky

## Testing

Tested manually using `yarn dev`

### DIFF
--- a/packages/browser-destinations/src/runtime/__tests__/plugin.test.ts
+++ b/packages/browser-destinations/src/runtime/__tests__/plugin.test.ts
@@ -1,0 +1,62 @@
+import { BrowserDestinationDefinition } from 'src/lib/browser-destinations'
+import { generatePlugins } from '../plugin'
+
+describe('generatePlugins', () => {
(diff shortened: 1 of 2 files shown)


chrisradek:
@mshwery Thanks, this is a really great update! I couldn't push directly to your branch but I did create a PR to your branch to add a test here:
https://github.com/koala-live/action-destinations/pull/1

Once the test is merged in I'd be happy to approve it!

chrisradek (approved):

chrisradek:
Merging into segmentio/initialize-once so CI can complete...

PR #537, opened 2022-03-23
Update support for new operators in types.ts, generate-fql.ts and parse-fql.ts

LLuque-twilio:

This PR addresses [JIRA-617](https://segment.atlassian.net/browse/BE-617).

This action-destination PR is the backbone for [this PR](https://github.com/segmentio/app/pull/13048) for the app repo.

In it we add the correct logic for the two new operators `is true` and `is false` to be handled correctly when generating/parsing FQL.

Unfortunately the previous editor of these files used tabs instead of spaces. So when saving the new changes it updated the files to be correctly formatted which basically shows that the whole file has been changed in the diff. Gurdas has given me a great hack that you can add `?w=1 to the end of the URL to see all the changes that don't include white space. Or just follow [this link](https://github.com/segmentio/action-destinations/pull/537/files?w=1) to have it done for you.

## Testing

Testing can be found in the app PR [here](https://github.com/segmentio/app/pull/13048#:~:text=true/false%20respectively.-,Testing,-Testing%20completed%20successfully)

In addition, 3 new tests have been written and all pass as expected.


packages/destination-subscriptions/src/parse-fql.ts
```diff
@@ -1,402 +1,439 @@
...
+          throw new Error('Operator token is missing')
+        }
+
+        const valueToken = tokens.shift()
+
+        if (!valueToken) {
+          throw new Error('Value token is missing')
+        }
+
+        const isTrue = operatorToken.value === '=' && checkBooleanTrue(valueToken.value) === true
```
mericsson, line 193: `checkBooleanTrue` feels a little too heavy. is this equivalent?

```suggestion
        const isTrue = operatorToken.value === '=' && valueToken.value === 'true'
```

packages/destination-subscriptions/src/parse-fql.ts
```diff
@@ -1,402 +1,439 @@
...
+        const isFalse = operatorToken.value === '=' && checkBooleanFalse(valueToken.value) === true
```
mericsson, line 194: Is this equivalent?

```suggestion
        const isFalse = operatorToken.value === '=' && valueToken.value === 'false'
```

dlasky: I actually suggested he make a method there, but _nit_ i wasn't expecting two methods, and I also thought we would have a few more cases to consider internally.

LLuque-twilio: These should work the same, will test it out to be sure.

LLuque-twilio: Per Dan's comment: Since we don't give the user the option to input how they want to represent the boolean that is saved we shouldn't have to check for other cases such as `0`, `1`, `'TRUE'`, `'FALSE'`, etc. We automatically manually save the value as either `'true'`/`'false'`  when the operator  used is `is_true`/`is_false` so I believe we only have to check for these cases.

LLuque-twilio: Also, testing with the recommended changes was successful

jinapark202 (approved):
LGTM 👏

PR #539, opened 2022-03-24
[Criteo Audiences] Fix error types and remove internal retry

pharika31:
- Removes internal retry logic.
- Replaces `Error` with `IntegrationError` class.
- [x] Local E2E and unit tests running correctly

### DIFF
--- a/packages/destination-actions/src/destinations/criteo-audiences/criteo-audiences.ts
+++ b/packages/destination-actions/src/destinations/criteo-audiences/criteo-audiences.ts
@@ -56,7 +56,7 @@ export const getAccessToken = async (
     const body = await res.json()
 
     if (res.status !== 200)
-        throw new Error(`Error while getting an access token: ${JSON.stringify(body)}`)
+        throw new IntegrationError(`Error while getting an access token: ${JSON.stringify(body)}`)
 
     return body.access_token
 }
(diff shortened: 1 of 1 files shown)


brennan:
@pharika31 Would rebase from master, please? Looks like there's only one new commit but the PR contains all original commits from the previous PR as well. Thanks!

brennan:
Thanks, @MarcDillar and @pharika31! This PR generally looks good to me. We should be able to deploy these changes on Monday, March 28 (since we don't deploy on Fridays). I'll follow up with the team on Slack once the deploy is complete.

brennan (approved):

MarcDillar:
Hi @brennan! We just spotted an issue in our code, due to the `throwHttpErrors` request option being `true` by default.
We fixed that and updated the unit tests to cover this case as well.
Can you review the changes please?
Thank you very much

brennan:
@MarcDillar Would you be able to rebase on master again, please, so I know we have only the most up-to-date changes in this branch? Thanks! cc @pharika31

MarcDillar:
Hi @brennan @mericsson 
We just pushed the following changes:
- removed the retry mechanism we implemented as Segment is already handling retries for HTTP 4xx 5xx errors
- replaced the Errors thrown by the criteo-audiences destination by IntegrationErrors

Can you review and merge these changes please?

PR #547, opened 2022-03-31
Upload normal and obfuscated destinations

arielsilvestri:
Adds obfuscated directories for action destinations.

Content was already being content hashed and probably does not need to be changed with this work.


chrisradek (approved):
So simple! Do we need to update any scripts to get the bundles uploaded?

arielsilvestri:
@chrisradek Nope! Should all work without any changes to the build script afaik. It's all part of the same webpack build flow.

PR #554, opened 2022-04-06
Talon.One - fix `updateCustomerProfilesAttributes` action slug

kaatinga:
- The `attributes` fields did have separate labels, so in the UI form we have two equal fields for attributes. The descriptions and labels are altered.
- The request mapper updated to allow using multiple data items in the request.

## Testing

- Additional test case was added in the `updateCustomerProfilesAttributes` slug.


packages/destination-actions/src/destinations/talon-one/updateCustomerProfilesAttributes/index.ts
```diff
@@ -7,21 +7,31 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      label: 'Data item to change customer profile attributes',
+      description:
+        'An array of JSON objects that contains customer profile identifier and list of attributes and their values. Customer profile ID is required.',
+      type: 'object',
+      multiple: true,
+      properties: {
+        customerProfileId: { ...customerProfileId },
+        attributes: { ...attribute }
+      },
+      required: true
```
nick-Ag, line 20: Is there a default Mapping that would make sense for data and customerProfileId since those are required fields ?

nick-Ag (approved):
Left a minor comment but otherwise looks good to me 👍

packages/destination-actions/src/destinations/talon-one/updateCustomerProfilesAttributes/index.ts
kaatinga: @nick-Ag Thank you for the review! What is default mapping? Could you elaborate?

dlasky: There is a default key on each field entry that allows you to specify a "sensible" mapping from the incoming SegmentEvent to the payload you receive in the perform method. Amplitude logEvent is a good destination to look at for some examples. 

The default can be any mapping kit directive (see the core package for a list of directives available  there)

But in most cases a simple path directive will suffice. 

```
default: { "@path":"$.userId"} // points at userId on the incoming segment event which is a root level property. 
```

kaatinga: thank you for explanation! we will plan this in our next PR

PR #562, opened 2022-04-08
Obfuscate bundle names 2.0

arielsilvestri:
Hooks into the webpack build process to generate predictable bundle names for us to consume on the frontend without needing to make database changes for the metadata of each action-destination.

## Testing

Manual testing - uploaded to S3, made API request from ajs 2.0 successfully.

### DIFF
--- a/packages/browser-destinations/webpack.config.js
+++ b/packages/browser-destinations/webpack.config.js
@@ -10,20 +10,9 @@ const isProd = process.env.NODE_ENV === 'production'
 
 const entries = files.reduce((acc, current) => {
   const [_dot, _src, _destinations, destination, ..._rest] = current.split('/')
-  const obfuscatedDestination = Buffer.from(destination).toString('base64').replace(/=/g, '');
   return {
     ...acc,
     [destination]: current,
-    [obfuscatedDestination]: current,
-  }
-}, {})
-
-const entriesToObfuscatedNameMap = files.reduce((acc, current) => {
-  const [_dot, _src, _destinations, destination, ..._rest] = current.split('/')
-  const obfuscatedDestination = Buffer.from(destination).toString('base64').replace(/=/g, '');
-  return {
-    ...acc,
-    [obfuscatedDestination]: destination,
   }
 }, {})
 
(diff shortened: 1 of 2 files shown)


packages/browser-destinations/webpack.config.js
```diff
@@ -18,6 +18,15 @@ const entries = files.reduce((acc, current) => {
   }
 }, {})
 
+const entiresToObfuscatedNameMap = files.reduce((acc, current) => {
```
danieljackins, line 21:
```suggestion
const entriesToObfuscatedNameMap = files.reduce((acc, current) => {
```

packages/browser-destinations/webpack.config.js
```diff
@@ -42,7 +51,12 @@ module.exports = {
   mode: process.env.NODE_ENV || 'development',
   devtool: 'source-map',
   output: {
-    filename: process.env.NODE_ENV === 'development' ? '[name].js' : '[name]/[contenthash].js',
+    filename: (file) => {
+      if (entiresToObfuscatedNameMap[file.chunk.name]) {
```
danieljackins, line 55:
```suggestion
      if (entriesToObfuscatedNameMap[file.chunk.name]) {
```

packages/browser-destinations/webpack.config.js
```diff
@@ -42,7 +51,12 @@ module.exports = {
...
+        return process.env.NODE_ENV === 'development' ? '[name].js' : '[name]/[name].js'
```
danieljackins, line 56: Should this be something like [obfuscatedName]/[contenthash]? I might be misunderstanding, but I imagine something like this

Normal:
fullstory/somecontenthash1.js
fullstory/somecontenthash2.js
...
Obfuscated:
ZnVsbHN0b3J5/somecontenthash1.js
ZnVsbHN0b3J5/somecontenthash2.js
...

arielsilvestri: Not quite in this case.

What we want to do at the end of the day is make sure that we can predictable translate whatever action we should be getting on from the frontend to something that exists on the backend.

If we use contenthashing for the obfuscated bundles, there won't be a way for us to construct the url ourselves from just the action name on the front end. 

To use your example, it makes it so that we can request `ZnVsbHN0b3J5/ZnVsbHN0b3J5.js` from the frontend since we know the action we're fetching is fullstory.

danieljackins: How will we differentiate between different versions of the fullstory package? 

I think we could still construct the url ourselves because the normal url with the content hash is in the database (and is in the settings endpoint)

arielsilvestri: The issue there is the content hash isn't consistent across both the obfuscated and unobfuscated files. 

What I could do here instead is use a map solution similar to what I'm doing to make sure the proper files are obfuscated, and store the contenthash that's been generated for that specific unobfuscated bundle and then just return that during build time

silesky: Will this affect browser caching at all?  For example, a user loads a page where the fullstory bundle has been upgraded — is there another way we are busting the cache if the content hash isn’t in the file name?

chrisradek: > The issue there is the content hash isn't consistent across both the obfuscated and unobfuscated files.

Why is this the case? I thought the contents of the bundle were unchanged and just the path where the bundle is stored was changed.

packages/browser-destinations/webpack.config.js
```diff
@@ -18,6 +18,15 @@ const entries = files.reduce((acc, current) => {
   }
 }, {})
 
+const entriesToObfuscatedNameMap = files.reduce((acc, current) => {
```
silesky, line 21: Just addressing the logic:

`entriesToObfuscatedNameMap` is just being used to check whether an element is in a set.

Would this make more sense -- 
```
const obfuscatedDestinationNames = files.reduce((acc, current) => {
  const [_dot, _src, _destinations, destination, ..._rest] = current.split('/')
  const obfuscatedDestination = Buffer.from(destination).toString('base64').replace(/=/g, '');
  return acc.add(obfuscatedDestination)
}, new Set([]))
```

And then

```
      if (obfuscatedDestinationNames.has(file.chunk.name)) {
```

packages/browser-destinations/webpack.config.js
arielsilvestri: @silesky The caching strategies for obfuscated and unobfuscated bundles will be the same for each experience individually, but there won;t be crossover there unfortunately.
 
 @chrisradek I've been trying to figure that out, my best guess so far has been that the filename also counts as the contents of the bundle, and in this case that's a base64 string instead of the original action name that we use.

packages/browser-destinations/webpack.config.js
arielsilvestri: That section of the code is a bit fluid right now as I figure out what the best path forward is for a solution, but if it remains I think that's a good suggestion!

packages/browser-destinations/webpack.config.js
```diff
@@ -37,73 +35,91 @@ plugins.push(
...
+
+const unobfuscatedOutputOptions = {
+  filename: (file) => process.env.NODE_ENV === 'development' ? `${file.chunk.name}.js` : `${file.chunk.name}/[contenthash].js`
+}
+
+const obfuscatedOutputOptions = {
+  filename: (file) => {
+    const obfuscatedOutput = Buffer.from(file.chunk.name).toString('base64').replace(/=/g, '');
+    return process.env.NODE_ENV === 'development' ? `${obfuscatedOutput}.js` : `${obfuscatedOutput}/[contenthash].js`
+  }
```
arielsilvestri, line 119: Looks a little clunky to also use the function version of creating the filename for unobfuscated builds, but this was the only way to get the content hashes to be consistent for both.

arielsilvestri:
With the solution as-is, the only thing we'll need to do make sure we do is copy over the existing versions of the unobfuscated action destination directories to the obfuscated directories so we can guarantee parity from this point on. Not really that big of a deal in the grand scope of things, and we should only need to do so once.

packages/browser-destinations/webpack.config.js
silesky: I feel like templetizing the config for just one option seems a bit overkill (IMO) — I thought the old version was a bit clearer IMO :-).

packages/browser-destinations/webpack.config.js
```diff
@@ -37,73 +35,91 @@ plugins.push(
...
 }
+
+const unobfuscatedOutput = outputTempalte(unobfuscatedOutputOptions)
```
silesky, line 122: typo — “outputTemplate” :-)

packages/browser-destinations/webpack.config.js
arielsilvestri: Yeah, now that I look at it again it does feel a little unnecessary. Being verbose for two configs seems fine, we can DRY things up if it ever gets bigger than this

packages/browser-destinations/webpack.config.js
```diff
@@ -107,3 +105,79 @@ module.exports = {
...
 }
+
+const obfuscatedOutput = {
+  entry: entries,
+  mode: process.env.NODE_ENV || 'development',
+  devtool: 'source-map',
+  output: {
+    filename: (file) => {
+      const obfuscatedOutput = Buffer.from(file.chunk.name).toString('base64').replace(/=/g, '');
+      return process.env.NODE_ENV === 'development' ? `${obfuscatedOutput}.js` : `${obfuscatedOutput}/[contenthash].js`
```
silesky, line 116: I like how these are dependencies of eachother -- much more DRY

packages/browser-destinations/webpack.config.js
```diff
@@ -37,12 +35,12 @@ plugins.push(
   })
 )
```
silesky, line 37: I like how two builds are dependencies of eachother -- much more DRY!

packages/browser-destinations/webpack.config.js
```diff
@@ -107,3 +105,79 @@ module.exports = {
...
+      const obfuscatedOutput = Buffer.from(file.chunk.name).toString('base64').replace(/=/g, '');
+      return process.env.NODE_ENV === 'development' ? `${obfuscatedOutput}.js` : `${obfuscatedOutput}/[contenthash].js`
+    },
+    path: path.resolve(__dirname, 'dist/web'),
+    publicPath: 'auto', // Needed for customers using custom CDNs with analytics.js
+    library: '[name]Destination',
+    libraryTarget: 'umd',
+    libraryExport: 'default'
+  },
+  module: {
```
silesky, line 124: Maybe we can make this more DRY -- but avoid turning this into a function (since webpack already allows config-as-a-function) like before. 
```ts
const obfuscatedOutput = {
 …unobfuscatedOutput,
 output: {
   …unobfuscatedOuput.output,
    filename: (file) => {
      // rename variable to avoid shadowing
      const obfuscatedOutputName = Buffer.from(file.chunk.name).toString('base64').replace(/=/g, '');
      return process.env.NODE_ENV === 'development' ? `${obfuscatedOutputName}.js` : `${obfuscatedOutputName}/[contenthash].js`
    },
 }
}
```
^ something like

packages/browser-destinations/webpack.config.js
```diff
@@ -107,3 +105,79 @@ module.exports = {
...
   plugins
 }
+
+const obfuscatedOutput = {
+  entry: entries,
+  mode: process.env.NODE_ENV || 'development',
+  devtool: 'source-map',
+  output: {
+    filename: (file) => {
+      const obfuscatedOutput = Buffer.from(file.chunk.name).toString('base64').replace(/=/g, '');
```
silesky, line 115: can we rename `obfuscatedOutput` to `obfuscatedOutputName` to avoid a naming conflict / shadowing?

packages/browser-destinations/webpack.config.js
arielsilvestri: I was thinking about doing this but was undecided at the time.

silesky: I think it makes a lot of sense -- it's definitely easier to understand the differences between the two configs!

silesky (approved):
LGTM

dlasky (approved):

PR #575, opened 2022-04-25
[Salesforce] - Bulk PoC

nick-Ag:
A PoC of Bulk API support.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/salesforce/account/index.ts
+++ b/packages/destination-actions/src/destinations/salesforce/account/index.ts
@@ -196,8 +196,6 @@ const action: ActionDefinition<Settings, Payload> = {
     const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
 
     if (payload[0].operation === 'bulkUpsert') {
-      console.log('routed correctly')
-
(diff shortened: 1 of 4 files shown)


nick-Ag:
This PoC code can bulk upsert to Salesforce based on a hardcoded CSV. Right now it doesn't include actually building that CSV as that is the next thing I'm going to explore doing.

packages/destination-actions/src/destinations/salesforce/account/index.ts
```diff
@@ -11,6 +11,12 @@ const action: ActionDefinition<Settings, Payload> = {
   fields: {
     operation: operation,
     traits: traits,
+    externalIdFieldName: {
+      label: 'External ID field name',
+      type: 'string',
+      required: true,
+      description: 'The name of the field that will be used to identify the account in Salesforce.'
```
mericsson, line 18: Suggest something like this

```suggestion
      description: 'API Name of the External ID field on Account'
```

https://help.salesforce.com/s/articleView?id=000325076&type=1

kdaswani: I think we can probably remove this though since we'll want customers to use the "Record Matchers" field and input the external ID there.

PR #580, opened 2022-05-05
Talon.One - added default mapping

kaatinga:
There is a default key on each field entry that allows to specify a "sensible" mapping from the incoming SegmentEvent to the payload you receive in the perform method.

example:
```
default: { "@path":"$.userId"}
```

The PR resulted from a comment in the previous Talon.One: PR github: https://github.com/segmentio/action-destinations/pull/554#pullrequestreview-934191250

## Testing

The default mapping must be seen chooen in the Segment UI forms.


kaatinga:
1. I have not found any possibility to indicate `parameter type` for `event type` here: https://github.com/talon-one/action-destinations/blob/sc-8437-add-default-mappings-to-our-integration-in/packages/destination-actions/src/destinations/talon-one/trackEvent/index.ts#L31. Is it possible to write something like: `default: { '@path': '$.event.parameter_type' }` ? Now I am using a constant string as a temporary solution.
2. Here we have `audience ID` that must be `audience ID` in Talon.One. https://github.com/talon-one/action-destinations/blob/sc-8437-add-default-mappings-to-our-integration-in/packages/destination-actions/src/destinations/talon-one/updateCustomerProfile/index.ts#L12. How do you store the parameters you receive from Talon.One responses? The response in the endpoint https://github.com/talon-one/action-destinations/tree/sc-8437-add-default-mappings-to-our-integration-in/packages/destination-actions/src/destinations/talon-one/createAudience contains that `audience ID`  we need in the request [above](https://github.com/talon-one/action-destinations/blob/sc-8437-add-default-mappings-to-our-integration-in/packages/destination-actions/src/destinations/talon-one/updateCustomerProfile/index.ts#L12).
3. Here https://github.com/talon-one/action-destinations/blob/sc-8437-add-default-mappings-to-our-integration-in/packages/destination-actions/src/destinations/talon-one/createAudience/index.ts#L9 we have something called `segment audience ID` . It must be an entity in Segment. It is like a group that customer (userId) belongs to. Something like `Flower lover`.

benhorowitz (approved):

packages/destination-actions/src/destinations/talon-one/trackEvent/index.ts
```diff
@@ -16,16 +18,23 @@ const action: ActionDefinition<Settings, Payload> = {
...
+        '@path': '$.event'
+      }
     },
     type: {
       label: 'Type',
       description: 'Type of event. Can be only `string`, `time`, `number`, `boolean`, `location`',
       type: 'string',
-      required: true
+      required: true,
+      default: TARGET_EVENT_TYPE
```
dlasky, line 31: I believe this is what you are looking for here:

```
method: {
      label: 'Method',
      description: 'HTTP method to use.',
      type: 'string',
      choices: [
        { label: 'POST', value: 'POST' },
        { label: 'PUT', value: 'PUT' },
        { label: 'PATCH', value: 'PATCH' },
        { label: 'DELETE', value: 'DELETE' },
        { label: 'GET', value: 'GET' }
      ],
      default: 'POST',
      required: true
    },
    ```

dlasky (approved):

dlasky:
1. responded inline
2. We have no storage capabilities on the hot path. My recommendation there would be to do a lookup call internally, but be careful with deadlines there. It's also possible to use localStorage for this on web based destinations (that run in browser vs cloud mode as implemented in this PR)
3. not sure what you are looking for here, probably maps to a segment group event?

packages/destination-actions/src/destinations/talon-one/trackEvent/index.ts
kaatinga: > 1. responded inline
> 2. We have no storage capabilities on the hot path. My recommendation there would be to do a lookup call internally, but be careful with deadlines there. It's also possible to use localStorage for this on web based destinations (that run in browser vs cloud mode as implemented in this PR)
> 3. not sure what you are looking for here, probably maps to a segment group event?

thank you for your review! Starting to work on this.

kaatinga: Hello @dlasky @benhorowitz ! 

After the last meeting with Segment we decided to rework our integration for the sake of simplicity. It will resolve the point 2 from my questions. We will provide one more PR which will change `updateCustomerProfile` and `updateMultipleCutomerProfilesAudiences` destinations in our integration following some changes in our API.

The current PR is good to merge to the main brach and bring to production. We will continue to work on mapping afterwards.

PR #583, opened 2022-05-06
[Actions-Server] Pass in mappings object for batching & noActionsTester flag

nick-Ag:
To enable testing batching enabled destinations locally this PR adds:
* Option to pass in a `mapping` object that applies to all payloads in a request. Previously we would use only the first payload as a mapping, resulting in that payload being cloned to every event in the batch. We will fallback to this if no `mapping` object is provided.
* A new `-n` `--noActionsTester` flag which turns off the actions tester. Useful for testing batching locally with Insomnia, and avoiding having to close the actions-tester popup every time.

## Testing
Tested the new flag with `./bin/run serve -n`. The actions-tester does not open but the server accepts local insomnia requests as expected.

Tested locally by successfully sending through this request shape. All mappings apply as expected.
Screen Shot 2022-05-06 at 10 13 46 AM

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -32,6 +32,10 @@ export default class Serve extends Command {
       char: 'b',
       description: 'destination actions directory',
       default: './packages/destination-actions/src/destinations'
+    }),
+    noActionsTester: flags.boolean({
+      char: 'n',
+      description: 'do not open actions tester'
     })
   }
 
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/serve.ts
```diff
@@ -32,6 +32,10 @@ export default class Serve extends Command {
       char: 'b',
       description: 'destination actions directory',
       default: './packages/destination-actions/src/destinations'
+    }),
+    noActionsTester: flags.boolean({
```
dlasky, line 36: would this make more sense as `no ui` since we are still technically testing? (genuinely asking)

packages/cli/src/lib/server.ts
```diff
@@ -222,9 +222,13 @@ function setupRoutes(def: DestinationDefinition | null): void {
           }
 
           if (Array.isArray(eventParams.data)) {
-            // We assume that the the first payload in the data array
-            // provided for testing contains all actions mappings
-            eventParams.mapping = eventParams.data[0] || {}
+            // If no mapping is provided default to using the first payload across all events.
+            if (!req.body.mapping) {
```
dlasky, line 226: isnt this functionally identically to `req.body.mapping ?? eventParams.data[0] ?? {}`

dlasky (approved):

packages/cli/src/commands/serve.ts
nick-Ag: agree with you here, changed it to `noUI`

packages/cli/src/lib/server.ts
nick-Ag: yes, updated and tested that logic :+1:

PR #593, opened 2022-05-17
[Salesforce] Bulk Upsert

nick-Ag:
This PR adds support for batching to Salesforce. It includes a new "Bulk Upsert" operation which is included in every action. This is an initial build that will only be usable by internal Segment users. More details on that can be found in the [(Segment Internal) Bulk API Mini-SDD](https://paper.dropbox.com/doc/Mini-SDD-Salesforce-Bulk-API--BhoQssc8VEjaaGVvnkxwnG0EAg-OW8Dn2fzQRDPlCa8trLLG).

Corresponding App PR which locks this to Segment Internal: https://github.com/segmentio/app/pull/13309

## Testing
Deployed successfully to stage. Batches of events arrive in the `performBatch` method, and are correctly delivered downstream to Salesforce using the bulk job flow.

Some pictures:
Screen Shot 2022-05-18 at 4 28 06 PM

Screen Shot 2022-05-18 at 4 27 54 PM


packages/destination-actions/src/destinations/salesforce/sf-object-to-shape.ts
```diff
@@ -95,3 +95,13 @@ export const mapObjectToShape = (payload: GenericPayload, sobject: string): Gene
 
   return shapeFunction(payload)
 }
+
+// Our key names are in snake case, but Salesforce's field names are in camel case.
+// I.E. 'last_name' is our key, but 'LastName' is the Salesforce field name.
+export const snakeCaseToCamelCase = (key: string): string => {
```
mericsson, line 101: I believe we have lodash in this repo so can potentially use `_.camelCase` https://lodash.com/docs/#camelCase

packages/destination-actions/src/destinations/salesforce/opportunity/index.ts
```diff
@@ -60,6 +60,17 @@ const action: ActionDefinition<Settings, Payload> = {
       }
       return await sf.upsertRecord(payload, 'Opportunity')
     }
+  },
+  performBatch: async (request, { settings, payload }) => {
+    const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
+
+    if (payload[0].operation === 'bulkUpsert') {
```
nick-Ag, line 67: An open question I have: Right now I'm settling for checking the first item in Payload[]. I would like to loop thru the array and return errors only for those missing required fields, and proceed with the other events. Not sure if that's currently possible

dlasky: If they are required and missing I believe the validation step will have already discarded them--see the batching unit tests for an example there.

packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts
```diff
@@ -254,4 +255,66 @@ describe('Salesforce', () => {
       })
     })
   })
+
+  describe.only('Bulk Operations', () => {
```
mericsson, line 259:
```suggestion
  describe('Bulk Operations', () => {
```

packages/destination-actions/src/destinations/salesforce/account/index.ts
```diff
@@ -185,6 +185,17 @@ const action: ActionDefinition<Settings, Payload> = {
       }
       return await sf.upsertRecord(payload, 'Account')
     }
+  },
+  performBatch: async (request, { settings, payload }) => {
```
mericsson, line 189: I wonder if adding this will cause existing traffic to start hitting `performBatch`? Should we add fallback here to single `perform` if `bulkUpsert` is not enabled?

mericsson: (And same for all other added `performBatch` calls)

packages/destination-actions/src/destinations/salesforce/cases/index.ts
```diff
@@ -33,6 +33,13 @@ const action: ActionDefinition<Settings, Payload> = {
     if (payload.operation === 'upsert') {
       return await sf.upsertRecord(payload, 'Case')
     }
+  },
+  performBatch: async (request, { settings, payload }) => {
+    const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
+
+    if (payload[0].operation === 'bulkUpsert') {
+      return await sf.bulkUpsert(payload, 'Case')
```
mericsson, line 41: I think cleaner / safer to define a static `const` in this file and re-use.

```js
const objectName = 'Case'
```

and 

```js
return await sf.bulkUpsert(payload, objectName)
```

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -1,8 +1,13 @@
...
 import type { GenericPayload } from './sf-types'
-import { mapObjectToShape } from './sf-object-to-shape'
+import { mapObjectToShape, snakeCaseToCamelCase } from './sf-object-to-shape'
 
 export const API_VERSION = 'v53.0'
+const isSettingsKey = new Map<string, boolean>([
+  ['operation', true],
+  ['traits', true],
+  ['externalIdFieldName', true]
+])
```
mericsson, line 10: Nitpick, can use a `Set` instead.

```suggestion
const isSettingsKey = new Set<string>(['operation', 'traits', 'externalIdFieldName'])
```

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -68,6 +77,110 @@ export default class Salesforce {
...
+  private closeBulkJob = async (jobId: string) => {
+    return this.request(`${this.instanceUrl}services/data/${API_VERSION}/jobs/ingest/${jobId}`, {
+      method: 'PATCH',
+      json: {
+        state: 'UploadComplete'
+      }
+    })
+  }
+
+  private buildCSVData = (payloads: GenericPayload[], externalIdFieldName: string): string => {
```
mericsson, line 144: I thin we should add some unit tests for these `build*` functions.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -68,6 +77,110 @@ export default class Salesforce {
...
+    })
+    row += `"${payload?.traits?.externalIdFieldValue}"`
+
+    return row
+  }
+
+  private buildHeader = (payloads: GenericPayload[], externalIdFieldName: string): string => {
+    let header = ''
+    Object.keys(payloads[0]).forEach((key) => {
+      console.log('CSV: ', key)
```
mericsson, line 174:
```suggestion
```

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
mericsson: Probably good idea to pull it out into its own file too for clarity e.g. csv-util.ts

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -68,6 +77,110 @@ export default class Salesforce {
...
+  private buildRow = (payload: GenericPayload): string => {
+    let row = ''
+
+    // Iterate over the keys in payload that map to a SF object field
+    // Add each value to the row, separated by a comma
+    // The last value in the row should be the externalIDField value
+
+    Object.entries(payload).forEach(([key, value]) => {
+      if (!isSettingsKey.get(key)) {
+        row += `"${value}",`
```
mericsson, line 163: May need to additionally escape `,` values that may reside inside of `value`.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -68,6 +77,110 @@ export default class Salesforce {
...
+
+    return headers
+  }
+
+  private buildRow = (payload: GenericPayload): string => {
+    let row = ''
+
+    // Iterate over the keys in payload that map to a SF object field
+    // Add each value to the row, separated by a comma
+    // The last value in the row should be the externalIDField value
```
mericsson, line 159: Could you add more info here about why the last value should be `externalIDField`? Not clear to me.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -68,6 +77,110 @@ export default class Salesforce {
...
+
+    return row
+  }
+
+  private buildHeader = (payloads: GenericPayload[], externalIdFieldName: string): string => {
+    let header = ''
+    Object.keys(payloads[0]).forEach((key) => {
+      console.log('CSV: ', key)
+      if (!isSettingsKey.get(key)) {
+        header += `"${snakeCaseToCamelCase(key)}",`
```
mericsson, line 176: May need to escape or error if `key` contains `,` already.

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
mericsson: I suppose that's because it's last in the header? Curious if there is a reason it is last or first?

packages/destination-actions/src/destinations/salesforce/sf-properties.ts
```diff
@@ -10,7 +10,8 @@ export const operation: InputField = {
   choices: [
     { label: 'Create', value: 'create' },
     { label: 'Update', value: 'update' },
-    { label: 'Upsert', value: 'upsert' }
+    { label: 'Upsert', value: 'upsert' },
+    { label: 'Bulk Upsert', value: 'bulkUpsert' }
```
mericsson, line 14: This will be hidden in the UI for now? Is that correct?

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
nick-Ag: I could put the `externalIDField` anywhere in the header, I put it last since it makes handling the last comma super simple. If I know the last item is always externalIDField, I just write that at the end without a comma

nick-Ag: (added to in code comment)

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
nick-Ag: I've added this validation to `customFields` only. Since we control the `key` value for the other keys and don't include any invalid names

packages/destination-actions/src/destinations/salesforce/account/index.ts
nick-Ag: I think that would be possible if we somehow get an array to arrive in [this code block](https://github.com/segmentio/integrations/blob/master/createActionDestination/index.js#L210) for existing traffic. But if I understand correct that's only possible if users enable `enable_batching` (will be locked to internal for now). But I will test this in stage Monday (pushed that code to the Opprotunity action for testing)

nick-Ag:
Addressed comments and refactored to use a new `sf-utils.ts` file for CSV generation methods. Also supports `customFields` in the header now.
A couple TODOs:
* Support incomplete data for CSV generation. Right now if payload[0] has a name value and payload[1] doesn't have a name value the CSV will be incorrectly formatted.
* Test `performBatch` hijacking existing data meant to arrive in the `perform` method in stage.
* Escape double quotes inside CSV values

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
mericsson: 👍

packages/destination-actions/src/destinations/salesforce/account/index.ts
mericsson: I think as soon as `performBatch` is included as a function and pushed with the register update it will start trying to send batch events.

https://github.com/segmentio/action-destinations/blob/160cc4f31908ee08820cd8ea5d4e429e2dee7465/packages/cli-internal/src/commands/push.ts#L171-L184

May need to patch the CLI to consider an additional field name like `batch_enabled` on the action definition. What do you think @dlasky ?

packages/destination-actions/src/destinations/salesforce/sf-utils.ts
```diff
@@ -0,0 +1,86 @@
...
+        header += `"${customFieldName}",`
+      })
+    }
+  })
+  header += `"${externalIdFieldName}"`
+
+  return header
+}
+
+// Our key names are in snake case, but Salesforce's field names are in Pascal case.
```
mericsson, line 81: Sounds good. Just an fyi -- Salesforce API for field names is case insensitive. So you don't strictly need PascalCase. (But doesn't hurt -- it is just a convention).

mericsson: (e.g. You could just get away with `camelCase` only)

packages/destination-actions/src/destinations/salesforce/account/index.ts
dlasky: thats what the code linked above does, that adds a user controllable MAPPING for enable_batching which shows up as a boolean in the mapping config.

packages/destination-actions/src/destinations/salesforce/sf-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+  return rows
+}
+
+/**
+ *
+ * @param payloads Each payload in the batch.
+ * @param externalIdFieldName The name of the field that contains the external ID.
+ * @returns The complete CSV to send to Salesforce.
+ */
+export const buildCSVData = (payloads: GenericPayload[], externalIdFieldName: string): string => {
```
nick-Ag, line 129: The entry point for CSV generation. Supporting incomplete payload data ended up being complex, so please leave plenty of feedback on this

nick-Ag:
Hey team, this new build includes supporting incomplete Payload data. I've tested in stage and locally, plus added new unit tests and the functionality looks good. That being said this adds complexity to the CSV generation, so any feedback is greatly appreciated!

packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts
```diff
@@ -254,4 +255,130 @@ describe('Salesforce', () => {
...
+        name: 'Squidward Tentacles',
+        phone: '1234567891',
+        description: 'Krusty Krab',
+        customFields: {
+          TickerSymbol: 'KRAB'
+        }
+      }
+    ]
+
+    it('should correctly upsert a batch of records', async () => {
```
rhall-twilio, line 314: I would expect testcases to have at least one expect() statement to provide a passing validation criteria. 

I think right now we are assuming that sf.bulkUpsert not throwing means that the code did what was intended, but that might not be necessarily true. e.g. is there a way to confirm that the nock endpoints got hit? or does the test automatically fail if they don't?

packages/destination-actions/src/destinations/salesforce/account/index.ts
```diff
@@ -170,20 +173,31 @@ const action: ActionDefinition<Settings, Payload> = {
...
         throw new IntegrationError('Missing name value', 'Misconfigured required field', 400)
       }
-      return await sf.upsertRecord(payload, 'Account')
+      return await sf.upsertRecord(payload, OBJECT_NAME)
+    }
+  },
+  performBatch: async (request, { settings, payload }) => {
+    const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
+
+    if (payload[0].operation === 'bulkUpsert') {
```
rhall-twilio, line 195: What happens if the operation is not bulkUpsert and we hit performBatch?

packages/destination-actions/src/destinations/salesforce/sf-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+        if (headerMap.has(pascalKey)) {
+          headerMap.get(pascalKey)?.push([actualValue, i])
+        } else {
+          headerMap.set(pascalKey, [[actualValue, i]])
+        }
+      }
+      if (key === 'customFields') {
+        const customFields = value as object
+        Object.entries(customFields).forEach(([customFieldName, customFieldValue]) => {
+          const fieldErr = validateHeaderField(customFieldName)
```
rhall-twilio, line 47: Is there a reason the validation method doesn't throw the exception? It's an interesting pattern to rely on the client to throw the error, particularly if we are not aggregating them.

packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts
rhall-twilio: Maybe https://michaelheap.com/nock-all-mocks-used/

packages/destination-actions/src/destinations/salesforce/sf-utils.ts
mericsson: +1 I think should just `throw` from `validateHeaderField`. Would simplify this a little.

packages/destination-actions/src/destinations/salesforce/sf-utils.ts
mericsson: I suppose the problem is that inside of `performBatch` we just get a list of payloads and not any metadata about the mappings or what the full event could be? Ideally we would just get the mappings that the customer configured in the UI and use that as the header, right? If so, could be a follow-up to provide that in some way.

packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts
nick-Ag: Enabled this and the case when a nock interceptor is not triggered throws a much more readable error than before. Thanks for the suggestion!

packages/destination-actions/src/destinations/salesforce/account/index.ts
mericsson: I think may be a good idea to add an so it doesn't "silently fail" just in case someone misconfigures it.

```js
else {
//  throw IntegrationError ...
}
```

packages/destination-actions/src/destinations/salesforce/opportunity/index.ts
```diff
@@ -45,20 +48,34 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      if (!payload[0].close_date || !payload[0].name || !payload[0].stage_name) {
+        throw new IntegrationError('Missing close_date, name or stage_name value', 'Misconfigured required field', 400)
+      }
+
+      return await sf.bulkUpsert(payload, OBJECT_NAME)
+    } else {
+      const errorMsg = 'Bulk Upsert action must be used with batching'
+      throw new IntegrationError(errorMsg, errorMsg, 400)
```
mericsson, line 78: Not blocking but I guess the second argument should is an optional code.

```suggestion
      const errorMsg = 'Bulk Upsert action must be used with batching'
      throw new IntegrationError(errorMsg, 'Bad Request', 400)
```

https://github.com/segmentio/action-destinations/blob/4a53c8bbb3a07d4b4aa0b391d377ef1ccd58ca3f/packages/destination-actions/src/destinations/slack/postToChannel/index.ts#L50

mericsson: (As a larger follow-up we should make `IntegrationError` easier to get right. I see a lot of different usage of it across action-destinations)

dlasky (approved):

packages/destination-actions/src/destinations/salesforce/sf-utils.ts
nick-Ag: The header ordering doesn't matter no. I'll go ahead and address this in a follow-up :+1:

PR #595, opened 2022-05-19
Stratconn 1260 - New Destination Action for Google Sheets

rhall-twilio:
This pull request covers a new DestinationAction for Google Sheets API. 

## Testing

- [x] Testing Completed in Staging
image

Test event:
```
		{
			"type": "track",
			"event": "new",
			"receivedAt": "2022-04-12T23:17:41.192501242Z",
			"channel": "server",
			"properties": {
				"CLOSE_DATE": "2022-07-08T00:00:00Z",
				"CLOSE_DATE_EOQ": "2022-07-08",
				"ENTRY_POINT": "Website Demo Request",
				"E_ARR_POST_LAUNCH_C": "100001.0",
				"FINANCE_ENTRY_POINT": "Inbound High STAGED"
			},
			"__segment_id": "0063q0000126KchAAE-Robert12"
		},
```

## Post-Merge

After code is merged to main, we will run `./bin/run push` on each environment to make sure that the settings are correctly propagated. After that, I will complete a manual test in production by going through the Google Sheets oauth flow and delivering a test event to my Google Sheets instance.


packages/core/src/destination-kit/types.ts
```diff
@@ -23,6 +24,8 @@ export interface ExecuteInput<Settings, Payload> {
   page?: string
   /** The data needed in OAuth requests */
   readonly auth?: AuthTokens
+  /** The input data in it's raw form without any transformations */
```
mericsson, line 27:
```suggestion
  /** The input data in its raw form without any transformations */
```

packages/destination-actions/src/destinations/google-sheets/googleapis/index.ts
```diff
@@ -0,0 +1,45 @@
...
+export default class GoogleSheets {
+  request: RequestClient
+
+  constructor(request: RequestClient) {
+    this.request = request
+  }
+
+  get = async (mappingSettings: MappingSettings) => {
+    return this.request(
+      `https://sheets.googleapis.com/${API_VERSION}/spreadsheets/${mappingSettings.spreadsheetId}/values/${mappingSettings.spreadsheetName}!A:A`,
```
rhall-twilio, line 14: `A:A` should be parameter

packages/destination-actions/src/destinations/google-sheets/googleapis/index.ts
```diff
@@ -0,0 +1,45 @@
...
+          valueInputOption: mappingSettings.dataFormat,
+          data: batchPayload
+        }
+      }
+    )
+  }
+
+  append = async (mappingSettings: MappingSettings, values: string[][]) => {
+    return this.request(
+      `https://sheets.googleapis.com/${API_VERSION}/spreadsheets/${mappingSettings.spreadsheetId}/values/${mappingSettings.spreadsheetName}!A2:append?valueInputOption=${mappingSettings.dataFormat}`,
```
rhall-twilio, line 36: `A2` should also be a parameter

packages/core/src/destination-kit/types.ts
```diff
@@ -23,6 +24,8 @@ export interface ExecuteInput<Settings, Payload> {
...
+  readonly rawData?: InputData | InputData[]
```
dlasky, line 28: I'm strongly opposed to allowing non mapped payloads. Once we allow this for partners we can never take it back, and it will generally encourage devs to ignore mappings in favor of non-transparent logic which is against some of the core principles of actions. Happy to discuss, but hopefully that makes sense.

packages/destination-actions/src/destinations/google-sheets/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const CONSTANTS = {
+  /**
+   * Based on benchmarks found while testing so we dont reach the timeout limit.
+   */
+  MAX_CELLS: 300000
```
marinhero, line 5: This is only used one time inside one file. What do you think about getting rid of the extra file and just leaving the constant close to the top?

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
```diff
@@ -0,0 +1,79 @@
+import { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+import { processData } from './operations'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Post Sheet',
+  description: 'Write values to a Google Sheets spreadsheet.',
+  defaultSubscription: 'type = "track" and event = "updated" or event = "new" or event = "deleted"',
```
brennan, line 10: I would caution against a default description with specific track event names—this may result in events flowing to Google Sheets that a customer does not intend.

packages/core/src/destination-kit/action.ts
```diff
@@ -126,8 +126,6 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
 
     // Construct the data bundle to send to an action
     const dataBundle = {
-      rawData: bundle.data,
```
marinhero, line 129: Is this strictly necessary? What benefits will we be seeing from it? will it make debugging harder?

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      description:
+        'The name of the spreadsheet. You can find this value on the tab at the bottom of the spreadsheet. Please provide a valid name of a sheet that already exists.',
+      type: 'string',
+      required: true,
+      default: 'Sheet1'
+    },
+    data_format: {
+      label: 'Data Format',
+      description:
+        'The way Google will interpret values. If you select raw, values will not be parsed and will be stored as-is. If you select user entered, values will be parsed as if you typed them into the UI. Numbers will stay as numbers, but strings may be converted to numbers, dates, etc. following the same rules that are applied when entering text into a cell via the Google Sheets UI.',
```
brennan, line 47: Does Google have documentation describing `strings may be converted to numbers, dates, etc. following the same rules that are applied when entering text into a cell via the Google Sheets UI.`? If so, it may be nice to link to those here.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * Utility function that converts the event properties into an array of strings that Google Sheets API can understand.
+ * Note that the identifier is forced as the first column. 
+ * @param identifier value used to imbue fields with a uniqueness constraint
+ * @param fields list of properties contained in the event
+ * @param columns list of properties that will be committed to the spreadsheet
+ * @returns a string object that has used the `fields` data to populate the `columns` ordering
+ * 
+ * @example
```
brennan, line 33: I love this example. Thank you, @rhall-twilio!

packages/destination-actions/src/destinations/google-sheets/generated-types.ts
```diff
@@ -0,0 +1,3 @@
+// Generated file. DO NOT MODIFY IT BY HAND.
+
+export interface Settings {}
```
marinhero, line 3: By reading the other actions files I can see that there are no settings but just want it to run it by you since everything seems to be operating according to mappingSettings. Right?

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
```diff
@@ -0,0 +1,79 @@
...
+  description: 'Write values to a Google Sheets spreadsheet.',
+  defaultSubscription: 'type = "track" and event = "updated" or event = "new" or event = "deleted"',
+  // TODO: Hide record_identifier and operation_type
+  fields: {
+    record_identifier: {
+      label: 'Record Identifier',
+      description: 'Property which uniquely identifies each row in the spreadsheet.',
+      type: 'string',
+      required: true,
+      default: { '@path': '$.__segment_id' }
```
brennan, line 18: It's not clear to me where `__segment_id` is assigned and/or used. Is this something we'll need in a future iteration of the integration?

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * Commits all passed events to the correct row in the spreadsheet, as well as the columns header row.
+ * @param mappingSettings configuration object detailing parameters for the call
+ * @param updateBatch array of events to commit to the spreadsheet
+ * @param gs interface object capable of interacting with Google Sheets API
+ */
+async function processUpdateBatch(
+  mappingSettings: MappingSettings,
+  updateBatch: { identifier: string; event: { [k: string]: string }; targetIndex: number }[],
```
marinhero, line 113: Consider making a type for this instead as it's a bit hard to read.

```ts
type processUpdateBatchInputParams = {
...
}
```

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * Takes an array of events and dynamically decides whether to append, update or delete rows from the spreadsheet.
+ * @param request request object used to perform HTTP calls
+ * @param events array of events to commit to the spreadsheet
+ */
+async function processData(request: RequestClient, events: Payload[]) {
+  // These are assumed to be constant across all events
+  const mappingSettings = {
+    spreadsheetId: events[0].spreadsheet_id,
```
brennan, line 207: I'd love to learn more about how we landed on this approach. This suggests that the pipeline is batching events based on a composite of event attributes `spreadsheetId`, `spreadsheetName`, and `dataFormat`, hence we would be able to simply look at the first event in the batch to infer these values for all subsequent events in the same batch.

As far as I understand, however, events are batched based on a `BatchKey`, which is a combination of the job's `Flow` attribute and the event's `Settings`: https://github.com/segmentio/integrations-consumer/blob/154e0481f7b4d4eadb432fd1fa75a36315ecaf8c/action_batch.go#L52-L53.

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      ]
+    },
+    fields: {
+      label: 'Fields',
+      description: `
+      The fields to write to the spreadsheet. 
+      On the left-hand side, input the name of the field as it will appear in the Google Sheet. 
+      On the right-hand side, select the field from your data model that maps to the given field in your sheet.
+      
+      ---
```
marinhero, line 63: Do we need this?

packages/destination-actions/src/destinations/index.ts
```diff
@@ -36,6 +36,7 @@ register('61dde0dc77eb0db0392649d3', './friendbuy')
 register('61f83101210c42a28a88d240', './gainsight-px-cloud-action')
 register('60ad61f9ff47a16b8fb7b5d9', './google-analytics-4')
 register('60ae8b97dcb6cc52d5d0d5ab', './google-enhanced-conversions')
+register('627ea052118e3cd530d28963', './google-sheets')
```
brennan, line 39: Should we sync the destination definition from stage to production now? Instruction are in step 2 under [this heading in Paper](https://paper.dropbox.com/doc/Destination-Actions-Interface--Bi97OHqmnYKZpOB4B3RWwQmkAg-BS1RKGj9VKTcjqLrCSuBV#:uid=217740603690332222029229&h2=How-to-deploy-new-Cloud-Destin).

brennan (changes requested):
Thank you for all your work here, @rhall-twilio! I left a few comments/questions. One thing I would like to chat about soon is the best approach to `mappingSettings`.

packages/destination-actions/src/destinations/google-sheets/constants.ts
rhall-twilio: We originally had it as a constant on the top but ran into issues mocking it with jest. Exporting it in a different file is the only workaround we've found so far.

packages/core/src/destination-kit/action.ts
rhall-twilio: Yes. Context for this change stems from this thread: https://github.com/segmentio/action-destinations/pull/595#discussion_r877470596

If we don't want to allow any action-destinations to take a dependency on it, then it should not be passed down to the aciton-destinations. I like the idea of doing it in a separate PR.

yarn.lock
```diff
@@ -2290,6 +2290,11 @@
   resolved "https://registry.yarnpkg.com/@polka/url/-/url-1.0.0-next.21.tgz#5de5a2385a35309427f6011992b544514d559aa1"
   integrity sha512-a5Sab1C4/icpTZVzZc5Ghpz88yQtGOyNqYXcZgOssB2uuAr+wF/MvN6bgtW32q7HHrvBki+BsZ0OuNv6EV3K9g==
 
+"@segment/a1-notation@^2.1.4":
```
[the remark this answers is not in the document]
rhall-twilio: This is a new package.

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
rhall-twilio: It's how RETL sources uniquely identify an event. We plan to make this field hidden and lock Google Sheets to only support RETL sources only in GA.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
rhall-twilio: The `spreadsheetId`, `spreadsheetName` and `dataFormat` are not part of the event payload. They are constants and are supplied by the user while setting up the action destination. In clean world, they would be part of the settings instead of the mappings, but because they may change per action instance, we opted to put them as mappings instead.

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
rhall-twilio: Yes. This is how customer can specify which properties of the payload will be written to the Google Sheet, as well as what name they will appear as.

packages/destination-actions/src/destinations/google-sheets/constants.ts
marinhero: Got it 👍

yarn.lock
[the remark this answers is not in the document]
marinhero: ooops! My bad

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
brennan: Makes sense - thanks for the clarification!

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
rhall-twilio: Should we just not have a defaultSubscription at all?

packages/destination-actions/src/destinations/index.ts
rhall-twilio: Isn't it from production to staging? @maryamsharif may have completed this step already.

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
rhall-twilio: I was not able to find additional extensive documentation but added some examples I found.

packages/destination-actions/src/destinations/index.ts
brennan: Ah, you are right - we do sync from production to stage. But I do want to call out that Google Sheets in prod is out of sync with stage at the moment. The settings in stage AFAICT are the ones we eventually want in production. Here's what I see in prod currently:

Screen Shot 2022-06-06 at 3 50 45 PM

rhall-twilio: This one is resolved by `./bin/run push` right?

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
rhall-twilio: I added some types, that are not necessarily easier to read. :) Let me know if it's too contrived

packages/core/src/middleware/after-response/prepare-response.ts
```diff
@@ -4,10 +4,17 @@ import type { ModifiedResponse } from '../../types'
 const prepareResponse: AfterResponseHook = async (_request, _options, response) => {
   const modifiedResponse = response as ModifiedResponse
 
-  // Clone the response before reading the body to avoid
-  // `TypeError: body used already` elsewhere
-  const clone = response.clone()
-  const content = await clone.text()
+  let content: string
+  if (_options.skipResponseCloning) {
```
mericsson, line 8: I think convention is if we are using a variable then it should no longer be `_options`. This convention comes from golang. https://stackoverflow.com/questions/27764421/what-is-underscore-comma-in-a-go-declaration

So you should rename `_options` to `options`.

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
```diff
@@ -0,0 +1,78 @@
...
+      label: 'Record Identifier',
+      description: 'Property which uniquely identifies each row in the spreadsheet.',
+      type: 'string',
+      required: true,
+      default: { '@path': '$.__segment_id' }
+    },
+    operation_type: {
+      label: 'Operation Type',
+      description:
+        "Describes the nature of the operation being performed. Only supported values are 'new', 'updated', and 'deleted'.",
```
mericsson, line 22: I guess we need some kind of a 'picklist' type to support this better.

packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
```diff
@@ -0,0 +1,78 @@
...
+        { label: 'Raw', value: 'RAW' },
+        { label: 'User Entered', value: 'USER_ENTERED' }
+      ]
+    },
+    fields: {
+      label: 'Fields',
+      description: `
+      The fields to write to the spreadsheet. 
+      On the left-hand side, input the name of the field as it will appear in the Google Sheet. 
+      On the right-hand side, select the field from your data model that maps to the given field in your sheet.
```
mericsson, line 60: This looks good. Left vs Right consistent with Salesforce https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/salesforce/sf-properties.ts#L62-L64

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  // Use the retrieved row identifiers along with the incoming events to decide which ones should be appended or updated.
+  const { appendBatch, updateBatch } = processGetSpreadsheetResponse(response.data, events, mappingSettings)
+
+  const promises = [
+    processUpdateBatch(mappingSettings, updateBatch, gs),
+    processAppendBatch(mappingSettings, appendBatch, gs)
+  ]
+
+  return await Promise.all(promises)
```
mericsson, line 229: Confirming it's OK for these promises to be processed in parallel?

brennan (approved):

brennan (approved):

packages/core/src/middleware/after-response/prepare-response.ts
mericsson: Is `_options` guaranteed to be defined? Wonder if should do something like:

```suggestion
  if (_options && _options.skipResponseCloning) {
```

packages/destination-actions/package.json
```diff
@@ -59,6 +60,10 @@
     },
     "setupFilesAfterEnv": [
       "<rootDir>/test/setup-after-env.ts"
+    ],
+    "coverageReporters": [
```
mericsson, line 64: What will this do?

rhall-twilio: This specifies how JEST should collect code coverage. 

The `.html` reporter creates a rendered page that shows code coverage details for user-readable consumption:
image

The `lcov` reporter is the standard VSCode extensions use to show coverage (see left green/yellow markers in code)
image

packages/core/src/middleware/after-response/prepare-response.ts
rhall-twilio: Does not seem that it can be undefined as per Typescript conventions:
image

rhall-twilio: ^ Manually changing the type to `NormalizedOptions | undefined` results in this linting rule so this check is automated by the framework already 
image

packages/destination-actions/package.json
mericsson: Cool, maybe worthwhile to have a README update as a follow-up or in this PR about how to use / access code coverage?

mericsson (approved):

packages/destination-actions/package.json
rhall-twilio: Added a note to myself to update README by end of week.

PR #597, opened 2022-05-20
Upgrade to Web SDK 4.0

spvismaya:
Adding support for braze SDK `4.0`. It should also be backwards compatible with `3.x` versions. A few methods, namespace, etc., have changed from version `3 -> 4`, and I am unsure of a few things in this PR. Please let me know if something does not look right.

## Testing


packages/browser-destinations/src/destinations/braze/__tests__/trackPurchase.test.ts
```diff
@@ -1,20 +1,22 @@
...
-  jest.spyOn(appboy, 'initialize').mockImplementation(() => true)
-  jest.spyOn(appboy, 'openSession').mockImplementation(() => true)
+  // we're not really testing that braze loads here, so we'll just mock it out
+  jest.spyOn(spyMethods, 'initialize').mockImplementation(() => true)
+  jest.spyOn(spyMethods, 'openSession').mockImplementation(() => true)
 })
 
 test('reports products when present', async () => {
-  const brazeLogPurchase = jest.spyOn(appboy, 'logPurchase').mockReturnValue(true)
+  const brazeLogPurchase = jest.spyOn(spyMethods, 'logPurchase').mockReturnValue(true)
```
davidbielik, line 14: would these module methods be called for v3.x? Wouldn't these only expect to be called with 4.x?

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -43,26 +46,26 @@ const presets: DestinationDefinition['presets'] = [
...
 
-export const destination: BrowserDestinationDefinition<Settings, typeof appboy> = {
+export const destination: BrowserDestinationDefinition<Settings, BrazeType> = {
   name: 'Braze Web Mode (Actions)',
   slug: 'actions-braze-web',
   mode: 'device',
   settings: {
     sdkVersion: {
-      description: 'The version of the SDK to use. Defaults to 3.3.',
+      description: 'The version of the SDK to use. Defaults to 4.0.',
```
davidbielik, line 55:
```suggestion
      description: 'The version of the SDK to use.',
```

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -43,26 +46,26 @@ const presets: DestinationDefinition['presets'] = [
...
       label: 'SDK Version',
       type: 'string',
       choices: [
-        {
```
davidbielik, line 56: i don't think we should remove 3.3 for customers using it

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -265,28 +268,38 @@ export const destination: BrowserDestinationDefinition<Settings, typeof appboy>
         subscriptions,
         ...expectedConfig
       } = settings
-      const version = sdkVersion ?? '3.5'
+      const version = sdkVersion ?? '4.0'
 
       resetUserCache()
 
-      await dependencies.loadScript(`https://js.appboycdn.com/web-sdk/${version}/appboy.no-amd.min.js`)
+      if (version.startsWith('3')) {
```
davidbielik, line 275:
```suggestion
      if (version.startsWith('3.')) {
```

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -52,10 +52,14 @@ export const destination: BrowserDestinationDefinition<Settings, BrazeType> = {
...
   settings: {
     sdkVersion: {
-      description: 'The version of the SDK to use. Defaults to 4.0.',
+      description: 'The version of the SDK to use.',
       label: 'SDK Version',
       type: 'string',
       choices: [
+        {
+          value: '3.1',
+          label: '3.5'
```
davidbielik, line 61:
```suggestion
          label: '3.1'
```

packages/browser-destinations/src/destinations/braze/__tests__/trackPurchase.test.ts
marinhero: You are setting the sdkVersion [here](https://github.com/segmentio/action-destinations/pull/597/files#diff-031587f0ade4d3e27d49e2417ab600362c80c4c821b14ddaff4db466bf68c1e0R19) to be 3.5. If you want to use 4.0 you should call the `brazeDestination` method with `4.0` as parameter instead.

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -272,13 +276,13 @@ export const destination: BrowserDestinationDefinition<Settings, BrazeType> = {
 
       resetUserCache()
 
-      if (version.startsWith('3')) {
+      if (version.startsWith('3.')) {
```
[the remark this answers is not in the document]
marinhero: Here's the information regarding Browser Support. Let us know if you have any questions https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/supported-browsers/

spvismaya:
@marinhero hello! we have no more concerns here and if everything looks good could you please help us get this through?

packages/browser-destinations/src/destinations/braze/debounce/index.ts
```diff
@@ -55,7 +55,7 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
     const shouldSend = shouldSendToBraze(event)
     ctx.updateEvent('integrations.Braze Web Mode (Actions)', shouldSend)
     ctx.updateEvent('integrations.Braze Cloud Mode (Actions)', shouldSend)
-    ctx.updateEvent('integrations.Appboy', shouldSend)
+    ctx.updateEvent('integrations.Braze', shouldSend)
```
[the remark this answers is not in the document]
davidbielik: @pooyaj is it possible to support changing this? what else would we need to change?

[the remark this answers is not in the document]
pooyaj: @davidbielik We call this the "creation name" of a destination, and it is not possible to change it. The reason this is not possible to change is that there might be customers relying on the "Appboy" name when instrumenting their web applications. If we rename from "Appboy" to "Braze" their instrumentation is going to break ( unless they make a change from Appboy --> Braze in their code ).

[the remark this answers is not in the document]
davidbielik: Understood thanks! Anything else we need to do to merge this pr?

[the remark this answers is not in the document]
davidbielik:
```suggestion
    ctx.updateEvent('integrations.Appboy', shouldSend)
```

davidbielik:
@pooyaj I reverted that namespace change. Let us know if anything else is needed to merge. Thanks!

marinhero (commented):
This looks good to me. Thanks for adding those extra validations. I'm going to let @pooyaj give the approval since this falls more within his wheelhouse. Nice work!

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
```diff
@@ -155,22 +156,21 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
 
     const user = client.getUser()
```
chrisradek, line 157: As far as I can tell, only `user` APIs are invoked after this line based on some conditions. If that's the case, I think you can do a fast exit `if (!user) return`, then you don't need to add the optional chaining operator wherever you use `user`.

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
```diff
@@ -155,22 +156,21 @@ const action: BrowserActionDefinition<Settings, typeof appboy, Payload> = {
...
 
-    payload.image_url !== undefined && user.setAvatarImageUrl(payload.image_url)
```
chrisradek, line 158: Did you mean to remove `setAvatarImageUrl`? I see `image_url` is still one of the action fields.

packages/browser-destinations/src/destinations/braze/__tests__/trackPurchase.test.ts
chrisradek, line 14: It looks like this spy isn't working properly - the action won't use the functions contained in the `spyMethods` object.

I was able to get the spy to work if I updated the `sdkVersion` to `4.1`, and then added this line right after `await trackPurchase(...)`

`const brazeLogPurchase = jest.spyOn(window.braze, 'logPurchase').mockReturnValue(true)`

It looks like this test is passing but not testing what we'd expect - I think `purchaseProperties` should just be rolled up into `properties`, then you'd see `brazeLogPurchase` called.

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
davidbielik: Hi @chrisradek yes we've removed the functionality in our dashboard so no need for this method in any SDK version. Where should we remove the `image_url` field?

chrisradek: You can remove it from here: https://github.com/segmentio/action-destinations/blob/9324bd47fbeee4f052f7abec20a361d65bb4c87a/packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts#L119-L127

However that does mean if a user passes in `avatar` as a trait, it will end up in `custom_attributes`. If you'd rather it be ignored completely then you don't need to make any changes.

chrisradek (approved):
The changes look good from my side!

pooyaj (approved):

davidbielik:
Hi @pooyaj @chrisradek thanks for reviewing. How long until this change is visible in the segment dashboard?

PR #599, opened 2022-05-20
Stratconn 1380: Support delete operation

rhall-twilio:
This change adds support for the 'deleted' event type in Google Sheets destination action

## Testing

Used insomnia to manually test. Full test suite will be included as part of STRATCONN-1260 merge into `main`

Manual cURL command:

```
curl --request POST \
  --url http://localhost:3000/postSheet \
  --header 'Content-Type: application/json' \
  --data '{
	"payload": [
		{
			"type": "track",
			"event": "new",
			"receivedAt": "2022-04-12T23:17:41.192501242Z",
			"channel": "server",
			"properties": {
				"CLOSE_DATE": "2022-07-08T00:00:00Z",
				"CLOSE_DATE_EOQ": "2022-07-08",
				"ENTRY_POINT": "Website Demo Request",
				"E_ARR_POST_LAUNCH_C": "100001.0",
				"FINANCE_ENTRY_POINT": "Inbound High CHANGE"
			},
			"__segment_id": "0063q0000126KchAAE-Robert209"
		},
		{
			"type": "track",
			"event": "deleted",
			"receivedAt": "2022-04-12T23:17:41.192501242Z",
			"channel": "server",
			"properties": {
				"CLOSE_DATE": "2022-07-10T00:00:00Z",
				"CLOSE_DATE_EOQ": "2022-07-10",
				"ENTRY_POINT": "Website Demo Response",
				"E_ARR_POST_LAUNCH_C": "100000.0",
				"FINANCE_ENTRY_POINT": "Inbound Low Change"
			},
			"__segment_id": "0063q0000126KchAAE-Robert204"
		}
	],
	"mapping": {
		"record_identifier": {
			"@path": "$.__segment_id"
		},
		"operation_type": {
			"@path": "$.event"
		},
		"spreadsheet_id": "1ORcFZ73VJXzj7rruKTrbUCgtRjqvKS4qW1uwDGP8tiY",
		"spreadsheet_name": "Sheet1",
		"data_format": "RAW",
		"fields": {
			"@path": "$.properties"
		}
	},
	"settings": {
	},
	"auth": {
		"clientId": "",
		"clientSecret": "",
		"accessToken": "ADD ME",
		"refreshToken": ""
	}
}'
```

### DIFF
--- a/packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
+++ b/packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
@@ -1,6 +1,6 @@
 import type { Payload } from './generated-types'
 import { RequestClient } from '@segment/actions-core'
-import GoogleSheets from '../googleapis/index'
+import { GoogleSheets, GetResponse } from '../googleapis/index'
 
 // TODO (STRATCONN-1379): Fork @flighter/a1-notation into @segment/a1-notation and add test cases
 import A1 from '@flighter/a1-notation'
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -120,7 +134,8 @@ function processUpdateBatch(
     values: [['id', ...mappingSettings.columns]]
   })
 
-  gs.batchUpdate(mappingSettings, batchPayload)
+  return gs
+    .batchUpdate(mappingSettings, batchPayload)
```
mericsson, line 138: As a followup (in this PR or another) can we move from `then` to `await` to be consistent in Promise usage?

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -179,12 +227,16 @@ function processData(request: RequestClient, events: Payload[]) {
...
+
+      const promises = [
+        processUpdateBatch(mappingSettings, updateBatch, gs),
+        processDeleteBatch(mappingSettings, deleteBatch, gs),
+        processAppendBatch(mappingSettings, appendBatch, gs)
+      ]
 
-      processUpdateBatch(mappingSettings, updateBatch, gs)
-      processAppendBatch(mappingSettings, appendBatch, gs)
+      return Promise.all(promises)
```
mericsson, line 239: Confirming it's OK for the promises to be evaluated concurrently?

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
rhall-twilio: Do we want to be sequential? I'm thinking running the update, append and delete in parallel will give a better chance of finishing under 5s

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
rhall-twilio: It _should_ be. The assumption is that we will get only one event with a given __segment_id per batch. 

Delete was implemented as clearing the row, rather than removing the row to allow for concurrency (otherwise the row that Update would change could be incorrect)

I'll add a note to do some load testing to make sure Google Sheets API can handle it on their side.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
mericsson: Have no preference on sequential in this case. I guess it is a moot point since this particular `.then` will go away when you remove the `console.log`.

If you wanted to, you can rewrite this to:
```js
await gs.batchUpdate(mappingSettings, batchPayload)
console.log('update')
```

rhall-twilio: ahhh, I see what you mean now. makes sense.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
rhall-twilio: aaaah, I need to go back to the drawing board. As it turns out, Append finds the first empty row and then starts to overwrite data from there which is not what I wanted:

image

image

May have to make delete actually remove the row and make all 3 operations sequential 😞

rhall-twilio: I started a section on the SDD for this: https://paper.dropbox.com/doc/SDD-Google-Sheets-Destination--Bh5f__Vl7GoUbLD3MivXrjAhAg-G2N5mhEIyrDMhqclqDwDB#:uid=093625986327327994378849&h2=Operation-Concurrency but things look a bit a grim at the moment.

rhall-twilio:
@mericsson We're going to proceed with the delete operation disabled for initial push. We will likely have to introduce locking to support it in order to deliver the best experience to the customer.

PR #603, opened 2022-05-24
PEREX-1329 fix: update SMS action to use api key

cfree:
Update personas sms action to use correct setting names and values

https://segment.atlassian.net/browse/PEREX-1329

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/personas-messaging-twilio/index.ts
+++ b/packages/destination-actions/src/destinations/personas-messaging-twilio/index.ts
@@ -84,15 +84,21 @@ const destination: DestinationDefinition<Settings> = {
   authentication: {
     scheme: 'custom',
     fields: {
-      twilioAccountId: {
-        label: 'Twilio Account ID',
-        description: 'Twilio Account ID',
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/personas-messaging-twilio/index.ts
```diff
@@ -84,15 +84,21 @@ const destination: DestinationDefinition<Settings> = {
   authentication: {
     scheme: 'custom',
     fields: {
-      twilioAccountId: {
-        label: 'Twilio Account ID',
-        description: 'Twilio Account ID',
+      twilioAccountSID: {
```
mericsson, line 87: I think renaming fields can be problematic if already deployed for existing customers. But my understanding is we do not have existing customers here?

cfree: Agree 😅  We have one or two customers, but with this change they'll have to update their credentials anyway which should trigger an actions' settings update. Breaking changes have been communicated.

mericsson (approved):

PR #605, opened 2022-05-24
Fix execution halting for Google Sheets when processing a large Spreadsheet

rhall-twilio:
Stop cloning the response object for Google Sheets to unblock STRATCONN-1260 release. I've filed tech debt ticket https://segment.atlassian.net/browse/STRATCONN-1396 to address underlying issue for all other action-destinations.

## Testing

Full test suite pending as part of STRATCONN-1260 merge into main.

### DIFF
--- a/packages/core/src/middleware/after-response/prepare-response.ts
+++ b/packages/core/src/middleware/after-response/prepare-response.ts
@@ -4,10 +4,17 @@ import type { ModifiedResponse } from '../../types'
 const prepareResponse: AfterResponseHook = async (_request, _options, response) => {
   const modifiedResponse = response as ModifiedResponse
 
-  // Clone the response before reading the body to avoid
-  // `TypeError: body used already` elsewhere
-  const clone = response.clone()
-  const content = await clone.text()
+  let content: string
+  if (_options.skipResponseCloning) {
+    // Skip cloning the response to avoid a Node crash in case the response payload is larger than 16KB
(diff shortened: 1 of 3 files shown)


dlasky (approved):

packages/core/src/middleware/after-response/prepare-response.ts
```diff
@@ -4,10 +4,15 @@ import type { ModifiedResponse } from '../../types'
 const prepareResponse: AfterResponseHook = async (_request, _options, response) => {
   const modifiedResponse = response as ModifiedResponse
 
-  // Clone the response before reading the body to avoid
-  // `TypeError: body used already` elsewhere
-  const clone = response.clone()
-  const content = await clone.text()
+  let content: string
+  if (_options.skipResponseCloning) {
```
marinhero, line 8: Should we add another comment here to explain why you can choose to skip the cloning?

rhall-twilio: I think skipping cloning should be the default unless something really needs it. I've added comments to indicate so.

marinhero (approved):
Thanks for addressing my comment. Awesome find. :shipit:

marinhero (changes requested):
__Looks like we need to fix some tests. Please address the issues and let me know when I can take another look.__

marinhero (approved):
I just noticed you are not merging into master. Feel free to 🚢

PR #611, opened 2022-05-26
STRATCONN-1375: Check for cells reaching max limit

maryamsharif:
This PR adds a check to make sure the number of cells in a sheet does not exceed 300,000. 

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/google-sheets/__tests__/postSheet.operations.test.ts
+++ b/packages/destination-actions/src/destinations/google-sheets/__tests__/postSheet.operations.test.ts
@@ -4,6 +4,12 @@ import { Payload } from '../postSheet/generated-types'
 import PostSheet from '../postSheet/index'
 import { GoogleSheets, GetResponse } from '../googleapis/index'
 
+jest.mock('../constants', () => ({
+  CONSTANTS: {
+    MAX_CELLS: 1
+  }
+}))
+
 const mockGoogleSheets = {
   get: jest.fn(),
   batchUpdate: jest.fn(),
@@ -77,8 +83,6 @@ describe('Google Sheets', () => {
     })
 
     it('should fail because number of cells limit is reached', async () => {
-      process.env.GOOGLE_SHEETS_MAX_CELLS = '1'
-
       // Make sure the spreadsheet contains the event from the payload
       const getResponse: Partial<GetResponse> = {
         values: [['id'], ['1234'], ['12345']]
@@ -88,7 +92,6 @@ describe('Google Sheets', () => {
         data: getResponse
       })
 
-      expect(GoogleSheets).toHaveBeenCalled()
       await expect(PostSheet.performBatch?.(jest.fn(), data as ExecuteInput<Settings, Payload[]>)).rejects.toThrowError(
         'Sheet has reached maximum limit'
       )
--- a/packages/destination-actions/src/destinations/google-sheets/__tests__/postSheet.operations.test.ts
+++ b/packages/destination-actions/src/destinations/google-sheets/__tests__/postSheet.operations.test.ts
@@ -75,5 +75,23 @@ describe('Google Sheets', () => {
       expect(mockGoogleSheets.append).not.toHaveBeenCalled()
       expect(mockGoogleSheets.batchUpdate).toHaveBeenCalled()
     })
+
+    it('should fail because number of cells limit is reached', async () => {
+      process.env.GOOGLE_SHEETS_MAX_CELLS = '1'
+
+      // Make sure the spreadsheet contains the event from the payload
+      const getResponse: Partial<GetResponse> = {
+        values: [['id'], ['1234'], ['12345']]
+      }
+
+      mockGoogleSheets.get.mockResolvedValue({
+        data: getResponse
+      })
+
+      expect(GoogleSheets).toHaveBeenCalled()
+      await expect(PostSheet.performBatch?.(jest.fn(), data as ExecuteInput<Settings, Payload[]>)).rejects.toThrowError(
+        'Sheet has reached maximum limit'
+      )
+    })
   })
 })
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -2292,7 +2292,7 @@
 
 "@segment/a1-notation@^2.1.4":
   version "2.1.4"
-  resolved "https://registry.yarnpkg.com/@segment/a1-notation/-/a1-notation-2.1.4.tgz#35a48a0688019c3ffff23b1ba890e864c891a11f"
+  resolved "https://registry.npmjs.org/@segment/a1-notation/-/a1-notation-2.1.4.tgz#35a48a0688019c3ffff23b1ba890e864c891a11f"
```
brennan, line 2295: I don't think the `yarn.lock` file should be updated in this PR. I'd recommend removing these changes from the PR.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -55,8 +57,13 @@ const generateColumnValuesFromFields = (identifier: string, fields: Fields, colu
  * @param events data to be written to the spreadsheet
  * @returns
  */
-function processGetSpreadsheetResponse(response: GetResponse, events: Payload[]) {
-  // TODO (STRATCONN-1375): Fail request if above row limit
+function processGetSpreadsheetResponse(response: GetResponse, events: Payload[], mappingSettings: MappingSettings) {
```
brennan, line 60: The logic here makes sense - but I wonder if it's possible or desirable to run this check before we deliver a batch? By checking the number of columns after a attempting to deliver a batch, I can see a situation where a partial batch is appended to the spreadsheet, and any items that would result in the spreadsheet exceeding 300,000 rows would fail.

I'll defer to @rhall-twilio and @kdaswani on this one.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -55,8 +57,13 @@ const generateColumnValuesFromFields = (identifier: string, fields: Fields, colu
...
+  const numColumns = mappingSettings.columns.length
+  const numRows = response.values.length
```
brennan, line 62: Could `response.values` ever be undefined? If so, we may want `response.values?.length`.

marinhero:
Make sure you remove all the template from the PR description:

> A summary of your pull request, including the what change you're making and why.

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -8,6 +8,8 @@ type Fields = {
   [k: string]: string
 }
 
+const MAXCELLS = 300000
```
marinhero, line 11: Why is 300000 the value? What do you think about adding a comment that tells us why this value was chosen?

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
rhall-twilio: It gets more complicated if some of the events in the batch are of the `delete` type but we don't currently support those. 

The `update` ones should not affect it so numItemsInCurrentBatch should be limited to `new` events only.  I'm not too worried about this given that a batch will have at maximum 1000 events which is orders of magnitude lower than 300,000 which is an approximation anyways, so might not be worth the extra complexity in the code.

kdaswani: I agree! @rhall-twilio we were thinking of failing an entire batch if it will cause the spreadsheet to exceed 300,000 cells right? Is it possible to accurately know that though, if it's possible that some items in the batch will update (not append) existing cells?

kdaswani: Robert beat me! Given @rhall-twilio's comment, I think it makes sense to not include an extra check. If we don't add that check but then the spreadsheet does end up exceeding 300,000 cells, will we still throw some sort of error? What's the behavior (part of the batch is still processed vs. the entire batch is ignored)?

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
kdaswani: I think this was based on @rhall-twilio's testing. More info here: https://segment.slack.com/archives/C03ADA99G7R/p1652292973665399

packages/destination-actions/src/destinations/google-sheets/postSheet/operations.ts
```diff
@@ -55,8 +57,13 @@ const generateColumnValuesFromFields = (identifier: string, fields: Fields, colu
...
  * @returns
  */
-function processGetSpreadsheetResponse(response: GetResponse, events: Payload[]) {
-  // TODO (STRATCONN-1375): Fail request if above row limit
+function processGetSpreadsheetResponse(response: GetResponse, events: Payload[], mappingSettings: MappingSettings) {
+  const numColumns = mappingSettings.columns.length
+  const numRows = response.values.length
+
+  if (numRows * numColumns > MAXCELLS) {
+    throw new IntegrationError('Sheet has reached maximum limit', 'INVALID_REQUEST_DATA', 400)
```
rhall-twilio, line 65: We should add a test-case to validate this branch.

packages/destination-actions/src/destinations/google-sheets/__tests__/postSheet.operations.test.ts
```diff
@@ -3,6 +3,13 @@ import { Settings } from '../generated-types'
 import { Payload } from '../postSheet/generated-types'
 import PostSheet from '../postSheet/index'
 import { GoogleSheets, GetResponse } from '../googleapis/index'
+import * as constants from '../constants'
```
rhall-twilio, line 6:
```suggestion
import { CONSTANTS } from '../constants'
```

packages/destination-actions/src/destinations/google-sheets/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const CONSTANTS = {
+  /**
```
marinhero, line 2: Nice!

marinhero (approved):
:shipit:

PR #612, opened 2022-06-01
Update heap.track calls

GedRap:
After some internal debates, we decided to update how we ingest events coming from this Segment integration.

Instead of passing `segment` as the third `heap.track` parameter, we decided to merge this value into the events property object. This change will make it easier to process events in our preferred way on our end.

In addition to unit tests, I verified that the events appear in Heap correctly using the local server.

### DIFF
--- a/packages/browser-destinations/src/destinations/heap/constants.ts
+++ b/packages/browser-destinations/src/destinations/heap/constants.ts
@@ -1,1 +1,1 @@
-export const HEAP_LIBRARY_NAME = 'segment'
+export const HEAP_SEGMENT_LIBRARY_NAME = 'destinations-actions'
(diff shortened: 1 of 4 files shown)


chrisradek (commented):
This looks great to me! Only comment is I think we can either remove the `constants.ts` file altogether or move `HEAP_SEGMENT_LIBRARY_NAME` to the constants file.

GedRap:
> This looks great to me! Only comment is I think we can either remove the `constants.ts` file altogether or move `HEAP_SEGMENT_LIBRARY_NAME` to the constants file.

@chrisradek thanks for pointing this out! Just updated the code.

marinhero (approved):

chrisradek (approved):
Thanks for making the update, looks good to me!

pooyaj (approved):

PR #615, opened 2022-06-02
STRATCONN-1350 : Bugfix - Support other datatypes as trait values in SF Actions

manali-bhosale:
We only supported String types for trait values so far which breaks some formatting related code in SOQL query generation code. This PR adds code to handle numeric and boolean data types along with String. We don't support object types yet. 

## Testing

#### Configured mappings for trait: 

Screen Shot 2022-06-01 at 5 10 17 PM

#### Issue reproduced in staging: 
Note the last two fields, values for No of Employees and Account Number fields. 
Screen Shot 2022-06-01 at 4 59 28 PM

#### After the fix: 
All String doesn't work after fixing the error. SF doesn't support String values in numeric fields. 
Screen Shot 2022-06-01 at 5 08 30 PM

SF doesn't support numeric data in String fields. 
Screen Shot 2022-06-01 at 5 09 51 PM

Values of correct types:
Screen Shot 2022-06-01 at 5 09 30 PM

### DIFF
--- a/packages/destination-actions/src/destinations/salesforce/sf-operations.ts
+++ b/packages/destination-actions/src/destinations/salesforce/sf-operations.ts
@@ -166,13 +166,31 @@ export default class Salesforce {
   // Salesforce field names should have only characters in {a-z, A-Z, 0-9, _}.
   private removeInvalidChars = (value: string) => value.replace(/[^a-zA-Z0-9_]/g, '')
 
+  // Pre-formats trait values based on datatypes for correct SOQL syntax
+  private typecast = (value: any) => {
+    switch (typeof value) {
+      case 'boolean':
+        return value
+      case 'number':
+        return value
+      case 'string':
+        return `'${this.escapeQuotes(value)}'`
+      default:
+        throw new IntegrationError(
+          'Unsupported datatype for record matcher traits - ' + typeof value,
+          'Unsupported Type',
+          400
+        )
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


brennan (commented):
The tests and code LGTM! I'll defer approval to @nick-Ag and @mericsson who are probably better equipped to determine whether Salesforce will play nice with this solution.

mericsson (approved):

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -161,18 +161,31 @@ export default class Salesforce {
...
   private removeInvalidChars = (value: string) => value.replace(/[^a-zA-Z0-9_]/g, '')
 
+  // Preformats trait values based on datatypes for correct SOQL syntax
+  private typecast = (value: any) => {
+    switch (typeof value) {
+      case 'boolean':
+        return value
+      case 'number':
+        return value
+      default:
```
nick-Ag, line 176: I wonder if we should throw an error here when users try to lookup by object and array types? I think those sort of lookups would fail with 404 anyways and so it would be better to have a descriptive error message: (400: "Invalid lookup type: Array/Object")

manali-bhosale: Done. I have added a test case for that as well.

nick-Ag (approved):

marinhero (approved):
:shipit:

PR #621, opened 2022-06-08
[BE-837] Sev Remediation

dlasky:
Added validation + sane defaults for settings.

To reproduce, remove default value from auth settings where type is `boolean` and run `./bin/run validate` locally.
Screen Shot 2022-06-13 at 10 27 53 AM

To reproduce, set default value to a different data type (for boolean and number types) and run `./bin/run validate` locally.
Screen Shot 2022-06-13 at 10 28 18 AM

## Testing

Executed push in stage and verified diffs. Executed validate locally with intentionally broken values.

### DIFF
--- a/packages/cli/src/commands/validate.ts
+++ b/packages/cli/src/commands/validate.ts
@@ -151,7 +151,7 @@ export default class Validate extends Command {
           }
           if (typeof fieldValues?.default !== typ) {
             errors.push(
-              new Error(`The default value for field "${field}" is not a ${typ}, but the type is set to ${typ}.`)
+              new Error(`The default value for field "${field}" is of type "${typeof fieldValues?.default}", but the type is set to "${typ}".`)
             )
           }
         }


jinapark202 (dismissed):

packages/cli/src/commands/validate.ts
```diff
@@ -122,6 +127,26 @@ export default class Validate extends Command {
...
 
+  validateSettings(destination: DestinationDefinition) {
+    const errors: Error[] = []
+    if (destination.mode === 'cloud') {
+      const dest = destination as CloudDestinationDefinition
+      Object.keys(dest.authentication?.fields ?? {}).forEach((field) => {
+        const typ = dest.authentication?.fields[field].type
+        //TODO: consider invalidating here -- for now we just warn
+        // this.isInvalid = true
+        if (typ === 'boolean' || typ === 'number') {
```
brennan, line 138: Maybe something like `if (typ === 'boolean' && typeof typ !== 'boolean')` and so on for other supported types?

packages/cli-internal/src/commands/push.ts
```diff
@@ -411,10 +411,17 @@ export function getOptions(
...
+      boolean: false,
+      string: '',
+      password: ''
+    }
+
     options[fieldKey] = {
       ...existing,
       tags,
-      default: schema.default ?? '',
+      default: schema.default ?? defaultValues[schema.type],
```
brennan, line 424: This should ensure that we always have some default value in case `defaultValues[schema.type]` is undefined.

```suggestion
      default: schema.default ?? (defaultValues[schema.type] || ''),
```

packages/cli/src/commands/validate.ts
```diff
@@ -122,6 +127,26 @@ export default class Validate extends Command {
...
+          errors.push(
```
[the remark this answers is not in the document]
dlasky: @jinapark202 @brennan somehow failed to push my most recent lol

packages/cli-internal/src/commands/push.ts
dlasky: I dont believe it can be undefined given typescript would throw of the schema type didnt match one of those values

packages/cli/src/commands/validate.ts
```diff
@@ -122,6 +127,28 @@ export default class Validate extends Command {
...
+    const errors: Error[] = []
+    if (destination.mode === 'cloud') {
+      const dest = destination as CloudDestinationDefinition
+      Object.keys(dest.authentication?.fields ?? {}).forEach((field) => {
+        const fieldValues = dest.authentication?.fields[field]
+        //TODO: consider invalidating here -- for now we just warn
+        // this.isInvalid = true
+        const literalType = fieldValues?.type === 'boolean' || fieldValues?.type === 'number'
+        const emptyDefault = fieldValues?.default === undefined || fieldValues?.default === null
+        if (literalType && emptyDefault) {
```
brennan, line 140: This is validating that a default is not empty for literal settings fields. Should we also validate that the defaults are sane? e.g. if `type` is `boolean` that the `default` is `true` or `false`?

packages/cli-internal/src/commands/push.ts
brennan: I see, the `GlobalSetting` type right now, where type can only be one of these four options - you are right. Thanks for clarifying!

packages/cli/src/commands/validate.ts
dlasky: [f9e1450](https://github.com/segmentio/action-destinations/pull/621/commits/f9e145006710905b608c45f8ac2a6c2611e56fcb)

packages/cli/src/commands/validate.ts
```diff
@@ -122,6 +131,35 @@ export default class Validate extends Command {
...
+          if (!isDefined(fieldValues?.default)) {
+            errors.push(
+              new Error(
+                `The authentication field "${field}" of type "${fieldValues?.type}" does not contain a default value. It is recommended to choose a sane default to avoid validation issues.`
+              )
+            )
+          }
+          if (typeof fieldValues?.default !== typ) {
+            errors.push(
+              new Error(`The default value for field "${field}" is not a ${typ}, but the type is set to ${typ}.`)
```
brennan, line 154: Nit on the language - maybe something like the below? Not a blocker at all.

```suggestion
              new Error(`The default value for field "${field}" is of type "${typeof fieldValues?.default}", but the type is set to "${typ}".`)
```

dlasky:
I was actually thinking it's fine to just leave until after the freeze since we won't be doing other pushes until then anyway (hopefully)

jinapark202 (approved):

PR #626, opened 2022-06-13
Ik/intercom actions

ibbykhazanchi:
This PR is for the new Intercom web actions destination. 

Here is the [SDD](https://paper.dropbox.com/doc/Intercom-Actions-Device-SDD--Bj_7YX48bG1TZVMx7tTHErOOAg-VqvXyRNqaDat9VmFOGsRD).

## Testing

I have tested this end-to-end, from the web actions destinations tester to the Intercom User's dashboard. 


packages/browser-destinations/src/destinations/intercom/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  },
+
+  initialize: async ({ settings }, deps) => {
+    //initialize Intercom
+    initScript({ appId: settings.appId })
+    initialBoot(settings.appId)
+
+    await deps.resolveWhen(() => window.Intercom.booted === true, 100)
+
+    //put settings on the Intercom object
```
brennan, line 56: Nit: I would delete this comment because it's clear from the code what's happening here. :)

packages/browser-destinations/src/destinations/intercom/index.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { BrowserDestinationDefinition } from '../../lib/browser-destinations'
+import { browserDestination } from '../../runtime/shim'
+import { initialBoot, initScript } from './init-script'
+
+import trackEvent from './trackEvent'
+import { Intercom } from './api'
+
+import updateUser from './updateUser'
+
+import updateCompany from './updateCompany'
```
brennan, line 11: Nit on order and spacing:

```suggestion
import { Intercom } from './api'
import trackEvent from './trackEvent'
import updateUser from './updateUser'
import updateCompany from './updateCompany'
```

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,101 @@
+import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
+import { Intercom } from '../api'
+import type { Settings } from '../generated-types'
+import { filterCustomTraits, isEmpty } from '../utils'
+import type { Payload } from './generated-types'
+
+// Change from unknown to the partner SDK types
+const action: BrowserActionDefinition<Settings, Intercom, Payload> = {
+  title: 'Track Event',
+  description: '',
```
brennan, line 10: Should we add a description?

packages/browser-destinations/src/destinations/intercom/updateCompany/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    company: {
+      description: "The user's company.",
+      label: 'Company',
+      type: 'object',
+      required: true,
+      properties: companyProperties,
+      default: {
+        company_id: { '@path': '$.groupId' },
+        name: { '@path': '$.traits.name' },
+        created_at: { '@path': '$.traits.createdAt' },
```
brennan, line 25: Should we default to `createdAt`, falling back to `created_at`?

packages/browser-destinations/src/destinations/intercom/updateCompany/index.ts
```diff
@@ -0,0 +1,73 @@
...
+        monthly_spend: { '@path': '$.traits.monthlySpend' },
+        size: { '@path': '$.traits.size' },
+        website: { '@path': '$.traits.website' },
+        industry: { '@path': '$.traits.industry' },
+        company_custom_traits: { '@path': '$.traits' }
+      }
+    },
+    hide_default_launcher: {
+      description:
+        'selectively show the chat widget. According to Intercom’s docs, you want to first hide the Messenger for all users inside their UI using Messenger settings. Then think about how you want to programmatically decide which users you’d like to show the widget to.',
```
brennan, line 36: Capitalize "S" here.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+import type { Settings } from '../generated-types'
+import { getCompanyProperties } from '../sharedCompanyProperties'
+import { convertDateToUnix, filterCustomTraits, getWidgetOptions, isEmpty } from '../utils'
+import type { Payload } from './generated-types'
+
+const companyProperties: Record<string, InputField> = getCompanyProperties()
+
+const action: BrowserActionDefinition<Settings, Intercom, Payload> = {
+  title: 'Update',
+  description: '',
```
brennan, line 13: Should we add a description?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        '@path': '$.traits.name'
+      }
+    },
+    first_name: {
+      description: "The user's first name.",
+      label: 'Name',
+      type: 'string',
+      required: false,
+      default: {
+        '@path': '$.traits.firstName'
```
brennan, line 49: Should we fall back to `traits.first_name`?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+      }
+    },
+    last_name: {
+      description: "The user's last name.",
+      label: 'Name',
+      type: 'string',
+      required: false,
+      default: {
+        '@path': '$.traits.lastName'
```
brennan, line 58: Fall back to `traits.last_name`?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        '@path': '$.traits.phone'
+      }
+    },
+    unsubscribed_from_emails: {
+      description: 'The email unsubscribe status for the user.',
+      label: 'Unsubscribed From Emails',
+      type: 'boolean',
+      required: false,
+      default: {
+        '@path': '$.traits.unsubscribedFromEmails'
```
brennan, line 76: Because this field is not required, and it's not in Segment's event spec, I probably wouldn't put a default. https://segment.com/docs/connections/spec/

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+      }
+    },
+    language_override: {
+      description: 'The messenger language (instead of relying on browser language settings).',
+      label: 'Language Override',
+      type: 'string',
+      required: false,
+      default: {
+        '@path': '$.traits.languageOverride'
```
brennan, line 85: Same for this field - not in Segment's spec.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        '@path': '$.traits.email'
+      }
+    },
+    created_at: {
+      description: 'A timestamp of when the person was created.',
+      label: 'Created At',
+      type: 'datetime',
+      required: false,
+      default: {
+        '@path': '$.traits.createdAt'
```
brennan, line 103: Same for this field - not in Segment's spec.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+      type: 'object',
+      required: false,
+      properties: {
+        image_url: {
+          description: 'The avatar/profile image URL.',
+          label: 'Image URL',
+          type: 'string',
+          required: true,
+          default: {
+            '@path': '$.traits.avatar'
```
brennan, line 118: Not in Segment's spec.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+    company: {
+      description: "The user's company.",
+      label: 'Company',
+      type: 'object',
+      required: false,
+      properties: companyProperties,
+      default: {
+        company_id: { '@path': '$.traits.company.id' },
+        name: { '@path': '$.traits.company.name' },
+        created_at: { '@path': '$.traits.company.createdAt' },
```
brennan, line 148: Should we fall back to `created_at`?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+        plan: { '@path': '$.traits.company.plan' },
+        monthly_spend: { '@path': '$.traits.company.monthlySpend' },
```
brennan, line 150: `monthly_spend` also isn't specced, but it might be strange UX if this is the only one of these fields without a default value.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+      multiple: true,
+      required: false,
+      properties: companyProperties,
+      default: {
+        '@arrayPath': [
+          '$.traits.companies',
+          {
+            company_id: { '@path': '$.id' },
+            name: { '@path': '$.name' },
+            created_at: { '@path': '$.createdAt' },
```
brennan, line 170: Should we fall back to `created_at`?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,278 @@
...
+            plan: { '@path': '$.plan' },
+            monthly_spend: { '@path': '$.monthlySpend' },
```
brennan, line 172: Should we fall back to `monthly_spend`?

packages/browser-destinations/src/destinations/intercom/updateCompany/index.ts
ibbykhazanchi: I think that's a good idea. Curious, do you know how these fallbacks appear in the UI?

brennan: I'm actually not sure off the top of my head - would you have time to test and share? I'm curious, too. :) cc @nick-Ag 

Also created this ticket for the future: https://segment.atlassian.net/browse/STRATCONN-1544

dlasky: if you use an `@if` directive they show up as a `coalesce function` in the UI, which has 2 args in this case, primary and fallback value.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
ibbykhazanchi: Should I provide a fallback to `monthlySpend` here too?

ibbykhazanchi: And in general, should we provide fallbacks wherever there's a multi-word property in which there could be confusion between camelCase and snake_case

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
ibbykhazanchi: decided to remove mapping for monthly_spend as per 
https://github.com/segmentio/action-destinations/pull/682#discussion_r925955016

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
brennan: @ibbykhazanchi Yeah, I think that's a good rule of thumb. I don't think we're necessarily doing that right now in other integrations, but I think we should start doing so until we can get automatic fallbacks in place.

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
+import { InputField } from '@segment/actions-core'
+
+export const getCompanyProperties = (): Record<string, InputField> => ({
+  company_id: {
+    description: 'The company id of the company.',
```
kdaswani, line 5:
```suggestion
    description: 'The unique identifier of the company.',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    label: 'Company Id',
```
kdaswani, line 6:
```suggestion
    label: 'Company ID',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+  },
+  name: {
+    description: 'The name of the company.',
+    label: 'Company Name',
+    type: 'string',
+    required: true
+  },
+  created_at: {
+    description: 'The time the company was created in your system.',
+    label: 'Created At',
```
kdaswani, line 18:
```suggestion
    label: 'Company Creation Time',
```

kdaswani: Should we add the same default as cloud based on the group spec? https://github.com/segmentio/action-destinations/pull/682/files#r925846413

```
default: {
        '@path': '$.traits.createdAt'`
}
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    type: 'datetime',
+    required: false
+  },
+  plan: {
+    description: 'The name of the plan the company is on.',
```
kdaswani, line 23:
```suggestion
    description: 'The name of the plan you have associated with the company.',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    label: 'Company Plan',
+    type: 'string',
+    required: false
+  },
+  monthly_spend: {
+    description: 'How much revenue the company generates for your business.',
```
kdaswani, line 29:
```suggestion
    description: 'The monthly spend of the company, e.g. how much revenue the company generates for your business.',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    label: 'Monthly Spend',
+    type: 'integer',
+    required: false
+  },
+  size: {
+    description: 'The number of employees in the company.',
+    label: 'Size',
```
kdaswani, line 36:
```suggestion
    label: 'Company size',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    required: false
+  },
+  website: {
+    description: 'The URL for the company website.',
+    label: 'Website',
+    type: 'string',
+    required: false
+  },
+  industry: {
+    description: 'The industry of the company.',
```
kdaswani, line 47:
```suggestion
    description: 'The industry that the company operates in.',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+  },
+  size: {
+    description: 'The number of employees in the company.',
+    label: 'Size',
+    type: 'integer',
+    required: false
+  },
+  website: {
+    description: 'The URL for the company website.',
+    label: 'Website',
```
kdaswani, line 42:
```suggestion
    label: 'Company Website',
```

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
```diff
@@ -0,0 +1,58 @@
...
+    description: 'The industry that the company operates in.',
+    label: 'Industry',
+    type: 'string',
+    required: false
+  },
+  company_custom_traits: {
+    description: 'The custom traits for the company object.',
+    label: 'Company Custom Traits',
+    type: 'object',
+    required: false
```
kdaswani, line 56: I would suggest adding:

`defaultObjectUI: 'keyvalue'`

Since we aren't assigning a default, might make sense to show customers that they can input separate key-value mappings. This was my suggestion for cloud too: https://github.com/segmentio/action-destinations/pull/682/files#r925997362

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,100 @@
+import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
+import { Intercom } from '../api'
+import type { Settings } from '../generated-types'
+import { filterCustomTraits, isEmpty } from '../utils'
+import type { Payload } from './generated-types'
+
+const action: BrowserActionDefinition<Settings, Intercom, Payload> = {
+  title: 'Track Event',
+  description: 'Track an event',
```
kdaswani, line 9:
```suggestion
  description: 'Submit an event to Intercom.',
```

pooyaj (approved):
🚢🚢🚢

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    event_name: {
+      description: 'The name of the event.',
+      label: 'Event Name',
+      type: 'string',
+      required: true,
+      default: {
+        '@path': '$.event'
+      }
+    },
+    price: {
```
kdaswani, line 21: @ibbykhazanchi Is there a reason we have separate mappings for "price" and "currency"? I don't see these fields explicitly called out in the Javascript API docs: https://developers.intercom.com/installing-intercom/docs/javascript-api-attributes-objects

packages/browser-destinations/src/destinations/intercom/updateCompany/index.ts
```diff
@@ -0,0 +1,82 @@
...
+            exists: { '@path': '$.traits.createdAt' },
+            then: { '@path': '$.traits.createdAt' },
+            else: { '@path': '$.traits.created_at' }
+          }
+        },
+        plan: { '@path': '$.traits.plan' },
+        size: { '@path': '$.traits.size' },
+        website: { '@path': '$.traits.website' },
+        industry: { '@path': '$.traits.industry' },
+        company_custom_traits: { '@path': '$.traits' }
```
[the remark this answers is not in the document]
kdaswani: Ah I now realize this might already be within an object and you probably can't have a key/value object in another object. Is this right?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+            exists: { '@path': '$.traits.company.createdAt' },
+            then: { '@path': '$.traits.company.createdAt' },
+            else: { '@path': '$.traits.company.created_at' }
+          }
+        },
+        plan: { '@path': '$.traits.company.plan' },
+        size: { '@path': '$.traits.company.size' },
+        website: { '@path': '$.traits.company.website' },
+        industry: { '@path': '$.traits.company.industry' },
+        company_custom_traits: { '@path': '$.traits.company' }
```
kdaswani, line 169: I think we should remove this default for `company_custom_traits` otherwise we risk sending a whole bunch of stuff that is nested, invalid, etc. What do you think?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+            website: { '@path': '$.website' },
+            industry: { '@path': '$.industry' },
+            company_custom_traits: { '@path': '$.' }
+          }
+        ]
+      }
+    },
+    hide_default_launcher: {
+      description:
+        'Selectively show the chat widget. According to Intercom’s docs, you want to first hide the Messenger for all users inside their UI using Messenger settings. Then think about how you want to programmatically decide which users you’d like to show the widget to.',
```
kdaswani, line 203:
```suggestion
        'Selectively show the chat widget. As per [Intercom docs](https://www.intercom.com/help/en/articles/189-turn-off-show-or-hide-the-intercom-messenger), you want to first hide the Messenger for all users inside the Intercom UI using Messenger settings. Then think about how you want to programmatically decide which users you would like to show the widget to.',
```

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+                exists: { '@path': '$.createdAt' },
+                then: { '@path': '$.createdAt' },
+                else: { '@path': '$.created_at' }
+              }
+            },
+            plan: { '@path': '$.plan' },
+            size: { '@path': '$.size' },
+            website: { '@path': '$.website' },
+            industry: { '@path': '$.industry' },
+            company_custom_traits: { '@path': '$.' }
```
kdaswani, line 196: Remove this default?

kdaswani (changes requested):

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
kdaswani: @ibbykhazanchi we added the default back to maintain parity with the classic destination. Let's make the default `monthly_spend` for the files that send the company object. See here: https://github.com/segmentio/action-destinations/pull/682#discussion_r927148490. Sorry for the back and forth!

packages/browser-destinations/src/destinations/intercom/sharedCompanyProperties.ts
ibbykhazanchi: Yes! Agree this will be a better implementation. Users define their custom attributes

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
ibbykhazanchi: Looks like it doesn't need to be sent with the keyword "price", but a nested object is required to be properly formatted. 

As in, it must be sent like this:

`....
  <keyword> {
         amount: <your_amount_multiplied_by_100_since_intercom_requires_in_cents>
         currency:  <your_currency>
  }
`

ibbykhazanchi: Also, I think cloud-mode should include this too, because users will likely only send in`revenue` & `currency` without the proper formatting:

https://segment.com/docs/connections/spec/track/#properties

kdaswani: cc @NolanChan

I agree - we should make sure we give customers a way to map these fields if they are sending order data especially to give parity with the classic destination: https://segment.com/docs/connections/destinations/catalog/intercom/#revenue-and-currency. I think we should just have separate mappings for `Revenue` and `Currency` (not within an object in the mappings) and then when we construct the outgoing JSON payload, pop them into an object so that Intercom picks them up

kdaswani: ^ @NolanChan can we add that to the cloud mode integration too?

ibbykhazanchi: I like that idea, it'll probably be more coherent to the user. 

Do you know what would be a good name to give to this object?

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    custom_traits: {
+      description: "The user's custom traits.",
+      label: 'Custom Traits',
+      type: 'object',
+      required: false,
+      default: {
```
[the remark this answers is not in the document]
ibbykhazanchi: So for web, custom attributes don't need to be pre-created. However, I think it'll be better that way, and will make it consistent between web & cloud.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+        '@if': {
+          exists: { '@path': '$.traits.firstName' },
+          then: { '@path': '$.traits.firstName' },
+          else: { '@path': '$.traits.first_name' }
+        }
+      }
+    },
+    last_name: {
```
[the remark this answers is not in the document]
ibbykhazanchi: I personally find it overkill too, but It looks like something that was [supported in the classic destination](https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/intercom/test/index.test.js#L143-L170).

Not necessarily a reason to keep it, but what do you think?

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
kdaswani: Let's call it `price` based on the docs example and what we did in the classic integration to be safe.

`Revenue`
- type = number
- description = `The amount associated with a purchase. Segment will multiply by 100 as Intercom requires the amount in cents.`
- default = `properties.revenue`

`Currency`
- type = string
- description = `The currency of the purchase amount.`
- default = `properties.currency`

kdaswani: Let me know if that makes sense @ibbykhazanchi! We'll need to update some of the logic below to remove it from the `price` object and only set currency if revenue is provided.

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+        image_url: {
+          description: 'The avatar/profile image URL.',
+          label: 'Image URL',
+          type: 'string',
+          required: true,
+          default: {
+            '@path': '$.traits.avatar'
+          }
+        },
+        type: {
```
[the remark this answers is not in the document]
ibbykhazanchi: agreed 100%.

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
ibbykhazanchi: Sounds good to me!

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
```diff
@@ -0,0 +1,298 @@
...
+      description: "The user's identity.",
+      label: 'Identity',
+      type: 'string',
+      required: false,
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    custom_traits: {
+      description: "The user's custom traits.",
```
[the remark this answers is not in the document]
ibbykhazanchi: @kdaswani, unlike server-side, the JS API technically doesn't require you to pre-create your custom attributes. It [says](https://www.intercom.com/help/en/articles/179-send-custom-user-attributes-to-intercom) you should use the dashboard to pre-create them, but it **doesn't throw an error** if you were to send a custom attribute at runtime, unlike the rest API.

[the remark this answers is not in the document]
kdaswani: Oh great @ibbykhazanchi! Your update is perfect then

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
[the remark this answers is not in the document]
kdaswani: @ibbykhazanchi I think we should just keep `Name`. With classic, the separate ones were probably needed because we hardcode the mappings but with Actions if a customer uses first and last name fields they could just put both in the mapping and I believe we'd concatenate. Thoughts?

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      required: false,
+      default: {
+        '@path': '$.properties.currency'
+      }
+    },
+    event_metadata: {
+      description: 'Optional metadata describing the event.',
+      label: 'Event Metadata',
+      type: 'object',
+      required: true,
```
brennan, line 45: Should this be `required: false` if the metadata are optional?

packages/browser-destinations/src/destinations/intercom/updateUser/__tests__/index.test.ts
```diff
@@ -0,0 +1,546 @@
...
+    name: 'Show',
+    enabled: true,
+    subscribe: 'type = "identify" or "page"',
+    mapping: {
+      user_id: { '@path': '$.userId' },
+      custom_traits: {
+        height: {
+          '@path': '$.traits.height'
+        },
+        tech: {
```
brennan, line 18: I know this is a test mapping, but I'm a little confused by this trait. Should it be `weight`?

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    // filter out reserved fields, drop custom objects & arrays
+    const filteredMetadata = filterCustomTraits(event_metadata)
+
+    // create price object
+    let price = {}
+    if (payload.revenue) {
+      price = {
+        amount: payload.revenue * 100,
+        currency: payload.currency ?? 'USD'
+      }
```
brennan, line 76: Not a blocker, but the below may be very minutely more efficient because you don't have to create a new object and overwrite the existing one initialized on line 71.

```suggestion
      price.amount = payload.revenue * 100
      price.currency = payload.currency ?? 'USD'
```

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
brennan: @ibbykhazanchi This is the only outstanding comment that I think would block approval/deploy to stage. Let me know what you think!

kdaswani: Agree let's make this optional (remove requiredness)!

ibbykhazanchi: Agree 100%

packages/browser-destinations/src/destinations/intercom/trackEvent/index.ts
ibbykhazanchi: the only problem with this is that `amount` and `currency` dont exist on the type {}. 

I could define an interface `Price` that has keys `amount?` & `currency?`? And define `price` like:

const price : Price = {}

@brennan

brennan: Makes sense - I would just ignore the comment, then, and let's move forward! 🚢

brennan (approved):

packages/browser-destinations/src/destinations/intercom/updateUser/index.ts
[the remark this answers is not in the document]
ibbykhazanchi: @kdaswani You're totally right. I tested out:

>  `     "name": {
>         "@template": "{{traits.firstName}} {{traits.lastName}}"
>       },`

which works! I will make remove the naming logic.

kdaswani (approved):
thank you for all your work on this @ibbykhazanchi !

PR #632, opened 2022-06-16
Update logEvent action to correct Functionality and create logPurchase action

LLuque-twilio:
This PR addresses [JIRA BE-316](https://segment.atlassian.net/browse/BE-316).

In it, we update the `logEvent` action to the correct functionality as well as create a new `logPurchase` action that acts as a direct copy of the _old_ functionality of `logEvent`

| 6/27/22 Requested Updates:  | Results |
| <ul><li>For both `logEvent` and `logPurchase`</li><li>Changes default path of products array from `@path` to `@arrayPath` => Making the `key/value` view the default</li><li>Allow users to insert additional key/value pairs</li></ul>  | Screen Shot 2022-06-27 at 7 29 35 PM |

#### logEvent

| Previous  | Now |
|  Generated `Product Purchased` events for each product in the `products` array when it was included in the payload  | Does not generate any additional events when a `products` array is included  |
| Had a `trackRevenuePerProduct` field  | No longer has a `trackRevenuePerProduct` field  |
| When ^ was set to `false`, would copy the all `revenue` fields (price, quantity, revenue, productId, revenueType) from `event_properties` to the top level of the event sent, making amplitude think there was revenue made on non-purchase events | Keeps all `revenue` properties in the `event_properties` so amplitude does not count it as revenue made |
| Preset default would trigger on ALL `track` calls  | Now preset only triggers on track calls when the event name is NOT `Order Completed`  |

#### logPurchase

- Is a direct copy of the old functionality of `logEvent`
  - Keeps `trackRevenuePerProduct` field
     - If True, will move each product's revenue fields to the top of each additional `Product Purchased` event so amplitude counts it as revenue
     - If False, will move the `Revenue` field found in `event_properties` to the top level of the main event so amplitude counts it as revenue
  - Will generate additional `Product Purchased` events for each product in the products array (if it is included)
  - Preset default triggers on all track called with event name `Order Completed`

### Local

Tested locally using actions-tester to send events to the amplitude endpoint. There were three main test events sent:
1. event sent using `logEvent` action containing `revenue` of $20 and `products` array fields in event properties
2. event sent using `logPurchase` action containing `revenue` of $20  and `products` array of objects that each have `revenue` fields. `trackRevenuePerProduct` set to FALSE
3. event sent using `logPurchase` action containing `revenue` of $20 and `products` array of objects that each have `revenue` fields. `trackRevenuePerProduct` set to TRUE

Expected:
1. No revenue field is tracked on the user. No additional events generated.
2. Revenue field in _event properties_ is tracked in amplitude. Additional `Product Purchased` event generated for each product in `products` array. ($20 revenue expected)
3. Additional `Product Purchased` event generated for each product in `products` array and the specific revenue fields for each product are moved to the top of the individual event and tracked in amplitude. ($183 from 1st product + $346 from second product = $529 revenue expected)

Results (as expected):
| 1  | 2 | 3 |
| ------------- | ------------- | ------------- |
| Screen Shot 2022-06-16 at 1 49 57 PM | Screen Shot 2022-06-16 at 1 50 12 PM | Screen Shot 2022-06-16 at 1 50 27 PM |

### Stage

Changes deployed to stage and tested there. 

- Created a new instance of Amplitude to ensure the preset actions worked as intended:  ✅
Screen Shot 2022-06-17 at 3 58 28 PM

- Ensured new `logPurchase` works the _exact_ same way that the previous version of `logEvent` used to ✅
- Ensured new version of `logEvent` removed the `trackRevenuePerProduct` field and never raises the `revenue`, `price`, `quantity`, `productId`, or `revenueType` fields to the top level of the event sent to amplitude. Also does not create & send additional events when a products array is included in the payload. ✅
- Saw all events coming through in amplitude's side as expected ✅

### Additional Team testing requests
- Sent events through a previous `logEvent` instance on stage that already had the `trackRevenuePerProduct` field set (even though it doesn't show in the UI anymore) => Does not affect outcome, still seeing expected results and the field is ignored ✅
- Sent exact same event through stage version of `logPurchase` and prod version of `logEvent` and checked results in amplitude to ensure they were the exact same ✅


packages/destination-actions/src/destinations/amplitude/index.ts
```diff
@@ -4,17 +4,24 @@ import identifyUser from './identifyUser'
...
 const presets: DestinationDefinition['presets'] = [
   {
     name: 'Track Calls',
-    subscribe: 'type = "track"',
+    subscribe: 'type = "track" and event != "Order Completed"',
     partnerAction: 'logEvent',
     mapping: defaultValues(logEvent.fields)
   },
+  {
+    name: 'Track Calls',
```
dlasky, line 20: Nit, we should probably change this to order completed similar to how it is for braze: https://github.com/segmentio/action-destinations/blob/f8e1dc00c15f357dd4c6cb927331798adc11915b/packages/destination-actions/src/destinations/braze/index.ts#L81

PR #634, opened 2022-06-17
STRATCONN-1413: New event parameters check 

maryamsharif:
This PR addresses [STRATCONN-1413](https://segment.atlassian.net/browse/STRATCONN-1413?focusedCommentId=281578). Google only allows parameters to be sent in as string or number and currently if events are sent with different variables like `null` GA4 silently drops these events. 

This PR adds a check to the params and user_properties and throws an error if the variables are not the right type. 

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/google-analytics-4/addToCart/index.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/addToCart/index.ts
@@ -2,7 +2,7 @@ import { ActionDefinition, IntegrationError } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
 import { ProductItem } from '../ga4-types'
-import { verifyCurrency } from '../ga4-functions'
+import { verifyCurrency, verifyParams } from '../ga4-functions'
 import {
   formatUserProperties,
   user_properties,
@@ -70,7 +70,7 @@ const action: ActionDefinition<Settings, Payload> = {
               items: googleItems,
               value: payload.value,
               engagement_time_msec: payload.engagement_time_msec,
-              ...payload.params
+              ...verifyParams(payload.params)
             }
           }
         ],
--- a/packages/destination-actions/src/destinations/google-analytics-4/addToWishlist/index.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/addToWishlist/index.ts
@@ -2,7 +2,7 @@ import { ActionDefinition, IntegrationError } from '@segment/actions-core'
 import { ProductItem } from '../ga4-types'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import { verifyCurrency } from '../ga4-functions'
+import { verifyCurrency, verifyParams } from '../ga4-functions'
 import {
   formatUserProperties,
   user_properties,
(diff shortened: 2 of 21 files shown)


packages/destination-actions/src/destinations/google-analytics-4/__tests__/addPaymentInfo.test.ts
```diff
@@ -571,5 +571,127 @@ describe('GA4', () => {
...
+          settings: {
+            apiSecret,
+            measurementId
+          },
+          mapping: {
+            client_id: {
+              '@path': '$.anonymousId'
+            },
+            user_properties: {
+              hello: null,
```
rhall-twilio, line 662: @kdaswani Can we confirm that the intention was indeed to prevent nulls in user_properties too? I'm confused by this comment:
https://segment.atlassian.net/browse/STRATCONN-1413?focusedCommentId=281427

maryamsharif: We were initially having trouble figuring out what was and wasn't accepted by ga4, but after looking at the [ga4 event builder](https://ga-dev-tools.web.app/ga4/event-builder/) we saw that they only give you the option to add user_properties as strings and numbers.
screen_shot_2022-06-16_at_11 44 22_am
.

rhall-twilio: My concern is that we may have customers that are passing null in user properties and the integration is working for them and suddenly this change would break their integration unprompted.

kdaswani: Hmm that's a fair concern! @maryamsharif since we've seen `null` for user properties go through successfully, could we have two different functions for verifying the data type? One function is for event properties and matches what you've already done. And then a separate function for user properties that allows nulls?

kdaswani: The error message for a `verifyEventProps` function could be: `GA4 only accepts string or number values for event parameters and item parameters. Please ensure you are not including null, array, or nested values.`

The error message for a `verifyUserProps` function could be: `GA4 only accepts string, number or null values for user properties. Please ensure you are not including array or nested values.`

rhall-twilio (approved):

rhall-twilio:
@maryamsharif Let's hold off on merging until we have the flagon plan resolved.

rhall-twilio (approved):

packages/destination-actions/src/destinations/google-analytics-4/addToWishlist/index.ts
```diff
@@ -87,7 +87,7 @@ const action: ActionDefinition<Settings, Payload> = {
               value: payload.value,
               items: googleItems,
               engagement_time_msec: payload.engagement_time_msec,
-              ...payload.params
+              ...verifyParams(payload.params)
```
brennan, line 90: Should this be gated behind a feature flag, also?

packages/destination-actions/src/destinations/google-analytics-4/addToCart/index.ts
```diff
@@ -57,6 +57,29 @@ const action: ActionDefinition<Settings, Payload> = {
       })
     }
 
+    if (features && features['actions-google-analytics-4-verify-params-feature']) {
```
brennan, line 60: This approach definitely works - but I wonder if there's a way to DRY up the code to make it more readable. Specifically, it isn't necessarily obvious at glance how this request differs from the request that isn't gated behind a feature. Can we DRY up the code in such a way that makes it immediately obvious that, behind the feature flag, we're passing `payload.params` to the `verifyParams` method?

PR #637, opened 2022-06-17
[BE-787] Prevent actual value coercion to string for '=' and '!=' operators

SyedWasiHaider:
Our current trigger UI only supports input for string values when using the `is` and `is not` operator

For instance below:
image

The value for `clv` above is actually interpreted and persisted as a string (`"5"`) instead of number (`5`) which is not at all obvious looking at the UI above. 

Unfortunately, our tester then also lets you pass in `{'clv': 5}` or `{'clv': "5"}` as valid matching event payload input. In reality, when you actually send events through our data pipeline, we will filter out events with `{'clv': 5}` because we're actually _just_ looking for `{clv: "5"}`.

This PR is a short term fix to prevent any user confusion but our long term fix will be to add more granular control over data type interpretations.

## Testing

- ~Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)~ N/A 

Trying to match on a number `5` will fail (whereas previously this would succeed)
Screen Shot 2022-06-17 at 12 08 31 PM

Trying to match on string `"5"` will pass:
Screen Shot 2022-06-17 at 12 07 56 PM

### DIFF
--- a/packages/destination-subscriptions/src/__tests__/validate.test.ts
+++ b/packages/destination-subscriptions/src/__tests__/validate.test.ts
@@ -131,7 +131,9 @@ test('operators - equals (numbers)', () => {
       ]
     }
 
-    expect(validate(ast, { properties: { value: 123 } })).toEqual(true)
+    // Since action tester UI only supports string type input, we assume the value in the ast is a string for now (i.e. 123 !== "123")
+    expect(validate(ast, { properties: { value: 123 } })).toEqual(false)
+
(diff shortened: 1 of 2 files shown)


jinapark202:
@SyedWasiHaider What if users actually want the event property to equal a number and not a string of a number? Is that not possible until we get a long-term fix?

jinapark202 (approved):
LGTM

PR #650, opened 2022-06-28
webhook: Add optional request signing

tysonmote:
This PR adds the ability to sign requests sent by the Webhook destination. The method used here is compatible with the existing Segment webhook destination, as documented here: https://segment.com/docs/connections/destinations/catalog/webhooks/#authentication

## Testing

Testing completed successfully locally:

### DIFF
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -89,12 +89,12 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
   readonly destinationName: string
   readonly schema?: JSONSchema4
   readonly hasBatchSupport: boolean
-  private extendRequest: RequestExtension<Settings> | undefined
+  private extendRequest: RequestExtension<Settings, any> | undefined
 
   constructor(
     destinationName: string,
     definition: ActionDefinition<Settings, Payload>,
-    extendRequest?: RequestExtension<Settings>
+    extendRequest?: RequestExtension<Settings, any>
   ) {
     super()
     this.definition = definition
--- a/packages/destination-actions/src/destinations/webhook/__test__/webhook.test.ts
+++ b/packages/destination-actions/src/destinations/webhook/__test__/webhook.test.ts
@@ -1,18 +1,15 @@
 import nock from 'nock'
 import { createTestEvent, createTestIntegration } from '@segment/actions-core'
 import Webhook from '../index'
+import { createHmac } from 'crypto'
 
 const testDestination = createTestIntegration(Webhook)
-const timestamp = new Date().toISOString()
 
 describe('Webhook', () => {
   describe('send', () => {
     it('should work with default mapping', async () => {
       const url = 'https://example.com'
-      const event = createTestEvent({
-        timestamp,
-        event: 'Test Event'
-      })
+      const event = createTestEvent()
 
       nock(url)
         .post('/', event as any)
@@ -32,10 +29,7 @@ describe('Webhook', () => {
 
     it('supports customizations', async () => {
       const url = 'https://example.build'
-      const event = createTestEvent({
-        timestamp,
-        event: 'Test Event'
-      })
+      const event = createTestEvent()
       const headerField = 'Custom-Header'
       const headerValue = 'Custom-Value'
       const data = { cool: true }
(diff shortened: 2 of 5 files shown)


packages/destination-actions/src/destinations/webhook/index.ts
```diff
@@ -1,12 +1,33 @@
...
+      sharedSecret: {
+        type: 'string',
+        label: 'Shared Secret',
+        description:
+          'If set, Segment will sign requests with an HMAC in the "X-Signature" request header. The HMAC is a hex-encoded SHA1 hash generated using this shared secret and the request body.'
+      }
+    }
+  },
+  extendRequest: ({ settings, payload }) => {
+    if (settings.sharedSecret && Object.prototype.hasOwnProperty.call(payload as object, 'data')) {
```
tysonmote, line 23: I'm a TypeScript noob so I'm not sure if there's a better way to do this. I ran into this linter error (https://github.com/satya164/eslint-browser/blob/master/docs/rules/no-prototype-builtins.md) and what I put here appears to be the recommended way to check if an object has a given property.

packages/destination-actions/src/destinations/webhook/index.ts
```diff
@@ -1,12 +1,33 @@
...
+      const digest = createHmac('sha1', settings.sharedSecret)
+        .update(JSON.stringify(payload.data), 'utf8')
```
tysonmote, line 25: One kind of unusual thing we're doing here is serializing the request body twice: once for the actual outbound request body and again here. I poked around a bit and I didn't see a good way to make the raw outbound request body available to `extendRequest` without doing a large refactor, and it looks like this is already on the team's radar: https://github.com/segmentio/action-destinations/blob/0dc437d4ef68b259d5adb3f632c295a086d6b8e3/packages/core/src/create-request-client.ts#L31

For now, given that `JSON.stringify(payload.data)` will be deterministic and consistent between the outbound request body and here, I think it's best to leave a larger refactoring of the request hooks to someone much more experienced with TypeScript and Node.js than I. :)

mericsson: fyi @rhall-twilio @dlasky

packages/destination-actions/src/destinations/webhook/index.ts
dlasky: i think you probably want "nullish coalescing operator" there:

`if (settings.sharedSecret && payload?.data) {`

tysonmote: Yeah, that's definitely cleaner though it can pull values from the prototype chain which feels weird. But maybe I'm overthinking it. I'll use the nullish coalescing operator!

dlasky (approved):

packages/core/src/destination-kit/action.ts
```diff
@@ -89,12 +89,12 @@ export class Action<Settings, Payload extends JSONLikeObject> extends EventEmitt
   readonly destinationName: string
   readonly schema?: JSONSchema4
   readonly hasBatchSupport: boolean
-  private extendRequest: RequestExtension<Settings> | undefined
+  private extendRequest: RequestExtension<Settings, any> | undefined
```
rhall-twilio, line 92: Can we add a line to ignore the warning as well as add a justification for it?

packages/destination-actions/src/destinations/webhook/__test__/webhook.test.ts
```diff
@@ -55,5 +49,34 @@ describe('Webhook', () => {
...
+      const responses = await testDestination.testAction('send', {
+        event,
+        mapping: {
+          url,
+          data: { '@path': '$.properties' }
+        },
+        settings: { sharedSecret },
+        useDefaultMappings: true
+      })
+
```
rhall-twilio, line 77: (nit) might be helpful for customers learning about usage to add a step to verify the payload with the shared secret. would also help validate the use case e2e.

packages/core/src/destination-kit/action.ts
rhall-twilio: Same for all modified files.

rhall-twilio (approved):

packages/core/src/destination-kit/action.ts
tysonmote: Fixed!

packages/destination-actions/src/destinations/webhook/__test__/webhook.test.ts
tysonmote: Good idea -- doing that now.

PR #652, opened 2022-06-28
Cordial Cart&Order Actions

espectrio:
Added 3 new actions to Cordial integration:
- addProductToCart
- removeProductFromCart
- upsertOrder

## Testing


packages/destination-actions/src/destinations/cordial/addProductToCart/__tests__/index.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const mapping = {
+      userIdentities: {'channels.email.address': 'contact@example.com'}
+    }
+
+    const settings = {
+      apiKey: 'cordialApiKey',
+      endpoint: 'https://api.cordial.io' as const
+    }
+
+    await testDestination.testAction('addProductToCart', {
```
rhall-twilio, line 39: It would be good to have at least one line of code validating behaviour in the testcase to protect against silent failures. 

This can be done by:
1.  Receiving the `response` object from this call and adding a few `expect` statements on it.
2. Validating that the endpoint got called (Nock [example](https://github.com/segmentio/action-destinations/blob/0ad65009907201bb717ca43e762aff457f77748a/packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts#L14); Mock [example](https://github.com/segmentio/action-destinations/blob/4c3eb45f5d92775cddb37687184a097904633705/packages/destination-actions/src/destinations/google-sheets/__tests__/postSheet.operations.test.ts#L81))

rhall-twilio: Same comment applies to removeProductFromCart and upsertOrder.

packages/destination-actions/src/destinations/cordial/addProductToCart/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Add Product to Cart',
+  description: 'Add product to Cordial contact cart',
+  defaultSubscription: 'type = "track" and event = "Product Added"',
+  fields: {
+    userIdentities: {
+      label: 'User Identities',
+      description:
+        'An ordered list of contact identifiers in Cordial. Each item in the list represents an identifier. For example, `channels.email.address -> userId` and/or `customerId -> traits.customerId`. At least one identifier should be valid otherwise the contact will not be identified and the request will be ignored.',
```
rhall-twilio, line 14: (nit) Can anything be used as a user identity, or does Cordial have a set number of properties that the API would allow? If so, it might be a good idea to link to some documentation here.

espectrio: Almost anything and our scheme is dynamic, so it would differ per customer. It would be resolved through our customer care department usually. We're preparing some brief documentation to be available in near-future.

packages/destination-actions/src/destinations/cordial/addProductToCart/__tests__/index.test.ts
espectrio: @rhall-twilio I've pushed an update to the test suite, please check if I've got you right.

Also, I've noticed that in test suites (e.g. packages/destination-actions/src/destinations/cordial/addProductToCart/__tests__/index.test.ts:121) not mapped properties are returned as undefined. Optional properties are nullable in generated-types, and while checking Dev Action Center I see that those properties are filtered out from request. On the other hand, when I put mapping in Dev Action Tester and then remove it, empty string is added as property.

Could you confirm that we would observe the same behavior  as in Dev Action Tester on Segment's production? Meaning that all null properties will be automatically filtered out from request and won't being sent, while empty objects or arrays will be preserved? Also, is there a way for a customer to remove mapping without property being set as empty string (so optional property would be null)?

espectrio: @rhall-twilio ping, just to check if above comment was seen

rhall-twilio: @dlasky for better context

espectrio: @rhall-twilio Could you please provide an ETA for when this PR may be merged?

dlasky: there were open issues so I wasn't sure if we were 100 to merge when I was looking at it.

espectrio: @dlasky Are there any right now? Please suggest what I should do to get this merged, thanks in advance!

rvadera12 (approved):

PR #654, opened 2022-06-29
[Customer.io] Use email if userId doesn't exist for createUpdatePerson

mike-engel:
We allow email addresses to be used as an identifier within our systems. We've gotten complaints from some users that they can't use the customer.io destination action if `userId` isn't present, but email is. This adds a small if/else block to `createUpdatePerson` that will use the email address trait if `userId` isn't present. If both aren't present, it should still fail on our end.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
+++ b/packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
@@ -15,7 +15,11 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       required: true,
       default: {
-        '@path': '$.userId'
+        '@if': {
+          exists: { '@path': '$.userId' },
+          then: { '@path': '$.userId' },
+          else: { '@template': '{{traits.email}}' }
+        }
       }
     },
     anonymous_id: {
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/customerio/createUpdatePerson/index.ts
```diff
@@ -15,7 +15,11 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       required: true,
       default: {
-        '@path': '$.userId'
+        '@if': {
+          exists: { '@path': '$.userId' },
+          then: { '@path': '$.userId' },
+          else: { '@template': '{{traits.email}}' }
```
rhall-twilio, line 21: (nit) It's not obvious to me why this is a template instead of a path directive.

mike-engel: I wasn't aware that traits could be paths (all other examples I looked at a while ago used templates)! If that's true, I can make this an update to be a path

rhall-twilio: Makes sense! I'd say let's make the change so that the code looks consistent and we reinforce the pattern. 👍

mike-engel: Done!

dlasky (approved):
lgtm, did you have anything else @rhall-twilio ?

rhall-twilio (approved):

PR #655, opened 2022-06-29
Allow builders to define their own enable_batching fields

nick-Ag:
This PR allows builders of batching destinations to override the automatically created `enable_batching` input field with their own. This will allow builders to customize the description, label, and default values of the field. This will be used in Salesforce to write new descriptions and labels, and make the field required.

## Testing
Tested by manually creating an `enable_batching` field in Salesforce, and pushing it to stage. The field is correctly added to the stage DB.

Screen Shot 2022-06-29 at 12 54 09 PM

### DIFF
--- a/packages/cli-internal/src/commands/push.ts
+++ b/packages/cli-internal/src/commands/push.ts
@@ -191,18 +191,6 @@ export default class Push extends Command {
               `The builder defined batching field is not a boolean. Please update the field type to "boolean".`
             )
           }
-
-          fields.push({
-            fieldKey: 'enable_batching',
-            type: 'boolean',
-            label: builderDefinedBatchingField.label,
-            description: builderDefinedBatchingField.description,
-            defaultValue: builderDefinedBatchingField.defaultValue,
-            required: builderDefinedBatchingField.required,
-            multiple: false,
-            dynamic: false,
-            allowNull: false
-          })
         }
 
         const base: BaseActionInput = {
--- a/packages/cli-internal/src/commands/push.ts
+++ b/packages/cli-internal/src/commands/push.ts
@@ -168,8 +168,11 @@ export default class Push extends Command {
           }
         })
 
+        const hasBuilderDefinedBatchingField = fields.filter((f) => f.fieldKey === 'enable_batching').length === 1
+        const isBatchingDestination = typeof action.performBatch === 'function'
+
         // Automatically include a field for customers to control batching behavior, when supported
-        if (typeof action.performBatch === 'function') {
+        if (isBatchingDestination && !hasBuilderDefinedBatchingField) {
           fields.push({
             fieldKey: 'enable_batching',
             type: 'boolean',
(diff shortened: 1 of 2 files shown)


packages/cli-internal/src/commands/push.ts
```diff
@@ -168,8 +168,11 @@ export default class Push extends Command {
           }
         })
 
+        const hasBuilderDefinedBatchingField = fields.filter((f) => f.fieldKey === 'enable_batching').length === 1
```
mericsson, line 171: Reason why this should be exactly `1`? nitpick could do

```suggestion
        const hasBuilderDefinedBatchingField = fields.some((f) => f.fieldKey === 'enable_batching')
```

packages/cli-internal/src/commands/push.ts
```diff
@@ -168,8 +168,11 @@ export default class Push extends Command {
...
+        const isBatchingDestination = typeof action.performBatch === 'function'
+
         // Automatically include a field for customers to control batching behavior, when supported
-        if (typeof action.performBatch === 'function') {
+        if (isBatchingDestination && !hasBuilderDefinedBatchingField) {
```
mericsson, line 175: Wonder if there is a way to capture this functionality in https://github.com/segmentio/action-destinations/blob/main/packages/cli-internal/src/__tests__/push.test.ts ?

nick-Ag:
> Wonder if there is a way to capture this functionality in https://github.com/segmentio/action-destinations/blob/main/packages/cli-internal/src/__tests__/push.test.ts ?

I don't think there is since `push.ts` doesn't output anything to check. I would want to check that the `fields` array includes a builder defined key but I don't see a clear way to do that

packages/cli-internal/src/commands/push.ts
```diff
@@ -181,6 +184,13 @@ export default class Push extends Command {
             dynamic: false,
             allowNull: false
           })
+        } else if (isBatchingDestination && hasBuilderDefinedBatchingField) {
+          const builderDefinedBatchingField = fields.filter((f) => f.fieldKey === 'enable_batching')[0]
```
marinhero, line 188: You already performed a lookup for this in line 171. Would it be possible to keep the object so you don't have to do this twice?

packages/cli-internal/src/commands/push.ts
```diff
@@ -181,6 +184,18 @@ export default class Push extends Command {
             dynamic: false,
             allowNull: false
           })
+        } else if (isBatchingDestination && builderDefinedBatchingField.length > 0) {
+          if (builderDefinedBatchingField[0].type !== 'boolean') {
```
marinhero, line 188: Nit: consider moving 188 to 198 to a `validateBatching` function so the parent block stays flat and easier to read.

marinhero (approved):

packages/cli-internal/src/commands/push.ts
```diff
@@ -310,6 +315,18 @@ export default class Push extends Command {
       await setSubscriptionPresets(metadata.id, presets)
     }
   }
+
+  validateBatching = (builderDefinedBatchingField: DestinationMetadataActionFieldCreateInput[]) => {
+    if (builderDefinedBatchingField[0].type !== 'boolean') {
```
jinapark202, line 320: For `builderDefinedBatchingField`, wonder if you can just do a `find` instead of a `filter` and do `builderDefinedBatchingField.type` here instead..

jinapark202 (approved):
small comment but otherwise LGTM

PR #657, opened 2022-06-30
do not escape forward slash for string directives

jinapark202:
This PR puts the logic to escape html characters for template string values behind a flag.

Further discussion here:
https://paper.dropbox.com/doc/Escape-Forward-Slash--BkYjgmWE22SsnHQMAJdt7CRJAg-17kgY9Ua5bQ7ZH7AfuxRD

## Testing

### DIFF
--- a/packages/core/src/mapping-kit/index.ts
+++ b/packages/core/src/mapping-kit/index.ts
@@ -7,8 +7,8 @@ import { removeUndefined } from '../remove-undefined'
 import validate from './validate'
 import { arrify } from '../arrify'
 
-type Directive = (options: JSONValue, payload: JSONObject) => JSONLike
-type StringDirective = (value: string, payload: JSONObject) => JSONLike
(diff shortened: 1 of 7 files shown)


dlasky (approved):

packages/core/src/mapping-kit/index.ts
```diff
@@ -26,17 +26,17 @@ function registerDirective(name: string, fn: Directive): void {
...
       throw new Error(`${name}: expected string, got ${realTypeOf(str)}`)
     }
 
-    return fn(str, payload)
+    return fn(str, payload, features)
   })
 }
 
-function runDirective(obj: JSONObject, payload: JSONObject): JSONLike {
+function runDirective(obj: JSONObject, payload: JSONObject, features?: { [key: string]: boolean }): JSONLike {
```
dlasky, line 39: only complaint would be we should consider making 'Features' a named type so you don't have to keep redefining this everywhere

PR #658, opened 2022-07-01
Changed keys for segment 2.0

srilok-engg-data:
Change the following keys for segment 2.0 

user_id : userID

anonymous_id : anonymousID

Currently, customer has to manually change this everytime while adding MoEngage as a destination in segment 2.0. 

## Testing


dlasky:
FYI this will only affect NEW setups as we don't retroactively alter mappings entries for existing keys to prevent data loss. 

Otherwise looks fine. I will note that you DO NOT explicitly need to change anything besides the [path values](https://github.com/segmentio/action-destinations/pull/658/files#diff-72a1750dc506d43a492984537d4a1157177e6cca9083b409a1298fc397285b3fR34) you changed on  to have the fix you are looking for work correctly, but in my opinion it does make sense to have the object keys match in payload as well.

dlasky (approved):

dlasky:
may want to take a look at the tests locally, looks like they failed on CI

PR #663, opened 2022-07-06
[Salesforce] - Simplify batching configuration

nick-Ag:
This PR implements a custom `enable_batching` field that provides unique info for users of the Salesforce Bulk API. It also removes the now redundant "Bulk Update" and "Bulk Upsert" operations. Instead users will be able to select `Operation: Upsert`, and `Enable Bulk API: true` to get the same configuration.

The flagon configuration in `app` that hides this from users during internal beta will not need any changes due to this PR.

## Testing

Tested successfully in stage. Event delivery still works and users will only have to flip the enable batching switch to use the bulk API.

Tested with the following configuration and python script:
Screen Shot 2022-07-08 at 10 19 23 AM

```python3
import analytics
import csv
import time

analytics.write_key = '<write-key>'
analytics.host = 'https://api.segment.build'

with open('./lead_id.csv') as file:
    reader = csv.reader(file)
    i = 0
    for row in reader:
        lead_id = row[0]
        print(i, lead_id)
        time.sleep(.01)
        analytics.track(str(i), 'Lead', {
            'recordId': lead_id,
            'company': 'Bulk Update 7.7 ' + str(i),
            'last_name': 'Stage 7.7 ' + str(i),
            'email': 'stage@test.7.7' + str(i)
        })

        i += 1

    print('flushing...')
    analytics.flush()
    print('done')

```
`lead_id.csv` is a list of 3000 record IDs for lead records in my testing Salesforce account.

Pics below show records updated today via stage test.
Screen Shot 2022-07-07 at 4 28 27 PM
Screen Shot 2022-07-07 at 4 28 14 PM

### DIFF
--- a/packages/destination-actions/src/destinations/salesforce/account/index.ts
+++ b/packages/destination-actions/src/destinations/salesforce/account/index.ts
@@ -7,7 +7,8 @@ import {
   customFields,
   operation,
   traits,
-  validateLookup
+  validateLookup,
+  enable_batching
 } from '../sf-properties'
 import type { Payload } from './generated-types'
 
@@ -19,6 +20,7 @@ const action: ActionDefinition<Settings, Payload> = {
   defaultSubscription: 'type = "group"',
   fields: {
     operation: operation,
+    enable_batching: enable_batching,
     traits: traits,
     bulkUpsertExternalId: bulkUpsertExternalId,
     bulkUpdateRecordId: bulkUpdateRecordId,
--- a/packages/destination-actions/src/destinations/salesforce/account/index.ts
+++ b/packages/destination-actions/src/destinations/salesforce/account/index.ts
@@ -202,7 +202,7 @@ const action: ActionDefinition<Settings, Payload> = {
   performBatch: async (request, { settings, payload }) => {
     const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
 
-    if (payload[0].operation === 'bulkUpsert') {
+    if (payload[0].operation === 'upsert') {
       if (!payload[0].name) {
         throw new IntegrationError('Missing name value', 'Misconfigured required field', 400)
       }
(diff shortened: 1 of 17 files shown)


brennan:
@nick-Ag How will this affect current beta users?

Since this also seems to be a pretty big change, I would suggest adding details on exactly how you tested. The goal should be to provide enough detail such that any engineer on the team could reproduce the test if needed. Thanks!

packages/destination-actions/src/destinations/salesforce/account/index.ts
```diff
@@ -200,7 +202,7 @@ const action: ActionDefinition<Settings, Payload> = {
   performBatch: async (request, { settings, payload }) => {
     const sf: Salesforce = new Salesforce(settings.instanceUrl, request)
 
-    if (payload[0].operation === 'bulkUpsert') {
+    if (payload[0].operation === 'upsert') {
```
rhall-twilio, line 205: Do these need to check that enable_batching is enabled?

nick-Ag:
> @nick-Ag How will this affect current beta users?
> 
> Since this also seems to be a pretty big change, I would suggest adding details on exactly how you tested. The goal should be to provide enough detail such that any engineer on the team could reproduce the test if needed. Thanks!

Existing configurations will need to re-enable the Enable Batching field since it defaults to false. Added some additional details to the writeup

packages/destination-actions/src/destinations/salesforce/account/index.ts
nick-Ag: I added a check for that in the bulkHandler here: https://github.com/segmentio/action-destinations/pull/663/files#diff-ce64a4366ba2a7982f3dffe832f75953866111b5d83fb6f18fec268635f6787cR86

I'm anticipating that this error will never fire as `integrations` only calls the performBatch method if `enable_batching` is true for the action

PR #664, opened 2022-07-07
Iterate destination

msingleton:
## Iterate Destination Overview
Thank you Segment team! We're excited to submit our device mode browser destination. Iterate provides survey software for our customers so we're taking advantage of the destination actions framework to allow our customers to easily deploy surveys by targeting events coming from Segment. You can learn more about our product here: [Iterate](https://iteratehq.com).

Our destination supports two actions: `trackEvent` and `identifyUser`

We're currently writing the documentation which we'll submit over on the segment-docs repo

## Testing

Instructions for testing:
1. Check out my branch
2. Using the local server, target the Iterate destination and use these settings (that API key is public and safe to use for demo purposes). We can send anyone on your team an invite to the Iterate account for this demo so you can verify the identify call works (we've tested and it does, but happy to share the invite). Once you fire the event below you will see the Iterate survey show up.

**Settings**
```
{
    "subscriptions": [
        {
            "partnerAction": "trackEvent",
            "name": "Track Event",
            "enabled": true,
            "subscribe": "type = \"track\"",
            "mapping": {
                "name": {
                    "@path": "$.event"
                }
            }
        },
        {
            "name": "Identify User",
            "subscribe": "type = \"identify\"",
            "partnerAction": "identifyUser",
            "enabled": true,
            "mapping": {
                "userId": {
                    "@path": "$.userId"
                },
                "traits": {
                    "@path": "$.traits"
                }
            }
        }
    ],
    "apiKey": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJjb21wYW55X2lkIjoiNjJjNTllNWQ3OTQ3MzgwMDAxZWU1NGYyIiwiaWF0IjoxNjU3MTE4MzAyfQ.vEPRW-lY8Hk6SpnBDRF7YP1k4zgtpRh-nCelKmwTEZs"
}
```

**Event**
```
{
    "name": "example",
    "userId": "43",
    "properties": {
        "foo": "bar"
    }, 
    "options": {}
}
```

**Survey displayed**
Screen Shot 2022-07-07 at 10 20 32 AM

--------

Let me know if you have any questions and thanks for making such an easy to use framework! It was a joy to build against.

### DIFF
--- a/packages/browser-destinations/src/destinations/iterate/identifyUser/index.ts
+++ b/packages/browser-destinations/src/destinations/iterate/identifyUser/index.ts
@@ -1,0 +1,42 @@
+import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { Iterate as IterateClient, Command } from '../types'
+
+const action: BrowserActionDefinition<Settings, IterateClient, Payload> = {
+  title: 'Identify User',
+  description: 'Sets user identity',
+  platform: 'web',
+  defaultSubscription: 'type = "identify"',
+  fields: {
+    userId: {
+      type: 'string',
+      required: false,
+      description: "The user's identity",
+      label: 'Identity',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    traits: {
+      type: 'object',
+      required: false,
+      description: 'The Segment traits to be forwarded to Iterate',
+      label: 'Traits',
+      default: {
+        '@path': '$.traits'
+      }
+    }
+  },
+  perform: (Iterate, event) => {
+    const traits = event.payload.traits || {}
+
+    if (event.payload.userId) {
+      traits['external_id'] = event.payload.userId
+    }
+
+    Iterate(Command.Identify, traits)
+  }
+}
+
+export default action
--- a/packages/browser-destinations/src/destinations/iterate/identifyUser/index.ts
+++ b/packages/browser-destinations/src/destinations/iterate/identifyUser/index.ts
@@ -12,7 +12,7 @@ const action: BrowserActionDefinition<Settings, IterateClient, Payload> = {
     userId: {
       type: 'string',
       required: false,
-      description: "The user's identity",
+      description: "The user's identity.",
       label: 'Identity',
       default: {
         '@path': '$.userId'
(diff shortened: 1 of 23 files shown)


packages/browser-destinations/src/destinations/iterate/identifyUser/index.ts
```diff
@@ -0,0 +1,42 @@
...
+const action: BrowserActionDefinition<Settings, IterateClient, Payload> = {
+  title: 'Identify User',
+  description: 'Sets user identity',
+  platform: 'web',
+  defaultSubscription: 'type = "identify"',
+  fields: {
+    userId: {
+      type: 'string',
+      required: false,
+      description: "The user's identity",
```
marinhero, line 15:
```suggestion
      description: "The user's identity.",
```

packages/browser-destinations/src/destinations/iterate/identifyUser/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      label: 'Identity',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    traits: {
+      type: 'object',
+      required: false,
+      description: 'The Segment traits to be forwarded to Iterate',
```
marinhero, line 24:
```suggestion
      description: 'The Segment traits to be forwarded to Iterate.',
```

marinhero (commented):
Well done! This looks good to me. I'm going to let @pooyaj give the final approval since Web Destinations are more on his wheelhouse.

packages/browser-destinations/src/destinations/iterate/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    // Initialize placeholder Iterate command
+    const iteratePlaceholder = <IterateApi>function (...args) {
+      iteratePlaceholder.command(args)
+    }
+    iteratePlaceholder.loaded = false
+    iteratePlaceholder.q = []
+    iteratePlaceholder.command = function (args: unknown[]) {
+      iteratePlaceholder.q.push(args)
+    }
+    window.Iterate = iteratePlaceholder
```
chrisradek, line 66: Does the `window.Iterate` object get replaced as part of loading?

If so, is it already replaced by the time `window.Iterate` is returned? Asking since I've seen race conditions in the past with some destinations that return `window.foo` before `window.foo` is replaced.

chrisradek (approved):
This is looking great! Just had a question around how Iterate is loaded.

packages/browser-destinations/src/destinations/iterate/index.ts
msingleton: @chrisradek Yea it does, down on line 69 we call

```
await deps.resolveWhen(() => !Object.prototype.hasOwnProperty.call(window.Iterate, 'loaded'), 100)
```

The "real" `window.Iterate` object has a `window.Iterate.loaded` property while the placeholder does not, that way we don't return the Iterate object until the real one is loaded

pooyaj (approved):

PR #666, opened 2022-07-07
publish browser package

dlasky:
This modifies the package json for browser-destinations to begin publishing this package. This will be consumed in some internal tooling that requires references to the types and manifest.

## Testing

Tested using `yalc` in the local environment.

### DIFF
--- a/packages/browser-destinations/tsconfig.build.json
+++ b/packages/browser-destinations/tsconfig.build.json
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "module": "esnext",
+    "module": "commonjs",
     "composite": true,
     "outDir": "dist",
     "rootDir": "src",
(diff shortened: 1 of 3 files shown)


packages/browser-destinations/tsconfig.build.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "module": "esnext",
+    "module": "commonjs",
```
dlasky, line 4: I confirmed with @pooyaj that this shouldn't impact anything web is doing with this code, however if there are concerns we could look at a dual export similar to actions core.

dlasky: interesting I would not have expected that but sure enough:

main:
Screen Shot 2022-07-07 at 4 19 12 PM

with change:

Screen Shot 2022-07-07 at 4 20 14 PM

PR #667, opened 2022-07-07
Add timeout option to webhook action

NolanChan:

## Testing

Screen Shot 2022-07-07 at 1 43 21 PM

Tested using actions-tester; putting in different Timeout numbers + url = `http://www.google.com:81/` led to differing amounts of time inbetween `etimedout` errors :) 

### DIFF
--- a/packages/destination-actions/src/destinations/webhook/send/generated-types.ts
+++ b/packages/destination-actions/src/destinations/webhook/send/generated-types.ts
@@ -16,6 +16,10 @@ export interface Payload {
     [k: string]: unknown
   }
   /**
+   * Time in milliseconds when a request should be aborted. Default is 10000
+   */
+  timeout?: number
+  /**
    * Payload to deliver to webhook URL (JSON-encoded).
    */
   data?: {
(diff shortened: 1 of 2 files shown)


tysonmote:
Centrifuge sets the request timeout for all destinations, and that's currently about 4 seconds across all destinations, with very rare exceptions for exceptional cases. We should be managing timeouts there, but I'm not aware of an exceptional reason for extending the timeout for webhook requests.

Timeouts are something we need to be strict about because long request times are expensive (costs money to sit around holding things in memory waiting for I/O) and wide variances in request time can dramatically change the properties of the system during unexpected increases in request time. For example, if you're sending 1,000 requests a second with 500ms round trip time, your average concurrency is 500. If your API has issues and now requests are timing out at 10 seconds, your average concurrency is now 10,000.

NolanChan:
Just throwing in some random input, but just want to corroborate that I experienced 4s timeouts when not setting it manually. This feature basically allows you to set a quicker timeout if desired :) I have a sneaking suspicion that amplitude is using timeouts for their action and setting it shorter, because they're experiencing improperly caught `etimedouterrors` (which I just fixed).

PR #670, opened 2022-07-12
Talon.One: new endpoints `update customer profiles v2` and `update customer sessions`

kaatinga:
- A new action `updateCustomerProfileV2` was added. It allows the customers to add a user to an audience without creating the audience before.
- A new action `updateCustomerSessions` was added. It allows the customers to update customer sessions.

## Testing

The tests pass only if `customerSessionId` is specified in `index.test.ts`/`should work` test of the `updateCustomerSession` slug. The current `should work` reports: `Nock: No match for request`.


packages/destination-actions/src/destinations/talon-one/updateCustomerSession/index.ts
```diff
@@ -0,0 +1,205 @@
...
+          multiple: true
+        },
+        referralCode: {
+          label: 'Referral Codes',
+          description: 'Any referral code entered.`',
+          type: 'string'
+        },
+        loyaltyCards: {
+          label: 'Referral Codes',
+          description: 'Any loyalty cards used. Up to 1 loyalty cards.`',
```
nick-Ag, line 62: If this only takes a maximum of 1 loyalty card, why make it an array type?

packages/destination-actions/src/destinations/talon-one/updateCustomerSession/index.ts
```diff
@@ -0,0 +1,205 @@
...
+        cardItems: {
+          label: 'Card Items',
+          description:
+            'The items to add to this sessions.\n' +
+            '\n' +
+            'If cart item flattening is disabled: Do not exceed 1000 items (regardless of their quantity) per request.\n' +
+            "If cart item flattening is enabled: Do not exceed 1000 items and ensure the sum of all cart item's quantity does not exceed 10.000 per request.`",
+          type: 'object',
+          multiple: true,
+          required: false,
```
nick-Ag, line 86: I would remove this line. The `required:true` in the properties items below (name, sku, etc) will conflict with this. If you want to make the `cardItems` object not required, while making those particular properties required only if a user provides a `cardItems` array you would have to do validation in the `perform` method to enforce that. 

Ex here: https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/facebook-conversions-api/viewContent/index.ts#L81

packages/destination-actions/src/destinations/talon-one/updateCustomerProfileV2/index.ts
```diff
@@ -0,0 +1,72 @@
...
+      properties: {
+        name: {
+          label: 'Name',
+          description: 'Attribute name',
+          type: 'string',
+          required: true
+        },
+        type: {
+          label: 'Type',
+          description: 'Attribute type. Can be only `string`, `time`, `number`, `boolean`, `location`',
```
nick-Ag, line 35: Perhaps you could use a choices object here to enforce the restrictions described?

packages/destination-actions/src/destinations/talon-one/updateCustomerSession/index.ts
kaatinga: @nick-Ag, good question, I just do not know how to fix our API in the destination. I would prefer to simplify this field in the destionation-action.

packages/destination-actions/src/destinations/talon-one/updateCustomerProfileV2/index.ts
kaatinga: added!

packages/destination-actions/src/destinations/talon-one/updateCustomerSession/index.ts
```diff
@@ -0,0 +1,129 @@
...
+    //
+    //     if (!item.name || !item.sku || !item.price || !item.quantity) {
+    //       throw new IntegrationError(
+    //         `card item must include an 'name', 'sku', 'price' and 'quantity' parameters.`,
+    //         'Misconfigured required field',
+    //         400
+    //       )
+    //     }
+    //   }
+    // }
```
kaatinga, line 110: please, @nick-Ag , look why it does not work. snapshot tests fail, unit tests do not trigger this error.

kaatinga:
@nick-Ag , I updated the PR.

- all the `required` fields inside 'customerSesstion' are removed. The are validated in talon.one, as temporary solution. 
- all the tests pass now
- added choices for attribute types.
- I tried to implement validation, but the tests fail. please, look at the commented implementation above.

let me know if something crucial must be fixed in this PR.

PS. Maybe we will merge our 2 PRs (https://github.com/segmentio/action-destinations/pull/643) into one for the Wednesday's release. Wdyt?

packages/destination-actions/src/destinations/talon-one/updateCustomerSession/index.ts
nick-Ag: Hi @kaatinga , there is a small import error with the `IntegrationError` code that is commented out here. It can be fixed with `import { ActionDefinition, IntegrationError } from '@segment/actions-core'`. 

Once I fixed that the unit test "Testing snapshot for TalonOne's updateCustomerSession destination action: › all fields" fails with:
`   IntegrationError: cart item must include an 'name', 'sku', 'price' and 'quantity' parameters.`

I'm not sure if your unit test here is expecting this error to be triggered, if so you could follow [this prior art](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/salesforce/__tests__/sf-operations.test.ts#L155). If you are not expecting that error, you would need to pass in all of those fields in your items array.

nick-Ag:
Once the tests pass this is ready for approval 👍

kaatinga:
> Once the tests pass this is ready for approval 👍

Hi @nick-Ag ! We have updated the PR. The additional validator of the cart times will be implemented in one of the next PRs. I do not see a reason to finish this right now as we have this validation in Talon.One (of course :) ). It is more important to release our new destination on Wednesday to let our partner use the new destinations.

Please, feel free to merge this PR to the main branch if there is no issue here.

packages/destination-actions/src/destinations/talon-one/updateCustomerSession/index.ts
kaatinga: @nick-Ag , we will work on this in one of the next PRs. Thank you!

nick-Ag (approved):

PR #678, opened 2022-07-13
BE-849 Add Track action + tests

NolanChan:
## Description

Adding the Track action!

Track Event API Docs - https://developers.intercom.com/intercom-api-reference/reference/submitting-events

This `track event` action is fairly straightforward, as we're passing in all attributes into Intercom's API. Unclear behavior is noted in the attribute descriptions, such as:

- Event names are case insensitive (they are automatically lowercased by intercom)
- There is a limit to 10 keys on metadata events

## Testing

Screen Shot 2022-06-15 at 1 22 39 PM

### DIFF
--- a/packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
+++ b/packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
@@ -1,0 +1,83 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import dayjs from '../../../lib/dayjs'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Track Event',
+  description: 'Submit a data event to Intercom.',
+  fields: {
+    event_name: {
+      type: 'string',
+      required: true,
+      description:
+        'The name of the event. Names are treated as case insensitive. Periods (.) and dollars ($) in event names are replaced with hyphens.',
+      label: 'Event Name',
+      default: {
+        '@path': '$.event'
+      }
+    },
+    created_at: {
+      type: 'datetime',
+      required: true,
+      description: 'A datetime in Unix timestamp format (seconds since Epoch).',
+      label: 'Created At',
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    user_id: {
+      type: 'string',
+      description: "The user's ID; required if no email provided.",
+      label: 'User ID',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    email: {
+      type: 'string',
+      description: "The user's email; required if no User ID provided.",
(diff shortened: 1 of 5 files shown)


packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,83 @@
...
+        '@path': '$.userId'
+      }
+    },
+    email: {
+      type: 'string',
+      description: "The user's email; required if no User ID provided.",
+      label: 'Email',
+      format: 'email',
+      default: {
+        '@path': '$.email'
```
brennan, line 43: Should this be `properties.email`, falling back to `traits.email`?

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,83 @@
...
+      default: {
+        '@path': '$.properties'
+      }
+    }
+  },
+  perform: (request, { payload }) => {
+    payload.created_at = convertValidTimestamp(payload.created_at)
+    return request('https://api.intercom.io/events', {
+      method: 'POST',
+      json: payload
```
brennan, line 60: What happens if `payload` contains more than 10 keys? Should we truncate properties to 10? Alternatively - and probably preferably - would could let customers map up to ten properties to this field in the UI, falling back to attempting to pass all properties.

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
NolanChan: Hey @brennan , thanks for the suggestion! Neils didn't have anything on his PRD about this, so I made a guess, but reviewing the [track spec](https://segment.com/docs/connections/spec/track/), your suggestion makes a lot more sense!

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
NolanChan: Hi @brennan , Intercom truncates the keys on their end -

> We currently limit the number of tracked metadata keys to 10 per event. Once the quota is reached, we ignore any further keys we receive. The first 10 metadata keys are determined by the order in which they are sent in with the event.

I'm also a bit confused about what you mean when you said 

> Alternatively - and probably preferably - would could let customers map up to ten properties to this field in the UI, falling back to attempting to pass all properties.

Are you referring to something syntactically that we're able to add inside the track mapping? I don't think I'm recognizing what you're referring to 😅

NolanChan: I see that customers are able to manually specify object key/value's inside a mapping's settings, but I don't think that's what you're talking about?

Screen Shot 2022-07-15 at 1 30 49 PM

PR #627, opened 2022-06-13
BE-805 Intercom Cloud Action Destination - Identify + Group

NolanChan:
The first PR for the Intercom Cloud Action Destination:  Includes code for `Identify` and `Group`

 Identify - Create or Update Contact 

-  We first try to create the user. If the creation failures due to a duplicate error, then find + update the user.  

 Group - Create or Update Company

- We first create the company, then attach the user. If the user doesn't exist yet, we throw a `retryable`

SDD - https://paper.dropbox.com/doc/Intercom-Actions-SDD--BiJ_eju4lu4bxVoNcxAnGsJCAg-GKk6vlMccszSpzbCCVPEi

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

PR #682, opened 2022-07-18
BE-805 Intercom Cloud Action-Destination

NolanChan:
## Description

[BE-805](https://segment.atlassian.net/browse/BE-805)
[PRD](https://paper.dropbox.com/doc/Intercom-Actions-Cloud-Mode-PRD--Bj01_GIuqAbCASv9dDcj1TPhAg-nMEizRG8DkHXTEp2fJQjr)
[SDD](https://paper.dropbox.com/doc/Intercom-Actions-SDD--BiJ_eju4lu4bxVoNcxAnGsJCAg-GKk6vlMccszSpzbCCVPEi)

Example workspace in staging with Intercom Cloud - https://app.segment.build/nolan/destinations/actions-intercom/sources/node-test/instances/62d1e8af0524a848e8cde209/event-tester

It contains an Identify mapping that creates users in this Intercom Workspace- https://app.intercom.com/a/apps/r9wsr2en

This PR was reviewed incrementally for [Identify, Group](https://github.com/segmentio/action-destinations/pull/627), and [Track](https://github.com/segmentio/action-destinations/pull/678). Nothing else has been added to this PR, but it'd be good to have one final check if possible!

## Testing


packages/destination-actions/src/destinations/index.ts
```diff
@@ -52,6 +52,7 @@ register('615cae349d109d6b7496a131', './tiktok-conversions')
 register('602efa1f249b9a5e2bf8a813', './twilio')
 register('614a3c7d791c91c41bae7599', './webhook')
 register('61dc4e96894a6d7954cc6e45', './voyage')
+register('62b5f2f40b10eaace2e12b23', './intercom')
```
marinhero, line 55: Confirmed ID matches our internal records 👍

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+    const contactId = payload.contact_id
+    delete payload.contact_id
+
+    const response = await createOrUpdateIntercomCompany(request, payload)
+    if (contactId) {
+      try {
+        const companyId = response.data.id
+        return await attachContactToIntercomCompany(request, contactId, companyId)
+      } catch (error) {
+        // Should be an HTTPError, but was failing instanceOf (?)
```
marinhero, line 113: I don't understand what this is trying to communicate. Can you be more specific?

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/__tests__/index.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    nock(`${endpoint}`).post(`/companies`).reply(200, {})
+    nock(`${endpoint}`).post(`/contacts/${userId}/companies`).reply(404, {})
+
+    await expect(
+      testDestination.testAction('groupIdentifyContact', {
+        event,
+        useDefaultMappings: true
+      })
+    ).rejects.toThrowError(new RetryableError(`Contact doesn't exist, retrying`))
```
marinhero, line 39: Shouldn't you test the subsequent success too?

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,75 @@
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
```
marinhero, line 1: This file is practically identical to `packages/destination-actions/src/destinations/intercom/__tests__/snapshot.test.ts` why do we duplicate that?

packages/destination-actions/src/destinations/intercom/__tests__/index.test.ts
```diff
@@ -0,0 +1,32 @@
+import nock from 'nock'
+import { createTestEvent, createTestIntegration, DecoratedResponse } from '@segment/actions-core'
+import Definition from '../index'
+
+const testDestination = createTestIntegration(Definition)
+const endpoint = 'https://api.intercom.io'
+
+describe('Intercom (actions)', () => {
+  describe('testAuthentication', () => {
```
marinhero, line 9: Should we test for failure too?

packages/destination-actions/src/destinations/intercom/identifyContact/__tests__/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    expect(responses.length).toBe(2)
+    expect(responses[0].status).toBe(200)
+    expect(responses[1].data).toMatchObject({})
+    expect(responses[1].options.body).toBe(
+      `{"role":"lead","external_id":"user1234","email":"user@example.com","name":"example user","last_seen_at":"${event.timestamp}"}`
+    )
+  })
+
+  it("should throw a retryable error if the user doesn't come up in a search but is a duplicate", async () => {
```
marinhero, line 50:
```suggestion
  it("should throw a retriable error if the user doesn't come up in a search but is a duplicate", async () => {
```

packages/destination-actions/src/destinations/intercom/identifyContact/__tests__/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    nock(`${endpoint}`).post(`/contacts/search`).reply(200, {})
+
+    await expect(
+      testDestination.testAction('identifyContact', {
+        event,
+        useDefaultMappings: true
+      })
+    ).rejects.toThrowError(
+      new RetryableError(
+        `Contact was reported duplicated but could not be searched for, probably due to Intercom search cache not being updated`
```
marinhero, line 63: Nice test!

packages/destination-actions/src/destinations/intercom/identifyContact/__tests__/snapshot.test.ts
```diff
@@ -0,0 +1,75 @@
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
```
marinhero, line 1: Same as `packages/destination-actions/src/destinations/intercom/__tests__/snapshot.test.ts` can we use only one? or why do we have to duplicate these files?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      description: 'The role of the contact. Accepted values are `user` or `lead`. Can only be updated if `lead`.',
+      label: 'Role',
+      default: 'lead'
+    },
+    external_id: {
+      type: 'string',
+      description: 'A unique identifier generated outside Intercom. Required if role=user and email is blank.',
+      label: 'External ID',
+      default: {
+        '@path': '$.userId'
```
marinhero, line 32: This will make external_id identical to user_id by default. Is that what you want? subsequent updates will generate more contacts in Intercom and that means our end-users will be paying more to Intercom.

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      }
+    }
+  },
+  perform: async (request, { payload }) => {
+    /**
+     * Tries to search and update the contact first. If no contact is found, then create.
+     * This is because we anticipate many more updates than creations happening in practice.
+     *
+     * Note: When creating a lead, Intercom doesn't accept an external_id (possibly a bug),
+     * it only accepts email.
```
marinhero, line 111: I don't think it's a bug. A lead often comes with less data, an `external_id` implies that they have been registered and that they are not a `lead` anymore but users.

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+        'The name of the event. Names are treated as case insensitive. Periods (.) and dollars ($) in event names are replaced with hyphens.',
+      label: 'Event Name',
+      default: {
+        '@path': '$.event'
+      }
+    },
+    created_at: {
+      type: 'datetime',
+      required: true,
+      description: 'A datetime in Unix timestamp format (seconds since Epoch).',
```
marinhero, line 23:
```suggestion
      description: 'A datetime in Unix timestamp format.',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    payload.created_at = convertValidTimestamp(payload.created_at)
+    delete payload.metadata?.email
+    return request('https://api.intercom.io/events', {
+      method: 'POST',
+      json: payload
+    })
+  }
+}
+
+function convertValidTimestamp(value: string | number): string | number {
```
marinhero, line 70: Might be worth adding unit tests for this method

packages/destination-actions/src/destinations/intercom/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  name: 'Intercom (Actions)',
+  slug: 'actions-intercom',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'oauth2',
+    fields: {},
+    testAuthentication: (request) => {
+      // Uses the admin route as a stand-in to test authentication creds.
+      // Should be a light request since there most likely won't be many admins
```
brennan, line 17: I'm not sure what this comment means. Is it risky to use the `admins` endpoint?

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    email: {
+      type: 'string',
+      description: "The user's email; required if no User ID provided.",
+      label: 'Email',
+      format: 'email',
+      default: {
+        '@if': {
+          exists: { '@path': '$.properties.email' },
+          then: { '@path': '$.properties.email' },
+          else: { '@path': '$.traits.email' }
```
brennan, line 46: `track` events won't have a `traits` object.

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+
+interface IntercomContact {
+  id: string
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Identify Contact',
+  description: 'Create or Update a Contact.',
+  defaultSubscription: 'type = "identify"',
+  platform: 'web',
```
brennan, line 18: Any reason the platform is `web`?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      default: {
+        '@path': '$.traits.avatar'
+      }
+    },
+    signed_up_at: {
+      label: 'Signed Up At',
+      type: 'datetime',
+      description: 'The timestamp when the contact was created.',
+      default: {
+        '@path': '$.createdAt'
```
brennan, line 74: I don't know if I would use a default for this field at all. `createdAt` is the timestamp the current event was created on the client device.

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      type: 'boolean',
+      description: 'Whether the contact is unsubscribed from emails.'
+    },
+    custom_attributes: {
+      label: 'Custom Attributes',
+      description:
+        'The custom attributes which are set for the contact. Note: Will throw an error if the object has an attribute that isn`t explicitly defined on Intercom.',
+      type: 'object',
+      default: {
+        '@path': '$.traits.customAttributes'
```
brennan, line 101: Is this in Segment's spec? If not, I would not use it as a default.

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  const response: ModifiedResponse<IntercomSearchData> = await request('https://api.intercom.io/contacts/search', {
+    method: 'POST',
+    json: { query }
+  })
+
+  if (response.data.total_count === 1) {
+    return response.data.data[0]
+  }
```
brennan, line 175:
```suggestion
  if (response?.data?.total_count === 1) {
    return response.data.data[0]
  }
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+async function updateIntercomContact(request: RequestClient, contactId: String, payload: Payload) {
+  return request(`https://api.intercom.io/contacts/${contactId}`, {
+    method: 'PUT',
+    json: {
+      ...payload
```
brennan, line 182: Why the spread operator here?

brennan:
@NolanChan Generally, how will RETL be able to use this destination if no `performBatch` handlers are defined?

brennan:
@NolanChan Generally, how will this destination be used with RETL with no `performBatch` handlers defined?

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+interface IntercomCreateCompanyData {
+  id: string
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Group Identify Contact',
+  description:
+    "Create or Update A Company. Note: Companies will be only visible in Intercom's Dashboard when there is at least one associated contact.",
+  defaultSubscription: 'type = "group"',
+  platform: 'web',
```
brennan, line 14: Should the platform be `cloud`?

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  fields: {
+    remote_created_at: {
+      type: 'datetime',
+      description: 'The time the company was created by you.',
+      label: 'Remote Created At',
+      default: {
+        '@path': '$.traits.createdAt'
```
brennan, line 21: Would recommend against a default here. `createdAt` refers to the time the current event took place, not the time at which the company was created.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      default: {
+        '@path': '$.traits.name'
+      }
+    },
+    monthly_spend: {
+      type: 'number',
+      description: 'The monthly spend of the company.',
+      label: 'Monthly Spend',
+      default: {
+        '@path': '$.traits.plan'
```
brennan, line 55: Looks like `monthly_spend` and `plan` both default to `traits.plan`. `monthly_spend` is type number, but `plan` is  type string. Also, `plan` is a Segment-specced group trait, which is usually type string: https://segment.com/docs/connections/spec/group/

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      description: 'The industry of the company.',
+      label: 'Industry',
+      default: {
+        '@path': '$.traits.industry'
+      }
+    },
+    custom_attributes: {
+      type: 'object',
+      description:
+        'Passing any traits not mapped to individual fields as Custom Attributes. Note: Will throw an error if you pass an attribute that isn`t explicitly defined.',
```
brennan, line 93:
```suggestion
        "Passing any traits not mapped to individual fields as Custom Attributes. Note: Will throw an error if you pass an attribute that isn't explicitly defined.",
```

brennan: I'm not sure I understand "Will throw an error if you pass an attribute that isn't explicitly defined."

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    const response = await createOrUpdateIntercomCompany(request, payload)
+    if (contactId) {
+      try {
+        const companyId = response.data.id
+        return await attachContactToIntercomCompany(request, contactId, companyId)
+      } catch (error) {
+        // Should be an HTTPError, but was failing instanceOf (?)
+        if (error?.response?.status === 404) {
+          throw new RetryableError(`Contact doesn't exist, retrying`)
```
brennan, line 115: Is there a risk of an infinite loop of retries here? What if a contact with that `contactId` is never created?

dlasky:
> @NolanChan Generally, how will this destination be used with RETL with no `performBatch` handlers defined?

as an aside -- if no batch handler is defined actions will call the default perform block in a loop with the events given. I can't comment on the intent in this specific PR, just wanted to spread that knowledge.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  request: RequestClient,
+  payload: Payload
+): Promise<ModifiedResponse<IntercomCreateCompanyData>> {
+  return request(`https://api.intercom.io/companies`, {
+    method: 'POST',
+    json: payload
+  })
+}
+
+function attachContactToIntercomCompany(request: RequestClient, contactId: string, companyId: string) {
```
brennan, line 133: What if `contactId` is undefined? `contactId` isn't a required property.

brennan:
@NolanChan This PR is pretty complicated, and I left a good number of questions/comments. I think it's worth going through and documenting manual test cases for all the functionality. This helps validate that everything is working as expected. Maryam recently did something similar for GA4: https://paper.dropbox.com/doc/Maryam-GA4-Testing-oEvVvYlc2u1ildRMlIGPt

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: You know what, I checked the [forums this time and I believe you're right](https://forum.intercom.com/s/question/0D52G00004tRJn2SAG/is-it-possible-to-change-a-contacts-externalid-via-the-api). Unfortunate they don't put this on the API docs :\

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
ibbykhazanchi: in the [classic destination](https://segment.com/docs/connections/destinations/catalog/intercom/#group), segment's `monthly_spend` defaults to intercom's `monthlySpend`

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
ibbykhazanchi: I believe the `sentAt` attribute or `timestamp` attribute refers to when the event took place.

default `createdAt` looks consistent with the [spec ](https://segment.com/docs/connections/spec/group/#traits)for group

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
ibbykhazanchi: should you default `custom_attributes` to `.traits`?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+        '@path': '$.traits.name'
+      }
+    },
+    avatar: {
+      label: 'Avatar',
+      description: 'URL of image to be associated with contact profile.',
+      type: 'string',
+      format: 'uri',
+      default: {
+        '@path': '$.traits.avatar'
```
ibbykhazanchi, line 66: so on [segment's side](https://segment.com/docs/connections/spec/identify/#traits), `avatar` is a string. However, on the [intercom side](https://developers.intercom.com/installing-intercom/docs/javascript-api-attributes-objects#avatar-object), `avatar` is an object that looks like this:

```
>   avatar: {
>     "type": "avatar", 
>     "image_url" :"https://yourwebsite.com/user_id/profile.png"
>   }
```
I think defaulting `avatar` to `.traits.avatar` may not work. Perhaps make `avatar` an object and have `image_url` default to `.traits.avatar`.

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
kdaswani: You could have the second field be `context.traits.email` since `track` calls can house `traits` within the `context` object.

```suggestion
          else: { '@path': '$.context.traits.email' }
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
kdaswani: ^ Agreed. I think we should remove any default for contact creation time as we don't know that this was their first identify call. The contact could already exist and we would then be overriding their true creation date.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: I'm not sure I understand what "platform" is for. I don't think we included this in every Actions destination. Is it an Intercom requirement, or an Actions framework requirement?

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        '@path': '$.groupId'
+      }
+    },
+    contact_id: {
+      type: 'string',
+      description:
+        'Attach this contact to the company. This ID is NOT the external_id or email; it is the Intercom unique identifier.',
+      label: 'Contact ID',
+      default: {
+        '@path': '$.userId'
```
kdaswani, line 39: I'm not sure this is the right default. You are setting `external_id` to `$.userId` in the identifyUser action. This says the `contact_id` is NOT the same as the `external_id` so setting this to `$.userId` seems incorrect. The description I found in Intercom docs for `contact_id` is "The unique identifier for the contact which is given by Intercom." Since this field is set by Intercom, we should not set a default and leave it blank.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+import type { Payload } from './generated-types'
+
+interface IntercomCreateCompanyData {
+  id: string
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Group Identify Contact',
+  description:
+    "Create or Update A Company. Note: Companies will be only visible in Intercom's Dashboard when there is at least one associated contact.",
```
kdaswani, line 12:
```suggestion
    "Create or update a company in Intercom.",
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
+import { ActionDefinition, ModifiedResponse, RequestClient, RetryableError } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+interface IntercomCreateCompanyData {
+  id: string
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Group Identify Contact',
```
kdaswani, line 10:
```suggestion
  title: 'Identify Company',
```
Open to other suggestions. Intercom doesn't use the language "group identify" so I think that would be a confusing way to explain this action.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+  description:
+    "Create or Update A Company. Note: Companies will be only visible in Intercom's Dashboard when there is at least one associated contact.",
+  defaultSubscription: 'type = "group"',
+  platform: 'web',
+  fields: {
+    remote_created_at: {
+      type: 'datetime',
+      description: 'The time the company was created by you.',
+      label: 'Remote Created At',
```
kdaswani, line 19:
```suggestion
      label: 'Company Creation Time',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      label: 'Company ID',
+      required: true,
+      default: {
+        '@path': '$.groupId'
+      }
+    },
+    contact_id: {
+      type: 'string',
+      description:
+        'Attach this contact to the company. This ID is NOT the external_id or email; it is the Intercom unique identifier.',
```
kdaswani, line 36:
```suggestion
        'The unique identifier for the contact which is given by Intercom. Setting a Contact ID will attach this contact to the company.',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: I would suggest removing a default here given the data type mismatch, and the fact that our spec doesn't technically have a field for monthly spend.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      type: 'number',
+      description: 'The monthly spend of the company.',
+      label: 'Monthly Spend',
+      default: {
+        '@path': '$.traits.plan'
+      }
+    },
+    plan: {
+      type: 'string',
+      description: 'The plan of the company.',
```
kdaswani, line 60:
```suggestion
      description: 'The name of the plan you have associated with the company.',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      label: 'Plan',
+      default: {
+        '@path': '$.traits.plan'
+      }
+    },
+    size: {
+      type: 'number',
+      description: 'The size of the company.',
```
kdaswani, line 68:
```suggestion
      description: 'The number of employees in this company.',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+        'Attach this contact to the company. This ID is NOT the external_id or email; it is the Intercom unique identifier.',
+      label: 'Contact ID',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    name: {
+      type: 'string',
+      description: 'The name of the company.',
+      label: 'Name',
```
kdaswani, line 45:
```suggestion
      label: 'Company Name',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      description: 'The monthly spend of the company.',
+      label: 'Monthly Spend',
+      default: {
+        '@path': '$.traits.plan'
+      }
+    },
+    plan: {
+      type: 'string',
+      description: 'The plan of the company.',
+      label: 'Plan',
```
kdaswani, line 61:
```suggestion
      label: 'Company Plan',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      default: {
+        '@path': '$.traits.plan'
+      }
+    },
+    size: {
+      type: 'number',
+      description: 'The size of the company.',
+      label: 'Size',
```
kdaswani, line 69:
```suggestion
      label: 'Company Size',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      default: {
+        '@path': '$.traits.employees'
+      }
+    },
+    website: {
+      type: 'string',
+      description: 'The website of the company.',
+      label: 'Website',
```
kdaswani, line 77:
```suggestion
      label: 'Company Website',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      type: 'number',
+      description: 'The size of the company.',
+      label: 'Size',
+      default: {
+        '@path': '$.traits.employees'
+      }
+    },
+    website: {
+      type: 'string',
+      description: 'The website of the company.',
```
kdaswani, line 76:
```suggestion
      description: 'The URL for this company's website.',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      label: 'Website',
+      default: {
+        '@path': '$.traits.website'
+      }
+    },
+    industry: {
+      type: 'string',
+      description: 'The industry of the company.',
```
kdaswani, line 84:
```suggestion
      description: 'The industry that this company operates in.',
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani:
```suggestion
        "A hash of key-value pairs containing any other data about the company you want Intercom to store.",
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
dlasky: the latter, it denotes PER ACTION if we intend to run on device mode (web) or cloud. The reason its per action is to support hybrid cases like session plugin.

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: I'm not sure what this meant either "Will throw an error if you pass an attribute that isn't explicitly defined." as the docs I'm reading imply you can pass any key-value as a custom attribute: https://developers.intercom.com/intercom-api-reference/reference/create-or-update-company.

My suggestion would be to set the default to `$.traits` and then since it is an object type, users will be able to toggle to key-value if they prepare to explicit set certain ones. It will look similar to Amplitude because type is object:
Screen Shot 2022-07-20 at 12 01 49 PM

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: ^ I agree. It seems like contact_id is set by Intercom upon contact creation. So if a contact is not found, then we should probably just skip trying to attach a contact to the company. I'm not sure the chances of the contact suddenly being created and the retry going through.

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  data: Array<IntercomContact>
+}
+
+interface IntercomContact {
+  id: string
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Identify Contact',
+  description: 'Create or Update a Contact.',
```
kdaswani, line 16:
```suggestion
  description: 'Create or update a contact in Intercom.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/__tests__/snapshot.test.ts
NolanChan: Happy to delete :) I think this was autogenerated?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: Woops, I think I was looking at a web example when I was writing out my identify - will change it!

packages/destination-actions/src/destinations/intercom/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  extendRequest({ auth }) {
+    return {
+      headers: {
+        authorization: `Bearer ${auth?.accessToken}`
+      }
+    }
+  },
+
+  onDelete: async (request, { payload }) => {
+    return request(`https://api.intercom.io/contacts/${payload.userId}`, {
```
kdaswani, line 30: We should not use `payload.userId` as the contact_id in the request path. The id is the unique identifier for the contact which is given by Intercom (seems to be autogenerated when a contact is first created). Intercom sets this value so it is very likely it is *not* equal to the userId in the Segment event: https://developers.intercom.com/intercom-api-reference/reference/delete-contact. What do we do in the classic version? We should probably reconsider how we approach delete functionality. cc @mericsson

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
NolanChan: It should be cloud! thanks for catching that

dlasky: i'm not aware of a top level prop there... can you link it?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: @ibbykhazanchi I believe that at least for Intercom's REST API, avatar is a string. Could you double check me there?
https://developers.intercom.com/intercom-api-reference/reference/create-contact

It's crazy that the API's are different like this :|

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: yes @dlasky: https://github.com/segmentio/action-destinations/pull/682/files#diff-e93882c54d0d493a4018e56d9aed588859fca7bd690b0d86478a3ee0618051b9R10. I've seen other integrations that don't have a `platform` set on the actions-level index.ts but they do have `mode` in the top-level index.ts and that seems to be enough?

For example, Salesforce (Actions): https://github.com/segmentio/action-destinations/tree/main/packages/destination-actions/src/destinations/salesforce

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
kdaswani:
```suggestion
  platform: 'cloud',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  defaultSubscription: 'type = "identify"',
+  platform: 'web',
+  fields: {
+    role: {
+      type: 'string',
+      required: true,
+      description: 'The role of the contact. Accepted values are `user` or `lead`. Can only be updated if `lead`.',
```
kdaswani, line 23:
```suggestion
      description: 'The role of the contact. Accepted values are `user` or `lead`.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Identify Contact',
+  description: 'Create or Update a Contact.',
+  defaultSubscription: 'type = "identify"',
+  platform: 'web',
+  fields: {
+    role: {
+      type: 'string',
```
kdaswani, line 21: I would if we should make this a picklist since we know only two values are accepted. For example:

```
choices: [
    { label: 'User', value: 'user' },
    { label: 'Lead', value: 'lead' }
  ]
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      required: true,
+      description: 'The role of the contact. Accepted values are `user` or `lead`. Can only be updated if `lead`.',
+      label: 'Role',
+      default: 'lead'
+    },
+    external_id: {
+      type: 'string',
+      description: 'A unique identifier generated outside Intercom. Required if role=user and email is blank.',
```
kdaswani, line 29:
```suggestion
      description: 'A unique identifier for the contact generated outside Intercom. External ID is required if the role is `user` and email is blank.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      label: 'External ID',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    email: {
+      type: 'string',
+      description: "The contact's email. Required if role=user and external_id is blank.",
```
kdaswani, line 37:
```suggestion
      description: "The contact's email address. Email is required if the role is `user` and external ID is blank.",
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      label: 'Email',
```
kdaswani, line 38:
```suggestion
      label: 'Email Address',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      type: 'string',
+      description: "The contact's name.",
+      label: 'Name',
+      default: {
+        '@path': '$.traits.name'
+      }
+    },
+    avatar: {
+      label: 'Avatar',
+      description: 'URL of image to be associated with contact profile.',
```
kdaswani, line 62:
```suggestion
      description: 'An image URL containing the avatar of a contact.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
kdaswani: As per Intercom server-side API docs, the data type for `avatar` is string which matches the Segment spec: https://developers.intercom.com/intercom-api-reference/reference/contacts-model. So I think this is fine!

@NolanChan what does `format: 'uri',` do?

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
dlasky: gotcha, sounds good, i'm curious to look at the types there at some point

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      type: 'string',
+      format: 'uri',
+      default: {
+        '@path': '$.traits.avatar'
+      }
+    },
+    signed_up_at: {
+      label: 'Signed Up At',
+      type: 'datetime',
+      description: 'The timestamp when the contact was created.',
```
kdaswani, line 72:
```suggestion
      description: 'The time specified for when a contact signed up.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      label: 'Avatar',
+      description: 'URL of image to be associated with contact profile.',
+      type: 'string',
+      format: 'uri',
+      default: {
+        '@path': '$.traits.avatar'
+      }
+    },
+    signed_up_at: {
+      label: 'Signed Up At',
```
kdaswani, line 70:
```suggestion
      label: 'Signed Up Timestamp',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
kdaswani: Also seems like this should be in UNIX timestamp. I see we clean up timestamps for trackEvent, but I don't see us checking that the identifyContact timestamps are unix strings?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      type: 'datetime',
+      description: 'The timestamp when the contact was created.',
+      default: {
+        '@path': '$.createdAt'
+      }
+    },
+    last_seen_at: {
+      label: 'Timestamp',
+      type: 'datetime',
+      description: 'The timestamp the contact was last seen.',
```
kdaswani, line 80:
```suggestion
      description: 'The time when the contact was last seen.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    signed_up_at: {
+      label: 'Signed Up At',
+      type: 'datetime',
+      description: 'The timestamp when the contact was created.',
+      default: {
+        '@path': '$.createdAt'
+      }
+    },
+    last_seen_at: {
+      label: 'Timestamp',
```
kdaswani, line 78:
```suggestion
      label: 'Last Seen Timestamp',
```

kdaswani: Also seems like this should be in UNIX timestamp. I see we clean up timestamps for trackEvent, but I don't see us checking that the identifyContact timestamps are unix ?

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      type: 'datetime',
+      description: 'The timestamp the contact was last seen.',
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    owner_id: {
+      label: 'Owner Id',
```
kdaswani, line 86:
```suggestion
      label: 'Owner ID',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+      type: 'number',
+      description: 'The id of an admin that has been assigned account ownership of the contact.'
```
kdaswani, line 88:
```suggestion
      description: 'The ID of an admin that has been assigned account ownership of the contact'
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    },
+    unsubscribed_from_emails: {
+      label: 'Unsubscribed From Emails',
+      type: 'boolean',
+      description: 'Whether the contact is unsubscribed from emails.'
+    },
+    custom_attributes: {
+      label: 'Custom Attributes',
+      description:
+        'The custom attributes which are set for the contact. Note: Will throw an error if the object has an attribute that isn`t explicitly defined on Intercom.',
```
kdaswani, line 98:
```suggestion
        'The custom attributes which are set for the contact.,
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
kdaswani: Correct, `traits.customAttributes` is not in the Segment spec so I suggest changing the default to the entire `$.traits` object. Since it is an object type, users will be able to toggle to key-value if they prefer to explicit set certain ones. It will look similar to Amplitude because type is object:

kdaswani:
```suggestion
        '@path': '$.traits'
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: Actually.. change of plans.  I read through Intercom docs and it seems like custom attributes need to be pre-created in Intercom first. Given this, I would suggest not setting a default at all. Instead, I think you should make the defaultUIView key-value so customers can input each key-value specifically. Would suggest adding this:

defaultObjectUI: 'keyvalue'

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Track Event',
+  description: 'Submit a data event to Intercom.',
+  fields: {
+    event_name: {
+      type: 'string',
+      required: true,
+      description:
+        'The name of the event. Names are treated as case insensitive. Periods (.) and dollars ($) in event names are replaced with hyphens.',
```
kdaswani, line 14:
```suggestion
        'The name of the event that occurred. Names are treated as case insensitive. Periods and dollar signs in event names are replaced with hyphens.',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import dayjs from '../../../lib/dayjs'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Track Event',
+  description: 'Submit a data event to Intercom.',
```
kdaswani, line 8: Should we add `defaultSubscription: 'type = "track"',`?

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
kdaswani, line 23:
```suggestion
      description: 'The time the event occurred as a UTC Unix timestamp',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      label: 'Event Name',
+      default: {
+        '@path': '$.event'
+      }
+    },
+    created_at: {
+      type: 'datetime',
+      required: true,
+      description: 'A datetime in Unix timestamp format (seconds since Epoch).',
+      label: 'Created At',
```
kdaswani, line 24:
```suggestion
      label: 'Event Timestamp',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    user_id: {
+      type: 'string',
+      description: "The user's ID; required if no email provided.",
```
kdaswani, line 31:
```suggestion
      description: "Your identifier for the user who performed the event. User ID is required if no email is provided.",
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      label: 'User ID',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    email: {
+      type: 'string',
+      description: "The user's email; required if no User ID provided.",
+      label: 'Email',
```
kdaswani, line 40:
```suggestion
      label: 'Email Address',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      type: 'string',
+      description: "The user's ID; required if no email provided.",
+      label: 'User ID',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    email: {
+      type: 'string',
+      description: "The user's email; required if no User ID provided.",
```
kdaswani, line 39:
```suggestion
      description: "The email address for the user who performed the event. Email is required if no User ID is provided.",
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+          exists: { '@path': '$.properties.email' },
+          then: { '@path': '$.properties.email' },
+          else: { '@path': '$.traits.email' }
+        }
+      }
+    },
+    metadata: {
+      type: 'object',
+      description:
+        'Metadata object describing the event. There is a limit to 10 keys. Intercom does not currently support nested JSON structures.',
```
kdaswani, line 53:
```suggestion
        'Optional metadata describing the event. Each event can contain up to ten metadata key-value pairs. If you send more than ten keys, Intercom will ignore the rest. Intercom does not support nested JSON structures within metadata.',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,88 @@
...
+      label: 'Metadata',
```
kdaswani, line 54:
```suggestion
      label: 'Event Metadata',
```

kdaswani:
Added a bunch of suggestions and comments - thanks @NolanChan for building this! One last question - I see that for submitting events, you can also include an `id` field (instead of user ID or email): https://developers.intercom.com/intercom-api-reference/reference/event-model. Seems like this is to support sending events for leads. My guess is that `id` = the `contact_id` that Intercom sets, but curious if you know more? We don't currently support mapping to `id` but that may be intentional.

Saw the note about Lead Events here: https://developers.intercom.com/intercom-api-reference/reference/submitting-events

Screen Shot 2022-07-20 at 1 23 45 PM

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
ibbykhazanchi: @NolanChan You're totally right!

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
NolanChan: I think that was an internal note, I was trying to cast the error as an HTTPError, but it wasn't going through. I'll remove the note since it's kinda random

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      type: 'string',
+      description: 'The name of the company.',
+      label: 'Name',
+      default: {
+        '@path': '$.traits.name'
+      }
+    },
+    monthly_spend: {
+      type: 'number',
+      description: 'The monthly spend of the company.',
```
kdaswani, line 52:
```suggestion
      description: 'The monthly spend of the company, e.g. how much revenue the company generates for your business.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/__tests__/index.test.ts
NolanChan: woah, i didn't even know that this was up for debate - https://english.stackexchange.com/questions/305273/retriable-or-retryable haha ... although I think from this post, it seems that both are ok?

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
NolanChan: Some context, I copied it over from CustomerIo; I'm of the opinion that since it's already been production tested, I don't think we need to? :)

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: Talked with Marin, resolving this comment :)

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
NolanChan: I'll go about removing the `platform` attributes then!

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: Actually based on Kiara's comments below, I'll just remove the platform attribute since the top level attribute has it

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: @kdaswani seems that track threw errors when the unix timestamp wasn't passed, but it was silently ignored on identify! Will fix this now, thanks for the catch!

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
kdaswani, line 8:
```suggestion
  description: 'Submit an event to Intercom.',
```

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: Changed it to not use it!

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
kdaswani: To keep parity with the classic destination let's make the default:
```suggestion
        '@path': '$.traits.monthly_spend'
```

NolanChan: Talked with @kdaswani and we'll be aligning with the classic docs - we'll make the default become `traits.monthly_spend`

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+    },
+    owner_id: {
+      label: 'Owner Id',
+      type: 'number',
+      description: 'The id of an admin that has been assigned account ownership of the contact.'
+    },
+    unsubscribed_from_emails: {
+      label: 'Unsubscribed From Emails',
+      type: 'boolean',
+      description: 'Whether the contact is unsubscribed from emails.'
```
kdaswani, line 93:
```suggestion
      description: "The contact's email unsubscribe status."
```

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
NolanChan: that's a great observation @brennan , I agree with you all. Taking it out!

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: @kdaswani If I'm understanding the code correctly, I believe [it requires that whatever attribute is entered into there is a full URL](https://github.com/segmentio/action-destinations#input-field-interface) (for example, no relative urls) . 

Braze [uses it for their Image_url](https://github.com/segmentio/action-destinations/blob/8c89af1060b852f4532d5c4c34bf56896d0a79a9/packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts#L119) and [friendbuy users it for their Page URL](https://github.com/segmentio/action-destinations/blob/8c89af1060b852f4532d5c4c34bf56896d0a79a9/packages/actions-shared/src/friendbuy/contextFields.ts#L12)

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
NolanChan: Thank you for bringing this up - agreed, I was hesitating between setting a default at all or not. I'll remove the default!

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/index.ts
NolanChan: sounds good to me!

packages/destination-actions/src/destinations/intercom/index.ts
NolanChan: @kdaswani [seems like we do use the `userId` in the delete method in classic](https://github.com/segmentio/integrations/blob/master/integrations/intercom/lib/mapper-delete.js#L9-L13). I totally understand that the terms are starting to get unwieldy (`contactId` != `userId`), open to thinking about this more. Looking at most `onDelete` methods, it seems that they utilize `userId`, so I'm happy just to say that this is the precedent more than a strict term

packages/destination-actions/src/destinations/intercom/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    testAuthentication: (request) => {
+      // Uses the admin route as a stand-in to test authentication creds.
+      // Should be a light request since there most likely won't be many admins
+      return request('https://api.intercom.io/admins')
+    }
+  },
+  extendRequest({ auth }) {
+    return {
+      headers: {
+        authorization: `Bearer ${auth?.accessToken}`
```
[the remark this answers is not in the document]
NolanChan: @marinhero pointed out that [Intercom has no refresh token logic](https://github.com/intercom/oauth2-intercom#refreshing-a-token), so the initial OAuth handshake giving us an accessToken is all the functionality that we can leverage

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: awesome, thanks for pointing this out!

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,187 @@
...
+  description: 'Create or Update a Contact.',
+  defaultSubscription: 'type = "identify"',
+  platform: 'web',
+  fields: {
+    role: {
+      type: 'string',
+      required: true,
+      description: 'The role of the contact. Accepted values are `user` or `lead`. Can only be updated if `lead`.',
+      label: 'Role',
+      default: 'lead'
```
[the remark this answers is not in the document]
NolanChan: Hmm, that's a good question. Niel's designated [that on the PRD](https://paper.dropbox.com/doc/Intercom-Actions-Cloud-Mode-PRD--BlxxtkdKY1euHxin43Tz1QbdAg-nMEizRG8DkHXTEp2fJQjr), but honestly i feel more comfortable if we force the user to choose, and that `choices` functionality makes it very clear

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: will do!

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
NolanChan: Correct!

packages/destination-actions/src/destinations/intercom/index.ts
```diff
@@ -0,0 +1,42 @@
+import type { DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+import identifyContact from './identifyContact'
+import groupIdentifyContact from './groupIdentifyContact'
+import trackEvent from './trackEvent'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Intercom (Actions)',
```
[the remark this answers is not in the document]
NolanChan: I'll need to go and delete the old one and make a new one, but ultimately that shouldn't be too bad! It'll be good to so Ibby and mine can be a pair

packages/destination-actions/src/destinations/intercom/index.ts
dlasky: we only have userId in delete that's all that's provided in the entire system

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
kdaswani: chatted with @NolanChan and we will leave the default as just `properties.traits`  (no coalesce)

kdaswani:
> @NolanChan Generally, how will this destination be used with RETL with no `performBatch` handlers defined?

Good question @brennan! I don't think it's a hard requirement to support batching for destinations that will be compatible with RETL. If Intercom's streaming API can handle a lot of data, should be fine? I believe there are other Actions destinations that do not have batch support but will probably end up being used by RETL (Braze,  Customerio, Conversions API destinations). Maybe in the future this will become necessary but for v1, I don't think we need to support any sort of batching API unless it's a simple thing to add.

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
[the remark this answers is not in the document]
kdaswani: I agree! We should make it required with choices of `lead` and `role` but no default.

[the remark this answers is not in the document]
kdaswani: Maybe Niels' didn't know we could define `choices` 😄

packages/destination-actions/src/destinations/intercom/index.ts
kdaswani: Got it! Thanks @dlasky @NolanChan. The problem is that the `id` in the request path for `delete` is assigned by Intercom so it is very unlikely this will be the same value as the `userId` set by the customer. This will result in most (all?) `delete` requests failing / not doing anything. I presume we will almost always get a `404 not found` error.

Screen Shot 2022-07-21 at 8 50 04 PM

https://developers.intercom.com/intercom-api-reference/reference/delete-contact

Not sure how or if this worked in the classic version, but I think before we add this in and say we support `delete` we should think a bit more about how we can correctly support deletions.  For example, could we [search for the contact](https://developers.intercom.com/intercom-api-reference/reference/search-for-contacts) using the `userId` - which should typically be set as the `external_id` in Intercom - and grab the Intercom contact `id` from there to process the deletion?

dlasky: Yeah thanks for the context there -- I did understand that discrepancy was just noting that that's all we get from the initial user submission. I think your approach could work -- question is do we block this on that or just "delete" the method for now. (Horrible pun intended)

kdaswani: Yeah I'm torn on what to do. Personally, I think we should try to get `delete` support in here before releasing to a Private Beta as our classic destination for Intercom supports `delete`. That said, I don't want to block this if there's a timeline we are trying to hit. Will let @NolanChan weigh in on if the approach I suggested could work and how long it would take.

NolanChan: Searching for the user if the delete fails to go through sounds like a good compromise in getting this to work. And I don't believe there is any real deadline besides Q3? (I'm also slightly confused about the deadline as well)

packages/destination-actions/src/destinations/index.ts
NolanChan: Will most likely be registering a new action-destination since we need to change the name, so we'll need to check this again

packages/destination-actions/src/destinations/intercom/groupIdentifyContact/__tests__/index.test.ts
NolanChan: removed since we're no longer retrying

packages/destination-actions/src/destinations/intercom/__tests__/index.test.ts
NolanChan: done!

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
NolanChan: Will be talking today about whether or not we want to default a choice

packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -0,0 +1,185 @@
...
+ * Note: contact leads can have duplicate emails (and so the creation can never throw a 409),
+ * but contact users can't.
+ *
+ * Intercom's API Docs - https://developers.intercom.com/intercom-api-reference/reference/search-for-contacts
+ */
+async function searchIntercomContact(request: RequestClient, payload: Payload) {
+  const searchFields = {
+    email: payload.email,
+    external_id: payload.external_id,
+    role: payload.role
```
kdaswani, line 150: remove role as search query to unblock being able to convert leads --> users

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      type: 'string',
+      description:
+        'The email address for the user who performed the event. Email is required if no User ID is provided.',
+      label: 'Email Address',
+      format: 'email',
+      default: {
+        '@path': '$.properties.email'
+      }
+    },
+    id: {
```
kdaswani, line 49: Remove `id` based on discussion. And add fields for `Revenue` and `Currency` as per this comment: https://github.com/segmentio/action-destinations/pull/626/files/746ff716635d9b51764c62f090958786c16d58ed#r927995415

NolanChan: After thinking about this, could we at least allow the customer the option to pass it in?

kdaswani: Yeah I think that's fine! Let's be clear in the description that this is the ID Intercom sets. Let's use this:

- label: `Contact ID`
- description: `Intercom's unique identifier for the contact. If no Contact ID is provided, Segment will use User ID or Email to find a user.` 
- no default set

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      description:
+        'The time the event occurred as a UTC Unix timestamp. Segment will convert to Unix if not already converted.',
+      label: 'Event Timestamp',
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    user_id: {
+      type: 'string',
+      description: 'Your identifier for the user who performed the event. User ID is required if no email is provided.',
```
kdaswani, line 33:
```suggestion
      description: 'Your identifier for the user who performed the event. User ID is required if no email or Contact ID is provided.',
```

packages/destination-actions/src/destinations/intercom/trackEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      label: 'User ID',
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    email: {
+      type: 'string',
+      description:
+        'The email address for the user who performed the event. Email is required if no User ID is provided.',
```
kdaswani, line 42:
```suggestion
        'The email address for the user who performed the event. Email is required if no User ID or Contact ID is provided.',
```

packages/destination-actions/src/destinations/index.ts
NolanChan: Ibby updated it!

packages/destination-actions/src/destinations/intercom/index.ts
NolanChan: @kdaswani and I talked and agreed that this path will only allow customers to delete users, not leads. Closing this as functionally resolved, will release this in a new PR!

NolanChan:
Closing this PR so that I can open up a non-messy new one

PR #456, opened 2022-02-08
add new field to postConversion GEC

rvadera12:
_A summary of your pull request, including the what change you're making and why._

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [ ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [ ] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

PR #688, opened 2022-07-19
[STRATCONN-1509] add new field pcc_games to GEC

rvadera12:
_This PR adds a new field in the `Post Conversion` action for Google Enhanced Conversions, called `PCC Game Flag`. This is not a required field however when set to true will include `pcc_game: 1` parameter in the request from Segment._ 

We are planning to only expose this setting to one customer with a [flagon](https://flagon.segment.com/families/destination-actions/gates/gec-pcc-games-field). 

Here is the app PR with the flagon code: https://github.com/segmentio/app/pull/13664

Note: The app PR will be merged first to ensure only workspaces added to the flagon will see this new field in the mappings. 

## Testing

Tested in stage via deploying to stage. [Testing Doc](https://paper.dropbox.com/doc/GEC-pcc_games-field-Testing--BlvHf5SCZFyu0dK6Rvj8tv8vAg-Sct7lxT58TYtujIkJWgLb)

- [ x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ x] [Segmenters] Tested in the staging environment

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/__tests__/google-enhanced-conversions.test.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/__tests__/google-enhanced-conversions.test.ts
@@ -105,6 +105,55 @@ describe('GoogleEnhancedConversions', () => {
       expect(responses.length).toBe(1)
       expect(responses[0].status).toBe(201)
     })
+    it('should send pcc_game:1 when PCC Game is set to true', async () => {
+      const event = createTestEvent({
+        timestamp,
+        event: 'Test Event',
+        properties: {
+          email: 'janedoe@gmail.com',
+          orderId: '123',
+          firstName: 'Bob John',
+          lastName: 'Smith',
+          phone: '14150000000',
+          address: {
+            street: '123 Market Street',
+            city: 'San Francisco',
+            state: 'CA',
+            postalCode: '94000',
+            country: 'USA'
+          }
+        }
+      })
+
+      nock('https://www.google.com/ads/event/api/v1')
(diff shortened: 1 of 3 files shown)


nick-Ag (approved):
Just wondering if you confirmed that these events arrive downstream as expected? Didn't see that in the testing doc. Only asking because I'm not sure if GEC is similar to GA4 in that it will always return 200. But looks great overall

rhall-twilio (approved):

rhall-twilio:
Probably want to shy away from using `/` for branch names because it breaks the image if you ever wanted to deploy the branch.

rvadera12:
> Just wondering if you confirmed that these events arrive downstream as expected? Didn't see that in the testing doc. Only asking because I'm not sure if GEC is similar to GA4 in that it will always return 200. But looks great overall

@nick-Ag I talked to @kdaswani and we don't have the ability to test end to end, however since this is an alpha field that will only be exposed through flagon & this is implemented the same way as another field (https://github.com/segmentio/action-destinations/pull/456) we are confident moving forward with the testing completed. We are also planning to have google validate on their end.

kdaswani:
@nick-Ag I also confirmed that when I am authenticated into our test Google Ads accounts and tied to a real conversion, I get a 200 success and the pcc_game = 1 is sent. The GEC API does not always return a 200 - it will throw a 400 if there's an error. So this should be enough to proceed, especially given this is behind a flagon and the customer is also going to test with their QA/dev source and have Google confirm things look good on their end before we flagon for their prod source. Thanks all!

PR #690, opened 2022-07-20
[Customer.io] Add support to send an event id for track event calls

jatinn:
The [customer.io track event api call](https://customer.io/docs/api/#operation/track) has support to pass an event id that may be used to deduplicate events.
This change makes that field available to the customer.io destination.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
+++ b/packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
@@ -47,7 +47,7 @@ const action: ActionDefinition<Settings, Payload> = {
     },
     event_id: {
       label: 'Event ID',
-      description: 'An optional identifier used to deduplicate events. [Learn more](https://customer.io/docs/api/#operation/track)',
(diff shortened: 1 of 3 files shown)


jatinn:
converted to draft while looking into if we can use the `meesageId` as the default value instead of requiring `properties.id`

packages/destination-actions/src/destinations/customerio/trackEvent/index.ts
```diff
@@ -44,6 +45,14 @@ const action: ActionDefinition<Settings, Payload> = {
         '@path': '$.event'
       }
     },
+    event_id: {
+      label: 'Event ID',
+      description: 'An optional identifier used to deduplicate events. [Learn more](https://customer.io/docs/api/#operation/track)',
```
marinhero, line 50:
```suggestion
      description: 'An optional identifier used to deduplicate events. [Learn more](https://customer.io/docs/api/#operation/track).',
```

marinhero (approved):

PR #697, opened 2022-07-25
Local Server - Run dynamicFields

nick-Ag:
This PR adds routing to dynamicFields in the local actions server. Any dynamicField defined on an action will have the route `/<action>/<dynamicFieldName>`. It also adds an `Auth` object as an input to the dynamicFields.

## Testing

Tested with a test field pointed to a cloudflare worker. Also tested against a sample Salesforce dynamic field which used the auth object being passed.

Using a passed in Auth object, running a request against the Salesforce REST API with this code: https://github.com/segmentio/action-destinations/blob/test-server-fields/packages/destination-actions/src/destinations/salesforce/customObject/index.ts#L52
Screen Shot 2022-07-25 at 12 39 30 PM

Running a request against a Cloudflare worker with this code: https://github.com/segmentio/action-destinations/blob/test-server-fields/packages/destination-actions/src/destinations/salesforce/customObject/index.ts#L42
Screen Shot 2022-07-25 at 12 28 21 PM

### DIFF
--- a/packages/cli/src/lib/server.ts
+++ b/packages/cli/src/lib/server.ts
@@ -256,11 +256,9 @@ function setupRoutes(def: DestinationDefinition | null): void {
                 auth: req.body.auth || {}
               }
               const action = destination.actions[actionSlug]
+              const result = await action.executeDynamicField(field, data)
 
-              await action.executeDynamicField(field, data)
-
-              const debug = await getExchanges(destination.responses)
-              return res.status(200).json(debug)
+              return res.status(200).json(result)
             } catch (err) {
               const output = marshalError(err as ResponseError)
               return res.status(200).json([output])
--- a/packages/cli/src/lib/server.ts
+++ b/packages/cli/src/lib/server.ts
@@ -52,11 +52,8 @@ const marshalError = (err: Error): ErrorOutput => {
 }
 
 const app = express()
-app.use(
-  express.json({
-    limit: '300mb'
-  })
-)
+app.use(express.json())
+
 app.use(
   cors({
     origin: [
--- a/packages/cli/src/lib/server.ts
+++ b/packages/cli/src/lib/server.ts
@@ -52,7 +52,11 @@ const marshalError = (err: Error): ErrorOutput => {
 }
 
 const app = express()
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/google-sheets/postSheet/generated-types.ts
```diff
@@ -23,11 +23,13 @@ export interface Payload {
   data_format: string
   /**
    *
-   *       The fields to write to the spreadsheet.
-   *       On the left-hand side, input the name of the field as it will appear in the Google Sheet.
-   *       On the right-hand side, select the field from your data model that maps to the given field in your sheet.
+   *   The fields to write to the spreadsheet.
```
nick-Ag, line 26: Not related, but it's autogenerating. Will remove if anyone thinks this should go in a separate PR

packages/cli/src/lib/server.ts
```diff
@@ -239,6 +240,35 @@ function setupRoutes(def: DestinationDefinition | null): void {
     )
   }
 
+  for (const actionSlug of Object.keys(destination.actions)) {
+    const definition = destination.actions[actionSlug].definition
+
+    if (definition.dynamicFields) {
+      for (const field in definition.dynamicFields) {
+        router.post(
+          `/${actionSlug}/${field}`,
```
dlasky, line 249: we should probably check to make sure we arent stepping on any partner action routes here

dlasky (approved):
looks good, 1 comment fine to address in a followon, but we should probably throw some sort of user visible error if they have stepped on themselves with naming.

you may also want to add a check for that to the validate command (in cli package) if this is the pathing we will be using on fab5 as well?

packages/cli/src/lib/server.ts
nick-Ag: I checked how the routing works for pipedrive and their routes look good, they have all the original routes + the new dynamic fields ones as well.

nick-Ag: Screen Shot 2022-07-27 at 9 46 39 AM

PR #699, opened 2022-07-25
Register Pardot (Actions)

kwang122:

## Testing


marinhero (changes requested):
Move this to compare against the main branch and not the stage one

packages/destination-actions/src/destinations/pardot/prospects/index.ts
```diff
@@ -0,0 +1,204 @@
+import { ActionDefinition, IntegrationError, HTTPError } from '@segment/actions-core'
+import type { ModifiedResponse } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import Pardot from '../pa-operations'
+import { customFields } from '../pa-properties'
+import type { Payload } from './generated-types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Prospect',
```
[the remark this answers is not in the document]
kdaswani: @kwang122 We should also add a default subscription of `identify`.

`defaultSubscription: 'type = "identify"',`

Here's an example in Salesforce (Actions): https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/salesforce/lead/index.ts#L20

packages/destination-actions/src/destinations/pardot/prospects/index.ts
```diff
@@ -0,0 +1,204 @@
...
+  description: 'Represents an opportunity, which is a sale or pending deal.',
+  fields: {
+    email: {
+      label: 'Email Address',
+      description: `The prospect's email address.   
+        Used to upsert prospect.  If multiple prospects have the given email, the prospect with the latest activity is updated. If there's no prospect with the given email, a prospect is created. Please note that Pardot treats email address as case sensitive and will create multiple prospects for casing differences.`,
```
kdaswani, line 15:
```suggestion
        Used to upsert a prospect in Pardot. If multiple prospects have the given email, the prospect with the latest activity is updated. If there's no prospect with the given email, a prospect is created. Please note that Pardot treats email address as case sensitive and will create multiple prospects for casing differences.`,
```

packages/destination-actions/src/destinations/pardot/pa-properties.ts
```diff
@@ -0,0 +1,11 @@
+import { InputField } from '@segment/actions-core/src/destination-kit/types'
+
+export const customFields: InputField = {
```
kdaswani, line 3: @kwang122 Since we only have one action, I wonder if we should just put this in the index.ts file for `upsert prospect`? Is there a reason we have it separated out? I know we do that for Salesforce (Actions), but that's because there are multiple actions to use the same field.

kdaswani: cc @marinhero in case you have a preference one way or the other. It works this way but might be cleaner to not have it separated out?

packages/destination-actions/src/destinations/pardot/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        label: 'Account ID',
+        description:
+          'You can find your Account ID (`piAId`) under **Marketing > Campaigns** in your [Pardot account](https://pi.pardot.com/campaign). After selecting your desired website campaign, press **View Tracking Code**. ',
+        type: 'string',
+        required: true
+      },
+      isSandbox: {
+        label: 'Sandbox Instance',
+        description:
+          'Enable to authenticate into a sandbox instance. You can log in to a sandbox by appending the sandbox name to your Salesforce username. For example, if a username for a production org is user@acme.com and the sandbox is named `test`, the username to log in to the sandbox is user@acme.com.test. If you are already authenticated, please disconnect and reconnect with your sandbox username.',
```
kdaswani, line 35: I think these were the instructions for Salesforce Sales Cloud sandboxes. Do we know it's the same instructions for Pardot sandboxes? 

I know our test account is a bit funky since it's a dev account with prod oauth and a sandbox api endpoint 😬  so it's probably not a good example of how to truly auth with a sandbox account.

kdaswani (changes requested):
great job @kwang122! added some small comments, but overall LGTM. 🎉

packages/destination-actions/src/destinations/pardot/index.ts
kdaswani: biggest thing to confirm - have we been able to test with a "true" sandbox account?

kwang122: It's the same instruction for Pardot since they both use the same Salesforce login endpoint. [Salesforce Sandbox Login](https://help.salesforce.com/s/articleView?id=sf.data_sandbox_create.htm&type=5) [Pardot Sandbox Instruction](https://help.salesforce.com/s/articleView?id=sf.pardot_sf_connector_sandbox.htm&type=5)

Unfortunately, we can not test with a "true" Sandbox account. Dev account does not allow the conversion to either Sandbox or Prod edition. Confirmed that Sandbox option does not exist in our Salesforce account.

packages/destination-actions/src/destinations/pardot/prospects/index.ts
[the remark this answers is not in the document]
kwang122: That make sense. Just added!

packages/destination-actions/src/destinations/pardot/pa-properties.ts
kwang122: I was thinking maybe we'll add more actions in the future. It would be good to add some functionalities in the future.

kdaswani: Makes sense! Good thinking. Let's leave it like this in case we do add more actions in the future 😄

packages/destination-actions/src/destinations/pardot/index.ts
kdaswani: sounds good! We can always rely on an early beta tester to confirm the sandbox functionality is working correctly.

packages/destination-actions/src/destinations/pardot/prospects/index.ts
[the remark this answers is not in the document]
kdaswani: @kwang122 Can we also update the name of this action to `Upsert Prospect`?

```suggestion
  title: 'Upsert Prospect',
```

PR #700, opened 2022-07-26
[Actions-Salesforce] Dynamic field for the customObjectName field

nick-Ag:
This PR includes:
1. Changes to `actions-core` which defines and handles a new `DynamicFieldError` type, which is used to pass error information back to app.
2. Changes to the local server to display the `DynamicFieldError` type if needed.
3. A dynamic field implementation for the Salesforce Custom Object action's name field. This dynamic field queries a users Salesforce account to provide them a list of Salesforce Objects which can be used as an input.

## Testing

Tested successfully in staging with the other PRs above.

Successful lookup of Salesforce Object names (there are >100 of these and scrolling/performance looks good):
Screen Shot 2022-09-28 at 5 06 07 PM

Lookup error triggered by stale authorization:
Screen Shot 2022-09-28 at 4 45 38 PM


LLuque-twilio (approved):
Wow - lgtm 🪄

packages/core/src/destination-kit/types.ts
```diff
@@ -37,6 +37,12 @@ export interface ExecuteInput<Settings, Payload> {
 export interface DynamicFieldResponse {
   choices: DynamicFieldItem[]
   nextPage?: string
+  error?: DynamicFieldError
+}
+
+export interface DynamicFieldError {
```
brennan, line 43: Nit: Does this need to be exported?

nick-Ag: I think there's a good case to leave it exported incase builders need to use it as a type. I'm creating an object of this type [here](https://github.com/segmentio/action-destinations/pull/700/files/5060509f6b4eb8da8eafc76bb24fbc53e3f94c3c#diff-ce64a4366ba2a7982f3dffe832f75953866111b5d83fb6f18fec268635f6787cR148) but I could have done something like this if needed:

```js
const sfError: DynamicFieldError = {
  code: `Err`,
  message: `Err`
}
```

PR #702, opened 2022-07-27
Rename new Heap destination

marinhero:
I'm renaming the new Heap destination from `Heap` to `Heap Cloud (Actions)` because there's an existing destination with that name. `Heap Cloud (Actions)` matches the current naming convention we and Heap have for `Heap Web (Actions)`.

Additionally, the linter detected that some type files from the new Qualtrics destination weren't regenerated so I'm adding them here. They only contain updates in comments.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/heap/index.ts
+++ b/packages/destination-actions/src/destinations/heap/index.ts
@@ -22,7 +22,7 @@ const presets: DestinationDefinition['presets'] = [
 
 const destination: DestinationDefinition<Settings> = {
   name: 'Heap Cloud (Actions)',
-  slug: 'actions-heap',
+  slug: 'actions-heap-cloud',
   mode: 'cloud',
   authentication: {
     scheme: 'custom',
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/heap/index.ts
```diff
@@ -21,7 +21,7 @@ const presets: DestinationDefinition['presets'] = [
 ]
 
 const destination: DestinationDefinition<Settings> = {
-  name: 'Heap',
+  name: 'Heap Cloud (Actions)',
   slug: 'actions-heap',
```
jinapark202, line 25: Is this a new destination? If so, should we also make the slug `actions-heap-cloud` or something so that it's more clear?

jinapark202 (approved):
Small comment but otherwise LGTM

packages/destination-actions/src/destinations/heap/index.ts
marinhero: Good idea. Let me do that as well

PR #703, opened 2022-07-27
Register Heap and Qualtrics as new destinations

marinhero:
Register new Heap and Qualtrics destinations.

## Testing

Testing not required because I'm only registering destinations.


nick-Ag (approved):

jinapark202 (approved):
Verified id, LGTM

PR #706, opened 2022-07-29
default enable_batching to true for Google Sheets

rhall-twilio:
This changes the implicit enable_batching field to be explicit and defaulted to `true` for Google Sheets

## Testing

Tested that changes looked okay in local server. They didn't. For some reason the `enable_batching` field shows at the top for action-tester. 
image

Confirmed that in staging is in the right place.
image

Used event:
```
{
    "type": "track",
    "event": "new",
    "receivedAt": "2022-06-10T21:09:23.487895578Z",
    "messageId": "message test",
    "channel": "server",
    "properties": {
      "P_BRAND": "Brand#21",
      "P_COMMENT": "nts sleep dog",
      "P_CONTAINER": "WRAP BAG",
      "P_MFGR": "Manufacturer#2",
      "P_NAME": "orange peru yellow blanched seashell",
      "P_PARTKEY": 399,
      "P_RETAILPRICE": 1299.39,
      "P_SIZE": 37,
      "P_TYPE": "LARGE BRUSHED BRASS"
    },
    "__segment_id": "399"
  }
```
against destination https://app.segment.build/rhall-test-workspace/destinations/actions-google-sheets/sources/javascript/instances/62b1e9d3063992217d4db16d/event-tester to confirm there were no regressions.

### DIFF
--- a/packages/destination-actions/src/destinations/google-sheets/postSheet/generated-types.ts
+++ b/packages/destination-actions/src/destinations/google-sheets/postSheet/generated-types.ts
@@ -36,4 +36,8 @@ export interface Payload {
   fields: {
     [k: string]: unknown
   }
+  /**
+   * Set as true to ensure Segment sends data to Google Sheets in batches. Please do not set to false.
+   */
+  enable_batching?: boolean
 }
--- a/packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
+++ b/packages/destination-actions/src/destinations/google-sheets/postSheet/index.ts
@@ -68,6 +68,12 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'object',
       required: true,
       defaultObjectUI: 'keyvalue:only'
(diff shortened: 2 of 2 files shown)


nick-Ag (approved):
This looks good to me 👍 non-blocking: I recall discussion of hiding this field completely such that Google Sheets would always be a batching destination. Is that still a possibility or did we decide otherwise?

rhall-twilio:
> This looks good to me 👍 non-blocking: I recall discussion of hiding this field completely such that Google Sheets would always be a batching destination. Is that still a possibility or did we decide otherwise?

Let me double check before merging. Might take it in with this one if possible. I'm not sure if it's only for strings or booleans can benefit from it too.

dlasky:
Unfortunately only strings due to how we implemented it as a "type" rather than add an additional field property. The type is converted to a string for validation purposes in the AJV schema generation phase.

PR #714, opened 2022-08-04
core: Add 'mappingRoot' option to support mapping against a nested field

tysonmote:
This PR is a WIP proposal for allowing custom mapping roots. This option would allow callers to specify that mappings should be done against a sub-object of the event, like `properties`. This supports use cases that use Segment events as a wrapper and in which we'd like to hide the event format implementation detail from people configuring mappings.

## Testing

### DIFF
--- a/packages/core/src/create-test-integration.ts
+++ b/packages/core/src/create-test-integration.ts
@@ -20,6 +20,10 @@ interface InputData<Settings> {
    */
   mapping?: JSONObject
   /**
+   * Optional path that mappings should be evaluated against.
+   */
+  mappingRoot?: string
+  /**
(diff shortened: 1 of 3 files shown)


packages/core/src/create-test-integration.ts
```diff
@@ -19,6 +19,10 @@ interface InputData<Settings> {
    * literal values as well as mapping-kit directives.
    */
   mapping?: JSONObject
+  /**
+   * Optional path that mappings should be evaluated against.
+   */
+  mappingRoot?: string
```
dlasky, line 25: would mappingRoot ultimately be stored in CPS as well?

PR #724, opened 2022-08-10
fix name + description in update

dlasky:
fixes an issue where name and description were not properly initialized

## Testing

### DIFF
--- a/packages/cli-internal/src/commands/push.ts
+++ b/packages/cli-internal/src/commands/push.ts
@@ -198,12 +198,14 @@ export default class Push extends Command {
         mobile: false
       }
 
+      const name = definition.name
+      const description = definition.description
       const options = getOptions(definition, metadata.options)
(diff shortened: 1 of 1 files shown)


packages/cli-internal/src/commands/push.ts
```diff
@@ -198,12 +198,14 @@ export default class Push extends Command {
         mobile: false
       }
 
+      const name = definition.name
```
dlasky, line 201: this actually fixes the update portion below on Line 271 as name wasn't actually defined at that point in the code

NolanChan (approved):
After the fun hour journey, the fact that we never were able to update the name through `push` seems odd; this definitely needs to be a PSA of sorts. But LGTM overall

dlasky:
agreed, i want to find out explicitly when that changed though first

LLuque-twilio (approved):
Approving so merging is available.

PR #725, opened 2022-08-10
[STRATCONN 1413] change check to use blocklist

maryamsharif:
This PR edits the original changes made for [STRATCONN-1413](https://segment.atlassian.net/browse/STRATCONN-1413) and implements a blocklist instead of an allowlist to make sure the filter is not too narrow and we are only blocking data types that we are sure GA4 does not accept. 

## Testing

We used Google's debug endpoint to verify that GA4 does not allow null, array, or objects as params but does allow nulls for user_properties. We also verified that GA4 does accept boolean variables for both params and user_properties.

[Testing Doc](https://paper.dropbox.com/doc/GA4-Parameters-Testing--BncVWRzcIaMUuSoFLrZ6K1FRAg-FMVqXKylO9KdoPkgtDtth).

### DIFF
--- a/packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
@@ -15,9 +15,10 @@ export function verifyParams(params: object | undefined): void {
   }
 
   Object.values(params).forEach((value) => {
-    if (typeof value == 'object' || value instanceof Array) {
+    const invalidEventType = typeof value == 'object' || value instanceof Array
+    if (invalidEventType) {
       throw new IntegrationError(
-        'GA4 does not accept null, array, or nested values for event parameters and item parameters. Please ensure you are using allowed data types.',
+        'GA4 does not accept null, array, or object values for event parameters and item parameters. Please ensure you are using allowed data types.',
         'Invalid value',
         400
       )
@@ -31,10 +32,10 @@ export function verifyUserProps(userProperties: object | undefined): void {
   }
 
   Object.values(userProperties).forEach((value) => {
-    // typeof null == 'object' and GA4 accepts nulls for user_properties
+    // Extra check to ensure null values are not filtered out since GA4 accepts nulls for user_properties
     if ((value != null && typeof value == 'object') || value instanceof Array) {
       throw new IntegrationError(
-        'GA4 does not accept array or nested values for user properties. Please ensure you are using allowed data types.',
+        'GA4 does not accept array or object values for user properties. Please ensure you are using allowed data types.',
         'Invalid value',
         400
       )
--- a/packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
@@ -15,9 +15,9 @@ export function verifyParams(params: object | undefined): void {
   }
 
   Object.values(params).forEach((value) => {
-    if (typeof value != 'string' && typeof value != 'number') {
+    if (typeof value == 'object' || value instanceof Array) {
       throw new IntegrationError(
-        'GA4 only accepts string or number values for event parameters and item parameters. Please ensure you are not including null, array, or nested values.',
+        'GA4 does not accept null, array, or nested values for event parameters and item parameters. Please ensure you are using allowed data types.',
         'Invalid value',
         400
       )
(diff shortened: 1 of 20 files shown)


packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
```diff
@@ -15,9 +15,9 @@ export function verifyParams(params: object | undefined): void {
   }
 
   Object.values(params).forEach((value) => {
-    if (typeof value != 'string' && typeof value != 'number') {
+    if (typeof value == 'object' || value instanceof Array) {
       throw new IntegrationError(
-        'GA4 only accepts string or number values for event parameters and item parameters. Please ensure you are not including null, array, or nested values.',
+        'GA4 does not accept null, array, or nested values for event parameters and item parameters. Please ensure you are using allowed data types.',
```
rhall-twilio, line 20: Is this supposed to check for null value?

marinhero, line 20: Does `nested` equals to `object` in this case or should we include `object` in the error message as well?

packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
```diff
@@ -15,9 +15,9 @@ export function verifyParams(params: object | undefined): void {
   }
 
   Object.values(params).forEach((value) => {
-    if (typeof value != 'string' && typeof value != 'number') {
+    if (typeof value == 'object' || value instanceof Array) {
```
marinhero, line 18: nit:
```suggestion
const invalidEventType = typeof value == 'object' || value instanceof Array
    if (invalidEventType) {
```

packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
```diff
@@ -31,9 +31,10 @@ export function verifyUserProps(userProperties: object | undefined): void {
   }
 
   Object.values(userProperties).forEach((value) => {
-    if (typeof value != 'string' && typeof value != 'number' && value != null) {
+    // typeof null == 'object' and GA4 accepts nulls for user_properties
```
marinhero, line 34: Can you reword this statement? As a potential maintainer, I don't quite follow its meaning.

packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
maryamsharif: In javascript the type of a null value evaluates to "object" so the check does include not allowing null values!

maryamsharif: yeah we are referring to objects when we say nested maybe we should say object instead of nested? @kdaswani

rhall-twilio: Can we confirm that if the user passes null, it will arrive here as null instead of undefined?

kdaswani: yup fine by me to change the wording to:

```suggestion
        'GA4 does not accept null, array, or object values for event parameters and item parameters. Please ensure you are using allowed data types.',
```

maryamsharif: yes it arrives as null and not undefined

packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
```diff
@@ -15,9 +15,10 @@ export function verifyParams(params: object | undefined): void {
   }
 
   Object.values(params).forEach((value) => {
-    if (typeof value != 'string' && typeof value != 'number') {
+    const invalidEventType = typeof value == 'object' || value instanceof Array
```
rhall-twilio, line 18: Is there a reason we're doing this here but not below as well?

rhall-twilio:
Let's make sure we update the Insomnia test cases to include a boolean somewhere before shipping.

rhall-twilio (changes requested):

packages/destination-actions/src/destinations/google-analytics-4/ga4-functions.ts
```diff
@@ -14,11 +14,22 @@ export function verifyParams(params: object | undefined): void {
...
   }
 
-  Object.values(params).forEach((value) => {
-    const invalidEventType = typeof value == 'object' || value instanceof Array
-    if (invalidEventType) {
+  Object.entries(params).forEach(([key, value]) => {
+    if (value instanceof Array) {
       throw new IntegrationError(
-        'GA4 does not accept null, array, or object values for event parameters and item parameters. Please ensure you are using allowed data types.',
+        `Param [${key}] has unsupported value. GA4 does not accept array values for event parameters and item parameters.`,
```
rhall-twilio, line 20: We should also include the type of the unsupported parameter.

rhall-twilio (approved):

PR #727, opened 2022-08-15
Add Mixpanel EU Data Residency Setting

jaredmixpanel:
This PR adds a setting for [Mixpanel's EU Data Residency feature](https://help.mixpanel.com/hc/en-us/articles/360039135652-Data-Residency-in-EU). When the setting is set to `EU` data will be sent to `api-eu.mixpanel.com`, the default is `US` which sends to `api.mixpanel.com`.

image

## Testing

- [ X ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [ X ] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/index.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/index.ts
@@ -7,6 +7,7 @@ import identifyUser from './identifyUser'
 import groupIdentifyUser from './groupIdentifyUser'
 
 import alias from './alias'
+import { ApiRegions } from './utils'
 
 /** used in the quick setup */
 const presets: DestinationDefinition['presets'] = [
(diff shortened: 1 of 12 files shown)


packages/destination-actions/src/destinations/mixpanel/index.ts
```diff
@@ -71,6 +71,13 @@ const destination: DestinationDefinition<Settings> = {
         description: 'Mixpanel project secret.',
         type: 'string',
         required: true
+      },
+      eu: {
+        label: 'EU Data Residency',
```
brennan, line 76: Will customers understand what this means, or should we have a label that is descriptive, like "Send Data to EU Endpoint"?

dlasky: I think its also worth noting that all the destinations we have so far in actions that do this use a dropdown with choices for US/EU so if its not too much trouble maybe we should consider changing it just for general consistency there?

brennan (approved):

brennan:
LGTM - any final comments before we merge @dlasky?

PR #729, opened 2022-08-18
Add batching support to the Mixpanel Actions Destination

zihejia:
This PR adds batching requests support to the Mixpanel Actions Destination.  A more efficient way to receive and process those large sets of data.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/index.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/index.ts
@@ -81,6 +81,12 @@ const destination: DestinationDefinition<Settings> = {
         choices: Object.values(ApiRegions).map((apiRegion) => ({ label: apiRegion, value: apiRegion })),
         default: ApiRegions.US,
         required: true
+      },
+      enable_batching: {
+        type: 'boolean',
+        label: 'Batch Data to Mixpanel',
+        description: 'Set as true to ensure Segment sends data to Mixpanel in batches. Please do not set to false.',
+        default: true
       }
     },
     testAuthentication: (request, { settings }) => {
--- a/packages/destination-actions/src/destinations/mixpanel/index.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/index.ts
@@ -81,12 +81,6 @@ const destination: DestinationDefinition<Settings> = {
         choices: Object.values(ApiRegions).map((apiRegion) => ({ label: apiRegion, value: apiRegion })),
         default: ApiRegions.US,
         required: true
-      },
-      enable_batching: {
-        type: 'boolean',
-        label: 'Batch Data to Mixpanel',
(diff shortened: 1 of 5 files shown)


packages/destination-actions/src/destinations/mixpanel/trackEvent/index.ts
```diff
@@ -1,10 +1,86 @@
...
+      utm_source: utm.utm_source,
+      utm_term: utm.utm_term,
+      ...payload.event_properties
+    }
+  }
+  return event;
+}
+
+const processData = async (request: RequestClient, settings: Settings, payload: Payload[]) => {
+  if (!settings.apiSecret) {
```
brennan, line 70: `apiSecret` is a required setting. Events without `settings.apiSecret` defined will be rejected before ever hitting this code block.

brennan:
@zihejia Would you also please run `./bin/run generate:types`? Adding a `performBatch` handler, then running the command should generate an `enable_batching` field.

packages/destination-actions/src/destinations/mixpanel/index.ts
```diff
@@ -71,6 +71,12 @@ const destination: DestinationDefinition<Settings> = {
         description: 'Mixpanel project secret.',
         type: 'string',
         required: true
+      },
```
brennan, line 74: @zihejia FYI the `enable_batching` field sits at the action level so has to be defined as a property on an `ActionDefinition` class. See [docs](https://segmentpartners.slack.com/archives/C02D9DFHFC2/p1659723668044869) here.

Since it looks as if you're implementing batching support for the `trackEvent` action only, I would recommend moving this `enable_batching` field into the `trackEvent` `ActionDefinition`, then running `./bin/run generate` one more time.

zihejia:
@brennan have you got a chance to have another look?

packages/destination-actions/src/destinations/mixpanel/trackEvent/index.ts
```diff
@@ -326,76 +398,21 @@ const action: ActionDefinition<Settings, Payload> = {
       default: {
         '@path': '$.context.userAgent'
       }
+    },
+    enable_batching: {
+      type: 'boolean',
+      label: 'Batch Data to Mixpanel',
+      description: 'Set as true to ensure Segment sends data to Mixpanel in batches. Please do not set to false.',
```
rvadera12, line 405: Why can't this value be false. If it is a breaking issue, might be worth adding extra validation to ensure 'false' is not passed, by throwing an `IntegrationError`.

zihejia: i will remove it.

rvadera12 (commented):
LGTM - any final comments before I approve @dlasky

dlasky:
looks fine to me!

dlasky (approved):

PR #734, opened 2022-08-22
Engage Destinations Trait Enrichment

nidubey:
[SDD](https://paper.dropbox.com/doc/Mini-SDD-Destination-Actions-Trait-Enrichment--Bnti4z293mAhJ6mQTyFWImVVAg-45LlW8bz0LspJbFIlxWB5)

In this PR, we are using the profile traits from segment events instead of making a profile api call to fetch all traits. Currently the change is gated (using traitEnrichment config property). The purpose of this change is NOT make profile api call since we already have profile traits in the segment event sent from upstream.

## Testing


packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -269,6 +276,13 @@ const action: ActionDefinition<Settings, Payload> = {
       description: 'Additional custom args that we be passed back opaquely on webhook events',
       type: 'object',
       required: false
+    },
+    traits: {
+      label: 'Traits',
+      description: 'A user profile traits',
```
lew-gordon, line 282:
```suggestion
      description: 'A user profile's traits',
```

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -288,7 +302,13 @@ const action: ActionDefinition<Settings, Payload> = {
...
     } else if (['subscribed', 'true'].includes(emailProfile?.subscriptionStatus)) {
       statsClient?.incr('actions-personas-messaging-sendgrid.subscribed', 1, tags)
-      const traits = await fetchProfileTraits(request, settings, payload.userId, statsClient, tags)
+
+      let traits
+      if (payload.traitEnrichment) {
+        traits = payload?.traits ? payload?.traits : JSON.parse('{}')
+      } else {
+        traits = await fetchProfileTraits(request, settings, payload.userId, statsClient, tags)
+      }
```
lew-gordon, line 311:
```suggestion
      let traits
      if (payload.traitEnrichment) {
        traits = payload?.traits ?? {}
      } else {
        traits = await fetchProfileTraits(request, settings, payload.userId, statsClient, tags)
      }
```

I think you might be able to simplify this but you may need to change the Profile interface on like 110.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/index.ts
```diff
@@ -151,11 +165,18 @@ const action: ActionDefinition<Settings, Payload> = {
       return
     } else if (['subscribed', 'true'].includes(externalId.subscriptionStatus)) {
       statsClient?.incr('actions-personas-messaging-twilio.subscribed', 1, tags)
-      const traits = await fetchProfileTraits(request, settings, payload.userId, statsClient, tags)
       const phone = payload.toNumber || externalId.id
       if (!phone) {
         return
       }
+
+      let traits
```
lew-gordon, line 173: Same as the sendgrid comment I had.

lew-gordon (approved):
Overall looks good! I just had a few suggestions to simplify some of the logic.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
nidubey: yeah we thought about that but since we are keeping the profile api call behind the gate for now, it did not make sense to change the interface for our use case. 

btw, is `payload?.traits ? payload?.traits : JSON.parse('{}')` and `payload?.traits ?? {}` same thing.

lew-gordon: Yeah at least that's what I was hoping. Just less verbose.

dlasky:
overall looks fine, i am worried that by putting your test util in core though that it could be confusing for 3rd party builders working on their own tests, should that be just a local util in destinations package?

dlasky:
yeah the majority of our users are our partner developers outside of segment. And yep dead on with the original intention of core. 

A couple options that would 'feel' better to me would be:

- adding some mixin type capabilities to the test event creation then utilizing those in your local (seems like more work)
- moving your existing to the actual action package as previously discussed

Feel free to export things you need from core though if they aren't already exported to make that work in the 'local' folder.

dlasky:
also have a look at the build, looks like some TS issues! :(

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
nidubey: it complains with that change 
`(property) Profile.traits: Record<string, string>
Type '{ [k: string]: unknown; }' is not assignable to type 'Record<string, string>'.
  'string' index signatures are incompatible.
    Type 'unknown' is not assignable to type 'string'.ts(2322)
index.ts(113, 3): The expected type comes from property 'traits' which is declared here on type 'Profile'`

I recall that this was the reason why I had to do the JSON.parse thing.

lew-gordon: If typescript is going to fight us we don't need to solve it in this PR.

dlasky (approved):

PR #740, opened 2022-08-24
Action Tester for Web

pooyaj:
Adding support for Action Tester for web destinations. Changes are rather minor: 
- passing `directoryName` property for each destination in the server.ts ( we need that name instead of the real slug for browser destinations ) 
- updating the `serve` command to accept a browser flag 
- update the launch.json to allow debugging with VS code. 

Note that the PR for the app portion is here: https://github.com/segmentio/app/pull/13980

## Testing
Tested locally with action tester. 

### DIFF
--- a/packages/cli/src/commands/serve.ts
+++ b/packages/cli/src/commands/serve.ts
@@ -1,5 +1,5 @@
 import { Command, flags } from '@oclif/command'
-import { ChildProcess, fork } from 'child_process'
+import { ChildProcess, fork, exec } from 'child_process'
 import { autoPrompt } from '../lib/prompt'
 import chalk from 'chalk'
 import chokidar from 'chokidar'
@@ -36,6 +36,10 @@ export default class Serve extends Command {
     noUI: flags.boolean({
       char: 'n',
       description: 'do not open actions tester UI in browser'
+    }),
+    browser: flags.boolean({
+      char: 'r',
+      description: 'serve browser destinations'
     })
   }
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/serve.ts
```diff
@@ -122,6 +126,10 @@ export default class Serve extends Command {
         child?.removeAllListeners()
         child = undefined
       })
+
+      if (flags.browser) {
+        exec('yarn --cwd packages/browser-destinations dev')
```
dlasky, line 131: we also have `execa` available if thats 'nicer'

packages/cli/src/lib/server.ts
```diff
@@ -152,7 +152,7 @@ function setupRoutes(def: DestinationDefinition | null): void {
   router.get(
     '/manifest',
     asyncHandler(async (_, res: express.Response) => {
-      res.json(destination.definition)
+      res.json({ ...destination.definition, directoryName: process.env.DESTINATION })
```
dlasky, line 155: is this just used by the webpack instance? i feel like im missing something

dlasky (approved):

packages/cli/src/lib/server.ts
dlasky: scratch previous after reading the app PR.

pooyaj: webpack bundles the web destinations under their directory name (e.g., `intercom.js`) and then assign them to the window under the `${directoryName}Destination` ... so in order to load the destination and be able to call them we need their directory name ( rather than the actual slug )

chrisradek (approved):

PR #748, opened 2022-08-31
Add 'Source Name' setting to Mixpanel Actions

zihejia:
This PR adds a setting 'Source Name' for Mixpanel Actions. Currently, the setting is only available in the Mixpanel Classic destination.

image

## Testing


packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
```diff
@@ -50,7 +50,8 @@ const action: ActionDefinition<Settings, Payload> = {
       $token: settings.projectToken,
       $group_key: group_key,
       $group_id: group_id,
-      $set: payload.traits
+      $set: payload.traits,
+      segment_source_name: settings.sourceName
```
rvadera12, line 54: Wondering why there is a `$` in front of all of the variables except for `segment_source_name`. If possible could we keep this consistent?

rvadera12 (commented):
The PR looks great! I left a question and once that is addressed, we can move forward!

packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/index.ts
zihejia: thanks @rvadera12 for the review. Unfortunately there are some inconsistencies in Mixpanel default property naming in terms of the `$` due to legacy reasons. It will break the existing customers if we change them.

rvadera12 (approved):

PR #751, opened 2022-09-01
Initialize UTM properties changes

LLuque-twilio:
This PR addresses [JIRA ACT-146](https://segment.atlassian.net/browse/ACT-146) & [Design Doc](https://paper.dropbox.com/doc/Amplitude-Transparency-and-control-over-setsetOnce--BocP0wAMd7029sN0gSA2eaSoAg-7PxGzLWcZJFtaA5pbAT7C)

In it we create three new object fields for the Amplitude `logEvent` action - `setAlways`, `setOnce` and `add`. The goal of these new fields is to replace the previous fields of `UTM_properties` and `referrer` since the values for these old fields undergo some hidden things in the background of the `perform` block that aren't clear to users. This change should make it much clearer to the user what is happening to their data, as well as allow for greater customizability as to what fields get passed to Amplitude in their respective `$set`, `$setOnce`, and `$add` fields.

To allow for backwards compatibility we decided to leave in the logic that handles what happens when the fields for `referrer` and `UTM_properties` are set. In the instance both the new fields _AND_ old field have values passed in, whatever values are in the older fields will win out. For example, say in the payload passed to the `perform` block the fields `setOnce`, `setAlways` and `UTM_properties` look like this respectively:
```
setOnce: {
  "initial_utm_source": "google"
}
setAlways: {
  "utm_source": "bing"
}
UTM_properties: {
  "utm_source": "yahoo"
}
```

The resulting payload sent to amplitude will look like
```
$set: {
  "utm_source": "yahoo"
}
$setAlways: {
  "utm_source": "yahoo"
}
```

(This override will only happen for the `utm_source`, `utm_medium`, `utm_campaign`, `utm_term`, `utm_content`, and `referrer` fields)

## Testing

Testing completed successfully via a stage deploy to Actions Amplitude and sending test events through `dispatch` + `event-tester`

### Test One
This test was to ensure the "overriding" was occurring correctly. Events sent with values for the `utm_properties` and `referrer` fields would override any matching fields sent in the `setOnce` and `setAlways` fields.

| Mapping  | Amplitude Result |
| Screen Shot 2022-09-12 at 11 39 19 AM | Screen Shot 2022-09-12 at 11 39 40 AM |

As seen, the values set in the `utm_properties` and `referrer` objects "win out" over those same fields set in the `setAlways` and `setOnce` objects.

### Test Two
This test was to ensure when there were no values in the `utm_properties` and `referrer` fields, the `setAlways` and `setOnce` fields worked correctly.

| Mapping  | Amplitude Result |
| Screen Shot 2022-09-12 at 11 41 44 AM | Screen Shot 2022-09-12 at 11 41 58 AM |

As seen, the corresponding fields are correctly set in Amplitude.

### Test Three
This test was to ensure that any fields in the `setOnce` object are truly only set on the first event, and ignored for any future events. In addition, it ensures that any fields in the `setAlways` object are updated on each new event.

| Mapping  | Amplitude Result |
| Screen Shot 2022-09-12 at 11 44 03 AM | Screen Shot 2022-09-12 at 11 44 35 AM |

As seen, only the fields in the `setAlways` object are updated on new events.

### Test Four
This test was to ensure functionality of the `add` object.

| Mapping  | Amplitude Result |
| Screen Shot 2022-09-12 at 11 48 23 AM | Screen Shot 2022-09-12 at 11 48 31 AM |

As seen, we set a variable `test` to 10 in the `setOnce` object, and then attempted to add 5 to it using the `add` object. The resulting field in amplitude is 15 showing it works as expected.

### Test Five
This test was to ensure all previous instances of the `LogEvent` action would still work correctly after the changes.
- The new fields do show up in the old action but there are no `defaults` set for them as seen in the image below.
- Events through dispatch were correctly delivered both when the new fields did and didn't have values for them.

Screen Shot 2022-09-12 at 11 45 58 AM

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
+++ b/packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
@@ -4,9 +4,6 @@ import { eventSchema } from '../event-schema'
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import { convertUTMProperties } from '../utm'
-import { convertReferrerProperty } from '../referrer'
-import { mergeUserProperties } from '../merge-user-properties'
 import { parseUserAgentProperties } from '../user-agent'
 import { getEndpointByRegion } from '../regional-endpoints'
 
@@ -108,6 +105,96 @@ const action: ActionDefinition<Settings, Payload> = {
         'Enabling this setting will set the Device manufacturer, Device Model and OS Name properties based on the user agent string provided in the userAgent field',
       default: true
     },
+    setOnce: {
+      label: 'Set Once',
+      description: 'The following fields will be set only once per session when using AJS2 as the source',
+      type: 'object',
+      additionalProperties: true,
+      properties: {
+        //should the types of these properties be string? Is there anything else they could be?
+        initial_referrer: {
+          label: 'Initial Referrer',
+          type: 'string',
+          description: 'The referrer of the web request'
+        },
+        initial_utm_source: {
+          label: 'Initial UTM Source',
+          type: 'string'
+        },
+        initial_utm_medium: {
+          label: 'Initial UTM Medium',
+          type: 'string'
+        },
+        initial_utm_campaign: {
+          label: 'Initial UTM Campaign',
+          type: 'string'
+        },
+        initial_utm_term: {
+          label: 'Initial UTM Term',
+          type: 'string'
+        },
+        initial_utm_content: {
+          label: 'Initial UTM Content',
+          type: 'string'
+        }
+      },
+      default: {
+        initial_referrer: { '@path': '$.context.page.referrer' },
+        initial_utm_source: { '@path': '$.context.campaign.source' },
+        initial_utm_medium: { '@path': '$.context.campaign.medium' },
+        initial_utm_campaign: { '@path': '$.context.campaign.name' },
+        initial_utm_term: { '@path': '$.context.campaign.term' },
+        initial_utm_content: { '@path': '$.context.campaign.content' }
+      }
+    },
+    setAlways: {
+      label: 'Set Always',
+      description: 'The following fields will be set every session when using AJS2 as the source',
+      type: 'object',
+      additionalProperties: true,
+      properties: {
+        //should the types of these properties be string? Is there anything else they could be?
+        referrer: {
+          label: 'Referrer',
+          type: 'string'
+        },
+        utm_source: {
+          label: 'UTM Source',
+          type: 'string'
+        },
+        utm_medium: {
+          label: 'UTM Medium',
+          type: 'string'
+        },
+        utm_campaign: {
+          label: 'UTM Campaign',
+          type: 'string'
+        },
+        utm_term: {
+          label: 'UTM Term',
+          type: 'string'
+        },
+        utm_content: {
+          label: 'UTM Content',
+          type: 'string'
+        }
+      },
+      default: {
+        referrer: { '@path': '$.context.page.referrer' },
+        utm_source: { '@path': '$.context.campaign.source' },
+        utm_medium: { '@path': '$.context.campaign.medium' },
+        utm_campaign: { '@path': '$.context.campaign.name' },
+        utm_term: { '@path': '$.context.campaign.term' },
+        utm_content: { '@path': '$.context.campaign.content' }
+      }
+    },
+    //Looking at the amplitude docs, do we expect this to be an object containing types of number?
+    add: {
+      label: 'Add',
+      description: 'The following fields will be set only once per session when using AJS2 as the source',
+      type: 'object',
+      additionalProperties: true
+    },
     utm_properties: {
       label: 'UTM Properties',
       type: 'object',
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
```diff
@@ -135,35 +183,80 @@ const action: ActionDefinition<Settings, Payload> = {
...
       default: {
+        referrer: { '@path': '$.context.page.referrer' },
         utm_source: { '@path': '$.context.campaign.source' },
         utm_medium: { '@path': '$.context.campaign.medium' },
         utm_campaign: { '@path': '$.context.campaign.name' },
         utm_term: { '@path': '$.context.campaign.term' },
         utm_content: { '@path': '$.context.campaign.content' }
       }
     },
+    //Looking at the amplitude docs, do we expect this to be an object containing types of number?
```
LLuque-twilio, line 194: See comment. [amplitude docs](https://www.docs.developers.amplitude.com/data/sdks/javascript/#user-properties)

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
```diff
@@ -108,11 +108,59 @@ const action: ActionDefinition<Settings, Payload> = {
...
+        initial_utm_content: { '@path': '$.context.campaign.content' }
+      }
+    },
+    setAlways: {
+      label: 'Set Always',
+      description: 'The following fields will be set every session when using AJS2 as the source',
+      type: 'object',
+      additionalProperties: true,
+      properties: {
+        //should the types of these properties be string? Is there anything else they could be?
```
LLuque-twilio, line 159: ditto prev comment

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
```diff
@@ -108,11 +108,59 @@ const action: ActionDefinition<Settings, Payload> = {
...
-    utm_properties: {
-      label: 'UTM Properties',
+    setOnce: {
+      label: 'Set Once',
+      description: 'The following fields will be set only once per session when using AJS2 as the source',
       type: 'object',
-      description: 'UTM Tracking Properties',
+      additionalProperties: true,
       properties: {
+        //should the types of these properties be string? Is there anything else they could be?
```
LLuque-twilio, line 117: Looks like they can be of any type [here](https://www.docs.developers.amplitude.com/data/sdks/javascript/#user-properties), not necessarily just string?

LLuque-twilio, line 117: Looks like they can be of any type [here](https://www.docs.developers.amplitude.com/data/sdks/javascript/#user-properties), not necessarily just string?

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
LLuque-twilio, line 194: ditto previous comment...

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
LLuque-twilio, line 159: ditto previous comment...

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
LLuque-twilio, line 194: Should these only be of type number [amplitude docs](https://www.docs.developers.amplitude.com/data/sdks/javascript/#add)? Not even sure if that's possible since we aren't defining any defaults for `add` ...

LLuque-twilio:
Diff doesn't show it correctly, but this PR removes the `defaults` for both the `referrer` and `UTM_properties` fields.

From my understanding, this should only affect **_new_** instances of this action. My thinking was we want users who set up new instances to use the new `setAlways`, `setOnce`, and `add` fields instead of the other fields. So removing the defaults for the old fields would be the best way to encourage that behavior.

dlasky:
I haven't looked at the PR yet, but I think we can safely remove the legacy fields entirely... Are you thinking that when they save it would potentially remove that functionality for them?

LLuque-twilio:
> I haven't looked at the PR yet, but I think we can safely remove the legacy fields entirely... Are you thinking that when they save it would potentially remove that functionality for them?

Ahh I see. So we can remove the fields but keep the override functionality in the `perform` block.

To answer your question, yes I was under the impression that (for all past instances of the `LogEvent` action that were initialized with the legacy fields) removing the legacy fields entirely would remove the field from the UI but their mappings would still be saved in the backend. I'm not too familiar what happens when we remove fields that users already have values saved for. Are they ignored completely or still included secretly in the payload passed to the perform block?

packages/destination-actions/src/destinations/amplitude/logEvent/index.ts
```diff
@@ -182,11 +272,21 @@ const action: ActionDefinition<Settings, Payload> = {
       properties.session_id = dayjs.utc(session_id).valueOf()
     }
 
+    if (Object.keys(Object.fromEntries(Object.entries(setOnce ?? {}).filter(([_, v]) => v))).length) {
```
dlasky, line 275: consider making this a utility method that is named in such a way to explain what its doing

dlasky (commented):
given the relative danger here lets also add some explicit tests for the 3 scenarios:

1) old settings
2) new settings
3) hybrid of both just to understand the outcome

dlasky:
per the discussion we had last week -- it actually makes a ton of sense to move this to a brand new action... maybe `logEventV2`? need to think on the naming, but that allows us to completely strip the legacy logic and makes this considerably safer overall I think.

PR #757, opened 2022-09-04
[Cordial] Identifiers changes, MergeContacts action

espectrio:
We've extracted mapping of segment's userId key in Cordial to settings and started to always send it (alongside with `userId` and `anonymousId` values) with every request like so:

`Object {
  "anonymousId": "246d1e7f-3dd3-4e4a-b7d0-5530bd9b5945",
  "groupId": "group1234",
  "segmentId": "user1234",
  "segmentIdKey": "segment_id",
  "userIdentities": Object {
    "channels.email.address": "contact@example.com",
  },
}`

## Testing


packages/destination-actions/src/destinations/cordial/addContactToList/index.ts
```diff
@@ -2,20 +2,14 @@ import type { ActionDefinition } from '@segment/actions-core'
...
   title: 'Add Contact to List',
   description: 'Add contact to a list. If the list does not exist in Cordial it will be created.',
   defaultSubscription: 'type = "group"',
   fields: {
-    userIdentities: {
-      label: 'User Identities',
-      description:
-        'An ordered list of contact identifiers in Cordial. Each item in the list represents an identifier. For example, `channels.email.address -> userId` and/or `customerId -> traits.customerId`. At least one identifier should be valid otherwise the contact will not be identified and the request will be ignored.',
-      type: 'object',
-      required: true,
```
LLuque-twilio, line 16: I see that `userIdentities` here has `required:true` whereas [here](https://github.com/segmentio/action-destinations/pull/757/files#diff-3ed702429b91ebdcf81a93ee9be4b18d3752e1709c8e742320aa465507bca54bR23) in the `userIdentityFields` it is `required:false`. Is this on purpose?

espectrio: Yes, as we exposing userId and anonymousId we may iedntify only based on those fields

espectrio: Maybe there is an elegant way to require_if in destination action UI somehow? We'd like to require at least one field filled in ('segment_id', 'segment_anonymous_id', or 'user_identities')

espectrio:
@LLuque-twilio please review the approach

jinapark202 (approved):

espectrio:
@jinapark202 please do not release it yet, I'll post here when it'll be ready on our side. Thanks in advance!

espectrio:
@dlasky @jinapark202 Ready for release, thanks in advance!

joe-ayoub-segment:
hi @espectrio , thanks for submitting this PR. 
Would you be available to catch up with me to run through the PR please? You can schedule a call with me via this Calendly link: http://calendly.com/joe_ayoub/integration-workshop 

Just want to get an understanding of what the potential impact could be to customers who are using the integration already (given there are changes to identifiers in this PR). 

I'm based in EMEA (Ireland), so if you can't find a timeslot which works for you let me know and I'll open up more slots.

joe-ayoub-segment:
Notes from PR review meeting with @espectrio and other team member: 

PR is to allow customers to specify which external identifier to use when sending data to Cordial. Also includes code to automatically collect Segment anonymousId and userId fields (as hidden fields). Also adds an Action to handle Segment alias() calls. 

**Impact of PR on customers who are already using the Integration**

- During our call we discussed risks of interrupting existing data flows
- Partner’s Engineers said that they did end to end testing where they fired payloads through the current ‘live’ integration on to the Cordial Platform, and then fired the same payloads through the event tester and on to the Cordial Platform. They are confident that the PR will not create data flow issues for existing customers
- There are only 2 customers using the Integration
- The Partner will notify the customers prior to deployment
- We discussed the risks of adding new required fields, and of making changes to how data is handled in general. 
-- No new required fields are added by this PR
-- New optional ‘hidden’ fields are added by this PR, and are not treated as required in the perform() function
- We discussed the steps to take if there is an interruption to customer data flow
-- 1. Ask customer to disable their existing Integration, then create and configure a new Integration
-- 2. If correct behavior is not restored then we should treat as an incident and roll back
- I did explain that the cost of rolling back is expensive and is a last resort only
- Unit tests pass
- We discussed Replay functionality, and how it relies on timestamps from the Segment payload being respected
Currently some (or all) Cordial Actions do not use the Segment timestamp, so a Replay would likely backfill historical events into Cordial, which would then be timestamped at the time of ingestion to Cordial
- We also discussed deduping events using the messageId. However Cordial already dedupes based on a hash of various payload fields

joe-ayoub-segment (approved):
Met with Partner's Engineers. Reviewed PR. See notes [here](https://github.com/segmentio/action-destinations/pull/757#issuecomment-1438351248) . More detailed notes shared via email.

joe-ayoub-segment:
Hi @espectrio , this PR is now deployed!

PR #765, opened 2022-09-08
[HGI-3] Introduce LinkedIn Audiences

brennan:
- This PR introduces LinkedIn Audiences to the `action-destinations` repo. This deployment will support internal testing only, and will not be rolled out to customers until later.
- Jira: https://segment.atlassian.net/browse/HGI-3
- SDD: https://paper.dropbox.com/doc/SDD-LinkedIn-Audiences-Action-Destination-90-Draft--Bozvb8fLlufSaRFyGg_79klYAg-spG2cNcs7SmoSKIJSL40m

## Testing


packages/destination-actions/src/destinations/linkedin-audiences/index.ts
```diff
@@ -0,0 +1,63 @@
...
+      })
+
+      return { accessToken: res?.data?.access_token }
+    }
+  },
+  extendRequest({ auth }) {
+    return {
+      headers: {
+        authorization: `Bearer ${auth?.accessToken}`,
+        'LinkedIn-Version': '202207',
```
nick-Ag, line 52: Maybe this could be pulled out into a constant?

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
```diff
@@ -0,0 +1,200 @@
...
+        '@path': '$.properties.audience_key'
+      }
+    },
+    enable_batching: {
+      label: 'Enable Batching',
+      description: 'Enable batching of requests to the LinkedIn DMP Segment.',
+      type: 'boolean',
+      default: true
+    },
+    email: {
```
nick-Ag, line 27: Just wondering why we don't want users to be able to map this themselves?

nick-Ag (approved):
Left a couple questions, looks good 🚀

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
brennan: Good question - because this is a Personas batch destination, all the events received by this integration will be formatted the same way, i.e. `email` will always be in `context.traits.email` for `track` events. That's why we also have the `email` toggle. Customers can toggle on to send `email` to LinkedIn - but we don't want them to change the mapping directive b/c they would just break the mapping.

nick-Ag: Makes sense thanks!

packages/destination-actions/src/destinations/linkedin-audiences/index.ts
brennan: I can get on board with that.

brennan:
@dlasky Curious if you have any reservations on [this change](https://github.com/segmentio/action-destinations/pull/765/files#diff-cb656e1718b849f54845213d6943bab099c89303c5431a810e0cbf489698d6eb) before I merge? Because I'm touching a core package, a codeowner review is required. 🚀

dlasky (approved):

PR #770, opened 2022-09-16
[STRATCONN-1649] - Adobe Target Cloud - Do not run getNestedObjects for null values

marinhero:
Errors surfaced showing that `getNestedObjects` was being called with `null` values. That makes sense because this is a recursive function that doesn't check if it runs on top of a `null` value. With this check we guarantee a safe execution of the code.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts
+++ b/packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts
@@ -1,15 +1,19 @@
 import { RequestClient, IntegrationError } from '@segment/actions-core'
 
 function getNestedObjects(obj: { [x: string]: any }, objectPath = '', attributes: { [x: string]: string } = {}) {
-  Object.keys(obj).forEach((key) => {
-    const currObjectPath = objectPath ? `${objectPath}.${key}` : key
-    if (typeof obj[key] !== 'object') {
-      attributes[currObjectPath] = obj[key].toString()
-    } else {
-      getNestedObjects(obj[key], currObjectPath, attributes)
-    }
-  })
-  return attributes
+  // Do not run on null or undefined
+  if (obj != null || obj != undefined) {
+    Object.keys(obj).forEach((key) => {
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts
```diff
@@ -1,15 +1,19 @@
...
-    const currObjectPath = objectPath ? `${objectPath}.${key}` : key
-    if (typeof obj[key] !== 'object') {
-      attributes[currObjectPath] = obj[key].toString()
-    } else {
-      getNestedObjects(obj[key], currObjectPath, attributes)
-    }
-  })
-  return attributes
+  // Do not run on null or undefined
+  if (obj != null || obj != undefined) {
```
kdaswani, line 5: Does this look at the entire object or just the individual keys within the object so they would get left out?

marinhero: Good question! This is a recursive function that operates on values inside the original object when `typeof obj[key] !== 'object'` so in this case `typeof null !== 'object'` returns `false` which sends `null` to `getNestedObjects(null)` which in turn will be evaluated in `Object.keys(null)` and this is what blows the code up.

kdaswani: Thank you for explaining!

rvadera12 (approved):
LGTM 🚀

rvadera12:
Tests are failing since `Argument of type '{ [x: string]: string; } | undefined' is not assignable to parameter of type '{ [x: string]: { toString: () => string; }; }'.` 

We should add a check [here](https://github.com/segmentio/action-destinations/blob/aa00483cbd159719c051c402aad2a5179e34b124/packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts#L41). To make sure `traits` is not null or undefined.

rvadera12 (approved):

dlasky (approved):

PR #786, opened 2022-09-28
[STRATCONN-1670] Add self managed bearer token to AT Cloud

marinhero:
This PR adds the capability to use authentication within Adobe Target Cloud. 

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts
+++ b/packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts
@@ -58,7 +58,9 @@ export default class AdobeTarget {
         { method: 'get' }
       )
     } catch (error) {
-      return new IntegrationError('No profile found in Adobe Target with this mbox3rdPartyId', 'Profile not found', 404)
+      if (error instanceof Error) {
+        return new IntegrationError(error.message, error.stack, error.message == 'Forbidden' ? 403 : 400)
+      }
     }
 
     return undefined
--- a/packages/destination-actions/src/destinations/adobe-target/generated-types.ts
+++ b/packages/destination-actions/src/destinations/adobe-target/generated-types.ts
@@ -5,4 +5,8 @@ export interface Settings {
    * Your Adobe Target client code. To find your client code in Adobe Target, navigate to **Administration > Implementation**. The client code is shown at the top under Account Details.
    */
   client_code: string
+  /**
+   * Self managed bearer token generated via [Adobe's authentication API](https://developers.adobetarget.com/api/#authentication-tokens). Expires every 90 days.
+   */
+  bearer_token?: string
 }
--- a/packages/destination-actions/src/destinations/adobe-target/generated-types.ts
+++ b/packages/destination-actions/src/destinations/adobe-target/generated-types.ts
@@ -6,7 +6,7 @@ export interface Settings {
    */
   client_code: string
   /**
-   * Self managed bearer token generated via [Adobe's authentication API](https://developers.adobetarget.com/api/#authentication-tokens). Expires every 90 days.
+   * If you choose to require authentication for Adobe Target's Profile API, you will need to generate an authentication token. Tokens can be generated in your Adobe Target account under the Implementation Settings tab or via the [Adobe.IO Authentication Token API](https://developers.adobetarget.com/api/#authentication-tokens). Input the authentication token here. Note: Authentication tokens expire so a new token will need to be generated and updated here prior to expiration.
    */
   bearer_token?: string
 }
--- a/packages/destination-actions/src/destinations/adobe-target/index.ts
+++ b/packages/destination-actions/src/destinations/adobe-target/index.ts
@@ -19,7 +19,7 @@ const destination: DestinationDefinition<Settings> = {
         required: true
       },
       bearer_token: {
-        label: 'Bearer Token',
+        label: 'Authentication Token',
         description:
           "If you choose to require authentication for Adobe Target's Profile API, you will need to generate an authentication token. Tokens can be generated in your Adobe Target account under the Implementation Settings tab or via the [Adobe.IO Authentication Token API](https://developers.adobetarget.com/api/#authentication-tokens). Input the authentication token here. Note: Authentication tokens expire so a new token will need to be generated and updated here prior to expiration.",
         type: 'string'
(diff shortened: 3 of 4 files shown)


dlasky (approved):

packages/destination-actions/src/destinations/adobe-target/index.ts
```diff
@@ -17,8 +17,24 @@ const destination: DestinationDefinition<Settings> = {
           'Your Adobe Target client code. To find your client code in Adobe Target, navigate to **Administration > Implementation**. The client code is shown at the top under Account Details.',
         type: 'string',
         required: true
+      },
+      bearer_token: {
```
rvadera12, line 21: Could you clarify why the `bearer_token` is not a required setting?

marinhero: It's not required because it's an opt-in feature. Not every customer will have it enabled so we can't enforce it.

rvadera12 (approved):

packages/destination-actions/src/destinations/adobe-target/index.ts
```diff
@@ -17,8 +17,24 @@ const destination: DestinationDefinition<Settings> = {
...
+        label: 'Bearer Token',
+        description:
+          "Self managed bearer token generated via [Adobe's authentication API](https://developers.adobetarget.com/api/#authentication-tokens). Expires every 90 days.",
```
marinhero, line 24: @kdaswani What do you think of this description?

packages/destination-actions/src/destinations/adobe-target/adobeTarget_operations.ts
```diff
@@ -58,7 +58,9 @@ export default class AdobeTarget {
         { method: 'get' }
       )
     } catch (error) {
-      return new IntegrationError('No profile found in Adobe Target with this mbox3rdPartyId', 'Profile not found', 404)
+      if (error instanceof Error) {
```
brennan, line 61: Do we need this check?

packages/destination-actions/src/destinations/adobe-target/generated-types.ts
```diff
@@ -5,4 +5,8 @@ export interface Settings {
    * Your Adobe Target client code. To find your client code in Adobe Target, navigate to **Administration > Implementation**. The client code is shown at the top under Account Details.
    */
   client_code: string
+  /**
+   * Self managed bearer token generated via [Adobe's authentication API](https://developers.adobetarget.com/api/#authentication-tokens). Expires every 90 days.
```
brennan, line 9:
```suggestion
   * Self-managed bearer token generated via [Adobe's authentication API](https://developers.adobetarget.com/api/#authentication-tokens). Expires every 90 days.
```

packages/destination-actions/src/destinations/adobe-target/index.ts
brennan, line 24:
```suggestion
          "Self-managed bearer token generated via [Adobe's authentication API](https://developers.adobetarget.com/api/#authentication-tokens). Expires every 90 days.",
```

brennan, line 24: Does Adobe document that the token expires every 90 days? If not, maybe we shouldn't mention it? The examples in their API docs seems to suggest the token expires after 86399990 seconds, or 1000 days.

packages/destination-actions/src/destinations/adobe-target/index.ts
```diff
@@ -17,8 +17,24 @@ const destination: DestinationDefinition<Settings> = {
...
+  },
+  extendRequest({ settings }) {
+    if (settings.bearer_token) {
+      return {
+        headers: {
+          Authorization: `Bearer ${settings.bearer_token}`
+        }
       }
     }
+    return {}
```
brennan, line 37: Do we need this return statement?

packages/destination-actions/src/destinations/adobe-target/index.ts
marinhero: Removing the expiration date is a good idea. I recall seeing the 90-day expiration mentioned in one of them though, plus my own tokens have a 90 days expiration date. In any case and given that the examples say something different, I'll remove it.

kdaswani: Hi @marinhero ! I put a description in the ticket that I'd like to use: https://segment.atlassian.net/browse/STRATCONN-1670. It sounds like we may not want to explicitly put 90 days in the description though so here's my updated suggestion:

```suggestion
          "If you choose to require authentication for Adobe Target’s Profile API, you will need to generate an authentication token. Tokens can be generated in your Adobe Target account under the Implementation Settings tab or via the [Adobe.IO Authentication Token API](https://developers.adobetarget.com/api/#authentication-tokens). Input the authentication token here. Note: Authentication tokens expire so a new token will need to be generated and updated here prior to expiration.",

Comment (kdaswani) [reply packages/destination-actions/src/destinations/adobe-target/index.ts:24]:
```suggestion
                   "If you choose to require authentication for Adobe Target’s Profile API, you will need to generate an authentication token. Tokens can be generated in your Adobe Target account under the Implementation Settings tab or via the [Adobe.IO Authentication Token API](https://developers.adobetarget.com/api/#authentication-tokens). Input the authentication token here. Note: Authentication tokens expire so a new token will need to be generated and updated here prior to expiration.",
```

kdaswani: Suggested description posted above! I removed the mention of 90 days but would still like to call out that it could expire and it must be updated if/when a new one is generated.

packages/destination-actions/src/destinations/adobe-target/index.ts
marinhero: Yes, the compiler isn't happy when the implicit `undefined` is present.

packages/destination-actions/src/destinations/adobe-target/index.ts
marinhero: Thank you @kdaswani. Really appreciate the help here!

packages/destination-actions/src/destinations/adobe-target/index.ts
```diff
@@ -17,8 +17,24 @@ const destination: DestinationDefinition<Settings> = {
           'Your Adobe Target client code. To find your client code in Adobe Target, navigate to **Administration > Implementation**. The client code is shown at the top under Account Details.',
         type: 'string',
         required: true
+      },
+      bearer_token: {
+        label: 'Bearer Token',
```
kdaswani, line 22: @marinhero Can we please rename this to "Authentication Token"?

```suggestion
        label: 'Authentication Token',
```

PR #800, opened 2022-10-07
Segment Utility Destination

pooyaj:
This is a new device mode "Destination", that offers few utilities for Segment. Currently it only offers a client-side throttling utility ( I'm not sure about SegmentUtility name, and open to ideas 😅 ) 

Throttle Action: It runs as a before plugin, and throttles events to Segment using a sliding window method. It offers two settings for controlling the throttling: 
- `throttleWindow` -- A window of time at which we will allow N events of type X to Segment 
- `passThroughCount` -- The value of N, the number of events we allow in a given window 

A new window starts when we see an event for the first time, of if the window for the given event is already expired. 

## Testing


packages/browser-destinations/src/destinations/segment-utility-web/__tests__/index.test.ts
```diff
@@ -0,0 +1,25 @@
+import { Analytics, Context } from '@segment/analytics-next'
+import segmentUtilityDestination, { destination } from '../index'
+
+describe('Segment Utility Web', () => {
+  test('loads hubspot analytics with just a HubID', async () => {
```
chrisradek, line 5: Copy-paste leftovers :smile:

packages/browser-destinations/src/destinations/segment-utility-web/throttle/__tests__/index.test.ts
```diff
@@ -0,0 +1,294 @@
...
+          name: 'Throttle',
+          enabled: true,
+          subscribe: 'type = "track"',
+          mapping: {}
+        }
+      ]
+    })
+
+    await throttle.load(Context.system(), ajs)
+    let passedThroguh = 0
```
chrisradek, line 114:
```suggestion
    let passedThrough = 0
```

packages/browser-destinations/src/destinations/segment-utility-web/throttle/__tests__/index.test.ts
```diff
@@ -0,0 +1,294 @@
...
+    for (let i = 0; i < 10; i++) {
+      const ctx = await throttle.track?.(new Context({ type: 'track', event: 'Video Played' }))
+      if (ctx?.event.integrations?.['Segment.io'] !== false) {
+        passedThroguh++
```
chrisradek, line 118:
```suggestion
        passedThrough++
```

packages/browser-destinations/src/destinations/segment-utility-web/throttle/__tests__/index.test.ts
```diff
@@ -0,0 +1,294 @@
...
+      }
+    }
+
+    expect(passedThroguh).toBe(3)
```
chrisradek, line 122:
```suggestion
    expect(passedThrough).toBe(3)
```

packages/browser-destinations/src/destinations/segment-utility-web/throttle/__tests__/index.test.ts
```diff
@@ -0,0 +1,294 @@
...
+          name: 'Throttle',
+          enabled: true,
+          subscribe: 'type = "track"',
+          mapping: {}
+        }
+      ]
+    })
+
+    await throttle2.load(Context.system(), ajs)
+    passedThroguh = 0
```
chrisradek, line 139:
```suggestion
    passedThrough = 0
```

packages/browser-destinations/src/destinations/segment-utility-web/throttle/__tests__/index.test.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    for (let i = 0; i < 10; i++) {
+      const ctx = await throttle2.track?.(new Context({ type: 'track', event: 'Video Played' }))
+      if (ctx?.event.integrations?.['Segment.io'] !== false) {
+        passedThroguh++
```
chrisradek, line 144:
```suggestion
        passedThrough++
```

packages/browser-destinations/src/destinations/segment-utility-web/throttle/__tests__/index.test.ts
```diff
@@ -0,0 +1,294 @@
...
+      }
+    }
+
+    expect(passedThroguh).toBe(8)
```
chrisradek, line 148:
```suggestion
    expect(passedThrough).toBe(8)
```

chrisradek (approved):
This looks good to me (very minor comments)! Very thorough tests!

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
```diff
@@ -0,0 +1,50 @@
...
+      } else {
+        eventMap[event] = {
+          // reset the window
+          windowStarted: Date.now(),
+          receivedCount: 1
+        }
+      }
+    } else {
+      // first time we've seen this event, start the window
+      eventMap[event] = {
```
mericsson, line 38: Is there risk of `eventMap` getting too large?

packages/browser-destinations/src/destinations/segment-utility-web/__tests__/index.test.ts
pooyaj: :lul: :disappear:

packages/browser-destinations/src/destinations/segment-utility-web/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  eventMap: Record<string, { windowStarted: number; receivedCount: number }>
+}
+
+// Switch from unknown to the partner SDK client types
+export const destination: BrowserDestinationDefinition<Settings, SegmentUtilityInstance> = {
+  name: 'Segment Utility Web',
+  slug: 'actions-segment-utility-web',
+  mode: 'device',
+
+  settings: {
```
mericsson, line 17: Should we also allow customer to define which events to throttle?

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
mericsson: Wonder if there is a way instead of defining it as `eventMap[subscription]`. That seems more appropriate?

mericsson: Because there could be many events matching the subscription.

dlasky: event is just the event type string though if i'm reading this correctly

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
```diff
@@ -0,0 +1,50 @@
...
+        windowStarted: Date.now(),
+        receivedCount: 1
+      }
+    }
+
+    if (eventMap[event].receivedCount > passThroughCount) {
+      context.updateEvent('integrations', { ...context.event.integrations, 'Segment.io': false })
```
dlasky, line 45: i wonder if there is any statistical danger from always front-loading the 'picked' event? maybe implementation becomes too complex otherwise.

dlasky (approved):

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
pooyaj: The `event` is the name of the event, e.g., `Product Purchased`. But given we don't provide a default preset for this destination, the action won't run on all events by default. User has to explicitly define the subscription for the throttle to have an effect. I can imagine object could potentially grow if the user configures the subscription so the throttle runs on all tracked events (instead of narrowing the scope of subscription to specific events), and even then the size of object key space would be all the distinct event names they track, which I can imagine should be in ~100s. 

Also if we do `eventMap[Subscription]`, then if we have something like: 'type = "track" and event = "Product Purchased" or event = "Product Returned"' we won't be distinguishing `Product Purchased` from `Product Returned` during the throttling.

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
pooyaj: Yeah, I can see cases where it may cause issues. Not sure if picking randomly in the window would help, or maybe some other method 🤔  Probably if we production-ize this, we should describe that we will pick the first N events in the window, and discard after that, so people use it with caution.

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
pooyaj: Alternatively we can put a cap on the size of the object as well ( maybe 1000 distinct event names ) just to guard against those rare edge cases 🤔

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
```diff
@@ -0,0 +1,50 @@
+import { SegmentUtilityInstance } from '..'
+import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: BrowserActionDefinition<Settings, SegmentUtilityInstance, Payload> = {
+  title: 'Throttle',
+  description: 'Throttle events sent to Segment.',
```
mericsson, line 8:
```suggestion
  description: 'Throttle events sent to Segment. Throttling is on a per-event name basis.',
```

I think something to distinguish that the throttling is not specific to this subscription, but per-events for all events that match this subscription.

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
```diff
@@ -0,0 +1,50 @@
+import { SegmentUtilityInstance } from '..'
+import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: BrowserActionDefinition<Settings, SegmentUtilityInstance, Payload> = {
+  title: 'Throttle',
```
mericsson, line 7:
```suggestion
  title: 'Throttle by event name',
```

packages/browser-destinations/src/destinations/segment-utility-web/throttle/index.ts
mericsson: Got it. Made some suggestions to have title/description better match functionality  https://github.com/segmentio/action-destinations/pull/800/files#r990506657

Maybe just add a comment for now about the distinct event names issue?

pooyaj: Good call 👌

mericsson (approved):

PR #806, opened 2022-10-11
[STRATCONN-1686] Google Ads pre-beta fixes

maryamsharif:
This PR addresses [STRATCONN-1686](https://segment.atlassian.net/browse/STRATCONN-1686) to do some pre-beta fixes.

1. Make “Conversion Action” a number type for all 3 of the new actions
2. Make “Call Timestamp” required in the Upload Call Conversion action
3. Fix error handling to solve  `Cannot read property 'code' of undefined` error when an event is valid/successful. 
Screen Shot 2022-10-11 at 3 18 14 PM

## Testing

Successfully tested in staging to get actions to go through and errors still show up.

Conversion Adjustment:
Screen Shot 2022-10-11 at 4 27 19 PM

Successful Conversion:
Screen Shot 2022-10-11 at 4 34 32 PM

Error Conversion:
Screen Shot 2022-10-11 at 4 47 00 PM

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
@@ -59,7 +59,7 @@ export async function getCustomVariables(
    See here: https://developers.google.com/google-ads/api/docs/best-practices/partial-failures
  */
 export function handleGoogleErrors(response: ModifiedResponse<PartialErrorResponse>) {
-  if (response.data.partialFailureError.code !== 0) {
+  if (response.data.partialFailureError) {
     throw new IntegrationError(response.data.partialFailureError.message, 'INVALID_ARGUMENT', 400)
   }
 }
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/__tests__/uploadCallConversion.test.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/__tests__/uploadCallConversion.test.ts
@@ -22,11 +22,11 @@ describe('GoogleEnhancedConversions', () => {
 
       nock(`https://googleads.googleapis.com/v11/customers/${customerId}:uploadCallConversions`)
         .post('')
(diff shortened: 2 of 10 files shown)


packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
```diff
@@ -59,7 +59,7 @@ export async function getCustomVariables(
    See here: https://developers.google.com/google-ads/api/docs/best-practices/partial-failures
  */
 export function handleGoogleErrors(response: ModifiedResponse<PartialErrorResponse>) {
-  if (response.data.partialFailureError.code !== 0) {
+  if (response.data.partialFailureError) {
```
brennan, line 62: I couldn't find a good schema and/or expected values for Google's partial errors in their docs here: https://developers.google.com/google-ads/api/docs/best-practices/partial-failures.

Curious if there are docs that clearly show the shape of `partialFailureError`? The prior code seems to indicate that the `partialFailureError` object can exist on the response even when no partial errors are present.

maryamsharif: I have not been able to find docs that clearly stated whether or not the `partialFailureError` object would be returned in case of fully successful call. From each of the conversions documentations it shows the result should be returned as:
Screen Shot 2022-10-11 at 5 23 32 PM

and the Status object of the `partialFailureError` states that the code returned is based on enum values of google.rpc.Code which can be found here: https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

since there is a code listed for `OK`, that is the code I used to compare the result to which I now found to be incorrect. I previously did not have a way to test a valid payload and could also not find any documentation to confirm what was actually returned. Now with a valid payload we see that the `partialFailureError` object is in fact not returned if the payload is valid.

marinhero (approved):
Had a sync-review session with Maryam for this one. Code is ready to go live. Well done @maryamsharif!

PR #811, opened 2022-10-12
Update generated-types for mixpanel and package.json

marinhero:
Updating files that were missing in previous PRs


nick-Ag (approved):

packages/destination-actions/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@segment/action-destinations",
   "description": "Destination Actions engine and definitions.",
-  "version": "3.110.0",
+  "version": "3.111.1-alpha.0",
```
rvadera12, line 4: Why are we merging an `alpha` version of the npm package into main?

marinhero: This should only live in `stage` right?

marinhero:
Closed by accident via Force-Push while cleaning the branch up. Will reopen soon. Sorry for the trouble!

PR #814, opened 2022-10-13
Add support for new 'blank' field on @if directive

LLuque-twilio:
This PR addresses [JIRA ACT-200](https://segment.atlassian.net/browse/ACT-200)

In it we update the action version of `mapping-kit` to handle the `blank` field on the `@if` directive. This new field will bypass the first field in the `coalesce(field1, field2)` if it has a value of null, undefined, OR empty string `""`. 

Instead of changing the functionality of the `exists` field on the `@if` directive, we decided to add the new field `blank`. This will allow partners who depend on the current functionality of the `@if` directive to keep things working exactly as as-is, while also allowing for partners who want to bypass empty string field to do so. To implement this feature an update to the `coalesce` function in `app` will need to be made to allow partners to choose whether they want to allow or skip empty string values in the coalesce function.

Corresponding `app` PR: https://github.com/segmentio/app/pull/14462

## Testing

Added unit tests to mimic the behavior we will expect.

This new functionality will not be used until proper support is added in `app`. Testing will be done there.

### DIFF
--- a/packages/core/src/mapping-kit/__tests__/index.test.ts
+++ b/packages/core/src/mapping-kit/__tests__/index.test.ts
@@ -99,7 +99,7 @@ describe('@literal', () => {
   })
 })
 
-describe.only('@if', () => {
+describe('@if', () => {
   const payload = { a: 1, b: true, c: false, d: null, e: '' }
 
   test('exists', () => {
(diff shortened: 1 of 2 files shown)


NolanChan (approved):
Able to talk about it in the office!

PR #818, opened 2022-10-14
update web destination builds to target window

chrisradek:
This PR updates the web action destinations library target from `umd` to `window`.

## UMD vs Window

The `umd` target is meant for bundles that support cjs, amd (require), and script installation (window). The bundles we build for web destinations are only loaded via script tags, and analytics-next expects the destination to exist on the `window` object.

## Why make this change?

When targeting `umd`, the library fails to load if the website it is loaded on is using `amd`. This is because the umd loader logic will conditionally choose to use cjs/amd/global exports depending on what it detects is available.

Example of the umd loader:
```js
"object" == typeof exports && "object" == typeof module ? 
  module.exports = t() : "function" == typeof define && define.amd ?
    define("heapDestination", [], t) : "object" == typeof exports ? 
      exports.heapDestination = t() : window.heapDestination = t() 
```

Since these bundles are only recognized by analytics-next when attached to `window`, we don't actually need to support `umd`. When we eventually allow importing these destination packages directly we can support cjs/esm as separate builds so this change now doesn't make that work harder in the future.

## Testing

I tested this change using the action-tester and in stage with my own website that has requirejs loaded on it.

Before these changes, you can see that the bundles are loaded, but the 3rd party libraries are not (Braze and Heap in this case):
broken

After these changes, 3rd party libraries are loading correctly and events are being sent (Braze and Heap):
works

### DIFF
--- a/packages/browser-destinations/webpack.config.js
+++ b/packages/browser-destinations/webpack.config.js
@@ -12,7 +12,7 @@ const entries = files.reduce((acc, current) => {
   const [_dot, _src, _destinations, destination, ..._rest] = current.split('/')
   return {
     ...acc,
-    [destination]: current,
+    [destination]: current
   }
 }, {})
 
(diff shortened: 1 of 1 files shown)


brennan (approved):

dlasky:
Can you also backport to actions-cli repo plz!!

NolanChan:
@chrisradek I can backport it!

danieljackins (approved):

pooyaj (approved):
not sure why this was umd in the first place 🤷

PR #827, opened 2022-10-19
[HGI-143] - Makes the upsertCustomObject Action's objectType field a dynamic field

varadarajan-tw:
This PR makes objectType field of  HubSpot action as dynamic field to help customers avoid mistakes while filling in. In addition to custom objects from hubspot, `Deals` and `Tickets` will be supported via this action. So, they have been added to the result of the API.

This change set also removes previous validations put in place to ensure customers were entering valid data. These validations are no longer needed as customers can now chose only the options displayed in dropdown.

Note:
HubSpot API doesn't support pagination nor sort

## Testing

Custom Objects in HubSpot instance
image

Options are loaded as expected

image

Event succeeded

image


packages/destination-actions/src/destinations/hubspot/errors.ts
```diff
@@ -1,7 +1,7 @@
 import { IntegrationError, RetryableError, HTTPError } from '@segment/actions-core'
 import { SEGMENT_UNIQUE_IDENTIFIER } from './properties'
 
-export class UpsertCompanyError extends HTTPError {
+export class HubSpotError extends HTTPError {
```
brennan, line 4: Love these custom error types for Hubspot to help us understand the expected shape of error responses.

packages/destination-actions/src/destinations/hubspot/upsertCustomObjectRecord/index.ts
```diff
@@ -54,4 +53,36 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      label: schema.labels.plural,
+      value: schema.fullyQualifiedName
+    }))
+    return {
+      choices: [...choices, ...defaultChoices]
+    }
+  } catch (err) {
+    return {
+      choices: [],
+      error: {
```
brennan, line 80: Out of curiosity, do you know where/how these errors would display in the Segment UI?

packages/destination-actions/src/destinations/hubspot/upsertCustomObjectRecord/__tests__/index.test.ts
```diff
@@ -51,166 +53,134 @@ describe('HubSpot.upsertCustomObjectRecord', () => {
     expect(responses[0].status).toBe(201)
   })
 
-  // Validate creation of custom object with Segment supported HubSpot Objects
-  it('should create a Deals object with Custom Objects Action', async () => {
```
brennan, line 55: Why are we removing this test? It suggests we're removing the ability to create a Deals object, but it doesn't look like we are in the code. I may be misunderstanding something here. :)

brennan (approved):
Left a few questions, but nothing concerning/blocking. Thanks for the great work @sayan-das-in and @varadarajan-tw!

packages/destination-actions/src/destinations/hubspot/upsertCustomObjectRecord/__tests__/index.test.ts
varadarajan-tw: Hey @brennan , previously there were validations to ensure the name typed is [either deals/tickets or not](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/hubspot/upsertCustomObjectRecord/index.ts#L40). The logic is no longer needed as values are restricted at input level now and so these tests will be very similar to the basic positive case test we already have.

nick-Ag (approved):
Nice 🚀

PR #829, opened 2022-10-20
LIBWEB-1233 Braze - allow deferring initialization

zikaari:
This patch adds a new setting for Braze web mode destination, `deferUntilIndentified` which defers calling Braze SDK's `initialize`. This is useful when customers want to save their Braze MAU's.

Testing method:

- Used the Actions tester to start a dev server that compiles and servers a chosen device mode destination (Braze in this case)
- Transplanted the URL at which braze was being served by action tester into the load-script file in analytics-next repo
- Analytics-next was now consuming the local dev build of braze instead of the remotely hosted version
- Tested against `analytics-next/examples/standalone-playground/index-local` to make sure the setting was behaving as expected

### DIFF
--- a/packages/browser-destinations/src/destinations/braze/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/index.ts
@@ -299,12 +299,19 @@ export const destination: BrowserDestinationDefinition<Settings, BrazeDestinatio
         await dependencies.loadScript(`https://js.appboycdn.com/web-sdk/${version}/braze.no-module.min.js`)
       }
 
-      const shouldInit = !deferUntilIdentified || (deferUntilIdentified && typeof analytics.user().id() === 'string')
+      let initialized = false
 
       const client: BrazeDestinationClient = {
         instance: version.indexOf('3.') === 0 ? window.appboy : window.braze,
-        initialized: shouldInit,
-        initialize: () => {
+        ready: () => {
+          if (initialized) {
+            return true
+          }
+
+          if (deferUntilIdentified && typeof analytics.user().id() !== 'string') {
+            return false
+          }
+
           client.instance.initialize(api_key, {
             baseUrl: window.BRAZE_BASE_URL || endpoint,
             ...expectedConfig
@@ -324,13 +331,11 @@ export const destination: BrowserDestinationDefinition<Settings, BrazeDestinatio
 
           client.instance.openSession()
 
-          client.initialized = true
+          return (initialized = true)
         }
       }
 
-      if (shouldInit) {
-        client.initialize()
-      }
+      client.ready()
 
       return client
     } catch (e) {
--- a/packages/browser-destinations/src/destinations/braze/index.ts
+++ b/packages/browser-destinations/src/destinations/braze/index.ts
@@ -8,7 +8,7 @@ import updateUserProfile from './updateUserProfile'
 import trackPurchase from './trackPurchase'
 import debounce, { resetUserCache } from './debounce'
 import { defaultValues, DestinationDefinition } from '@segment/actions-core'
-import { BrazeType } from './braze-types'
+import { BrazeDestinationClient } from './braze-types'
 
 declare global {
   interface Window {
@@ -49,7 +49,7 @@ const presets: DestinationDefinition['presets'] = [
   }
 ]
 
-export const destination: BrowserDestinationDefinition<Settings, BrazeType> = {
+export const destination: BrowserDestinationDefinition<Settings, BrazeDestinationClient> = {
   name: 'Braze Web Mode (Actions)',
   slug: 'actions-braze-web',
   mode: 'device',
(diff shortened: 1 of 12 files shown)


mericsson:
@zikaari can you include information in the PR description with how you have tested this? Thanks!

zikaari:
Added testing method to PR description

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
```diff
@@ -149,12 +149,17 @@ const action: BrowserActionDefinition<Settings, BrazeType, Payload> = {
   },
 
   perform: (client, { payload }) => {
+    // client is initialized here when tracking unidentifed user has been disabled
+    if (!client.initialized) {
+      client.initialize()
```
pooyaj, line 154: Few issues with the approach here: 
* This `perform` method can be re-mapped to a different call that is not necessarily `identify` or `group`, so we can't assume that the user is known at this point 
* We can also have group or identify calls without setting the `analytics.user().id()`, so no guarantee of a defined user at this point either 

I think we need to change to either of these approaches: 
* Subscribe to AJS's event dispatcher and look for `identify` events 
* Or maybe in all of the `perform` calls check if `analytics.user().id()` is defined, and if so, then initialize the library

pooyaj (commented):
Added one comment, and one nit.

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
chrisradek: +1 to hooking into A.JS's event dispatcher listening for an identify event.

packages/browser-destinations/src/destinations/braze/index.ts
```diff
@@ -290,28 +299,40 @@ export const destination: BrowserDestinationDefinition<Settings, BrazeType> = {
...
-      const brazeObject: BrazeType = version.indexOf('3.') === 0 ? window.appboy : window.braze
+      const shouldInit = !deferUntilIdentified || (deferUntilIdentified && typeof analytics.user().id() === 'string')
 
-      brazeObject.initialize(api_key, {
-        baseUrl: window.BRAZE_BASE_URL || endpoint,
-        ...expectedConfig
-      })
+      const client: BrazeDestinationClient = {
+        instance: version.indexOf('3.') === 0 ? window.appboy : window.braze,
+        initialized: shouldInit,
```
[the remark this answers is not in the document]
zikaari: Good catch!

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
zikaari: Using `identify` event is not a reliable option. The `identify` event is emitted after the `this.dispatch` of the actual event, in the `then` clause. This means that there is a chance where some events immediately after the identify call can be missed by Braze.

Please review the changes I pushed which uses Pooya's second advice.

zikaari: ...unless we add another event named `will-identify`, or `before-identify` - which would be instantaneous and not wait for `dispatch` to finish.

What do you think of that?

chrisradek: I think the 2nd option is actually better - userId can be defined without identify being called - if the cookie is read or it's present as a query parameter - or it can even be set directly via `analytics.user().id('value')`

chrisradek:
Implementation looks good! Can you add some tests that ensure `openSession()` is deferred? We have tests that check that `braze` APIs get called when expected
example: https://github.com/segmentio/action-destinations/blob/7a4903222e0aca172f994a04ac91166380c477c1/packages/browser-destinations/src/destinations/braze/__tests__/trackEvent.test.ts#L7

We should be able to add a test that shows openSession()/other APIs aren't called until after `analytics.user().id()` exists.

packages/browser-destinations/src/destinations/braze/updateUserProfile/index.ts
zikaari: > userId can be defined without identify being called

In this case I should call `braze.changeUser` within `ready` function, since as you say that `identify`, thus `updateUserProfile` can be bypassed.

zikaari: And if the traits can be set by short-circuiting identify, do we need to worry about that?

zikaari:
Added test for `openSession` call

chrisradek (approved):
Implementation looks good to me! Thanks for adding that test!

pooyaj (approved):

mericsson (approved):
approving on behalf of buildexp teams for this file only `packages/destination-actions/src/destinations/twilio/sendSMS.types.ts` (since that was blocked by  https://github.com/segmentio/action-destinations/blob/main/.github/CODEOWNERS#L12)

PR #831, opened 2022-10-21
Add document to global object

pooyaj:
One of our Partners is using the npm package of their library, and the package accesses `document` directly but our CLI commands such as generate types fail to import the package as `document` is undefined on the global context. Including document in the global context to solve the issue. 

## Testing
Tested with the Partner's branch of code, and type generation works as expected with the fix.


dlasky (approved):
Interesting

PR #832, opened 2022-10-21
Change action to be a create or update, add additional fields

askkaz:
We updated the action for the Loops action destination to be a create or update action as that is going to be more useful for our users. We have no production use so far so there should be no impact on users.

## Testing


packages/destination-actions/src/destinations/loops/createContact/index.ts
```diff
@@ -1,35 +0,0 @@
-import type { ActionDefinition } from '@segment/actions-core'
-import type { Settings } from '../generated-types'
-import type { Payload } from './generated-types'
-
-const action: ActionDefinition<Settings, Payload> = {
```
rhall-twilio, line 5: Acknowledging that this is fine due to there being no customers using this integration in Production.

rhall-twilio (approved):

PR #834, opened 2022-10-24
Add new @case directive that supports lower() and upper() functions

LLuque-twilio:
This PR addresses [JIRA ACT-92](https://segment.atlassian.net/browse/ACT-92)

In it, we add a new `@case` directive that will allow mapping a string from its value in the incoming event to its equivalent `toLowercase()` or `toUppercase()` value. This change will have to be accompanied by a change in `app` that supports these new directives, but that is non-blocking since this PR needs to be implemented first.

App PR `need approval!` https://github.com/segmentio/app/pull/14554

## Testing

Added unit tests to ensure correct functionality.

Did stage deploy of integrations with these changes bundled in & ran a `yarn add @segment/actions-core`. Manually ran a `graphql` request to update the `city` field to use the new `@case` directive. Clearly the UI is not currently equipped to know how to render this new directive, but the event-preview can give you the correct value as shown below. The `city` field is not using the new directive and thus its value is "ATLANTA", whereas `city` is using the new `@case` with `operator:lower` so its value is "atlanta"

Screen Shot 2022-10-25 at 7 09 05 PM

I also did the same to an `Actions Slack` destination and then sent an event through dispatch to see if it was handled correctly. It was!

| Event sent to slack destination  | Resolved message appearing in slack |
| Screen Shot 2022-10-25 at 7 26 47 PM | Screen Shot 2022-10-25 at 7 27 18 PM |

-------

### DIFF
--- a/packages/core/src/mapping-kit/__tests__/index.test.ts
+++ b/packages/core/src/mapping-kit/__tests__/index.test.ts
@@ -203,22 +203,20 @@ describe('@if', () => {
   })
 })
 
-describe('@case', () => {
-  const payload = { a: 1, b: 'MAKE ME LOWER CASE', c: 'make me upper case', d: null, e: '', f: false }
+describe.only('@case', () => {
+  const payload = {
+    a: 1,
+    b: 'MAKE ME LOWER CASE',
+    c: 'make me upper case',
+    d: null,
+    e: '',
+    f: false,
+    g: 'mIXeD StRinG CasING W/ NuMBErs 190wdB',
+    h: { key1: 'with values', key2: ['hi', true, null] }
+  }
 
-  test('operator', () => {
+  test('handle string-input transformations', () => {
     let output = transform(
-      {
-        '@case': {
-          operator: 'lower',
-          value: { '@path': '$.a' }
-        }
-      },
-      payload
-    )
-    expect(output).toStrictEqual(undefined)
-
-    output = transform(
       {
         '@case': {
           operator: 'lower',
@@ -244,17 +242,6 @@ describe('@case', () => {
       {
         '@case': {
           operator: 'lower',
-          value: { '@path': '$.d' }
-        }
-      },
-      payload
-    )
-    expect(output).toStrictEqual(undefined)
-
-    output = transform(
-      {
-        '@case': {
-          operator: 'lower',
           value: { '@path': '$.e' }
         }
       },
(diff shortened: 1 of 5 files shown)


packages/core/src/mapping-kit/validate.ts
```diff
@@ -105,6 +105,16 @@ function validateDirectiveOrString(v: unknown, stack: string[] = []) {
   }
 }
 
+function validateString(v: unknown, stack: string[] = []) {
+  const type = realTypeOrDirective(v)
```
NolanChan, line 109: I think if it's just one case, it'd be a bit clearer to do an `if/else`!

packages/core/src/mapping-kit/__tests__/index.test.ts
```diff
@@ -203,6 +203,78 @@ describe('@if', () => {
...
+        '@case': {
+          operator: 'lower',
+          value: { '@path': '$.f' }
+        }
+      },
+      payload
+    )
+    expect(output).toStrictEqual(undefined)
+  })
+})
```
NolanChan, line 276: Could we have a test handling an error case? :)

NolanChan (commented):
Besides two suggestions, looks good overall! Will wait for a staging test and then will approve!

dlasky (changes requested):

dlasky:
Very solid initial start, I have a few nits around the testing specifically, as well as my slightly more important comment around what to do in the case where we have an unexpected type. In general I think mapping kit should be pretty forgiving so we dont lose data potentially due to a strangely formatted incoming segment event.

packages/core/src/mapping-kit/validate.ts
NolanChan: but that should go into `registerDirective` right?

LLuque-twilio: Think you are right here Nolan, the check for the specific operator will go in `registerDirective`, this should just be to ensure that the operator is a string in general!

packages/core/src/mapping-kit/__tests__/index.test.ts
```diff
@@ -203,6 +203,165 @@ describe('@if', () => {
   })
 })
 
+describe.only('@case', () => {
```
NolanChan, line 206: left an `.only` :)

packages/core/src/mapping-kit/__tests__/index.test.ts
```diff
@@ -203,42 +203,87 @@ describe('@if', () => {
...
+    let output = transform(
+      {
+        '@case': {
+          operator: 'lower',
+          value: { '@path': '$.a' }
+        }
+      },
+      payload
+    )
+    console.log('number', output)
```
NolanChan, line 285: and `console.log`'s

packages/core/src/mapping-kit/__tests__/index.test.ts
```diff
@@ -203,6 +203,162 @@ describe('@if', () => {
...
+          operator: 'upper',
+          value: { '@path': '$.g' }
+        }
+      },
+      payload
+    )
+    expect(output).toStrictEqual('MIXED STRING CASING W/ NUMBERS 190WDB')
+  })
+
+  test('handle non-string-input correctly', () => {
```
NolanChan, line 275: minor nit: clarify the test description so we know exactly what should happen, like "handle non-string-input by not modifying the value"

packages/core/src/mapping-kit/index.ts
```diff
@@ -74,6 +74,34 @@ registerDirective('@if', (opts, payload) => {
   }
 })
 
+registerDirective('@case', (opts, payload) => {
+  // let condition = false
```
NolanChan, line 78: do we need this comment?

ricardo-rossi (approved):
lgtm

PR #840, opened 2022-10-27
Fixes the push-browser-destinations command for amplitude/fullstory

chrisradek:
This PR fixes publishing browser destinations using the `./bin/run push-browser-destinations` command. Currently Amplitude/Fullstory either throw an error or changes the published name.

The command does a check to ensure that a destination's name matches the destination's creation name. If they don't match, an error is thrown. This is typically desired behavior because the `integrations` object uses the creation name.

2 destinations currently fail this check:
- Amplitude
- Fullstory

Unfortunately users may already be using the destination name to enable/disable them in analytics-next. For example, users could disable Fullstory's trackEvent action by setting `{ integrations: { 'Fullstory (Actions) trackEvent': false } }` in analytics-next since the destination name was used as the plugin name.

To maintain this compatibility for Amplitude/Fullstory, I made them part of an exception set so the creationName check doesn't apply to them.

### Why did you change Fullstory's name?

Fullstory is currently published as `Fullstory (Actions)`. This was changed in #629 to `Fullstory` in order to match the creation name but that change was never published - due to concerns about customers relying on `Fullstory (Actions)` in the integration object with analytics-next. This PR reverts that change as well.

## Testing

I tested these changes by running `./bin/run push-browser-destinations` selecting all of our destinations. With these changes, the only diff is the bundle URLs:
image

### DIFF
--- a/packages/browser-destinations/src/destinations/fullstory/index.ts
+++ b/packages/browser-destinations/src/destinations/fullstory/index.ts
@@ -17,7 +17,7 @@ declare global {
 export const segmentEventSource = 'segment-browser-actions'
 
 export const destination: BrowserDestinationDefinition<Settings, FS> = {
-  name: 'Fullstory',
+  name: 'Fullstory (Actions)',
   slug: 'actions-fullstory',
   mode: 'device',
   presets: [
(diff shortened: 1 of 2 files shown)


danieljackins (approved):
🎉

pooyaj (approved):
🚢

mericsson (approved):

PR #843, opened 2022-10-27
[STRATCONN-1664] SFMC Post Testing Improvements 

rvadera12:
1. Updated the following field labels: 

- `client_id `→ Client ID

- `client_secret` → Client Secret

- `account_id` → Account ID 

2. Updated the description and label for values for dataExtension Action 

- Label: `Data Extension Fields`

- Description: `The fields in the data extension that contain data about an event, such as Product Name, Revenue, Event Time, etc. Fields must be created in the data extension before sending data for it. On the left-hand side, input the SFMC field name exactly how it appears in the data extension. On the right-hand side, map the Segment field that contains the corresponding value.`

3. Updated `values` for both `dataExtension` & `contactDataExtension` Action to `required:true` 

## Testing


brennan:
@rvadera12 Do we need to run `./bin/run generate:types`?

brennan (approved):

PR #850, opened 2022-11-04
Add Snapchat error condition

joe-ayoub-segment:
https://segment.atlassian.net/browse/ACT-233
- Updated InputField descriptions to indicate which combination of identifier fields are required in a payload. 
- Updated perform() to prevent payload from being sent if necessary identifiers are missing

### DIFF
--- a/packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/index.ts
+++ b/packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/index.ts
@@ -68,9 +68,6 @@ const action: ActionDefinition<Settings, Payload> = {
     click_id: click_id
   },
   perform: (request, data) => {
-    const payload: Object = formatPayload(data.payload)
-    const settings: Settings = conversionType(data.settings, data.payload.event_conversion_type)
-
     if (data.payload.currency && !CURRENCY_ISO_4217_CODES.has(data.payload.currency.toUpperCase())) {
       throw new IntegrationError(
         `${data.payload.currency} is not a valid currency code.`,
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/index.ts
```diff
@@ -78,7 +78,15 @@ const action: ActionDefinition<Settings, Payload> = {
         400
       )
     }
-
+    
+    if(!data.payload.email && !data.payload.phone_number && !data.payload.mobile_ad_id && (!data.payload.ip_address || !data.payload.user_agent)) {
```
brennan, line 82: I think it may be a good idea to link to the Snapchat documentation with these requirements.

brennan:
Thanks, @joe-ayoub-segment!

Remember to run `./bin/run generate:types` to update the `generated-types` file and reflects your new field descriptions.

As a next step, let's update the unit tests to validate the new error.

packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/index.ts
brennan: Also, I think one way to improve the code would be to move this validation and the currency validation before transforming/formatting the payloads on line 71. If we're going to fail the event anyway, we might as well save the CPU time.

brennan:
Thanks, @joe-ayoub-segment!

Remember to run `./bin/run generate:types` to update the `generated-types` file and reflects your new field descriptions.

As a next step, let's update the unit tests to validate the new error.

packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/index.ts
```diff
@@ -68,25 +68,31 @@ const action: ActionDefinition<Settings, Payload> = {
...
-    
-    if(!data.payload.email && !data.payload.phone_number && !data.payload.mobile_ad_id && (!data.payload.ip_address || !data.payload.user_agent)) {
+
+    if (
+      !data.payload.email &&
+      !data.payload.phone_number &&
+      !data.payload.mobile_ad_id &&
+      (!data.payload.ip_address || !data.payload.user_agent)
+    ) {
+      console.log('GOT TO HERE------')
```
brennan, line 85: Delete `console.log`.

brennan (approved):

PR #862, opened 2022-11-09
Remove requiredness of Pardot sandbox toggle

kdaswani:
It appears that having the sandbox toggle as required causes some bugginess in the UI where you are unable to save the destination without toggling sandbox to on.  We should remove requiredness of this field since it already as a default of `false`.


marinhero (approved):

kdaswani:
Thanks for approving @marinhero! Do you think we need to update or add any tests for this, and/or test in stage?

marinhero:
@kdaswani I reviewed the rest of the codebase for this integration and the places that use it won't be needing changes. Also, the test suite explicitly passes values to this setting and it does not rely on default values. Therefore, changes in the suite are not required. 

Moving this to PR as it's ready to be merged.

marinhero:
No problem! The `Ready for Release` tag sets it in the queue for next week's release.

PR #864, opened 2022-11-10
Joe ayoub segment hgi 173

joe-ayoub-segment:
PR for https://segment.atlassian.net/browse/HGI-173 

Ticket title: "Amplitude Actions always passes Library as 'segment' regardless of the mapped value" 

## Changes

Updated logic in logPurchase, logEventV2, logEvent, identifyUser and groupIdentifyUser Actions to the following: 
 - removed default library field mapping directive
 - for mappings where no library field mapping directive is set, the library value will be set to 'segment'
 - for mappings where the library field directive is set, it will be respected

Updated logic in logPurchase, logEventV2, logEvent and identifyUser Actions to the following: 
 - pre-existing logic was setting platform field to 'Web' if the label field value returned 'analytics.js'. 
 - this logic was going to break due to deleting the default directive on the library field
 - so a new 'hidden' field called library_hidden was introduced
 - platform field is not set to 'Web' via the library_hidden field if $.context.library.name === 'analytics.js'. 

## Testing

Added unit tests in all of the affected Actions, to test for the following scenarios where applicable: 
 - 1) should set library value if library directive is configured
 - 2) should set library value to "segment" if library directive is not configured
 - 3) should set platform = "Web" if library_hidden = analytics.js

I ran all new tests and they passed. 
I also tested each change using the action-tester
I did **not** test in the Staging Environment.

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
+++ b/packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
@@ -237,10 +237,8 @@ const action: ActionDefinition<Settings, Payload> = {
     library: {
       label: 'Library',
       type: 'string',
-      description: 'The name of the library that generated the event.',
-      default: {
-        '@path': '$.context.library.name'
-      }
+      description:
+        'The name of the library that generated the event. If nothing is provided, Segment will send "segment" as the Library.'
     }
   },
 
@@ -275,7 +273,7 @@ const action: ActionDefinition<Settings, Payload> = {
       ...(userAgentParsing && parseUserAgentProperties(userAgent)),
       // Make sure any top-level properties take precedence over user-agent properties
       ...removeUndefined(properties),
-      library: 'segment'
+      library: library ?? 'segment'
     })
 
     return request(getEndpointByRegion('identify', settings.endpoint), {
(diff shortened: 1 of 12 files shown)


packages/destination-actions/src/destinations/amplitude/identifyUser/index.ts
```diff
@@ -254,8 +261,8 @@ const action: ActionDefinition<Settings, Payload> = {
       properties.platform = properties.platform.replace(/ios/i, 'iOS').replace(/android/i, 'Android')
     }
 
-    if (library === 'analytics.js') {
-      properties.platform = 'Web'
+    if (library_hidden) {
```
dlasky, line 264: extra check is redundant there

packages/destination-actions/src/destinations/amplitude/logPurchase/index.ts
```diff
@@ -243,7 +244,7 @@ const action: ActionDefinition<Settings, Payload> = {
         ...removeUndefined(properties),
         // Conditionally track revenue with main event
         ...(products.length && trackRevenuePerProduct ? {} : getRevenueProperties(payload)),
-        library: 'segment'
+        library: library ?? 'segment'
```
brennan, line 247: Does Amplitude require that we send 'segment' here? Not sure if we should reach out to validate.

PR #873, opened 2022-11-14
Create new browser destination for Wisepops

boris-hocde:
This PR adds the Wisepops browser destination.
* Initialize the Wisepops SDK
* Identify & Group to Wisepops' [custom properties](https://support.wisepops.com/article/yrdyv1tfih-set-up-custom-properties)
* Track Event to Wisepops [custom event](https://support.wisepops.com/article/zbpq1z0exk-set-up-custom-events-to-trigger-popups) & [goal tracking](https://support.wisepops.com/article/mx3z8na6yb-set-up-goal-tracking)
* Track Page to Wisepops' [page view](https://support.wisepops.com/article/uymb5lywhi-wisepops-on-single-page-applications)

## Testing


packages/browser-destinations/src/destinations/wisepops/index.ts
```diff
@@ -0,0 +1,89 @@
...
+    window.WisePopsObject = 'wisepops'
+    window.wisepops =
+      window.wisepops ||
+      function (...arg) {
+        ;(window.wisepops.q = window.wisepops.q || []).push(arg)
+      }
+    window.wisepops.l = Date.now()
+    window.wisepops('options', { autoPageview: false })
+    // Can load asynchronously, no need to wait
+    void deps.loadScript(`https://loader.wisepops.com/get-loader.js?v=1&site=${settings.websiteId}`)
```
pooyaj, line 77: This could be a non-issue depending on how get-loader.js works. But want to double check: Given that we are returning the `window.wiseops` immediately, it seems that the `wiseops` SDK would be the Wiseops stub rather than the actual SDK at the time of return. Given that we store the reference to the stub, would this be a problem in later calls to that reference? Basically the concern is while `get-loader.js` will replace the `window.wiseops` with the actual SDK, the `wiseops` reference stored by the actions engine may remain the stub forever.

packages/browser-destinations/src/destinations/wisepops/trackEvent/index.ts
```diff
@@ -0,0 +1,27 @@
+import type { BrowserActionDefinition } from '../../../lib/browser-destinations'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import type { Wisepops } from '../types';
+
+const action: BrowserActionDefinition<Settings, Wisepops, Payload> = {
+  title: 'Track Event',
+  description: "Send a [custom event](https://support.wisepops.com/article/zbpq1z0exk-set-up-custom-events-to-trigger-popups) to Wisepops. Keep in mind that events are counted as page views in your Wisepops' monthly quota.",
+  defaultSubscription: 'type = "track"',
```
pooyaj, line 9: Just a note that this will also respond to `Order Completed` events. So for each `Order Completed` events we will run two actions: `trackEvent` and `trackGoal`. Just checking if that's ok?

pooyaj (commented):
Looks great. Double checking couple of points before merging this in.

packages/browser-destinations/src/destinations/wisepops/index.ts
boris-hocde: This shouldn't be a problem. Once initialized, `get-loader.js` will replace the `window.wisepops.q.push` function, not `window.wisepops`.

pooyaj (approved):

PR #894, opened 2022-11-21
Mixpanel trackPurchase: generatePurchaseEventPerProduct default to true and two other fixes

zihejia:
This PR will change `generatePurchaseEventPerProduct` field default to true in the `trackPurchase` action. This PR also has the below changes:
- mp_lib: can now indicate the event is from Segment Actions
- $source: add better support for Mixpanel partner

## Testing


packages/destination-actions/src/destinations/mixpanel/trackPurchase/__tests__/index.test.ts
```diff
@@ -72,8 +72,12 @@ describe('Mixpanel.trackPurchase', () => {
 
     nock('https://api.mixpanel.com').post('/import?strict=1').reply(200, {})
 
+    const mapping = {
```
rvadera12, line 75: You might want to make this a global variable since it's used all over the file.

packages/destination-actions/src/destinations/mixpanel/trackPurchase/__tests__/index.test.ts
```diff
@@ -99,8 +103,6 @@ describe('Mixpanel.trackPurchase', () => {
     })
     expect(responses.length).toBe(1)
     expect(responses[0].status).toBe(200)
-    expect(responses[0].data).toMatchObject({})
-    expect(responses[0].options.json).toMatchObject([expectedOrderCompleted])
```
rvadera12, line 103: Is there a reason why the following expect statements were removed?

packages/destination-actions/src/destinations/mixpanel/trackPurchase/__tests__/index.test.ts
```diff
@@ -118,18 +120,19 @@ describe('Mixpanel.trackPurchase', () => {
...
```
rvadera12, line 122: Is there a reason why the following expect statements were removed?

packages/destination-actions/src/destinations/mixpanel/trackPurchase/__tests__/index.test.ts
zihejia: The purpose of this test is to check whether the Action will hit the correct endpoint based on the settings(EU), as long as the nock works and get 200 status, we should be good. Testing the payload will be covered by other tests.

packages/destination-actions/src/destinations/mixpanel/trackPurchase/__tests__/index.test.ts
zihejia: The same reason as above.

rvadera12 (approved):

PR #897, opened 2022-11-24
[Amplitude] calculate `revenue` based on price and quantity

prayansh:
For Amplitude `logPurchase` events, when `trackRevenuePerProduct=true`, we will now compute revenue based on the incoming `price` and `quantity` rather than ignore when `revenue` is not present 

## Testing
- [x] [@prayansh] Tested in the staging environment

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude/logPurchase/index.ts
+++ b/packages/destination-actions/src/destinations/amplitude/logPurchase/index.ts
@@ -30,12 +30,17 @@ interface EventRevenue {
 }
 
 function getRevenueProperties(payload: EventRevenue): EventRevenue {
-  if (typeof payload.revenue !== 'number') {
+  let revenue = payload.revenue
+  if (typeof payload.quantity === 'number' && typeof payload.price === 'number') {
+    revenue = payload.quantity * payload.price
+  }
+
+  if (!revenue) {
     return {}
   }
 
   return {
-    revenue: payload.revenue,
+    revenue,
     revenueType: payload.revenueType ?? 'Purchase',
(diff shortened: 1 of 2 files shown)


LLuque-twilio (approved):
I believe this new revenue computation will happen for **_any_** event, even when `trackRevenuePerProduct = false` because of [this line](https://sourcegraph.segment.com/github.com/segmentio/action-destinations/-/blob/packages/destination-actions/src/destinations/amplitude/logPurchase/index.ts?L245). 

Approving because I think it makes sense this new logic should work the same regardless of the value of `trackRevenuePerProduct`.

packages/destination-actions/src/destinations/amplitude/logPurchase/index.ts
```diff
@@ -30,12 +30,17 @@ interface EventRevenue {
 }
 
 function getRevenueProperties(payload: EventRevenue): EventRevenue {
-  if (typeof payload.revenue !== 'number') {
+  let revenue = payload.revenue
+  if (typeof payload.quantity === 'number' && typeof payload.price === 'number') {
```
mericsson, line 34: Should we additionally check that revenue is not populated? @loganluque 

  if (!revenue && (typeof payload.quantity === 'number' && typeof payload.price === 'number')) {

prayansh: i dont think we need to check for if `revenue` exists, since if  quantity and price exist then their calculation should override the `revenue` value as per our docs here https://github.com/segmentio/action-destinations/blob/00371a6464a35399a3b7fb5a7608785ea0c88f7d/packages/destination-actions/src/destinations/amplitude/event-schema.ts#L204

PR #901, opened 2022-11-29
[HGI-155] Braze/batching support

sayan-das-in:
This PR adds batching support for `Track Event` in Braze Cloud Mode (Actions)

## Testing
Testing completed successfully in local and staging.

Detailed test document is available here
https://docs.google.com/document/d/1wKwsk4nIkpFLu8IttOXK5CQoyBFoFMW6ORHKGikCvL0/edit#heading=h.e5g41ggybm7d

### DIFF
--- a/packages/destination-actions/src/destinations/braze/__tests__/braze.test.ts
+++ b/packages/destination-actions/src/destinations/braze/__tests__/braze.test.ts
@@ -1,6 +1,8 @@
 import nock from 'nock'
-import { createTestEvent, createTestIntegration, DecoratedResponse } from '@segment/actions-core'
+import { createTestEvent, createTestIntegration, DecoratedResponse, SegmentEvent } from '@segment/actions-core'
 import Braze from '../index'
+
+beforeEach(() => nock.cleanAll())
 
 const testDestination = createTestIntegration(Braze)
 const receivedAt = '2021-08-03T17:40:04.055Z'
@@ -10,7 +12,7 @@ const settings = {
   endpoint: 'https://rest.iad-01.braze.com' as const
 }
 
-describe(Braze.name, () => {
+describe('Braze Cloud Mode (Actions)', () => {
   describe('updateUserProfile', () => {
     it('should work with default mappings', async () => {
       nock('https://rest.iad-01.braze.com').post('/users/track').reply(200, {})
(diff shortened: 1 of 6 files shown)


packages/destination-actions/src/destinations/braze/__tests__/braze.test.ts
```diff
@@ -148,6 +150,89 @@ describe(Braze.name, () => {
...
+            time: '2021-08-03T17:40:04.055Z',
+            properties: {},
+            _update_existing_only: true
+          }
+        ]
+      })
+    })
   })
 
   describe('trackPurchase', () => {
```
rvadera12, line 238: Could you clarify why the `trackPurchase` unitTest needed to be updated, since from my understanding the code is still doing the same thing but we created a new function `sendTrackPurchase` that is in the `utils.ts` file. 

Could you also include two separate unit tests -> one with properties & payload and one without it (might be the answer to my question above since there is `products` added to the unit test event)

rvadera12 (commented):
Great Job! Have a couple questions about the `trackPurchase` event and might be helpful to add a couple screenshots in the testing doc to account for the changes to the `trackPurchase` action.

packages/destination-actions/src/destinations/braze/__tests__/braze.test.ts
```diff
@@ -181,22 +276,21 @@ describe(Braze.name, () => {
...
+            external_id: 'user1234',
+            app_id: 'my-app-id',
+            time: '2021-08-03T17:40:04.055Z',
+            properties: { products: [{ product_id: 'test-product-id', currency: 'USD', price: 99.99, quantity: 1 }] },
+            _update_existing_only: false,
+            product_id: 'test-product-id',
+            currency: 'USD',
+            price: 99.99,
+            quantity: 1
+          }
```
rvadera12, line 291: Could you clarify why the `trackPurchase` unitTest needed to be updated, since from my understanding the code is still doing the same thing but we created a new function `sendTrackPurchase` that is in the `utils.ts` file. 

Could you also include two separate unit tests -> one with properties & payload and one without it (might be the answer to my question above since there is `products` added to the unit test event)

sayan-das-in: I noticed that `Track Purchase` had very low Code Coverage. On further investigation it was found that there was a bug in that particular unit test which was incorrectly testing on `trackEvent` instead of `trackPurchase`.

https://github.com/segmentio/action-destinations/blob/914cbc81e561f985d050653b16a7be1c47ffc408/packages/destination-actions/src/destinations/braze/__tests__/braze.test.ts#L163

After fixing the event the test case was throwing additional errors as the old test payload was for `trackEvent`, had to change that.

packages/destination-actions/src/destinations/braze/utils.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (!braze_id && !user_alias && !external_id) {
+    throw new IntegrationError(
+      'One of "external_id" or "user_alias" or "braze_id" is required.',
+      'Missing required fields',
+      400
+    )
+  }
+
+  // Skip when there are no products to send to Braze
+  if (payload.products.length === 0) {
```
rvadera12, line 70: Could we add an extra unit test to make sure we are handling a payload with no products correctly, and just for any future updates to the code we have a unit test in place.

packages/destination-actions/src/destinations/braze/__tests__/braze.test.ts
rvadera12: Thanks for clarifying!

packages/destination-actions/src/destinations/braze/utils.ts
sayan-das-in: Added unit test to check the the condition when products array is empty.

rvadera12 (approved):

PR #913, opened 2022-12-02
[Amplitude] Fix session_id formatting in amplitude actions

hvardhan-unth:
JIRA Ticket - https://segment.atlassian.net/browse/HGI-203

Before passing the session_id in DayJs npm function to format, I have added a check that if the session_id is a string and contains only numbers then typecast it from string to number and return without formatting

Why: Because if the session_id is a string and contains only numbers then it's probably already a unix timestamp. DayJS doesn't parse unix timestamps correctly outside of the `.unix()` initializer.

This fix is similar to formatting used in [Intercom](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/intercom/util.ts#L52)

## Testing

Screenshot 2022-12-05 at 3 33 28 PM

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
+++ b/packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
@@ -348,7 +348,7 @@ describe('Amplitude', () => {
         event: 'Test Event',
         anonymousId: 'julio',
         integrations: {
-          // @ts-expect-error integrations should accept complext objects;
+          // @ts-expect-error integrations should accept complex objects;
           'Actions Amplitude': {
             session_id: '1234567890'
           }
@@ -401,7 +401,7 @@ describe('Amplitude', () => {
         event: 'Test Event',
         anonymousId: 'julio',
         integrations: {
-          // @ts-expect-error integrations should accept complext objects;
+          // @ts-expect-error integrations should accept complex objects;
           'Actions Amplitude': {
             session_id: 1234567890
           }
@@ -910,7 +910,7 @@ describe('Amplitude', () => {
         event: 'Test Event',
         anonymousId: 'julio',
         integrations: {
-          // @ts-expect-error integrations should accept complext objects;
+          // @ts-expect-error integrations should accept complex objects;
           'Actions Amplitude': {
             session_id: '1234567890'
           }
@@ -963,7 +963,7 @@ describe('Amplitude', () => {
         event: 'Test Event',
         anonymousId: 'julio',
         integrations: {
-          // @ts-expect-error integrations should accept complext objects;
+          // @ts-expect-error integrations should accept complex objects;
           'Actions Amplitude': {
             session_id: 1234567890
           }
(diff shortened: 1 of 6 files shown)


hvardhan-unth:
JIRA Ticket: https://segment.atlassian.net/browse/HGI-203

packages/destination-actions/src/destinations/amplitude/convert-timestamp.ts
```diff
@@ -0,0 +1,12 @@
+import dayjs from '../../lib/dayjs'
+
+export function formatSessionId(session_id: string | number): number {
+  // Timestamps may be on a `string` field, so check if the string is only
+  // numbers. If it is, convert it into a Number since it's probably already a unix timestamp.
+  // DayJS doesn't parse unix timestamps correctly outside of the `.unix()`
+  // initializer.
+  if (typeof session_id === 'string' && /^\d+$/.test(session_id)) {
```
marinhero, line 8: Nit: Consider checking the length of the string too. We know valid UNIX timestamps are 10 digits long so something like this would do `/^\d{10}$/`

packages/destination-actions/src/destinations/amplitude/__tests__/convert-timestamp.test.ts
```diff
@@ -0,0 +1,18 @@
+import { formatSessionId } from '../convert-timestamp'
+
+describe('Amplitude - Convert timestamp - format session_id', () => {
```
marinhero, line 3: Nice test cases! Should we consider testing for something like `'2021-06-08'` or we never see these type of dates in here?

packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
```diff
@@ -395,6 +395,59 @@ describe('Amplitude', () => {
       `)
     })
 
+    it('supports session_id from `integrations.Actions Amplitude.session_id` in number format', async () => {
+      const event = createTestEvent({
+        timestamp,
+        event: 'Test Event',
+        anonymousId: 'julio',
+        integrations: {
+          // @ts-expect-error integrations should accept complext objects;
```
marinhero, line 404:
```suggestion
          // @ts-expect-error integrations should accept complex objects;
```

Is this comment required? Should we disable a linter rule instead?

marinhero (commented):
Nice work! Left a few comments and questions that I'd like to see answered before shipping.

packages/destination-actions/src/destinations/amplitude/__tests__/amplitude.test.ts
hvardhan-unth: @marinhero Even if we disable a rule then we will be doing it for a single line temporarily. And to disable that rule we will need to add a comment. So, I think it will not make any difference.

We have corrected the spelling mistake in the comment as per your suggestion.

packages/destination-actions/src/destinations/amplitude/convert-timestamp.ts
hvardhan-unth: UNIX timestamp is usually a 10 digit integer, representing the number of seconds passed since 1970-01-01 00:00:00 UTC (midnight). Less commonly you may find them with 13 digits (3 extra digits for millisecond precision) or 16 digits (microsecond precision).

@marinhero  Since it is a possibility for it to be 10 digit or 13 digit or 16 digit, I have not added a length condition here.

marinhero (approved):

PR #923, opened 2022-12-07
Upload manifest of browser destinations

danieljackins:
This PR creates a manifest file of the browser action destinations, which will be uploaded to S3 by the CLI and then consumed by ajs-renderer to determine which versions of browser-destinations to serve. It maps the destination name to the webpack hash, and will create a manifest-<SHA> (github commit SHA) as well as a manifest-latest.

It also adds versioning to the actions-core chunk. I believe this may have caused an issue in the past where a user could load multiple destinations that were using different versions of actions-core causing conflicts, but this issue shouldn't be present in the new versioning system. It's solved by
- No longer using the url field in the actions database to determine which version to serve
- One build (manifest) will always be using the same actions-core version, and since the manifest determines which versions to serve for every action destination, it won't get out of sync
- We always upload all browser actions to S3 every time there's a push, and since we're using the manifest as the source of truth, it doesn't rely on updating every destination definition like it did previously.

Example manifest:
image

Related PRs:
https://github.com/segmentio/action-cli/pull/81
https://github.com/segmentio/analytics.js-versions/pull/782
https://github.com/segmentio/ajs-renderer/pull/391

## Testing


packages/browser-destinations/webpack.config.js
```diff
@@ -40,6 +63,7 @@ const unobfuscatedOutput = {
   mode: process.env.NODE_ENV || 'development',
   devtool: 'source-map',
   output: {
+    chunkFilename: '[name]/[contenthash].js',
```
pooyaj, line 66: Not sure about impact of this one ... but if you tested in both dev/stage then all good 😅

pooyaj (approved):
Looks good to me! I think we eventually need to update the new CLI as well: https://github.com/segmentio/action-cli/blob/main/src/browser_config.ts

packages/browser-destinations/webpack.config.js
danieljackins: This line adds versioning to the actions-core. I think it could cause a problem if this was merged and then actions-core was updated and some but not all destinations were updated to use the new actions-core (since users could then possibly load different destinations using multiple versions of actions-core and cause conflicts), but as long as the other changes are merged it should be okay (cli/ajs-renderer).

If we think it's too risky it's not completely necessary (but it does give us the ability to rollout any changes to actions-core the same way).

chrisradek: I'm trying to understand how this part of the webpack docs impact us:
> Note that these filenames need to be generated at runtime to send the requests for chunks. Because of this, placeholders like [name] and [chunkhash] need to add a mapping from chunk id to placeholder value to the output bundle with the webpack runtime. This increases the size and may invalidate the bundle when placeholder value for any chunk changes.

To be honest I'm not sure if that's a warning or just telling us our bundle might increase in size...

dlasky (approved):

PR #939, opened 2022-12-15
Update validate() method to use correct types & update tests accordingly

LLuque-twilio:
This PR addresses JIRA [ACT-74](https://segment.atlassian.net/jira/software/projects/ACT/boards/510?selectedIssue=ACT-74)

In it we update the `destination-subscriptions` `validate()` method to ensure that the operators `<` `>` `<=` `>=` are only given fields with `type === 'number'`, otherwise we return false.

This pattern is consistent with how we currently check the string operators such as `begins with`, `ends with`, `contains` are only passed a field of `type === 'string'`

## Testing

Published new pacakge of `destination-subscriptions` => installed in `app` => deployed `app` feature branch to stage => noticed that only valid events with `numbers` for the value were matching instead of ones with `strings` 🎉

### DIFF
--- a/packages/destination-subscriptions/src/validate.ts
+++ b/packages/destination-subscriptions/src/validate.ts
@@ -76,13 +76,13 @@ const validateValue = (actual: unknown, operator: Operator, expected?: string | 
     case '!=':
       return actual !== String(expected)
     case '<':
-      return Number(actual) < Number(expected)
+      return typeof actual === 'number' && Number(actual) < Number(expected)
     case '<=':
-      return Number(actual) <= Number(expected)
+      return typeof actual === 'number' && Number(actual) <= Number(expected)
     case '>':
-      return Number(actual) > Number(expected)
+      return typeof actual === 'number' && Number(actual) > Number(expected)
     case '>=':
-      return Number(actual) >= Number(expected)
+      return typeof actual === 'number' && Number(actual) >= Number(expected)
     case 'contains':
       return typeof actual === 'string' && actual.includes(String(expected))
     case 'not_contains':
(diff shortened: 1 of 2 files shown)


packages/destination-subscriptions/src/validate.ts
```diff
@@ -76,13 +76,13 @@ const validateValue = (actual: unknown, operator: Operator, expected?: string |
     case '!=':
       return actual !== String(expected)
     case '<':
-      return Number(actual) < Number(expected)
+      return typeof actual === 'number' && Number(actual) < Number(expected)
```
NolanChan, line 79: I have a few open-ended questions about this, happy to pair if you'd like, but:

1. It seems like we're changing the validation from a more lenient (forced type conversion) to a more strict. This area of the code is a bit foreign to me ~ do we run the risk of invalidating other people's already-created actions? (It seems like it never worked with strings in the first place though, so is the answer no? If so, then I'm confused how this `validate` function ever worked...)
2. How do customers know that their event's are faulty because the event trigger isn't doing a type conversion? Is it just that action-tester doesn't find any events? Or is there something more explicit? (Doesn't have to be solved here, but I'm just bemoaning the fact that it might be too obscure an error for our users)

NolanChan: Logan and I walked through the questions!

1. Logan demonstrated on the nifty [transformation tester](https://app.segment.com/logan-test-two/transformations-tester) that the FQL for comparing a string to a number was never valid in the first place. He also showed me that the `validate` function was primarily used in `app` to verify test events. Therefore, there'll be no backwards compatibility issues
2. Logan showed me that on the mappings page, the customer can use the "is test event valid" button to make sure if their events are valid. That is a reasonable way for customers to check if their events are working

NolanChan (approved):

PR #940, opened 2022-12-15
[ACT-362] adds additional bracket support to path conversion in get

dlasky:
Currently the actions path directive does not support objects with brackets in their key:

`const example = {'[im valid]':123456}`

This pull request adds support for arbitrary [] in keys. However keys that contain [] but also a numeric index or a quoted string will still be treated as individual keys:

a[txt] - treated as a single key `a[txt]`
a['txt'] treated as the keys `a`,`txt`
a[0] treated as the keys `a`, `0`

## Testing
Screen Shot 2022-12-16 at 9 34 55 AM

### DIFF
--- a/packages/core/src/get.ts
+++ b/packages/core/src/get.ts
@@ -2,6 +2,7 @@
  * Lightweight alternative to lodash.get with similar coverage
  * Supports basic path lookup via dot notation `'foo.bar[0].baz'` or an array ['foo', 'bar', '0', 'baz']
  */
+
 export function get<T = unknown>(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   obj: any,
@@ -15,7 +16,13 @@ export function get<T = unknown>(
 
   // Check if path is string or array. Regex : ensure that we do not have '.' and brackets.
   // Regex explained: https://regexr.com/58j0k
-  const pathArray = Array.isArray(path) ? path : (path.match(/([^[.\]])+/g) as string[])
+  const pathArray = Array.isArray(path)
+    ? path
+    : path
+        .split(/\[(?='|\d)|\.|(?<='|\d)]+/g)
+        .filter((f) => f)
+        .map((s) => s.replace(/'/g, ''))
+  // const pathArray = Array.isArray(path) ? path : (path.match(/([^[.\]])+/g) as string[])
 
   // Find value if exist return otherwise return undefined value
   return pathArray.reduce((prevObj, key) => prevObj && prevObj[key], obj)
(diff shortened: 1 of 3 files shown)


packages/core/src/get.ts
```diff
@@ -2,6 +2,17 @@
  * Lightweight alternative to lodash.get with similar coverage
  * Supports basic path lookup via dot notation `'foo.bar[0].baz'` or an array ['foo', 'bar', '0', 'baz']
  */
+
+let arrayRe: RegExp
+
+try {
+  arrayRe = /\[(?="|'|\d)|\.|(?<="|'|\d)]+/g
+} catch (e) {
+  //safari does not support lookbehind operator so we will default
```
dlasky, line 11: @pooyaj @chrisradek any thoughts here?

valerieernst (approved):
👏

packages/core/src/get.ts
pooyaj: Might not be a good idea if it doesn't work on Safari 🤔 could we do something like this for Safari to avoid the look behind (very untested, but we don't match the closing bracket in the first Regex, but later filter it out in the `map` part if we don't see an associated opening bracket: 
```
string.split(/\[(?="|'|\d)|\.+/g).map(i => i.replace(/(^[^\[]*)\]$/, '$1')).filter((f) => f)
        .map((s) => s.replace(/'|"/g, ''))
```

dlasky: yeah i can try to figure something out, my original approach which i think will be much slower would be to just write a lexer for this format that uses an FSM to map the various states which i know would work as well.

that said, the original issue we are trying to solve for is that for certain source types (see ticket for details) we include some text in brackets that is meant to be part of the key. I doubt we would encounter that particular issue in normal sources.

pooyaj: Ah! I see ... agree not a common case with traditional AJS sources, so I think falling back to the old behavior for Safari should be fine.

nmackey: @dlasky @pooyaj 
I don't believe this approach works for Safari.  The lookbehind causes a `SyntaxError` which prevents the script from even loading.  This doesn't appear to be something that a `try/catch` can handle.  I've been seeing a ton of errors from our Safari users the last few days and I believe I have traced it back to this.

image

You can see this error by going to the segment docs in Safari: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/persistence/ and opening up the console

nmackey: Also I commented here, b/c there doesn't seem to be a way to open issues against this repo.

dlasky: thanks for bringing this up, we're on it!

PR #942, opened 2022-12-16
Updating amplitude session to use universal storage

pooyaj:
This PR uses AJS storage layer instead of directly using localStorage. As a result, the sessionId will be stored in the cookie, and therefore, works between sub-domains. 

This is mostly subbing the `window.localStorage.{get,set}Item` with `storage.{get,set}`. Also removed conversion back and forth between `number` <--> `string` as AJS storage preserves the type of stored data. 

## Testing


packages/browser-destinations/src/destinations/amplitude-plugins/sessionId/__tests__/sessionId.test.ts
```diff
@@ -9,7 +10,7 @@ expect.extend({
     }
 
     if (typeof expected === 'string') {
-      got = parseInt(expected, 10)
+      expected = parseInt(expected, 10)
```
pooyaj, line 13: bug in the tests 😬

silesky (approved):

PR #943, opened 2022-12-19
Handles ES5 transpilation

massinissanm:
This PR allows ES5 transpilation. 
We've recently been using Sentry on our apps and it caught this exception on multiple occasions 

`SyntaxError: Unexpected token '.'
  at definition.actionssessionIdperform (webpack://[name]Destination/./src/destinations/amplitude-plugins/sessionId/index.ts:60:23`

To fix this issue I suggest to convert the code into ES5 compatible browser.

## Testing

### DIFF
--- a/packages/browser-destinations/tsconfig.build.json
+++ b/packages/browser-destinations/tsconfig.build.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "target": "ES5",
     "module": "esnext",
     "composite": true,
     "outDir": "dist",
--- a/packages/browser-destinations/tsconfig.build.json
+++ b/packages/browser-destinations/tsconfig.build.json
@@ -1,7 +1,6 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "target": "ES5",
     "module": "esnext",
     "composite": true,
     "outDir": "dist",
--- a/packages/actions-shared/tsconfig.build.json
+++ b/packages/actions-shared/tsconfig.build.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "target": "ES5",
     "composite": true,
     "rootDir": "src",
     "outDir": "dist"
--- a/packages/actions-shared/tsconfig.build.json
+++ b/packages/actions-shared/tsconfig.build.json
@@ -1,7 +1,6 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "target": "ES5",
     "composite": true,
     "rootDir": "src",
     "outDir": "dist"
--- a/packages/browser-destinations/webpack.config.js
+++ b/packages/browser-destinations/webpack.config.js
@@ -39,6 +39,7 @@ const unobfuscatedOutput = {
   entry: entries,
   mode: process.env.NODE_ENV || 'development',
   devtool: 'source-map',
+  target: ['web', 'es5'],
   output: {
     filename: (file) =>
       process.env.NODE_ENV === 'development' ? `${file.chunk.name}.js` : `${file.chunk.name}/[contenthash].js`,
@@ -123,5 +124,4 @@ const obfuscatedOutput = {
     }
   }
 }
-
 module.exports = [unobfuscatedOutput, obfuscatedOutput]
--- a/packages/destination-subscriptions/tsconfig.build.json
+++ b/packages/destination-subscriptions/tsconfig.build.json
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "target": "ES5",
+    "target": "es2018",
     "lib": ["es2018"],
     "composite": true,
     "module": "es2015",
--- a/packages/destination-subscriptions/tsconfig.build.json
+++ b/packages/destination-subscriptions/tsconfig.build.json
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "target": "es2018",
+    "target": "ES5",
     "lib": ["es2018"],
     "composite": true,
     "module": "es2015",


silesky:
Interesting that this wasn't being transpired to ES5 before. On the analytics SDK, we already transpire to ES5, so it makes sense to also transpire destinations to the same target.

packages/browser-destinations/tsconfig.build.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "target": "ES5",
```
silesky, line 4: I imagine we would want to compile destination-actions to es5 as well. If so, changing the target  in the base config https://github.com/segmentio/action-destinations/blob/master/tsconfig.json#L4 instead would make sense, as well as removing any overridden target for any deps such as https://github.com/segmentio/action-destinations/blob/main/packages/destination-subscriptions/tsconfig.build.json

massinissanm: i added it here 427444281c9b2fd03d4d493b393d9e38047b68c4

silesky (dismissed):

silesky:
@dlasky I'm approving, but the full CI test suite should be run for a change like this.

chrisradek:
Sorry to be chiming in on this late (been on vacation the last 2 weeks).

Can you share what browser environment is encountering this error? I'm wondering if it is really necessary to go all the way back to ES5, or if there's a compromise where we can go with ES2015.

I also think we should be handling the transpilation via webpack, not the tsconfig. This would only impact the web bundles then, and would not require cloud mode destinations to run ES5 code. I also think this is safer long-term in case we have dependencies outside of our control that drop ES5 support (assuming that's really needed), and lets cloud mode destinations continue to take advantage of runtime improvements that target newer ES standards.

massinissanm:
> Sorry to be chiming in on this late (been on vacation the last 2 weeks).
> 
Hi!
> Can you share what browser environment is encountering this error? I'm wondering if it is really necessary to go all the way back to ES5, or if there's a compromise where we can go with ES2015.
> 
it happened with Chrome mobile, Safari Mobile, Edge, Firefox, Chrome, Headless Chrome... basically every common browser possible.

> I also think we should be handling the transpilation via webpack, not the tsconfig. 
To be honest, that is the first thing i tried but i failed in doing so (not very familiar with webpack). But i'll give it another try today.

dlasky:
> @dlasky I'm approving, but the full CI test suite should be run against this branch for a change like this (for example, the tsc build will break if some action tries to use certain features like es6 private fields). How should we handle this?

I'm concerned about how this will impact cloud as well, I would prefer if we figured out the webpack approach since that only affects the bundles built for browser + CDN and not the actual published npm libs which are internally consumed in several other places and have a current expectation of being 2018.

We can of of course test that, but it's a fairly concerning change given the surface area that is potentially impacted.

dlasky (changes requested):
would like to see a webpack only approach here

massinissanm:
> would like to see a webpack only approach here
Hi Dan, here it is ;) 87bcc3b06c1b986997b4b2deb41a30f5e9b9f160

massinissanm:
Hi ! does the webpack only approach fits the bill ?  
Best

dlasky:
Yep, i'll approve, but given the nature of the change we'll need to do some fairly extensive testing prior to merging.

Thanks for your patience!

chrisradek:
Yes I'll take a look as well. I do think we should be able to get away with transpiling to 2015 instead of es5 - it seems like the issue this is directly fixing is around our use of the nullish operator which isn't supported until later versions of ES.

massinissanm:
> Yes I'll take a look as well. I do think we should be able to get away with transpiling to 2015 instead of es5 - it seems like the issue this is directly fixing is around our use of the nullish operator which isn't supported until later versions of ES.

Hi @chrisradek ! i was going to fix my PR and saw in Webpack documentation that

> When no information about the target or the [environment](https://webpack.js.org/configuration/output/#outputenvironment) features is provided, then ES2015 will be used.

(https://webpack.js.org/configuration/target/#:~:text=When%20no%20information%20about%20the%20target%20or%20the%20environment%20features%20is%20provided%2C%20then%20ES2015%20will%20be%20used.)

So i guess that won't fix the issue. Have you had any time to run the extensive tests you had in mind ?

massinissanm:
> > Yes I'll take a look as well. I do think we should be able to get away with transpiling to 2015 instead of es5 - it seems like the issue this is directly fixing is around our use of the nullish operator which isn't supported until later versions of ES.
> 
> Hi @chrisradek ! i was going to fix my PR and saw in Webpack documentation that
> 
> > When no information about the target or the [environment](https://webpack.js.org/configuration/output/#outputenvironment) features is provided, then ES2015 will be used.
> 
> (https://webpack.js.org/configuration/target/#:~:text=When%20no%20information%20about%20the%20target%20or%20the%20environment%20features%20is%20provided%2C%20then%20ES2015%20will%20be%20used.)
> 
> So i guess that won't fix the issue. Have you had any time to run the extensive tests you had in mind ?

Hi @dlasky , @chrisradek ! How are you doing ? Do you guys have had time to look at that PR and messages here ?
Best

massinissanm:
Hi @dlasky , @chrisradek ! How are you doing ? Do you guys have had time to look at that PR and messages here ?
Best

massinissanm:
Hi @dlasky , @chrisradek ! Could you tell me please if there's anything else i can do ?

vthibault:
Is this going to be merged ?
We missed events from 16k users because of this. We are thinking to move to server-side calls instead to avoid problems :/

silesky:
cc @SyedWasiHaider

silesky:
> @chrisradek This would only impact the web bundles then, and would not require cloud mode destinations to run ES5 code.
Since only browser-destinations/tsconfig.json, it shouldn't affect cloud mode destinations?

Can we just go back to the fix of making browser-destinations/tsconfig.json target ES2015?

massinissanm:
> > @chrisradek This would only impact the web bundles then, and would not require cloud mode destinations to run ES5 code.
> > Since only browser-destinations/tsconfig.json, it shouldn't affect cloud mode destinations?
> 
> Can we just go back to the fix of making browser-destinations/tsconfig.json target ES2015?

Can we ?

silesky:
I remember why we need to use webpack -- there are shared / core dependencies that we also want to bundle and just changing the target of this one repo won't do that -- and we don't want to have to change the tsconfig build targets of these shared dependencies.

Jackman3005:
Hey all, we're [still having issues](https://github.com/segmentio/analytics-next/issues/734#issuecomment-1672490687) that this PR was hoping to resolve. Things look like they have stalled with this PR, is it possible to get it across the finish line or is another approach required?

PR #945, opened 2022-12-20
SalesWings (Actions) Destination

denis-egorushkin-sw:
The PR adds new SalesWings (Actions) Destination to allow Segment users to connect their Segment Data to SalesWings. The destination contains a single action: `submitEvent`. The action uses SalesWings public API to send events to SalesWings. Most of the destination code is concerned with converting Segment events to SalesWings event format.

Future plans: This is the first iteration that is integrated with general purpose SalesWings API. In the next iteration, we plan to implement Segment specific endpoints in our backend, so the logic of this destination component would be greatly simplified. 

## Testing


packages/destination-actions/src/destinations/saleswings/api.ts
```diff
@@ -0,0 +1,46 @@
+export const apiBaseUrl = 'https://helium.saleswings.pro/api/core'
```
marinhero, line 1: Hey! I appreciate the types and encapsulation of variables shown in this file. Thank you for making the review process easier!

packages/destination-actions/src/destinations/saleswings/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  name: 'Saleswings (Actions)',
+  slug: 'actions-saleswings',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'custom',
+    fields: {
+      apiKey: {
+        label: 'API Key',
+        description: 'Segment.io API key for your SalesWings project',
```
marinhero, line 17: Descriptions should end with a `.`.
```suggestion
        description: 'Segment.io API key for your SalesWings project.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+import { apiBaseUrl } from '../api'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Submit Event',
+  description: '',
+  fields: {
+    customEventPropertyMapping: {
+      label: 'Custom Event Properties',
+      description:
+        'Control how Segment Track events are mapped to SalesWings custom events. In SalesWings, custom events are displayed and evaluated in the Rule Engine as "[[Name]] Payload", where "Name" is the name of the event and "Payload" is a string representing any event specific information. SalesWings Falcon Engine allows you to define rules based on this representation. To control how it is formed, provide the Segment Track event name on the left-hand side and the Track event property name on the right side. For example, Segment Track event "User Registered" with property "plan" set to "Pro Annual" will be formatted as SalesWings custom event "[[User Registered]] Pro Annual" if you configure "User Registered" on the left-side and "plan" on the right side',
```
marinhero, line 14:
```suggestion
        'Control how Segment Track events are mapped to SalesWings custom events. In SalesWings, custom events are displayed and evaluated in the Rule Engine as "[[Name]] Payload", where "Name" is the name of the event and "Payload" is a string representing any event specific information. SalesWings Falcon Engine allows you to define rules based on this representation. To control how it is formed, provide the Segment Track event name on the left-hand side and the Track event property name on the right side. For example, Segment Track event "User Registered" with property "plan" set to "Pro Annual" will be formatted as SalesWings custom event "[[User Registered]] Pro Annual" if you configure "User Registered" on the left-side and "plan" on the right side.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      type: 'object',
+      defaultObjectUI: 'keyvalue:only'
+    },
+    userId: {
+      label: 'Segment User ID',
+      description: 'Permanent identifier of a Segment user the event is attributed to',
```
marinhero, line 20:
```suggestion
      description: 'Permanent identifier of a Segment user the event is attributed to.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      type: 'string',
+      dynamic: true,
+      default: {
+        '@path': '$.userId'
+      }
+    },
+    anonymousId: {
+      label: 'Segment Anonymous User ID',
+      description: 'A pseudo-unique substitute for a Segment user ID the event is attributed to',
```
marinhero, line 29:
```suggestion
      description: 'A pseudo-unique substitute for a Segment user ID the event is attributed to.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Email',
+      description: 'Identified email of the Segment User',
+      type: 'string',
+      default: {
+        '@path': '$.traits.email'
+      }
+    },
+    type: {
+      label: 'Event Type',
+      description: 'Type of the event',
```
marinhero, line 45:
```suggestion
      description: 'Type of the event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Segment Anonymous User ID',
+      description: 'A pseudo-unique substitute for a Segment user ID the event is attributed to',
+      type: 'string',
+      default: {
+        '@path': '$.anonymousId'
+      }
+    },
+    email: {
+      label: 'Email',
+      description: 'Identified email of the Segment User',
```
marinhero, line 37:
```suggestion
      description: 'Identified email of the Segment User.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Event Type',
+      description: 'Type of the event',
+      type: 'string',
+      default: {
+        '@path': '$.type'
+      }
+    },
+    url: {
+      label: 'URL',
+      description: 'URL associated with the Page event',
```
marinhero, line 53:
```suggestion
      description: 'URL associated with the Page event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      type: 'string',
+      default: {
+        '@path': '$.properties.url'
+      }
+    },
+    referrerUrl: {
+      label: 'Referrer URL',
+      description: 'Referrer URL associated with the Page event',
```
marinhero, line 61:
```suggestion
      description: 'Referrer URL associated with the Page event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Context URL',
+      description: 'URL associated with the event in the event context',
+      type: 'string',
+      default: {
+        '@path': '$.context.page.url'
+      }
+    },
+    contextReferrerUrl: {
+      label: 'Context Referrer URL',
+      description: 'Referrer URL associated with the event in the event context',
```
marinhero, line 77:
```suggestion
      description: 'Referrer URL associated with the event in the event context.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Referrer URL',
+      description: 'Referrer URL associated with the Page event',
+      type: 'string',
+      default: {
+        '@path': '$.properties.referrer'
+      }
+    },
+    contextUrl: {
+      label: 'Context URL',
+      description: 'URL associated with the event in the event context',
```
marinhero, line 69:
```suggestion
      description: 'URL associated with the event in the event context.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Context Referrer URL',
+      description: 'Referrer URL associated with the event in the event context',
+      type: 'string',
+      default: {
+        '@path': '$.context.page.referrer'
+      }
+    },
+    userAgent: {
+      label: 'User Agent',
+      description: 'User Agent associated with the event',
```
marinhero, line 85:
```suggestion
      description: 'User Agent associated with the event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Event timestamp',
+      description: 'When the event was sent',
+      type: 'datetime',
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    eventName: {
+      label: 'Event Name',
+      description: 'Name of the Track event',
```
marinhero, line 101:
```suggestion
      description: 'Name of the Track event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'User Agent',
+      description: 'User Agent associated with the event',
+      type: 'string',
+      default: {
+        '@path': '$.context.userAgent'
+      }
+    },
+    timestamp: {
+      label: 'Event timestamp',
+      description: 'When the event was sent',
```
marinhero, line 93:
```suggestion
      description: 'When the event was sent.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Event Name',
+      description: 'Name of the Track event',
+      type: 'string',
+      default: {
+        '@path': '$.event'
+      }
+    },
+    screenName: {
+      label: 'Screen Name',
+      description: 'Screen name of the Screen event',
```
marinhero, line 109:
```suggestion
      description: 'Screen name of the Screen event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Event Properties',
+      description: 'Properties of the Track event',
+      type: 'object',
+      default: {
+        '@path': '$.properties'
+      }
+    },
+    traits: {
+      label: 'Event Traits',
+      description: 'Traits of the Identify event',
```
marinhero, line 125:
```suggestion
      description: 'Traits of the Identify event.',
```

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      label: 'Screen Name',
+      description: 'Screen name of the Screen event',
+      type: 'string',
+      default: {
+        '@path': '$.name'
+      }
+    },
+    properties: {
+      label: 'Event Properties',
+      description: 'Properties of the Track event',
```
marinhero, line 117:
```suggestion
      description: 'Properties of the Track event.',
```

marinhero (changes requested):
Fantastic PR! Thanks for going the extra mile with your tests, and types. Most of the comments I left are regarding missing `.` please address and then I can approve. Well done!

packages/destination-actions/src/destinations/saleswings/api.ts
yevsw: Thank you for acknowledging the efforts we put into the project1

yevsw: @marinhero I accepted all proposed changes

yevsw (approved):

packages/destination-actions/src/destinations/saleswings/submitEvent/index.ts
marinhero: There are still a few missing dots in this file. Can you address them as well?

packages/destination-actions/src/destinations/saleswings/api.ts
marinhero: Thanks for addressing the comments. I've identified a few more missing dots that I'd like to see fixed. After that, I'll approve the PR. Your changes will go live next Wednesday as there's a standing company-wide code freeze for the Holidays.

yevsw: I approved your recent suggestions. We could name this PR "Connecting the dots";)

marinhero: LOL Yes, I might just rename the PR 🤣  Thank you!

rhall-twilio:
Hello, this PR is marked Ready for Release but it looks like some testcases are failing. Could you correct those?

dlasky:
It appears your snapshot tests have timestamps in them so those will fail every time. Recommend hardcoding a timestamp in your test event for those snapshot tests

yevsw:
Added one more commit with hardcoded timestamp values. A frequently asked question "how it worked before" echoed staying unanswered ;)

yevsw:
@rhall-twilio It seems we deserve `Ready for Release` label again!

rhall-twilio (approved):

marinhero (approved):

joe-ayoub-segment:
Hi @yevsw, 
I hope you are doing well! 

We're making a slight change to our PR process starting this week in order to be able to provide enhanced feedback for PRs.  

If possible, I'd like to meet with you to discuss the use-case for this new Integration and to get an understanding of how SalesWings works. 

I know you are keen to get your new Destination deployed ASAP, so would it be possible to meet with me some time this week? 

I'm available Thu 5th from 14:00 to 16:00 UK time and on Fri 6th from 10:00 to 18:00 UK time. Would either of these slots work for you? 

Kind regards,
Joe

yevsw:
Hi @joe-ayoub-segment !
Yes, sure let's meet if it is necessary. Any time works for me - I am on holidays this week, so my days are basically free from meetings, and I would be happy to meet with you and discuss the questions that stayed unclear.
Let me know the time and place of the meeting.
Regards,
Yevgeny

yevsw:
H @joe-ayoub-segment !
How and when we could proceed with it?
Kind Regards,
Yevgeny

joe-ayoub-segment:
Notes from meeting with Saleswings folks: 

Recommendations: 

Handle email in both identify() and track() calls
Traits.email and properties.email 
See below for example code for doing an if then else for your defaults: 
 default: {
       '@if': {
         exists: { '@path': '$.traits.email' },
         then: { '@path': '$.traits.email' },
         else: { '@path': '$.properties.email' }
       }
     }


Add description for the Submit Event action
Currently it’s empty. Should be a 1 line description
 (description: '',)

Needs to handle defaultSubscription value for track, page, identify, screen 
Please see code example below which should be added into your submitEvent/index.ts file, into the Action 
defaultSubscription: 'type = "identify" or type = "track" or type = "screen" or type = "page"'


Need to add a Preset for the single Action. Having a Preset will save the customer time and will likely reduce Support queries. 
See below for code for how to add some Presets. This needs to go in the main index.ts file. There should be 1 X preset required in the array as you only have 1 X Action built.  
 presets: [
   {
     name: 'Track Calls',
     subscribe: 'type = "track" and event != "Order Completed"',
     partnerAction: 'trackEvent',
     mapping: defaultValues(trackEvent.fields)
   },
   {
     name: 'Order Completed Calls',
     subscribe: 'event = "Order Completed"',
     partnerAction: 'trackPurchase',
     mapping: defaultValues(trackPurchase.fields)
   },
   {
     name: 'Identify Calls',
     subscribe: 'type = "identify"',
     partnerAction: 'updateUserProfile',
     mapping: defaultValues(updateUserProfile.fields)
   }
 ]

Merge url and contextUrl fields into a single field referencing context and properties 
default: {
       '@if': {
         exists: { '@path': '$.context.url' },
         then: { '@path': '$.context.url' },
         else: { '@path': '$.properties.url }
       }
     }


We didn’t get to properly review the customEventPropertyMapping field.
It seems that this field should be used to provide event name mappings of some sort. It’s possible that this could be done in a different / better / less confusing way. 
Denis/Yevgeny to explain in more detail what this field is for. Maybe record a screenshare to explain. 
Joe to ping Tim to make sure Saleswings still gets ‘perks’ as they submitted the initial PR in December. 
Joe to attempt to review the rest of the code offline. May request meeting if needed.

joe-ayoub-segment:
Further feedback shared by email: 

1. It looks like you are not handling properties or traits which are arrays or objects. You are only handling strings, numbers and booleans. Is this deliberate? A lot of Segment customers will send data with properties containing arrays. If you don't want to support these data types that's fine. We'll just need to call this out in the documentation. 
for example here are some very common things customers put in properties and traits:  

properties: {
  item_ids_added_to_cart: ["2345", "3546557", "46567"]
}

or 

traits: {
  address: {
    city: "Paris", 
    country: "France",
    street: "Rue des Bruyers 6"
  }
}

or 

properties: {
  products: [
    {product_id: "1234", product_name: "corn flakes"},
    {product_id: "4546", product_name: "rice crispies"},
  ]
}
 
2. I noticed that your identify() handler requires the email address as an input. This is another argument for splitting out identify() into a separate Action so that you can make traits.email a required field for identify() calls. It's up to you of course but that's how I'd approach it. You could have an Action for page, track, screen, and then another Action for identify.

denis-egorushkin-sw:
Hi @joe-ayoub-segment!

I've implemented all the recommendations from https://github.com/segmentio/action-destinations/pull/945#issuecomment-1378685236.

Regarding the recommendations from https://github.com/segmentio/action-destinations/pull/945#issuecomment-1378709920, as shared via email:
1. Unfortunately, this is a limitation of our current event model, the custom properties of SalesWings events could only have primitive values (strings / numbers / booleans), so we are not able to store and process structured property values like arrays or objects. With that in mind, skipping these properties was the only solution we could come up with at the moment.
2. I agree that it could be a reasonable approach to have a dedicated action per event type, but since we already plan to greatly simplify the component code in the near future (by moving all the event conversion logic to our backend), I'd prefer if we could keep the single action for now. In the next iteration, we would definitely like to have one generic action that would send all the supported events to SalesWings BE, so I would like to avoid starting with multiple actions and then replacing them with yet another new action in the next version. So in this iteration, an Identify event without the email trait would trigger an action invocation, but wouldn't trigger any SalesWings API call. I hope it's acceptable, considering our plans. If not, we definitely can change our approach, but then maybe I'd like to talk with you a bit more about the future migration path for the users.

I've also sent you an email with a screencast link where I try to explain why we need the "Custom Property Mapping".

Please let me know what you think. Looking forward to your further feedback.

joe-ayoub-segment:
Saleswings team agreed to implement the following changes: 

Recommended changes: 
 - There would be 4 different Actions to handle track, page, screen, and identify. Some of the Actions will share common fields. 
 - There would be specific mapping fields 'kind', 'data' and 'values' fields. These are concepts that Saleswings customers understand, so they should find it intuitive to configure.  

COMMON FIELDS 
- userId
- anonymousId
- email
- url
- referrer url
- user agent
- event timestamp

Action 1: TRACK
- COMMON FIELDS
- kind ("track")
- data (event)
- values (properties)

Action 2: PAGE
- userId
- anonymousId
- url
- referrer url
- user agent
- event timestamp

Action 3: SCREEN
- COMMON FIELDS
- kind ("screen")
- data (event)
- values (properties)

Action 4: IDENTIFY
- ALL EVENTS fields
- kind ("identify")
- data (traits.email)
- values (traits)

denis-egorushkin-sw:
Hi @joe-ayoub-segment, I've implemented all the changes we've agreed on, please have a look at the last revision.

One follow-up question: compared to the initial implementation, some actions have required fields now. What's going to happen if there is an event that matches action's subscription, but the data for the required field is missing? Is it going to be silently ignored? Would the customers be able to inspect the events that matched the subscription but did not have the data?

I've looked at the codebase and found a few examples where a mapping with a required field was added, but the default value was coming from from an event property that is not considered to be required from the Segment spec point of view. So I assumed it's a normal practise. Please let me know if it should be changed in this PR.

packages/destination-actions/src/destinations/saleswings/submitIdentifyEvent/__tests__/index.test.ts
```diff
@@ -0,0 +1,135 @@
...
+          { type: 'email', value: event.traits?.email }
+        ],
+        kind: 'Identify',
+        data: 'peter@example.com',
+        timestamp: expectedTs(event.timestamp),
+        values: {}
+      })
+    })
+
+    it('should not skip an event without ids', async () => {
```
joe-ayoub-segment, line 69: quick note @denis-egorushkin-sw 
A Segment API call will always contain a userId or an anonymousId. Are you expecting Segment to forward events to Saleswings even if there is no userId and no anonymousId in the inbound payload?

denis-egorushkin-sw: Yes, I've read it in the docs. However, I've decided to add a test case, because both fields are optional, and theoretically, we could still handle an event if the email is provided. But it's not like we explicitly need it. If you think we should not even consider this behavior in tests, I can remove it.

joe-ayoub-segment (approved):

joe-ayoub-segment:
unit tests ran locally passed. 
Actions Tester looks good.

joe-ayoub-segment:
hi @denis-egorushkin-sw I've merged and flagged this for deployment tomorrow.

PR #958, opened 2023-01-05
Do not load heap if it's already loaded

murphpdx:
The heap snippet will fail if we load it more than once. This does a check to see if it's already been loaded and will bail if that's the case. 

## Testing

I don't think we can really test this one locally.


pooyaj (approved):
Looks great!

PR #976, opened 2023-01-17
Pipedrive pe 20

joe-ayoub-segment:

## Summary

This fix addresses a number of issues raised by multiple customers regarding the Pipedrive Action Destination.
See https://segment.zendesk.com/agent/tickets/493476 
See https://segment.atlassian.net/browse/PE-20 

Issue 1: Pipedrive Destination's Dynamic Fields were not working. Fix applied to point the Dynamic Fields to correct URLs
Issue 2: Pipedrive was creating new objects unnecessarily, when an upsert should have happened. This was because external identifiers were not being set when objects were first created. Fix was to ensure that external identifiers got included when creating an object
Issue 3: Default field mappings were not provided, and those which were provided were problematic. For example multiple identifier fields were mapping to the same identifier. Fix was to provide some basic default field mappings.

## Testing

Manual testing was done with Postman (to figure out the Pipedrive API), then with the Actions Tester. 
Unit tests were also updated. 
All Actions were tested against a trial version of Pipedrive.
Manual tests were done for all create and update operations for all 6 Actions. 

This has not been tested in Staging yet.


joe-ayoub-segment:
Do not merge this to Main. Another PR will be raised instead.

PR #978, opened 2023-01-18
Run actions-core tests in real browser

silesky:
#### Flow:
Webpack to bundle test and dependencies ->  polyfill test framework -> use karma to inject bundle into browser
- use jasmine to polyfill "describe" 
- use jest-browser-globals to polyfill "test.each", "expect", "jest.fn" 
- use jest-mock to polyfill "spyOn"

#### Target Environments:
- Webkit (Safari)
- Chrome

This library contains a mix of utilities used in both node or the browser, so not everything will be runnable. Only the test with the "*.iso.test.ts" mask will get run in the browser.


dlasky:
This is awesome!!!


I wonder if it makes more/additional sense to do this for browser package? Thinking then only stuff from core that's intended for browsers gets executed?

dlasky (approved):

silesky:
> This is awesome!!!
> 
> I wonder if it makes more/additional sense to do this for browser package? Thinking then only stuff from core that's intended for browsers gets executed?

Yes, this is just the first step 😉

PR #987, opened 2023-01-20
Channels 329

alfrimpong:
[CHANNELS-329](https://segment.atlassian.net/browse/CHANNELS-329)
This PR adds WhatsApp support for Twilio Engage

## Testing

- Deployed to staging
- Triggered test sms
- Triggered test whatsApp

- [x] Added unit tests
- [x] Tested in the staging environment

[CHANNELS-329]: https://segment.atlassian.net/browse/CHANNELS-329?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


joe-ayoub-segment:
Hi @alfrimpong please let me know when you'd like me to review the PR (i.e when you are done with changes).

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
```diff
@@ -0,0 +1,80 @@
...
+import { Liquid as LiquidJs } from 'liquidjs'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { IntegrationError } from '@segment/actions-core'
+import { StatsClient, StatsContext } from '@segment/actions-core/src/destination-kit'
+import { MessageSender, RequestFn } from '../utils/message-sender'
+
+const Liquid = new LiquidJs()
+
+export class SmsMessageSender extends MessageSender<Payload> {
```
dlasky, line 11: any particular reason to use a class here?

packages/destination-actions/src/destinations/engage-messaging-twilio/sendWhatsApp/whatsapp-sender.ts
```diff
@@ -0,0 +1,84 @@
...
+    try {
+      const profile = {
+        profile: {
+          traits: this.payload.traits
+        }
+      }
+
+      const mapping: Record<string, string> = {}
+      for (const [key, val] of Object.entries(this.payload.contentVariables)) {
+        mapping[key] = await Liquid.parseAndRender(val as string, profile)
```
dlasky, line 72: does liquid have any caching where we arent parsing every time?

packages/destination-actions/src/destinations/engage-messaging-twilio/utils/message-sender.ts
```diff
@@ -0,0 +1,141 @@
...
+type SendabilityPayload = { sendabilityStatus: SendabilityStatus; phone: string | undefined }
+
+export type RequestFn = (url: string, options?: RequestOptions) => Promise<Response>
+
+type MinimalPayload = Pick<
+  Payload,
+  'from' | 'toNumber' | 'customArgs' | 'externalIds' | 'traits' | 'send' | 'eventOccurredTS'
+>
+
+export abstract class MessageSender<SmsPayload extends MinimalPayload> {
```
dlasky, line 25: ditto on class usage

dlasky (approved):

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
alfrimpong: I used a class here for readability. Given that there will be more actions that use the same shared code that exist in MessageSender, it could get out of control having a bunch of branching logic in the perform function. To me, the readability benefit is worth it.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendWhatsApp/whatsapp-sender.ts
alfrimpong: contentVariables looks like this:
`{ "1": "{{ profile.traits.first_name}}", "2": "{{profile.traits.last_name}}" }`

end goal for mapping needs to be like this:
`{ "1": "John", "2": "Marston" }`

To eliminate parsing multiple times, I could probably join all the values to a string (or stringify the object), call liquid.parse on that string, then iterate through all the templates, find the corresponding key to the template in the original contentVariables object, then calling render on the template and store its value to the corresponding key.

I'm not sure how expensive is the parse function is, do you have any data on it?

packages/destination-actions/src/destinations/engage-messaging-twilio/utils/message-sender.ts
alfrimpong: A bit more context on how we arrived at this point may be useful here. You can see it by looking at the commit messages from the [the initial repo](https://github.com/segmentio/action-destinations/pull/977/commits)
We were initially going to put everything in one action, so I applied the template pattern that you see here with the classes. Then we decided to separate them into multiple actions within the same destination.

If you have an alternative strategy, i'm happy to refactor later.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendWhatsApp/whatsapp-sender.ts
alfrimpong: Actually, the tradeoff with the alternative may be higher in complexity than simply using parseAndRender in this way. I don't believe there's a caching layer available. Happy to look into other alternatives if you have any ideas.

A minor optimization i could make here with the current solution, would be to run all the parseAndRenders in one shot with Promise.all

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
dlasky: its fine to extract, but do it functionally vs class based.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendWhatsApp/whatsapp-sender.ts
dlasky: I guess i didn't realize there were multiple templates at play, if so I dont think there is much we can do there as you said.

packages/destination-actions/src/destinations/engage-messaging-twilio/utils/message-sender.ts
dlasky: extract to a set of smaller composable functions and call those as needed to avoid duplication. This removes any chance that class state will come back to bite you later, particularly around things like the request object

alfrimpong: I will refactor. Is this a blocker for being merged in? There are planned updates to this action on my plate for the upcoming few weeks.

dlasky: absolutely not, merge away! (its also really just style preference, but i do feel like there are some tangible benefits to functional)

dlasky: well, coordinate with @NolanChan prior to merging, purely around the order of stuff hes trying to do right now, but definitely not a blocker.

alfrimpong:
I added a small change around the contentVariables -- our team needed a way to bypass the reconciliation of those variables, so we use the "traitEnrichment" field as the indicator for this. Whenever we pass in enriched traits we expect the action to reconcile the contentVariables. Otherwise, we use the contentVariables directly, as is.

PR #994, opened 2023-01-24
Add browser destination tests with saucelabs 

silesky:
This configures smoke tests to run in CI on a several year old version of iOS Safari, as well as the latest Safari, Chrome, and Firefox.

image

### Goal
- Avoid breaking every destination all at once due to regressions in bundling / syntax (see https://github.com/segmentio/app/pull/14981)

#### What this tests
- Obvious Reference errors / syntax errors in our bundle resulting from missing browser-support for modern JS features (e.g. private fields, optional chaining). 

#### What this _does not_ test (intentionally)
- Behavior / business logic

#### Saucelabs Testing Flow

1. Browser destinations are webpacked via `yarn build`.
3. Static server is started on localhost port 5555, serving all destinations. For example, navigating to `http://localhost:5555?destination=braze` will serve the braze .js brundle via script tag.
4. Saucelabs Connect Proxy starts via`"@wdio/sauce-service`, proxying that localhost from 5555 -> Saucelabs Cloud.
6.  A webdriver session starts, spawning a webdriver client. The tests contain client   commands which are sent to the remote webdriver server on Sauce Labs, which are runs against the proxied local server. The destinations are served from a page containing `<script src="http://localhost-but-proxied:5555?name=${destination}"></script>`.
    - The browsers are chosen from `wdio.conf.*` 
7. Webriver commands return, and the test suite interprets the results of these commands to determine ultimate test success or failure. Test suite results are reported back to CI and Saucelabs.

#### Areas for improvement
- Testing Runtime errors that can result from breakage in our library (e.g. deps.loadScript, deps.resolveWhen). Can be addressed by either:
   - Running one of the existing suites of of one of the "isomorphic-friendly" destinations in the browser (such as adobe-target -- something that _doesn't_ use mocks or nock.)
   - unit testing these in something like Karma (see previous isomorphic test PR) with the assistance of 
   -  or creating a "test destination" where we interact with parts of the actual API via `await event.track?.(). 

## Running the tests on your machine 
1. Build dependencies and start local server

```sh
yarn browser-destinations:build &&
yarn start-destination-server
```

2. Run tests locally

```sh
yarn run test
```

3. Run tests on Sauce Labs

```sh
SAUCE_USERNAME=??? SAUCE_ACCESS_KEY=??? yarn run test:sauce
```

### DIFF
--- a/packages/browser-destinations-integration-tests/README.md
+++ b/packages/browser-destinations-integration-tests/README.md
@@ -1,0 +1,33 @@
+# Browser Destinations Integration Tests
+
+These are for tests that target the final, webpacked output of browser-destinations, and can ideally be run in multiple browsers via a device farm like Sauce Labs.
+
+These tests are meant to run in real browsers and issues that might be missed in the package unit tests, which run in a node environment.
+
+Browser targets:
+
+- latest verison of chrome
+- latest version of safari
+- latest version of firefox
+- iOS 13 (older version of safari)
+
+## Running tests
+
+1. Build dependencies and start local server
+
+```sh
+yarn browser-destinations:build &&
+yarn start-destination-server
+```
+
+2. Run tests locally
+
+```sh
+yarn run test:local
+```
+
+3. Run tests on Sauce Labs
+
+```sh
+SAUCE_USERNAME=??? SAUCE_ACCESS_KEY=??? yarn run test:sauce
+```
--- a/packages/browser-destinations-integration-tests/src/server/utils.ts
+++ b/packages/browser-destinations-integration-tests/src/server/utils.ts
@@ -1,0 +1,21 @@
+import { readdirSync } from 'fs'
+import path from 'path'
+
+const ls = (dirPath: string) =>
+  readdirSync(dirPath, { withFileTypes: true })
+    .filter((dirent) => dirent.isDirectory())
+    .map((dirent) => dirent.name)
+
+export const DESTINATIONS_DIST_WEB = path.join(__dirname, '../../../browser-destinations', 'dist', 'web')
+
+export const listDestinations = () =>
+  ls(DESTINATIONS_DIST_WEB).map((dirPath) => {
+    const destinationDirPath = path.join(DESTINATIONS_DIST_WEB, dirPath)
+    const fileName = readdirSync(destinationDirPath).find((el) => el.endsWith('js'))
+    if (!fileName) throw new Error('Invariant: no .js file found.')
+    return {
+      dirPath: dirPath,
+      fileName,
+      filePath: path.join(destinationDirPath, fileName)
+    }
+  })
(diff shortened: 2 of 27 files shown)


NolanChan:
Hi Seth, I saw the merge conflict. I think the only thing I did to the file was change [browser-test timeout minutes to 10 minutes](https://github.com/segmentio/action-destinations/commit/d3cd4fbd48b29760fff6bced2ad401694e793087#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR71)

silesky:
> Hi Seth, I noticed the merge conflict; I think the only thing I did to the file was change [browser-test timeout minutes to 10 minutes](https://github.com/segmentio/action-destinations/commit/d3cd4fbd48b29760fff6bced2ad401694e793087#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR71)

That's fine, I also tweaked it.

packages/browser-destinations/tsconfig.json
```diff
@@ -12,6 +12,12 @@
       "@segment/destination-subscriptions": ["../destination-subscriptions/src"]
     }
   },
-  "include": ["src", "test"],
```
silesky, line 15: By leave the "include" blank, we add proper support for files such as the `config.ts` files at the root. Should be safe, as does not affect build at all.

packages/browser-destinations-integration-tests/README.md
```diff
@@ -0,0 +1,33 @@
+# Browser Destinations Integration Tests
```
dlasky, line 1: 🔥

packages/browser-destinations-integration-tests/src/server/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { readdirSync } from 'fs'
+import path from 'path'
+
+export const ls = (dirPath: string) =>
+  readdirSync(dirPath, { withFileTypes: true })
```
dlasky, line 5: i think we have glob as a dev dependency, might clean this up a tiny bit.

packages/destination-subscriptions/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-	"extends": "./tsconfig.build.json",
-	"exclude": [],
-	"include": ["src"]
+  "extends": "./tsconfig.build.json",
+  "include": ["src"],
+  "exclude": ["dist"]
```
dlasky, line 4: @pooyaj i wonder if this is why core was getting bundled? nice catch @silesky

dlasky (approved):

packages/destination-subscriptions/tsconfig.json
silesky: Thanks. I noticed these issues because the root "yarn lint" was taking minutes instead of seconds.

PR #996, opened 2023-01-24
Pipedrive actions PE-20

joe-ayoub-segment:
## Summary
 - Updating all 6 Actions to add default field mappings where reasonable. Corrected some default mappings which would have caused customer confusion. 
 - Fixed the broken Dynamic Fields on Actions which had Dynamic Fields. They were requesting to incorrect URLs. 
 - Updated payloads being sent to Pipedrive to handle when users provide Pipedrive entity IDs or when users provide external identifiers for entities.  

## Testing
 - Tested locally with Actions Tester
 - Signed up for trial Pipedrive account and did manual tests for all of the following: 
 -- Creating a Node, Updating a Node 
 -- Creating a Person, Updating a Person
 -- Creating an Organization, Updating an Organization
 -- Creating a Deal, Updating a Deal
 -- Creating a Lead, Updating a Lead
 -- Creating a Activity, Updating a Activity
 
- For all of the above tested the Dynamic Fields and made sure that related entities were connected correctly. 

- Deployed to Stage and Tested. Noticed some basic issues which are now fixed in this PR.

### DIFF
--- a/packages/destination-actions/src/destinations/pipedrive/createUpdateDeal/index.ts
+++ b/packages/destination-actions/src/destinations/pipedrive/createUpdateDeal/index.ts
@@ -9,9 +9,9 @@ import { addCustomFieldsFromPayloadToEntity } from '../utils'
 const fieldHandler = PipedriveClient.fieldHandler
 
 const action: ActionDefinition<Settings, Payload> = {
-  title: 'Create or Update a Deal',
+  title: 'Create or update a Deal',
   description: "Update a Deal in Pipedrive or create it if it doesn't exist yet.",
-  defaultSubscription: 'type = "track"',
+  defaultSubscription: 'type = "track" and event = "Deal Upserted"',
   fields: {
     deal_match_field: {
       label: 'Deal match field',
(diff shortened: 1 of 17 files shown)


packages/destination-actions/src/destinations/pipedrive/createUpdateDeal/index.ts
```diff
@@ -188,6 +207,10 @@ const action: ActionDefinition<Settings, Payload> = {
         400
       )
     }
+    if (!deal.id)
+      if (payload.deal_match_field && payload.deal_match_value)
```
rhall-twilio, line 211: /nit Would prefer:
    `if (!deal.id && payload.deal_match_field && payload.deal_match_value)`

rhall-twilio: Same for similar pattern in remaining files.

rhall-twilio (approved):
LGTM but may want partner's sign-off before merging

joe-ayoub-segment:
Also did a review with Nolan.

PR #1000, opened 2023-01-26
refactor get() tests

silesky:
Building on some work that @dlasky merged in earlier today, I thought it would be good to re-unify the get() tests and, in my opinion, bring a bit more clarity to what's actually happening with this function.


packages/core/src/get.ts
```diff
@@ -27,15 +27,18 @@ export function get<T = unknown>(
   // Check if path is string or array. Regex : splits the path into valid array of path chunks
   // we support an extra edge case that lodash does not around brackets
   // a['b'] = ['a','b']
-  // a[b] = ['a[b]'] //brackets without numeric index or a string are considered part of the key
```
silesky, line 30: ['a[b]'] Doesn't seem to work, so I removed this -- cc @dlasky

packages/core/src/get.ts
```diff
@@ -27,7 +27,6 @@ export function get<T = unknown>(
...
```
silesky, line 30: `a[b] = ['a[b]']` was wrong, so I removed the whole line (though the comment itself kind of seems correct, it's worded slightly confusing and doesn't seem to relate to the expression in a way that I understood).

dlasky (approved):

PR #1002, opened 2023-01-26
fixing a couple of issues with new Ironclad destination

joe-ayoub-segment:
## Description
Ironclad is a new Destination which was deployed yesterday. No customer is using it. 
After deployment we noticed that there were some mapping issues which needed fixing in the settings and Action mappings. 

These are: 

Change group_id type from string to integer 
 
Fix event_name field 
 - correct the default from  
default: { '@path': '$event' },
 To  
default: { '@path': '$.event' },
 
Fix sid setting field
 - remove the default line 

Fix staging_endpoint setting field
 - remove the required field line 

Fix test_mode setting field
 - remove the required field line 

## Testing
 - Testing will be done in production once deployment has taken place.

### DIFF
--- a/packages/destination-actions/src/destinations/ironclad/index.ts
+++ b/packages/destination-actions/src/destinations/ironclad/index.ts
@@ -16,7 +16,6 @@ const destination: DestinationDefinition<Settings> = {
         description:
           'Site Access ID. An ID that’s unique for each site within your account. Information on finding your sid can be found in the authentication section.',
         type: 'string',
-        default: '',
         required: true
       },
       staging_endpoint: {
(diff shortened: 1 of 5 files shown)


packages/destination-actions/src/destinations/ironclad/index.ts
```diff
@@ -16,23 +16,20 @@ const destination: DestinationDefinition<Settings> = {
...
         type: 'string',
-        default: '',
         required: true
       },
       staging_endpoint: {
         label: 'Staging Endpoint',
         description:
           'Turn this ON, to send requests to the staging server, ONLY if Clickwrap support instructs you to do so.',
         type: 'boolean',
-        required: true,
```
dlasky, line 27: curious why we are removing these? just because its falsey?

sirosenf: it is a checkbox that does not have to be true, however requiring it forced the value to be true in order to save the destination settings.

PR #1004, opened 2023-01-26
fix for pe-52

joe-ayoub-segment:
## Description
The code in this change was created by a Heap Engineer ( @murphpdx ). 
I added a minor tweak for "Page Viewed". 

This PR fixes an issue where events to Heap may not contain an event name. This caused an incident sev-2-crooked-squirrel. 

## Testing
Testing was done locally with Actions Tester. The Heap Engineer was present when tested. 
Note: there is only 1 customer using this Integration, and the Integration is in Private Beta.

### DIFF
--- a/packages/destination-actions/src/destinations/heap/trackEvent/index.ts
+++ b/packages/destination-actions/src/destinations/heap/trackEvent/index.ts
@@ -148,15 +148,26 @@ const action: ActionDefinition<Settings, Payload> = {
 }
 
 const getEventName = (payload: Payload) => {
+  let eventName: string | undefined
   switch (payload.type) {
     case 'track':
-      return payload.event
+      eventName = payload.event
+      break
     case 'page':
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -148,15 +148,26 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      return payload.name
+      eventName = payload.name ? payload.name : 'Screen Viewed'
+      break
     default:
-      return undefined
+      eventName = 'track'
+      break
   }
+
+  if (!eventName) {
```
marinhero, line 167: Good with having the extra check given that the lack of `eventName` could make or break a request. 👍

marinhero (approved):

marinhero (changes requested):
Let's fix the broken tests first. Sorry for the confusion.

marinhero (approved):

PR #1011, opened 2023-01-31
[HEAP-38485] Move to the integrations endpoint

murphpdx:
We want to move to the integrations API. This allows us to leverage additional features that our customer has requested. 

## Testing

I've done some manual tests using the `./bin/run serve`


packages/destination-actions/src/destinations/heap/heapUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (anonymous_id) {
+    return {
+      anonymous_id
+    }
+  }
+  throw new IntegrationError('Either identity or anonymous id are required.')
+}
+
+export const getEventName = (payload: Payload) => {
```
murphpdx, line 25: There shouldn't be any changes to this function. I'm just moving to the utils file.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -78,15 +83,6 @@ const action: ActionDefinition<Settings, Payload> = {
         '@path': '$.timestamp'
       }
     },
-    session_id: {
```
murphpdx, line 81: We're intentionally no longer supporting session_id. Our customer isn't using it so it should be safe to remove.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -113,61 +109,33 @@ const action: ActionDefinition<Settings, Payload> = {
...
     }
 
-    if (payload.anonymous_id && !payload.identity) {
-      heapPayload.user_id = getHeapUserId(payload.anonymous_id)
-    }
-
-    if (payload.identity) {
-      heapPayload.identity = payload.identity
-    }
+    event.user_identifier = getUserIdentifier({ identity: payload.identity, anonymous_id: payload.anonymous_id })
```
murphpdx, line 122: This PRs intention is to support merging users from the web destination with the cloud destination.  I've taken out setting the user id because we weren't using it. The user_identifier object only supports one property. We would like to update that in the future.

joe-ayoub-segment:
acknowledging this PR. Will review ASAP.

joe-ayoub-segment:
hi @murphpdx , 
Given this is is a significant change with the potential to break stuff I think we should do the following: 

Check the customer's appetite for having a breaking change deployed. i.e check if they are OK with losing data for an unspecified period of time. If the customer is not OK with any interruption to their data flow then I recommend we create a completely new Integration which will run alongside the existing one. We'll then test the new Integration and if it's working well we'll ask the customer to switch over to using it. We'll then decommission the pre-existing Integration. 

However if the customer is OK with having their data flow interrupted for an unspecified period of time then we can merge your PR and deploy. 

Regardless of which option we go with I'd like to have a face to face code review with you where we can walk the code and fully test out it out using sample payloads from the customer's Segment Workspace. For example we could take some representative payloads from the customer's Workspace and run them through the Tester tool and into an instance of Heap. We can then check in Heap if the correct entities are getting updated, etc. 

As a next step would you be OK clarifying with the customer which approach we should go with please? 
After that we can schedule a testing session. Feel free to book that using this [Calendly link](http://calendly.com/joe_ayoub/integration-workshop).

joe-ayoub-segment:
As discussed with Ling Toh and Amanda we'll reject this PR and open a new PR for a new Heap Integration.

PR #1012, opened 2023-01-31
Ankit/ga4 web new

rvadera12:

## Testing


packages/browser-destinations/src/destinations/index.ts
```diff
@@ -45,7 +45,4 @@ register('635ada35ce269dbe305203ff', './logrocket')
 register('6372e18fb2b3d5d741c34bb6', './sabil')
 register('6372e1e36d9c2181f3900834', './wisepops')
 register('637c192eba61b944e08ee158', './vwo')
-register('638f843c4520d424f63c9e51', './commandbar')
```
SyedWasiHaider, line 48: Is this intentional?

rvadera12: Nope, just added back `commandbar` and `ripe`

packages/browser-destinations/src/destinations/google-analytics-4-web/setConfigurationFields/index.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+    if (payload.campaign_content) {
+      gtag('set', { campaign_content: payload.campaign_content })
+    }
+    if (payload.user_id) {
+      gtag('set', { user_id: payload.user_id })
+    }
+    if (payload.user_properties) {
+      gtag('set', { user_properties: payload.user_properties })
+    }
```
zikaari, line 144:
```suggestion
    const properties = [
      'screen_resolution',
      'page_title',
      'page_referrer',
      '...'
    ]

    gtag('set', Object.fromEntries(
      properties
        .map((key) => [key, payload[key]])
        .filter(([, value]) => !value)
    ))
```

packages/browser-destinations/src/destinations/google-analytics-4-web/types.ts
```diff
@@ -0,0 +1,4 @@
+export type GA = {
+    //dataLayer: any
```
zikaari, line 2: Should we just remove it?

zikaari (commented):
Nice work!

Can someone help me understand why the tests create a "local" subscription model instead of testing against the "real" default subscription that the users are likely going to be facing?

Also, I'm not seeing any event filtering _(based on the `event` property)_ across the actions' `defaultSubscription` that'd discern one action from another at runtime, as seen [here ](https://github.com/segmentio/action-destinations/blob/main/packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts#L9). How does the destination express which partner should be triggered out of the many available?

packages/browser-destinations/src/destinations/google-analytics-4-web/setConfigurationFields/index.ts
zikaari: The `properties` key array might as well be type decorated assuming that type is available

packages/browser-destinations/src/destinations/google-analytics-4-web/index.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (settings.enableConsentMode) {
+      window.gtag('consent', 'default', {
+        ad_storage: settings.defaultAdsStorageConsentState,
+        analytics_storage: settings.defaultAnalyticsStorageConsentState,
+        wait_for_update: settings.waitTimeToUpdateConsentStage
+      })
+    }
+    const script = `https://www.googletagmanager.com/gtag/js?id=${settings.measurementID}`
+    await deps.loadScript(script)
+    return window.gtag
```
pooyaj, line 173: When `loadScript` completes, does it replace `window.gtag` with a different object? If so, we may need to make sure that `window.gtag` points to the full `gtag` library rather than the stub using `await deps.resolveWhen ...`. Might not be necessary, but worth double checking

packages/browser-destinations/src/destinations/google-analytics-4-web/setConfigurationFields/index.ts
silesky: do we have a polyfill for fromEntries? I think we might want to use reduce instead...

pooyaj (approved):
Looks great 🙌  Only one minor comment to verify.

packages/browser-destinations/src/destinations/google-analytics-4-web/__tests__/addPaymentInfo.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+]
+describe('GoogleAnalytics4Web.addPaymentInfo', () => {
+  test('Basic Event with Default Mappings', async () => {
+    const [event] = await googleAnalytics4Web({
+      ...setting,
+      subscriptions
+    })
+    jest.spyOn(destination, 'initialize')
```
pooyaj, line 51: In all of the tests can we do this instead: 
```
jest.spyOn(destination, 'initialize').mockImplementation(() => {
... // see examples of these in other web destinations
})
```

Because currently we are running the initialize method, and actually trying to load the GA tag from Google in each run of each test 😬 Makes tests flaky, and they fail with no network connects!

packages/browser-destinations/src/destinations/google-analytics-4-web/index.ts
```diff
@@ -0,0 +1,199 @@
...
+      cookie_update: settings.cookieUpdate,
+      cookie_domain: settings.cookieDomain,
+      cookie_prefix: settings.cookiePrefix,
+      cookie_expires: settings.cookieExpirationInSeconds,
+      cookie_path: settings.cookiePath,
+      allow_ad_personalization_signals: settings.allowAdPersonalizationSignals,
+      allow_google_signals: settings.allowGoogleSignals
+    }
+
+    window.dataLayer = window.dataLayer || []
```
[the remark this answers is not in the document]
rvadera12: Going to leave this the same so it matches exactly to google's implementation.

rvadera12:
Thanks for reviewing the PR @zikaari! 
> 
> Can someone help me understand why the tests create a "local" subscription model instead of testing against the "real" default subscription that the users are likely going to be facing?
>
Good point! There is not specific reason, the mappings actually do all match up for the most part. However we were using the google analytics 4 cloud mode tests as a framework and implemented the tests in the same method. I will make a ticket to fix this, but will move forward without changing it in this PR. 

> Also, I'm not seeing any event filtering _(based on the `event` property)_ across the actions' `defaultSubscription` that'd discern one action from another at runtime, as seen [here ](https://github.com/segmentio/action-destinations/blob/main/packages/browser-destinations/src/destinations/braze/trackPurchase/index.ts#L9). How does the destination express which partner action should be triggered out of the many available?

Great catch, I updated the event properties for the default subscriptions, for all of the actions that have a corresponding event name that matches the Segment Spec!

PR #1016, opened 2023-02-01
[HEAP-38485] Move to the integrations endpoint

murphpdx:
We want to move to the integrations API. This allows us to leverage additional features that our customer has requested.

## Testing

I've done some manual tests using the ./bin/run serve


packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -78,15 +83,6 @@ const action: ActionDefinition<Settings, Payload> = {
         '@path': '$.timestamp'
```
joe-ayoub-segment, line 83: Let's have a chat about the anonymousId: 
* It has a default mapping, but should it? 
* Heap only accepts a single identifier, so what happens if a user starts off with only an anonymousId, but then they sign in or register and now have a userId and anonymousId? Will this become 2 separate user profiles in Heap?

packages/destination-actions/src/destinations/heap/heapUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const getEventName = (payload: Payload) => {
+  let eventName: string | undefined
+  switch (payload.type) {
+    case 'track':
+      eventName = payload.event
+      break
+    case 'page':
+      eventName = payload.name ? payload.name : 'Page Viewed'
```
joe-ayoub-segment, line 32: What if we allowed customers to do something along these lines? Reason for this would be to give the customer flexibility to specify how we treat page and screen names. 

Options for a 'Home' page view eventName could be: 
 1) Home Page Viewed
 2) Home
 3) Page Viewed  

The choice of format could be via an additional field in the Action (dropdown list called 'Page Name Format', with a default selected).

Some customers have thousands of different page names, so they could end up with a lot of different events being tracked to Heap. They might want to keep it simple and just have 'Page Viewed' instead. Or they might want to be able to differentiate between a real interaction style events (.i.e track() events) and a page events, which could be accomplished by appending the string 'Page Viewed' to the end of the event name.  

crappy pseudo code below: 

let name : "";
if(!payload.name)
  name = 'Page Viewed'
else {
  switch(payload.pageNameFormat){
    case : 'name + Page Viewed': 
       name = `{payload.name} Page Viewed`
       break;
    case : 'name':
       name = `payload.name`
       break;
    case : 'Page Viewed':
       name = 'Page Viewed'
    break; 
  }
}

eventName = name

joe-ayoub-segment (dismissed):
Hi @murphpdx , I listed out some items for us to chat about when we meet. 
I'll also record a Loom Video and attach it to this PR to point out a couple of things with the identify() Action (which is not part of this PR).

joe-ayoub-segment:
Hi @murphpdx , I recorded a quick Loom video to highlight a couple of things in the identify() Action. Let's chat about it later during our call. 

https://www.loom.com/share/81b18432a9ae4cf2a3f45c1377c519d3

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
murphpdx: We chatted about this. The current code is the intended behavior.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -39,10 +47,7 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       allowNull: true,
       description:
-        'An identity, typically corresponding to an existing user. If no such identity exists, then a new user will be created with that identity. Case-sensitive string, limited to 255 characters.',
-      default: {
-        '@path': '$.userId'
-      }
+        'An identity, typically corresponding to an existing user. If no such identity exists, we will set the anonymous id property on the user.'
```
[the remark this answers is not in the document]
murphpdx: We chatted about this, I've updated the description.

murphpdx:
> Summary from video:
> 
> * will it be confusing to customers if the identify Action has a userId field, and the track Action has an identity field?
> * the logic in the identify Action seems confusing and possibly incorrect (though Amanda said the code needs to be changed)

The logic in the track call is correct. I plan to update the identify action.

joe-ayoub-segment:
Hi @murphpdx , 

Identifiers - in the new track() Action there's a single identifier field called 'user_identifier'. Is this field supposed to be for an anonymous identifier, or for a known user identifier? With most Integrations I've worked with there are always separate fields used for collecting anonymous Vs known user identifiers. It looks like you're using the 'identity' field if it's populated, with a fallback to the 'anonymousId' field. Could you help me understand this better please? 

Testing - we didn't do end to end testing to demonstrate that the new track() Action won't negatively impact the current customer's data collection. Would you be open to having a session to cover this?

There's an overhead for making changes to the repository (code reviews, etc). Would it be preferable to get the changes for the identify() call into the same PR, or are you not planning to make these changes in the near future?

murphpdx:
> Hi @murphpdx ,
> 
> Identifiers - in the new track() Action there's a single identifier field called 'user_identifier'. Is this field supposed to be for an anonymous identifier, or for a known user identifier? With most Integrations I've worked with there are always separate fields used for collecting anonymous Vs known user identifiers. It looks like you're using the 'identity' field if it's populated, with a fallback to the 'anonymousId' field. Could you help me understand this better please?

This is intentional business logic on the Heap side. We talked about this already in our last call. It's not scalable to explain the logic of every PR. Is there potential for this change to have an adverse impact on the segment side?

> Testing - we didn't do end to end testing to demonstrate that the new track() Action won't negatively impact the current customer's data collection. Would you be open to having a session to cover this?
This integration is still in Beta and it's a POC for the customer. They are ok with interruptions. I'd love it if you tested the PR's asynchronously but doing it over zoom is going to slow things down to an unacceptable level for the customer. 

> There's an overhead for making changes to the repository (code reviews, etc). Would it be preferable to get the changes for the identify() call into the same PR, or are you not planning to make these changes in the near future?

I would prefer to make additional changes in a subsequent PR. It's better to have smaller PRs. Smaller PRs are proven to ensure fewer bugs as the writer and reviews are able to catch them easier. It's generally better practice to release smaller changes more often. It's easier to isolate problematic changes and roll them back.

joe-ayoub-segment (approved):
Logan H reviewed this and approved it to be merged.

murphpdx:
> Partner declined to apply recommendations.

Hi Joe, what recommendations did we decline? I didn't see any actionable feedback?

schung507:
@joe-ayoub-segment was this deployed?

joe-ayoub-segment:
Hi @schung507 , @murphpdx this has been deployed now.

PR #1019, opened 2023-02-03
[HGI-306] - Add new error class and introduce error codes enum

varadarajan-tw:
In an effort to improve error handling and traceability for HGI destinations (more info [here](https://docs.google.com/document/d/1Q18pNSFVDvuvqacHg2y7CCoO1T2spaKl5To5Bzefuqw/edit#)), I did an audit on errors in the codebase, classified them to various categories and introduced new error classes for builders to build.

To bring in consistency to error codes and enable faster debugging, this PR adds:

- `PayloadValidationError` - Indicates error with fields in the payload. This can be used for validations performed within perform or performBatch. This 
- `Error Codes` - The `code` field of error is important as it shows up in Datadog and in Event Delivery tab. `Error Type` field is written to traces by `centrifuge` and `code` is set to  [`Error Type`](https://github.com/segmentio/integrations/blob/e833bc7508711d407ac7f95f7644f6c7af29dadf/Service/cloudevents.js#L873) field if not explicitly defined which happens in most of the cases. The enum values provide standard error code string for some of the common errors. 
- Updates some of the HGI/STRATCONN destinations that can use these new error codes.
- In addition to this, the descriptions for these new error codes have been added to Contentful. These descriptions should give additional context in debugging. Refer [this](https://docs.google.com/document/d/1BeMsDu19yCZdJ1WZzM31JadOpvdq2xyFyhrBeiJbuRU/edit) doc to view the new descriptions.

**Notes**:

- For AJVValidationErrors, the code field is set to 400 in [integrations](https://github.com/segmentio/integrations/blob/e833bc7508711d407ac7f95f7644f6c7af29dadf/createActionDestination/index.js#L316). If we can classify it as `INVALID PAYLOAD` error, then we will be able to clearly distinguish between segment errors and HttpErrors caused by partner APIs.  - Will verify if the assumption is true.
Here is the related PR - https://github.com/segmentio/integrations/pull/2521

## Testing

All unit tests are passing. There are no functional changes. Only the error codes have been refactored and these new codes should provide additional details for debugging due to to new descriptions set in [contentful](https://docs.google.com/document/d/1BeMsDu19yCZdJ1WZzM31JadOpvdq2xyFyhrBeiJbuRU/edit). Some of these errors are difficult to reproduce in staging. Here is a snapshot of `payload_validation_failed`.

image


packages/core/src/errors.ts
```diff
@@ -62,9 +63,36 @@ export class RetryableError extends CustomError {
...
     super(message)
   }
 }
+
+/**
+ * Error for invalid field values
+ * Should include a user-friendly message.
+ * These errors will not be retried and the user has to fix the incorrect field mapping
+ */
+export class ValidationError extends IntegrationError {
```
mericsson, line 78: Nice, great add!

packages/core/src/errors.ts
```diff
@@ -62,9 +63,36 @@ export class RetryableError extends CustomError {
...
+  /**
+   * @param message - a human-friendly message to display to users
+   * @param code - an optional error code/reason
+   * @param status - an optional http status code (e.g. 400)
+   */
+  constructor(message = '') {
```
mericsson, line 84: Should we default an actual message like 'Validation error' ?

mericsson: (Or is there a way in typescript to require uses of this to put a message? I'm unsure.)

packages/core/src/errors.ts
rhall-twilio, line 78: Wonder if there's a way to pass a hint to the user of what field precisely failed validation.

packages/core/src/errors.ts
varadarajan-tw: Thanks for the comment. I removed the default empty string value. It is better to leave it as required param for the builder provide one.

packages/core/src/errors.ts
varadarajan-tw: Thanks for the comment Robert. Could you give an example on how the hints should be?
I added this class is to bring in consistency for `code` value and leave the error messages for builders to provide based on context.

rhall-twilio: I believe the way the error message display works is a mapping of the error code. i.e, for `MISCONFIGURED_FIELD` it would show `Segment was unable to perform this request because a field in your settings is not configured correctly. Please revise them and try again` So the builder doesn't really get to pass out an error message that gets displayed in the UI, it only displays in the logs (someone can double check me?).

If so, it would be great if could allow the builder to pass `new ValidationError("The value of personas_computation_key and personas_audience_key must match.", "personas_computation_key, personas_audience_key")` so that the error in the UI can be upgraded to look like: 

`Segment was unable to perform this request because the field "personas_computation_key, personas_audience_key" in your settings is not configured correctly. Please revise them and try again`. 

I'm based this comment on a lot of assumptions about the Event Delivery errors, but at least that's what the _implementation looks like to me_. Think it's worth reviewing the implementation.

varadarajan-tw: Thanks Robert! You are correct! The content for the generic description in the delivery overview is retrieved from [`contentful`](https://github.com/segmentio/gateway-api/blob/a023e5c619c0d278e119d4ffb03f16cf706adfb6/src/graphql/integrations/Integration.ts#L351). The customer given specific error messages can be only seen in `Response from Destination` tab. 
Some notes on Event Delivery [here](https://docs.google.com/document/d/1F2JxTZNk-MBmieCIe0wmvwdLr0n9yH2qMfcMBgAnMbM/edit#). (WIP) 

image

This means when we introduce new error codes, we also have to update the contentful with appropriate description to ensure proper error message is shown. Thanks for pointing this out.

On interpolating fields with contentful error messages: 
- We have to persist these fields in `clickhouse` (We will have to make changes to `integration`,  `centrifuge-visibility` and few other services in the persistence path)
- Alter the queries to retrieve the messages by errorType and fields 
- interpolate with contentful messages

This doesn't look straightforward and we can look at this option in the long term as it would require co-ordination across multiple teams. I ll this to our backlog.

packages/destination-actions/src/destinations/google-analytics-4/addPaymentInfo/index.ts
```diff
@@ -42,11 +42,7 @@ const action: ActionDefinition<Settings, Payload> = {
     // Google requires that `add_payment_info` events include an array of products: https://developers.google.com/analytics/devguides/collection/ga4/reference/events
     // This differs from the Segment spec, which doesn't require a products array: https://segment.com/docs/connections/spec/ecommerce/v2/#payment-info-entered
     if (payload.items && !payload.items.length) {
-      throw new IntegrationError(
```
rhall-twilio, line 45: I think it might be worth checking if we lose a lot of detail from the error message as we currently do not have MisconfiguredField in `contentful`. If that is the case, maybe we hold off merging and using the new error types until we populate some error message there.

varadarajan-tw: Sure Robert. I am trying to get access to the right `contentful` project hosting these templates. On the other hand, even the current error codes do not have description set. For eg; `Misconfigured required field` shows up like this in event-delivery.
image

I am trying to get access to the right project in contentful and I doubt the descriptions have been updated in sometime.

packages/core/src/errors.ts
varadarajan-tw: I have added new descriptions to contentful. [This](https://docs.google.com/document/d/1BeMsDu19yCZdJ1WZzM31JadOpvdq2xyFyhrBeiJbuRU/edit) doc contains the new error codes and their description that has been set in contentful. For displaying hints to the user, i am working on a [change](https://github.com/segmentio/app/compare/error-handling?expand=1) in the app.

packages/destination-actions/src/destinations/google-analytics-4/addPaymentInfo/index.ts
varadarajan-tw: Marking this as resolved. Now that i have added Contentful messages, we should have additional information with these changes.

rhall-twilio:
We should update the README to provide new guidance on how to use these errors.

marinhero (commented):
Code looks good. Make sure to update the README and update the interface document with the guidelines that indicate how to use this.

packages/core/src/errors.ts
```diff
@@ -62,9 +63,35 @@ export class RetryableError extends CustomError {
...
 
   constructor(message = '') {
     super(message)
   }
 }
+
+/**
+ * Error to indicate the payload is missing fields that are required.
+ * Should include a user-friendly message.
+ * These errors will not be retried and the user has to fix the payload or .
```
brennan, line 76:
```suggestion
 * These errors will not be retried and the user has to fix the payload.
```

packages/core/src/destination-kit/index.ts
```diff
@@ -581,7 +581,7 @@ export class Destination<Settings = JSONObject> {
       const oauthSettings = getOAuth2Data(settings)
       const newTokens = await this.refreshAccessToken(destinationSettings, oauthSettings)
       if (!newTokens) {
-        throw new InvalidAuthenticationError('Failed to refresh access token')
+        throw new IntegrationError('Failed to refresh access token', ErrorCodes.OAUTH_REFRESH_FAILED)
```
brennan, line 584: When thinking about how to scale integrations moving forward, the more we can abstract away implementation from third-party builders, the better. So, I wonder if we should continue to use `InvalidAuthenticationError` here? This may be a good idea - because if we use the `IntegrationError` class, we would additionally need to pass in `401` as the status code.

packages/core/src/errors.ts
```diff
@@ -62,9 +63,35 @@ export class RetryableError extends CustomError {
  */
 export class InvalidAuthenticationError extends CustomError {
   status = 401
-  code = 'invalid_authentication'
+  code = ErrorCodes.INVALID_AUTHENTICATION
```
brennan, line 66: We could do something like the below. This would allow us to continue using the `InvalidAuthenticationError` class for all of our authentication errors. This approach also makes it clear that all auth-related errors should have a status code of 401.

```suggestion
  code = this.code || ErrorCodes.INVALID_AUTHENTICATION
```

brennan, line 66: We could do something like the below. This would allow us to continue using the `InvalidAuthenticationError` class for all of our authentication errors. This approach also makes it clear that all auth-related errors should have a status code of 401.

```suggestion
  code = this.code || ErrorCodes.INVALID_AUTHENTICATION
```

packages/core/src/destination-kit/index.ts
varadarajan-tw: Thanks for catching this. Fixed the issue by allowing users to set error code  for  `InvalidAuthenticationError`. The status code will remain `401`.

packages/core/src/errors.ts
varadarajan-tw: Made ErrorCodes.INVALID_AUTHENTICATION as the default value and it can be overriden through constructor param.

varadarajan-tw:
Thank you for all your suggestions - @marinhero , @rhall-twilio and @brennan . I have addressed the comments and also added an error handling guide as suggested.

rhall-twilio (approved):

docs/error-handling.md
```diff
@@ -0,0 +1,46 @@
...
+Couple of examples are
+
+- A field is marked as `required` in the `ActionDefinition` and the event payload doesn't contain the field.
+- A field's type is marked as `datetime` in the `ActionDefinition` but the field has a value of different type in the event payload.
+
+### Http errors
+
+All Http Errors resulting from API calls are automatically captured by the framework. The `request` object used for making HTTP API calls has an option called `throwHttpErrors`. In its default state of `true`, the `request` object throws an `HttpError` which are captured and handled automatically by the framework. Http Errors are retried depending on the status code. Refer [error.ts](../packages/core/src/errors.ts) for retryable status codes.
+
+In case you wish to override this behavior, you can set `throwHttpErrors` explicitly to `false`. The `request` object would then not throw HttpError and return the `response` along with status code. You can then chose to throw your own version of the error or proceed with your open implementation
```
brennan, line 22:
```suggestion
In case you wish to override this behavior, you can set `throwHttpErrors` explicitly to `false`. The `request` object would then not throw HttpError and return the `response` along with status code. You can then choose to throw your own version of the error or proceed with your open implementation.
```

docs/error-handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Authentication errors
+
+Errors due to invalid access tokens, refresh tokens or api keys are captured as `InvalidAuthenticationError`. If the API you are integrating with returns the standard Authentication error codes, they are handled as defined in [Http Errors](error-handling.md#http-errors) section.
+
+## Custom error handling
+
+The inbuilt error handling should help you with most of the error scenarios you face. In cases where the validation capabilities provided by the framework is not sufficient or the error messages returned by the API you are interacting with are not helpful, you can chose to capture and throw your own custom errors. It is important to follow the following guidelines to ensure your errors are structured properly and displayed to your destination users for appropriate action.
```
brennan, line 30:
```suggestion
The inbuilt error handling should help you with most of the error scenarios you face. In cases where the validation capabilities provided by the framework are not sufficient or the error messages returned by the API you are interacting with are not helpful, you can chose to capture and throw your own custom errors. It is important to adhere to the following guidelines to ensure your errors are structured properly and displayed to your destination users for appropriate action.
```

docs/error-handling.md
```diff
@@ -0,0 +1,46 @@
...
+
+- DO NOT throw Javascript `Error` objects. Any error thrown from an action MUST contain a `message` describing the error, an `error code` indicating the type of error and a `status code` indicating the http status of the action. You MUST use predefined error classes defined in [error.ts](../packages/core/src/errors.ts). These classes help in capturing the necessary information in appropriate format. For example, assume that your action needs one of product id or product name. This kind of validation is not currently supported in `Action Definition`. Instead of `throw new Error('One of product id or name is required')`, use `throw new PayloadValidationError('One of product id or name is required').
```
brennan, line 32:
```suggestion
- DO NOT throw Javascript `Error` objects. Any error thrown from an action MUST contain a `message` describing the error, an `error code` indicating the type of error and a `status code` indicating the http status of the action. You MUST use predefined error classes defined in [error.ts](../packages/core/src/errors.ts). These classes help in capturing the necessary information in appropriate format. For example, assume that your action needs one of product id or product name. This kind of validation is not currently supported in `Action Definition`. Instead of `throw new Error('One of product id or name is required')`, use `throw new PayloadValidationError('One of product id or name is required')`. 
```

NolanChan (approved):
This is going to be a huge step towards better observability, thank you @varadarajan-tw !

PR #1023, opened 2023-02-06
Twilio Studio as a Segment Action Destination

aradhakrishnan-twilio:
First PR for Twilio Studio as an Action Destination

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/twilio-studio/triggerStudioFlow/index.ts
+++ b/packages/destination-actions/src/destinations/twilio-studio/triggerStudioFlow/index.ts
@@ -40,6 +40,12 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'hidden',
       default: { '@path': '$.userId' }
     },
+    anonymousId: {
+      label: 'Anonymous ID',
+      description: 'A Distinct External ID',
+      type: 'hidden',
+      default: { '@path': '$.anonymousId' }
+    },
     eventType: {
       label: 'Event type',
       type: 'hidden',
@@ -58,16 +64,20 @@ const action: ActionDefinition<Settings, Payload> = {
         400
       )
     }
-    if (!payload.userId) {
+    if (!payload.userId && !payload.anonymousId) {
       throw new IntegrationError(
-        'Unable to trigger Studio Flow. No User ID identifier found for this Segment profile!',
+        'Unable to trigger Studio Flow. No User identifier found for this Segment profile!',
         `No userId found in the Segment Event`,
         400
       )
     }
     if (
       stateContext &&
-      !shouldSendToStudio(`${payload.flowSid}_${payload.userId}`, stateContext, payload.coolingOffPeriod!)
+      !shouldSendToStudio(
+        `${payload.flowSid}_${payload.userId || payload.anonymousId}`,
+        stateContext,
+        payload.coolingOffPeriod!
+      )
     ) {
       throw new IntegrationError(
         `Unable to trigger Studio Flow. Only 1 request allowed per Flow SID - User ID combination in ${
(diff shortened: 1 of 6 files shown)


packages/destination-actions/src/destinations/twilio-studio/triggerStudioFlow/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!payload.userId) {
+      throw new IntegrationError(
+        'Unable to trigger Studio Flow. No User ID identifier found for this Segment profile!',
+        `No userId found in the Segment Event`,
+        400
+      )
+    }
+    if (
+      stateContext &&
+      !shouldSendToStudio(`${payload.flowSid}_${payload.userId}`, stateContext, payload.coolingOffPeriod!)
```
joe-ayoub-segment, line 70: hi @aradhakrishnan-twilio , 
When we eventually go live with the ID Sync capability, it's entirely possible and commonplace that API calls will be sent to this Integration which do not have the userId field populated. In this case the cache capability will probably not work. 

For this reason I suggest we fall back to the anonymousId also. Or, we could use the phone number instead of the userId in the cache key.

joe-ayoub-segment (changes requested):
hi @aradhakrishnan-twilio , 

Thanks for the PR! 

During our last call we discussed a few things, but it looks like you've made the call to not include them, which is totally fine. 

These are: 
1. Need to handle EU or US Engage Space  - I think it's OK to exclude this as we probably won't launch when using Personas as the phone source. 

2. Need to handle userId fallback to anonymousId when fetching phone from Engage - While I think the anonymousId isn't required for fetching the phone from Personas, I think it might be needed as part of your cache key when there is no userId in the payload. Please see the inline comment. 

Other than that I think we're good. I didn't actually test the code myself but I know that you'll be testing it extensively.

packages/destination-actions/src/destinations/twilio-studio/triggerStudioFlow/index.ts
aradhakrishnan-twilio: Thanks @joe-ayoub-segment , we are doing a userId check in the line 61, so it won't reach here unless the payload has the userId.
Also, are you suggesting that it is a possibility that with ID Sync feature, the userid field might not be present anymore in the event payload ? Won't that be a backward compatibility issue for other destinations as well?
We considered keeping phone as the cache key but our PM brought up a scenario where multiple profiles can have same phone numbers and we didn't want to have cooldown between different profiles triggering the action, that's why we went with the userId.
With that being  said, if userId is going to be removed from the payload as part of the ID Sync for sure, then we might need to think of a different cache key for sure.
Do you think we can do a deployment with this right now and during the private beta once we are sure the ID sync is going to affect the userId, we can take a call? Otherwise I am totally fine to take this up with the team tomorrow for further discussion? Thoughts?

joe-ayoub-segment: Hi @aradhakrishnan-twilio , 

Yes it's possible (and likely) that some Segment events won't contain a userId at all. Those user profiles are considered 'anonymous', and only have an anonymousId. 

This sort of thing is very comment. For example, imaging if a user visits a website for the first time ever, they will be assigned an anonymousId. That user might then fill in a form and provide their phone number in the form. Now the user profile in Segment has an anonymousId and a phone number. But they still do not have a userId. However the customer may still want to be able to opt that user into a Twilio flow.

aradhakrishnan-twilio: Made the changes.

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @aradhakrishnan-twilio , 
Just FYI the deployment didn't go out today due to an issue unrelated to your PR. It should be deployed later today.

joe-ayoub-segment:
Hi @aradhakrishnan-twilio , 
Unfortunately the deployment yesterday was cancelled (nothing related to your PR). The team plans to release next week. I'll confirm the date when I know more. Apologies for the inconvenience.

PR #1032, opened 2023-02-08
Try to only pass in NODE_OPTIONS for node18+

SyedWasiHaider:
In certain environments, we don't quite yet have access to node18 so we have to use what we've got. Unfortunately, if you provide `NODE_OPTIONS=--openssl-legacy-provider`, then node will complain with `--openssl-legacy-provider is not allowed in NODE_OPTIONS` if not supported (as opposed to just ignoring it).

This PR adds the NODE_OPTIONS env var conditionally by detecting the node version being used. Not the cleanest approach but this should get us unblocked in the short term.

## Testing


packages/destination-subscriptions/package.json
```diff
@@ -17,7 +17,7 @@
     "test": "jest",
     "typecheck": "tsc -p tsconfig.build.json --noEmit",
     "prepare": "yarn build",
-    "size": "size-limit"
+    "size": "bash scripts/size.sh"
```
LLuque-twilio, line 20: Im confused on why we are changing this one - it doesn't look like we use `NODE_OPTIONS=--openssl-legacy-provider` in it to begin with?

LLuque-twilio (approved):
One comment - otherwise all the copy looked right to me

NolanChan:
To clarify (and for posterity) - Will it be until we can use Node 18 on the workbenches that we can delete these scripts? Or are there other environments you know of that are blocking?

packages/destination-subscriptions/package.json
SyedWasiHaider: yeah it should've actually had it but because [this](https://github.com/segmentio/action-destinations/blob/main/.github/workflows/ci.yml#L64-L66) runs conditionally, I never actually ran into the error I'm seeing now in the CI without it

SyedWasiHaider:
Yeah this is mostly for workbench @NolanChan

PR #1031, opened 2023-02-08
anoymousId in segment to be tracked as device_id in mixpanel

dneralla:
just as the way user information is dependent on anonymousID and userId in segment, in the new version of idmerge its dependent on device_id and user_id in mixpanel

PR #1033, opened 2023-02-08
Mixpanel - identify null id fix

spjtls9:
`identify` for Mixpanel destination calls Mixpanel's `/identify` and `/engage` APIs. `/engage` requires `distinct_id` in order to modify user property. In the mapping, we had `user_id` directly assigned to `distinct_id`.

 Some `identify` calls are made without `user_id` hence this call is not captured in user's Mixpanel project. Since some users are intentionally using `identify` without `user_id` and expect to use Segment's `anonymous_id` as unique identifier, their events were not captured.

This PR updates the mapping to accommodate such use case. 

## Testing


zihejia (approved):
LGTM

joe-ayoub-segment:
hi @spjtls9 , asknowledging your PR! I'll schedule time to get a review done ASAP.

spjtls9:
@joe-ayoub-segment Is there any ETA for a review? This PR fixes a customer escalation and we'd like to communicate a timeline to them

joe-ayoub-segment:
Hi @spjtls9 , I'm just getting to this PR and the other Mixpanel PRs ([1031](https://github.com/segmentio/action-destinations/pull/1031) and [1035](https://github.com/segmentio/action-destinations/pull/1035) now. 

Given this is a change to an identifier would you be available to catch up with me to discuss the change, the potential impact on customers, etc? Would you be OK scheduling a call with me via this [Calendly URL](http://calendly.com/joe_ayoub/integration-workshop)? 

Perhaps we could sync with @dneralla at the same time, as they have a PR open too ([1031](https://github.com/segmentio/action-destinations/pull/1031))? I'll comment on the 1031 PR also so @dneralla is across this request.

spjtls9:
@joe-ayoub-segment I booked some time on your calendar. Since this is an urgent change for us I booked the earliest available spot and this does not work well for PT in US, so you might need a separate meeting with @dneralla

packages/destination-actions/src/destinations/mixpanel/identifyUser/index.ts
```diff
@@ -91,7 +103,7 @@ const action: ActionDefinition<Settings, Payload> = {
       }
       const data = {
         $token: settings.projectToken,
-        $distinct_id: payload.user_id,
+        $distinct_id: payload.distinct_id,
```
joe-ayoub-segment, line 106:
```suggestion
        $distinct_id: payload.user_id ? payload.user_id : payload.anonymous_id
```

joe-ayoub-segment (changes requested):
Hi @spjtls9 , thanks for taking the time to catch up just now. Here are note from our call: 

- Customer escalation - more urgent 
- PR is to allow traits to be collected and upserted to profiles in Mixpanel even when a userId is not present in the payload (of the identify() call). 
- Current behaviour is that an anonymous identify() call is sent to Mixpanel by Segment, but is rejected by Mixpanel. 
- This PR will change behaviour for customers insofar as anonymous identify() calls will now result in new profiles being created in Mixpanel
- Sonya did explain that a subsequent identify() call containing the anonymousId and userId will not result in split profiles in Mixpanel, as Mixpanel will merge the userId and anonymousId into the same profile.  
- This PR would cause a incident if approved as it adds a new defacto ‘required’ field called distinctId
- Recommend instead leveraging the pre-existing userId and anonymousId fields in the perform function to set the distinct_id field in the outbound payload. Something like the code below: 

`$distinct_id: payload.user_id ? payload.user_id : payload.anonymous_id`

packages/destination-actions/src/destinations/mixpanel/identifyUser/index.ts
spjtls9: addressed this comment in the latest commit -- used null coalescing

joe-ayoub-segment (approved):
Met with Sonya and discussed this PR. Sonya removed a new Field and instead will use values from existing fields. The only change is to do with logic for determining which value to send as the main identifier to Mixpanel (distinct_id). This change will result in anonymous events being accepted by Mixpanel. Sonya indicated that this would be OK.

joe-ayoub-segment:
Hi @spjtls9 , this PR has now been deployed. Please verify and let me know if there are any issues!

spjtls9:
our support team was able to verify that the fix is working as expected. Thank you!

PR #1037, opened 2023-02-09
Prevent trapping calls to `then` on returning CommandBar's SDK object

tkainrad:
## Change description

Unfortunately, our earlier PR #1009 does not fix all cases of problems with initializing our browser destination that we have seen during private beta.

The problem is that some customers do not initialize CommandBar with the snippet that is included in the destination but use an older version of the CommandBar SDK initialization snippet. Back then, we initialized a `Proxy` that trapped all field access. This is a problem when the SDK object is the return value of a promise, as that will implicitly access the object's `then` field and cause the Promise to never resolve. 

The implemented solution is to explicitly set `then` to `undefined` on the returned SDK object.

## Testing

### DIFF
--- a/packages/browser-destinations/src/destinations/commandbar/index.ts
+++ b/packages/browser-destinations/src/destinations/commandbar/index.ts
@@ -58,8 +58,9 @@ export const destination: BrowserDestinationDefinition<Settings, CommandBarClien
 
     await deps.resolveWhen(() => Object.prototype.hasOwnProperty.call(window, 'CommandBar'), 100)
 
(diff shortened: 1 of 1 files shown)


lchoward:
Hey @joe-ayoub-segment ! Just checking if anything is holding this up or would be helpful to share here.

tkainrad:
@pooyaj Anything we can do to help streamline the review of this PR? It's currently blocking some mutual customers to start using our integration.

silesky (approved):
LGTM

PR #1048, opened 2023-02-15
Implementation for Outfunnel Action Destination

tammesalu:
Implementation for Outfunnel Action Destination.
- unit tests added
- tested manually with local server

## Testing


joe-ayoub-segment:
hi @tammesalu , 

It looks like this is for a new Integration. We've recently changed our review procedure for new Integration PRs to include a face to face meeting. The reason for the call is to help the reviewer to get a better understanding the following:
- use-cases for this Integration
- how identifiers should be mapped
- which types of data will be collected and sent to the Destination

This background context will help when reviewing the code in the PR.

Would you be OK scheduling a call with me via [Calendly](http://calendly.com/joe_ayoub/integration-workshop) please?

joe-ayoub-segment:
Hi @tammesalu , 

Thanks for your time just now. I sent you an email with recommended changes. I've summarised below: 

- remove console.log
- Strong recommendation to add a field to collect the action type or action name
- for trackEvent Action - change the email default mapping
- for trackEvent Action - Recommend adding anonymousId field into track() events
- for trackEvent Action - Recommend upserting a user profile via a track() event which contains email address
- for identifyContact Action - Recommend changing the email default mapping
- for groupIdentifyContact Action - Recommend changing the email default mapping

We also discussed an alternate approach to trait and property fields where you could specify exactly which fields in traits or properties which you'd like to collect.

tammesalu:
Hi @joe-ayoub-segment 

Ive submitted improvements. Sent detailed list of changes via email.

joe-ayoub-segment:
Hi @tammesalu , 
Thanks for applying the changes. 

I'd recommend the following: 

1. for your forwardIdentifyEvent Action - I could change the default path from '@path': '$.email' to '@path': '$.traits.email'
2. Up to you, but I'd advice the following: For all 3 of your Actions I would make traits and properties objects visible. This will allow the customer to opt to send you only the data they want/need to send. It will also give them more visibility over the data you are collecting. Also, if the customer decides to send an identify() call to your forwardTrackEvent Action then they'll be able to specify which object they want to set as traits. 

Apart from that it looks great!

tammesalu:
Hey Joe.
I've implemented the suggestions and also added presets. This should definitely help the user to get started with the destination faster. 
I also fixed the tests.

By they way, yarn install is failing:

error An unexpected error occurred: "ENOENT: no such file or directory, open '/home/raul/.cache/yarn/v6/npm-@segment-control-plane-service-client-9.3.0-e5cd6e7cd335cf00640e43a5ece17a290e003ddd/node_modules/@segment/control-plane-service-client/.yarn-metadata.json'".

That does not seem to be a problem caused by my code changes.

packages/destination-actions/src/destinations/outfunnel/forwardTrackEvent/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  fields: {
+    action: {
+      type: 'hidden',
+      required: true,
+      description: 'Indicates which action was triggered',
+      label: 'Action name',
+      default: 'track'
+    },
+    event_name: {
+      type: 'hidden',
```
joe-ayoub-segment, line 19: Probably best to make this visible, because: 
1. Only track() events ever have a $.event field. 
2. If the customer changes the subscription from 'type = track' to something else, data collection will break because there won't be anything at the $.event location in the payload.

packages/destination-actions/src/destinations/outfunnel/forwardIdentifyEvent/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    action: {
+      type: 'hidden',
+      required: true,
+      description: 'Indicates which action was triggered',
+      label: 'Action name',
+      default: 'identify'
+    },
+    user_id: {
+      type: 'hidden',
+      required: true,
```
joe-ayoub-segment, line 20: Should userId be a required field? 
Sometimes customers fire identify() calls containing just traits, an anonymousId and no userId. 
If you have userId as being required then these 'anonymous' identify() calls will be rejected.

joe-ayoub-segment: Remember, with Segment you will always either get a userId, or an anonymousId or both in every API call

packages/destination-actions/src/destinations/outfunnel/forwardGroupEvent/index.ts
```diff
@@ -0,0 +1,90 @@
...
+    action: {
+      type: 'hidden',
+      required: true,
+      description: 'Indicates which action was triggered',
+      label: 'Action name',
+      default: 'group'
+    },
+    user_id: {
+      type: 'hidden',
+      required: true,
```
joe-ayoub-segment, line 20: Should userId be a required field?
Sometimes customers fire group() calls containing just the groupId, traits, an anonymousId and no userId.
If you make userId required then your Integration will reject any group() call that doesn't have an anonymousId (which can happen).

packages/destination-actions/src/destinations/outfunnel/forwardGroupEvent/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      description: 'The time the event occured in UTC',
+      label: 'Event timestamp',
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    traits: {
+      type: 'object',
+      required: true,
```
joe-ayoub-segment, line 62: Querying if traits is a required field. Do you intend to reject the payload if no traits are provided?

joe-ayoub-segment (changes requested):
Hi @tammesalu , 
Couple minor questions - please see inline. 
If you can get the items addressed we should be able to deploy this Wedneday. 
Confirmed tests pass. 
Thank you! 
Joe

packages/destination-actions/src/destinations/outfunnel/forwardTrackEvent/index.ts
tammesalu: Fixed

packages/destination-actions/src/destinations/outfunnel/forwardIdentifyEvent/index.ts
tammesalu: Thanks. user_id should definitely not be a required field. Fixed.

packages/destination-actions/src/destinations/outfunnel/forwardGroupEvent/index.ts
tammesalu: Great suggestion. I removed required for all traits, properties, and context fields

packages/destination-actions/src/destinations/outfunnel/forwardGroupEvent/index.ts
tammesalu: Also removed required option from here.

joe-ayoub-segment:
hi @tammesalu this PR has been deployed.

PR #1051, opened 2023-02-16
Improve docs about starting the actions tester for browser destinations

stephennancekivell:
Hi, this PR just tweaks some of the developer documentation about running the actions tester for browser destinations.

I dont know if im doing something wrong, but the actions tester would always crash loading a destination If I didnt run `yarn dev` in `packages/browser-destinations`. It would be nice if this happened automatically but it didnt seem to.

Heres a screenshot of me running the actions tester without running `yarn dev`. After I click "Send Test Event" I tries to load `http://localhost:9001/dist/web/adobe-target.js` and fails.
image

Anyway, this PR just mentions that you need to start `yarn dev`. 
Thanks

### DIFF
--- a/packages/browser-destinations/README.md
+++ b/packages/browser-destinations/README.md
@@ -28,6 +28,15 @@ bin/run generate:types
 
 You can use our Action Tester environment to test your browser actions:
 
+First start webpack in destinations.
+
+```
+cd packages/browser-destinations
+yarn dev
+```
+
+Then start the Action Test environment
+
 ```
 ./bin/run serve --directory ./packages/browser-destinations/src/destinations --browser
 ```


varadarajan-tw:
Hey @stephennancekivell, thank you for raising this PR. 

Apologies, it took us some while to get back to this. It is not required to run `yarn dev` to start the webpack server. The `./bin/run serve` command should start it automatically. We noticed that the webpack server start was failing silently after Node 18 upgrade. We have fixed this now via this [PR](https://github.com/segmentio/action-destinations/pull/1256).

Do let us know if you have any questions or issues using the browser dev command after the update.

stephennancekivell:
Thanks.

PR #1076, opened 2023-02-28
[Mixpanel] Remove unused mapping in /track

spjtls9:
This PR removes `user_properties` from mixpanel properties. This property is not used in `/track` and causes user confusion. Since this field was essentially a no-op, no unit test was broken; there is no change in user experience either, except that `user_properties` won't show up in mapping UI.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
@@ -40,8 +40,7 @@ export const eventProperties: Record<string, InputField> = {
   insert_id: {
     label: 'Insert ID',
     type: 'string',
-    description:
-      'A random id that is unique to an event. Mixpanel uses $insert_id to deduplicate events.',
+    description: 'A random id that is unique to an event. Mixpanel uses $insert_id to deduplicate events.',
     default: {
       '@path': '$.messageId'
     }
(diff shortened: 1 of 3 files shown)


joe-ayoub-segment:
Hi @spjtls9 , acknowledging this PR. I'll get to is ASAP, however it will likely be next week before this is deployed.

joe-ayoub-segment (commented):
Hi @spjtls9 , 
Looks like this change is just removing user_properties field from the trackEvent and trackPurchase Actions, which makes sense. 

Looks like a safe change to me. 

Quick question before I approve it: 

Is the reason why you are removing this because:
1. The Mixpanel API is unable to accept user profile details via the /import API call?
OR
2. Segment's track() event specification doesn't make it possible to pass traits in a Segment track() call?

Reason 2 wouldn't be a valid reason, as it is possible to pass traits in a Segment track() call, however they need to be passed in the $.context.traits object. 

I'm assuming your reason for removing this field is reason 1. Is that correct?

Thank you!

spjtls9:
Hi @joe-ayoub-segment -- thanks for a quick reply!

You are correct. The reason for this change is that Mixpanel's `/import` API does not accept user properties. User properties are only handled in "identify."

Which reminds me, I will also have to update the help doc (https://segment.com/docs/connections/destinations/catalog/actions-mixpanel/). I will do a follow up for that in a separate task.

joe-ayoub-segment (approved):
Approved.

joe-ayoub-segment:
hi @spjtls9 this PR has been deployed. Please confirm that you are happy with the changes.

spjtls9:
Confirmed that things are as expected. Thank you!

PR #1092, opened 2023-03-06
[STRATCONN-1895] [Facebook-CAPI] Add partner_id and partner_name fields to user_data

varadarajan-tw:
This PR adds partner_id and partner_name fields to all Facebook CAPI actions. Meta requested us to add these fields so that customers who use identity partners of Meta like Liveramp can calcualte EMQ more accurately.

## Testing

[Testing Doc](https://docs.google.com/document/d/1mbspfKSsejHXIjhFq5ifUs5w0DtCGYjjqNSNEWP_K_0/edit#
) - Partner_id and Partner_name weren't visible in FB events manager. So, mailed META to confirm if they are being received as expected. They confirmed they are being received as expected.

### DIFF
--- a/packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-user-data.ts
+++ b/packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-user-data.ts
@@ -107,12 +107,12 @@ export const user_data_field: InputField = {
     },
     partner_id: {
       label: 'Partner ID',
-      description: 'The facebook partner ID',
+      description: 'The ID issued by Facebook identity partner',
       type: 'string'
     },
     partner_name: {
-      label: 'Partner name',
-      description: 'The facebook partner name',
+      label: 'Partner Name',
+      description: 'The name of the Facebook identity partner',
       type: 'string'
     }
   },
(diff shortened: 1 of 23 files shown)


sayan-das-in (approved):
LGTM !!!

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-user-data.ts
```diff
@@ -105,6 +104,16 @@ export const user_data_field: InputField = {
       label: 'Facebook Login ID',
       description: 'The ID issued by Facebook when a person first logs into an instance of an app.',
       type: 'integer'
+    },
+    partner_id: {
```
nick-Ag, line 108: For partner ID and name, are there any relevant docs we can include in the description for users?

nick-Ag (approved):
Left a question

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-user-data.ts
varadarajan-tw: Hey @nick-Ag . For now, no. The feature is in alpha state at Meta's end and that is why this doesn't show up in their docs. However, dropbox wants to use it right away.

PR #1098, opened 2023-03-08
[HEAP]: Fix 'Screen event' event names

murphpdx:
- Updated the screen view and page view event names to reduce cardinality. The cardinality of these events was causing issues for us.
- Updated the identity check to ensure that we do not set identity if userId is set to '0'. This was a customer request. 
- Updated the identify call to allow customers to configure identity. Another customer request. 
- I've also removed the unnecessary request made to our identify endpoint. The add user properties endpoint can achieve the same result as the two calls. 

## Testing

I've sent test events using the event tester `./bin/run serve` and added additional unit tests. 


packages/destination-actions/src/destinations/heap/heapUtils.ts
```diff
@@ -8,31 +8,35 @@ export const getUserIdentifier = ({
   identity?: string | null
   anonymous_id?: string | null
 }): { [k: string]: string } => {
-  if (identity) {
+  if (identity && isDefined(identity)) {
```
murphpdx, line 11: keeping the identity null check to keep the linter happy.

joe-ayoub-segment:
hi @murphpdx , acknowledging this PR. I'll get it reviewed ASAP.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -98,9 +98,19 @@ const action: ActionDefinition<Settings, Payload> = {
...
         '@path': '$.name'
       }
+    },
+    traits: {
+      label: 'User Properties',
+      type: 'object',
+      description:
+        'An object with key-value properties you want associated with the user. Each property must either be a number or string with fewer than 1024 characters.',
+      default: {
+        '@path': '$.traits'
```
joe-ayoub-segment, line 108: Usually you won't find customers adding a traits object at the root of the payload of a page(), screen() or track(). Instead they usually get added in the context object. 

```suggestion
        '@path': '$.context.traits'
```

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -125,16 +137,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
       app_id: settings.appId,
       events: [event],
       library: 'Segment'
     }
 
-    return request('https://heapanalytics.com/api/integrations/track', {
-      method: 'post',
-      json: payLoad
-    })
+    if (isDefined(payload.identity)) {
```
joe-ayoub-segment, line 146: Hi @murphpdx , should this call to add_user_properties be necessary if the traits field is empty?

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -98,9 +98,19 @@ const action: ActionDefinition<Settings, Payload> = {
       default: {
         '@path': '$.name'
       }
+    },
+    traits: {
```
joe-ayoub-segment, line 102: Adding a new optional field is safe, as long as it's referenced defensively (which it appears to be).

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
```diff
@@ -2,20 +2,28 @@ import type { ActionDefinition } from '@segment/actions-core'
...
   fields: {
-    user_id: {
-      label: 'User ID',
+    identity: {
+      label: 'Identity',
       type: 'string',
       allowNull: true,
       description:
-        'An identity, typically corresponding to an existing user. If no such identity exists, then a new user will be created with that identity. Case-sensitive string, limited to 255 characters.',
+        'A unique identity and maintain user histories across sessions and devices under a single profile. If no identity is provided we will add the anonymous_id to the event. More on identify: https://developers.heap.io/docs/using-identify',
```
joe-ayoub-segment, line 26: hi @murphpdx the description is a little tricky to understand from a grammatical point of view. "A unique identity and maintain user histories..."

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
```diff
@@ -2,20 +2,28 @@ import type { ActionDefinition } from '@segment/actions-core'
...
+}
 
 const action: ActionDefinition<Settings, Payload> = {
   title: 'Identify User',
   description: 'Set the user ID for a particular device ID or update user properties.',
   defaultSubscription: 'type = "identify"',
   fields: {
-    user_id: {
-      label: 'User ID',
+    identity: {
```
joe-ayoub-segment, line 21: hi @murphpdx this is going to create a new optional field called 'identity', which is generally safe. 

However in the perform() function below you are throwing an error if this field value is falsy. 
Customers who already have the Heap Destination configured won't have this field setting saved to our database, so this field will always resolve to undefined which will result in an error being thrown (for those customers). This will break all identify() events for those specific customers.   

Some options we have here are: 
1. If your customers are OK with this, they can 'fix' their instances of the Destination by editing and then saving the identifyUser Action. This will ensure that the field settings are saved to the DB. 
2. rework your code so that you don't need to throw an error if the identity field is undefined. I'm not sure you'll be able to do this though. 

Happy to discuss other options if you like - let me know.

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
```diff
@@ -44,35 +52,34 @@ const action: ActionDefinition<Settings, Payload> = {
       throw new IntegrationError('Missing Heap app ID.', 'Missing required field', 400)
     }
 
-    const responses = []
+    if (!payload.identity || !isDefined(payload.identity)) {
```
joe-ayoub-segment, line 55: As mentioned above, this new optional 'identity' field will resolve as undefined for pre-existing instances of the Integration, which will result in errors being thrown for all events which are processed by this Action.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -125,16 +137,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
-    return request('https://heapanalytics.com/api/integrations/track', {
-      method: 'post',
-      json: payLoad
-    })
+    if (isDefined(payload.identity)) {
+      const userPropertiesPayload = {
+        app_id: settings.appId,
+        identity: payload.identity,
+        properties: {
+          anonymous_id: payload.anonymous_id,
```
joe-ayoub-segment, line 151: In the identifyUser action you check if the payload.anoymous_id field is truthy before adding it to the payload for the add_user_properties API call. 

Should that also be done here?

joe-ayoub-segment (changes requested):
Hi @murphpdx thanks for submitting these changes. 
There is a breaking change in there which is worth addressing or it will cause an incident. I've left inline comments for you to review.

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
murphpdx: I chatted with the customer to let them know that this was a possibility and they still wanted the change. Will this still result in errors being thrown if a userId exists? I've kept the default mapping to userId.

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
murphpdx: The identity field defaults to userId. Is this still going to cause a breaking change? I've talked to the customer about that possibility and they're ok with it.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
murphpdx: We can, it was my assumption that it's always on the payload. Is that not true?

joe-ayoub-segment: hi @murphpdx 
The anonymousId is simply a GUID created by the client side Segment libraries which is stored in a cookie/local storage. 
It's only in the payload if the event is coming from a front end device such as a browser or a mobile iOS / Android app. If the event is coming from a back end service or raw HTTP POST request it's not going to contain an anonymousId, unless the customer explicitly adds it.

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
```diff
@@ -44,35 +52,34 @@ const action: ActionDefinition<Settings, Payload> = {
       throw new IntegrationError('Missing Heap app ID.', 'Missing required field', 400)
     }
 
-    const responses = []
+    if (!payload.user_id || !isDefined(payload.user_id)) {
```
joe-ayoub-segment, line 55: Hi @murphpdx - am I correct in assuming that you want this Action to start rejecting identify() events which do not contain a user_id? If so is this OK with the current customer(s) using the Integration?

joe-ayoub-segment (changes requested):
Hi @murphpdx - some comments/questions on the code. I didn't look at the unit tests.

packages/destination-actions/src/destinations/heap/trackEvent/index.ts
```diff
@@ -125,16 +137,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
       app_id: settings.appId,
       events: [event],
       library: 'Segment'
     }
 
-    return request('https://heapanalytics.com/api/integrations/track', {
-      method: 'post',
-      json: payLoad
-    })
+    if (isDefined(payload.identity) && (isDefined(payload.anonymous_id) || isDefined(payload.traits))) {
```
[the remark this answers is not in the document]
murphpdx: Yes, the logic is intentional. That's correct, you cannot call add user properties without identity. Identity is already on the track payload so you must have identity and anon id or traits.

packages/destination-actions/src/destinations/heap/identifyUser/index.ts
murphpdx: Yes this is correct. We can't associate user properties or identity with a user without an identity. The previous logic was incorrect and assigned an incorrect identity to the user. Adding incorrect identities will cause problems for our customers.

joe-ayoub-segment:
FYI @murphpdx - the deploy didn't go out yesterday. Expecting it to happen today.

joe-ayoub-segment:
Hi @murphpdx  - this PR has been deployed. Could you let me know if it works as expected please? I'm late notifying you as it was a public holiday here on Friday.

PR #1102, opened 2023-03-09
[Actions Braze Cloud] Updated batch failure logic

sayan-das-in:
This PR addresses [HGI-402](https://segment.atlassian.net/browse/HGI-402) and [HGI-405](https://segment.atlassian.net/browse/HGI-405)

The current logic for batched events to `trackEvent` discards the entire batch if a single event is invalid. This PR removes this check sending all events to Braze API and let it discard invalid events. This would enable partial batch failures.

## Testing
Tested in local environment.

[HGI-402]: https://segment.atlassian.net/browse/HGI-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HGI-405]: https://segment.atlassian.net/browse/HGI-405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/destination-actions/src/destinations/braze/utils.ts
+++ b/packages/destination-actions/src/destinations/braze/utils.ts
@@ -32,7 +32,6 @@ export function sendTrackEvent(request: RequestClient, settings: Settings, paylo
 
   return request(`${settings.endpoint}/users/track`, {
     method: 'post',
-    headers: { 'X-Braze-Batch': 'true' },
     json: {
       events: [
         {
--- a/packages/destination-actions/src/destinations/braze/utils.ts
+++ b/packages/destination-actions/src/destinations/braze/utils.ts
@@ -18,19 +18,52 @@ function toISO8601(date: DateInput): DateOutput {
   return d.isValid() ? d.toISOString() : undefined
 }
 
-export function sendTrackEvent(request: RequestClient, settings: Settings, payloads: TrackEventPayload[]) {
+export function sendTrackEvent(request: RequestClient, settings: Settings, payload: TrackEventPayload) {
+  const { braze_id, external_id } = payload
+  const user_alias = getUserAlias(payload.user_alias)
+
+  if (!braze_id && !user_alias && !external_id) {
+    throw new IntegrationError(
+      'One of "external_id" or "user_alias" or "braze_id" is required.',
+      'Missing required fields',
+      400
+    )
+  }
+
+  return request(`${settings.endpoint}/users/track`, {
+    method: 'post',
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/braze/utils.ts
```diff
@@ -18,19 +18,52 @@ function toISO8601(date: DateInput): DateOutput {
...
+    throw new IntegrationError(
+      'One of "external_id" or "user_alias" or "braze_id" is required.',
+      'Missing required fields',
+      400
+    )
+  }
+
+  return request(`${settings.endpoint}/users/track`, {
+    method: 'post',
+    headers: { 'X-Braze-Batch': 'true' },
```
varadarajan-tw, line 35: Should we remove this as we are sending a single event here?

varadarajan-tw (commented):
Looks good. Added a small comment. One question before i approve - I understand Chime asked for this update but should call out this behaviour in our docs?. Unaccounted event loss is possible here and it is possible we have to do support to triage such issues in case customers raise complaints.

packages/destination-actions/src/destinations/braze/utils.ts
sayan-das-in: Removed the header for single events.

varadarajan-tw (approved):
Discussed with @sayan-das-in  offline. We don't have any other way to track failures and customer who initially is okay with this implementation. Lets just add a datadog stat to track such silent failures if possible. Lets document this behaviour in Segment docs if possible. But we should definitely have some internal documentation on how to identify and troubleshoot issues with batching.

packages/destination-actions/src/destinations/braze/utils.ts
```diff
@@ -18,19 +18,51 @@ function toISO8601(date: DateInput): DateOutput {
...
 
-    if (!braze_id && !user_alias && !external_id) {
-      throw new IntegrationError(
-        'One of "external_id" or "user_alias" or "braze_id" is required.',
-        'Missing required fields',
-        400
-      )
-    }
+    // Disable errors until Actions Framework has a multistatus support
+    // if (!braze_id && !user_alias && !external_id) {
```
rhall-twilio, line 59: I agree we need some way to track this other than fully silently. Probably some datadog ErrorAsSuccess metric.

PR #1108, opened 2023-03-14
HGI-399 Actions Pardot - status of undefined

immanojkumar:
## Summary
This PR is to resolve HGI-399: Actions Pardot - status of undefined.
Throwing a default error if response key doesn't exist in the error object.

JIRA Ticket : [Jira](https://segment.atlassian.net/browse/HGI-399?atlOrigin=eyJpIjoiOTVhOGM0YjEyYTM1NGZjMTgwNDkyZDQwMmVhM2E4NWQiLCJwIjoiaiJ9)

### Local Testing
**Error:**
image

**After Fix:**
image


packages/destination-actions/src/destinations/actions-pardot/prospects/index.ts
```diff
@@ -188,28 +188,31 @@ const action: ActionDefinition<Settings, Payload> = {
       return await pa.upsertRecord(payload)
     } catch (err) {
       const error = err as HTTPError
+      if (!error.response) {
+        throw new IntegrationError(`Something went wrong.`, 'PARDOT_ERROR', 500)
```
rhall-twilio, line 192: I think we want to log `err` otherwise we still don't know what the real error is. Think `response` object will be null when err is not an HTTPError.

marinhero (commented):
Great work! Now, would it be possible to add a unit test that looks for this error?

packages/destination-actions/src/destinations/actions-pardot/prospects/index.ts
immanojkumar: Added logger. Thanks!

packages/destination-actions/src/destinations/actions-pardot/prospects/index.ts
```diff
@@ -188,28 +189,32 @@ const action: ActionDefinition<Settings, Payload> = {
       return await pa.upsertRecord(payload)
     } catch (err) {
       const error = err as HTTPError
+      if (!error.response) {
+        logger.error(`Pardot error: `, error)
```
varadarajan-tw, line 193: I am not sure if this logger will work but the logger client passed through perform should work.  You can change the signature of perform to `perform: async (request, { settings, payload , logger }))` and use the logger object passed in as param. Also, the logger will only work if it is enabled for the destination in `integrations` like [this](https://github.com/segmentio/integrations/blob/116cdfcd011de9e15781046d12e35248d6339016/integrations/index.js#L228) . Error logs are also not shown in production by default as the log level is `CRIT' ([Production log level](https://github.com/segmentio/eks-configuration/blob/0b59da074e7da2d34a8c70d8c96f3f40488f1d3a/clusters/centrifuge-destinations/production-us-west-2.yaml#L1212)). So, we may have to do logger.crit here.

packages/destination-actions/src/destinations/actions-pardot/prospects/index.ts
varadarajan-tw: Think it is better if we through the err as such if it is not an instance of HTTPError. Logger is tricky to work with due to the reason mentioned in the previous comment.
```suggestion
        throw err
```

packages/destination-actions/src/lib/logger.ts
```diff
@@ -0,0 +1,8 @@
+import { Logger } from 'ecs-logs-js'
```
varadarajan-tw, line 1: We can remove this logger and use the Logger recommended in the previous comment if at all we want to use it.

immanojkumar:
> Great work! Now, would it be possible to add a unit test that looks for this error?

Hey @marinhero, I do not think we can add a unit test case for this change since we can't delete the response from error object in the test case.

packages/destination-actions/src/destinations/actions-pardot/prospects/index.ts
```diff
@@ -188,28 +188,31 @@ const action: ActionDefinition<Settings, Payload> = {
       return await pa.upsertRecord(payload)
     } catch (err) {
       const error = err as HTTPError
+      if (!error.response) {
+        throw error
```
varadarajan-tw, line 192: I think we should throw `err` and not `error` here for the exception to be not considered as HTTPError.

varadarajan-tw:
Thanks for doing the suggested changes. Now, i hope we'll be able to capture the correct error in Sentry and probably fix it. Let me create a follow up ticket for the same.

varadarajan-tw (approved):

PR #1116, opened 2023-03-17
Bring back autopublish

SyedWasiHaider:
We want to bring back autopublish on our npm packages. Let's see if we can do it? (Had to close a previous PR #1115  because the branch name cannot have underscores).

Usage:

* yarn run release 
  * This will check that you're on the main branch and then run `lerna version minor` which will create a commit on main and push up the new version tags
* The CI should then pick up these new pushed tags and automatically publish to NPM

## Testing

Successful CI run (I had to change the branch name to this one): https://github.com/segmentio/action-destinations/actions/runs/4452213760/jobs/7819658996

NPM Published package: https://www.npmjs.com/package/@segment/action-destinations/v/3.139.0


SyedWasiHaider:
Probably a little weird that I have a publish commit on here but it's pretty harmless IMO

.github/workflows/publish.yml
```diff
@@ -0,0 +1,74 @@
...
+    timeout-minutes: 15
+
+    strategy:
+      matrix:
+        node-version: [18.x]
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          persist-credentials: false
```
silesky, line 24: Can't you pass Github PAT here?

.husky/pre-commit
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+if [ "$CI" = "true" ]; then
+    echo "Skipping precommit hook because we're in the CI"
```
silesky, line 4: We can also just do:

env:
  HUSKY: 0

.github/workflows/publish.yml
```diff
@@ -0,0 +1,74 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  push:
+    branches:
+      - npmautopublish
```
silesky, line 9: Instead of creating special branch, could you create a magic merge commit message like "autopublish" whenever you merge to master?

silesky, line 9: Would it make sense to filter on a magic commit message like "Publish"? That way, you can push the tags and the new version commit to main?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,74 @@
...
+
+      - name: Set NPM Token
+        run: |
+          npm set '//registry.npmjs.org/:_authToken' ${{ secrets.NPM_PUBLISH_TOKEN }}
+          npm whoami
+
+      - name: Publish
+        run: |
+          lerna publish minor \
+            --allowBranch=npmautopublish \
```
silesky, line 72: I assume this does the version bumping and creating tags as well? Do we need to push the tags and stuff to main afterward?

.husky/pre-commit
```diff
@@ -23,4 +28,4 @@ then
     brew install gitleaks
 fi
 
-yarn lint-staged -c lint-staged.config.js
\ No newline at end of file
+yarn lint-staged -c lint-staged.config.js
```
silesky, line 31: .editorconfig has "insert final new line = true" ?

.husky/pre-commit
SyedWasiHaider: Oh this is good to know!

.husky/pre-commit
SyedWasiHaider: no idea but I'll revert this anyway given the handy env var trick you mentioned

.github/workflows/publish.yml
SyedWasiHaider: Good idea!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - uses: actions/cache@v2
+        id: yarn-cache # use this to check for `cache-hit` (`steps.yarn-cache.outputs.cache-hit != 'true'`)
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Use Github Personal Access Token
```
silesky, line 44: You should need to the PAT here if you're passing it during checkout, right?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Build
+        run: NODE_ENV=production yarn build
+
+      # set to the author of the most recent commit
+      - name: Set author
+        run: |
+          git config --global user.email "`git log --format='%ae' HEAD^!`"
+          git config --global user.name "`git log --format='%an' HEAD^!`"
+          git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN  }}@github.com/${{ github.repository }}
```
silesky, line 60: Hmm is this actually necessary with the PAT?

silesky, line 60: Not sure I understand this line. It's introducing another token, and the PAT should be sufficient for giving you push permissions

silesky:
@SyedWasiHaider Issues with this publishing pipeline:
- tags are generally important-- I'd rather do things the manual way than not have tags 
- an autopublish branch is something i've never seen on any publishing pipeline - a more typical flow would be locally, to:
1.    checkout master
2.  "npx lerna version minor -y" (creates a publish / release commit and push)
3. CI runs / builds with lerna publish

silesky:
> > tags are generally important
> 
> Are they used elsewhere or in some other process?
> 

[See how tags are used for us to create github releases](
https://github.com/segmentio/analytics-next/releases ) -- they make it easy/possible to map npm versions back to specific commits, otherwise, npm publishing is black box.

silesky:
FYI, if you want to see usage examples in github actions, I sometimes use: https://github.com/search?q=lerna+publish+path%3A.github%2Fworkflows&type=code&ref=advsearch

SyedWasiHaider:
Those are great examples

I'm actually re-assessing whether this effort is actually worth doing now since it still requires manual intervention to some degree. I think I had envisioned a completely automated path but it seems like that might not be the best idea.

@valerieernst do you have any strong opinions?

silesky:
> I'm actually re-assessing whether this effort is actually worth doing now since it still requires manual intervention to some degree. I think I had envisioned a completely automated path but it seems like that might not be the best idea.

If it makes you feel better, it's totally normal to make the versioning part manual (unless you use something like changeset or semantic commits). Automating the build and publish to npm part and you have 95% of the benefit, as it saves users from having to deal with npm credentials and prevents mistakes having to do with bad local builds (like, someone forgot to pull the latest from master, or someone forgot to use the correct node version, or there are some weird build artifacts in dist).

valerieernst:
One of the primary goals is to ensure everything that's published to npm has gone through a full code review and been merged into main. Currently a dev could publish any code on their machine. So the process of having a dev push up a publish commit, then having CI actually do the publishing from the main branch _does_ achieve that goal.

We could also add a script in the package.json to generate the actual publish commit so it's formatted correctly

.github/workflows/publish.yml
```diff
@@ -0,0 +1,73 @@
...
+          git fetch --tags
+
+      - name: Set NPM Token
+        run: |
+          npm set '//registry.npmjs.org/:_authToken' ${{ secrets.NPM_PUBLISH_TOKEN }}
+          npm whoami
+
+      - name: Publish
+        env:
+          HUSKY: 0
```
silesky, line 71: I would set this at the top of the file so the preinstall hooks are not installed

scripts/release.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+branch=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD);
+if [[ $branch != "main" ]];
+then
+  echo "You must be on the main branch to release"
+  exit
+fi;
+
```
silesky, line 8: I would also do a `git pull --ff-only` here

.github/workflows/publish.yml
```diff
@@ -0,0 +1,73 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  push:
+    branches:
+      - main
```
silesky, line 9: If this runs on every merge to main, won't this mean every merge going forward will have an exit 1 (since that's the result of lerna publish?). Do we want to filter by tag instead?
```
    tags:
      - "@segment/*"
```

scripts/release.sh
```diff
@@ -0,0 +1,11 @@
...
+
+echo "Running lerna version minor..."
+lerna version minor -y
```
silesky, line 11: should we do `lerna version --no-private` so we don't bump private packages (like integration tests)?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Set NPM Token
+        run: |
+          npm set '//registry.npmjs.org/:_authToken' ${{ secrets.NPM_PUBLISH_TOKEN }}
+          npm whoami
+
+      - name: Publish
+        env:
+          HUSKY: 0
+        run: |
+          lerna publish from-git --yes --allowBranch=main
```
silesky, line 73: Do we need to add `--no-git-tag-version --no-push` -- my sense is that `lerna version` already pushes all that stuff, so we don't actually need to any committing or pushing here at all (unless lerna is generating a changelog?)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: ${{ matrix.node-version }}
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - uses: actions/cache@v2
```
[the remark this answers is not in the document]
SyedWasiHaider: Do you know if I still need the previous step?

.github/workflows/publish.yml
SyedWasiHaider: Yeah we shouldn't need to set the git author anymore because this particular command will _only_ publish to npm. It doesn't push any tags or commits.

We do still need git I think to fetch the latest tags.

.github/workflows/publish.yml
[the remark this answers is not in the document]
silesky: no, you don't

.github/workflows/publish.yml
```diff
@@ -0,0 +1,55 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
```
silesky, line 4: should the name just be Publish

silesky:
We might want to add some internal documentation on how to release somewhere: even as simple as "git checkout main && yarn release"

silesky (approved):

PR #1127, opened 2023-03-22
[HEAP-41596]: Add Identity and traits to track event

murphpdx:
I've added support for user properties and identity to the track event. Now if the event has the userId or context.traits they will be sent to heap. We explicitly check for 0 values in the identity field because one of out customers sends in a 0 for failed logins. We don't want to identify the user in this case. 

## Testing


joe-ayoub-segment:
hi @murphpdx thanks for the PR. I'll schedule it for review.

pooyaj (approved):

PR #1128, opened 2023-03-23
Add batch_size config to webhook action

peterdemartini:
This change adds the `batch_size` to the webhook destination so it can be adjusted depending on the capabilities of the webhook service.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/webhook/send/index.ts
+++ b/packages/destination-actions/src/destinations/webhook/send/index.ts
@@ -27,6 +27,13 @@ const action: ActionDefinition<Settings, Payload> = {
       default: 'POST',
       required: true
     },
+    batch_size: {
+      label: 'Batch Size',
+      description: 'The target number of events to batch together in a single request, max size 4000.',
+      type: 'number',
+      required: false,
+      default: 0
+    },
     headers: {
       label: 'Headers',
       description: 'HTTP headers to send with each request.',
--- a/packages/destination-actions/src/destinations/webhook/send/generated-types.ts
+++ b/packages/destination-actions/src/destinations/webhook/send/generated-types.ts
@@ -10,6 +10,10 @@ export interface Payload {
    */
   method: string
   /**
+   * The target number of events to batch together in a single request, max size 4000.
+   */
+  batch_size?: number
+  /**
    * HTTP headers to send with each request.
    */
   headers?: {


tysonmote (approved):
Needs Actions team approval, of course, but LGTM.

packages/destination-actions/src/destinations/webhook/send/index.ts
```diff
@@ -27,6 +27,13 @@ const action: ActionDefinition<Settings, Payload> = {
       default: 'POST',
       required: true
     },
+    batch_size: {
```
SyedWasiHaider, line 30: Would this become a reserved keyword? Wondering whether it makes sense to move it outside of `fields` too.

peterdemartini: It is reserved, similar to `enable_batching` and that is why I placed it at the same level as `enable_batching`. I am not sure if it can be moved up (I am a bit unfamiliar with the actions design/configuration). I am open to other ideas or different name change if needed.

SyedWasiHaider: I think this approach is fine. We just need to do a little more work on the cli side like this: https://sourcegraph.segment.com/github.com/segmentio/action-destinations/-/blob/packages/cli-internal/src/commands/push.ts?L162:24

packages/destination-actions/src/destinations/webhook/send/index.ts
```diff
@@ -27,6 +27,13 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      label: 'Batch Size',
+      description: 'The target number of events to batch together in a single request, max size 4000.',
+      type: 'number',
+      required: false,
+      default: 0
```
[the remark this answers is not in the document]
peterdemartini: Batch size of 0 means, fall back to default configuration (which is configured to [1000](https://github.com/segmentio/integrations-go/blob/691be7c9cbc155a58782db7dc457eab7f5e77ee2/models/actions/integration.go#L91))

packages/destination-actions/src/destinations/webhook/send/index.ts
SyedWasiHaider: Created a follow up ticket for this in our backlog: https://segment.atlassian.net/browse/ACT-601

SyedWasiHaider (approved):

packages/destination-actions/src/destinations/webhook/send/index.ts
valerieernst: would it make sense to have batch_size defined in the `properties` property in the enable_batching field? https://sourcegraph.segment.com/github.com/segmentio/action-destinations/-/blob/packages/core/src/destination-kit/types.ts?L133:3

That way, we wouldn't need to add a new reserved keyword, and we could tightly couple batch size the whether batching is enabled or not

peterdemartini: Would that change `enable_batching` to an object?

peterdemartini: I know it is not ideal having a reserved property but we already having the `enable_batching` property and this is correlated

valerieernst: it would add a child property to the field like this. 
```
    enable_batching: {
      type: 'boolean',
      label: 'Batch Data',
      description:
        'If true, Segment will batch events before sending to API',
      default: false,
      properties: {
        batchSize: {
          label: 'Batch Size',
          description: 'The target number of events to batch together in a single request, max size 4000.',
          type: 'number',
          required: false,
          default: 0
        }
      }
    }

peterdemartini: @valerieernst do you thinks is okay to move forward with? Maybe we can improve on this later?

valerieernst: nvm. just reread the comment describing the `properties` field "Optional definition for the properties of `type: 'object'` fields".

I'm good with the current implementation

valerieernst (approved):

marinhero (approved):

PR #1129, opened 2023-03-23
fixing june presets

joe-ayoub-segment:
## Summary

Fixing June Integration presets. 

## Testing

None needed. This integration won't run in the app because presets are incorrectly configured.  
Preset objects require a mapping attribute. This isn't enforced in the action-destination repo, but is required at runtime (by Integrations).


marinhero (approved):
For future reference, please elaborate in the description on why this fixes the issue. Once that's covered, you are good to :shipit:

PR #1130, opened 2023-03-24
fixing presets for acoustic

joe-ayoub-segment:
## Summary 

Fixing presets which are breaking the UI from rendering for acoustic Destination. Note: this is a new Destination - no customers. 

## Testing

None needed


marinhero (approved):
Same as the other PR. Let's keep a good troubleshooting reference in the PR description and elaborate on why this fixes the issue.

PR #1134, opened 2023-03-27
TikTok Audiences Post PROD updates

maryamsharif:
This PR changes 2 things with the new TikTok Audiences destination.
1. It requires users to input `audience id` to update audiences in TikTok instead of the `audience name` so we don't have to get all audiences and filter through them to match which audience to send users to. 
2. It adds a `createAudience` action to provide users a way to create their audience.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/tiktok-audiences/createAudience/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/tiktok-audiences/createAudience/__tests__/index.test.ts
@@ -1,0 +1,176 @@
+import nock from 'nock'
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
+import Destination from '../../index'
+import { BASE_URL, TIKTOK_API_VERSION } from '../../constants'
+
+const testDestination = createTestIntegration(Destination)
+
+interface AuthTokens {
+  accessToken: string
+  refreshToken: string
+}
+
+const auth: AuthTokens = {
+  accessToken: 'test',
+  refreshToken: 'test'
+}
+
+const event = createTestEvent({
+  event: 'Audience Entered',
+  type: 'track',
+  properties: {
+    audience_key: 'personas_test_audience'
+  },
+  context: {
+    device: {
+      advertisingId: '123'
+    },
+    traits: {
+      email: 'testing@testing.com'
+    }
+  }
+})
+
+const urlParams = {
+  advertiser_id: '123',
+  page: 1,
+  page_size: 100
+}
+
+const updateUsersRequestBody = {
+  advertiser_ids: ['123'],
+  action: 'add',
+  id_schema: ['EMAIL_SHA256', 'IDFA_SHA256'],
+  batch_data: [
+    [
+      {
+        id: '44d206f60172cd898051a9fb2174750aee1eca00f6f63f12801b90644321e342',
+        audience_ids: ['1234345']
+      },
+      {
+        id: 'a665a45920422f9d417e4867efdc4fb8a04a1f3fff1fa07e998e86f7f7a27ae3',
+        audience_ids: ['1234345']
+      }
+    ]
+  ]
+}
+
+const createAudienceRequestBody = {
+  custom_audience_name: 'personas_test_audience',
+  advertiser_id: '123',
+  id_type: 'EMAIL_SHA256',
+  action: 'create'
+}
+
+describe('TiktokAudiences.addUser', () => {
+  // it('should fail if `personas_audience_key` field does not match the `custom_audience_name` field', async () => {
+  //   await expect(
+  //     testDestination.testAction('addUser', {
+  //       event,
+  //       settings: {
+  //         advertiser_ids: ['123']
+  //       },
+  //       useDefaultMappings: true,
+  //       auth,
+  //       mapping: {
+  //         personas_audience_key: 'mismatched_audience',
+  //         id_type: 'EMAIL_SHA256',
+  //         selected_advertiser_id: '123'
+  //       }
+  //     })
+  //   ).rejects.toThrow('The value of `custom_audience_name` and `personas_audience_key` must match.')
+  // })
+
+  it('should succeed if an exisiting audience is found', async () => {
+    // nock(`${BASE_URL}${TIKTOK_API_VERSION}/dmp/custom_audience/list/`)
+    //   .get(/.*/)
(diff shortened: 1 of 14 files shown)


packages/destination-actions/src/destinations/tiktok-audiences/createAudience/__tests__/index.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const createAudienceRequestBody = {
+  custom_audience_name: 'personas_test_audience',
+  advertiser_id: '123',
+  id_type: 'EMAIL_SHA256',
+  action: 'create'
+}
+
+describe('TiktokAudiences.createAudience', () => {
+  it('should successfully create a new audience', async () => {
```
marinhero, line 31: I thought we were not creating audiences for this destination. What changed?

packages/destination-actions/src/destinations/tiktok-audiences/removeUser/generated-types.ts
```diff
@@ -6,9 +6,9 @@ export interface Payload {
    */
   selected_advertiser_id: string
   /**
-   * Custom audience name of audience to be synced. This audience must already exist in your TikTok Advertising account
+   * Audience ID for the TikTok Audience you want to sync your Engage audience to. This is returned after you create an audience and can also be found in the TikTok Audiences dashboard.
    */
-  custom_audience_name: string
+  audience_id: string
```
marinhero, line 11: You are renaming a required field. Do we need to run a migration for that??

packages/destination-actions/src/destinations/tiktok-audiences/functions.ts
```diff
@@ -159,3 +105,11 @@ export function extractUsers(payloads: GenericPayload[], audienceID: string): {}
   })
   return batch_data
 }
+
+export async function createAudience(
```
marinhero, line 109: Why is this `async` if you are not calling `await` within the function body?

marinhero (changes requested):
Overall the code looks great, just left some questions that I'd like to answer before moving forward.

packages/destination-actions/src/destinations/tiktok-audiences/createAudience/__tests__/index.test.ts
maryamsharif: In order to create an audience the users would have to create a curl request and get an `access-token` which would require a lot of other requests that they would have to do. I thought maybe we could utilize the resources we have to give them a way to create the audience manually since we already have the information needed to make the request for them. They would just have to fill in the information and use the test action method. Just a work around I thought of to try to make the process easier for users.

packages/destination-actions/src/destinations/tiktok-audiences/removeUser/generated-types.ts
brennan: I don't think we need a migration in this case b/c no customers are using the destination yet. Caveat this with the fact that existing instances set up for testing won't work any longer because they would be missing an existing field IIRC. cc @maryamsharif

packages/destination-actions/src/destinations/tiktok-audiences/functions.ts
```diff
@@ -139,7 +82,10 @@ export function extractUsers(payloads: GenericPayload[], audienceID: string): {}
       let email_id = {}
       if (payload.email) {
         payload.email = payload.email.replace(/\+.*@/, '@').replace(/\./g, '').toLowerCase()
-        email_id = { id: createHash('sha256').update(payload.email).digest('hex'), audience_ids: [audienceID] }
+        email_id = {
+          id: createHash('sha256').update(payload.email).digest('hex'),
+          audience_ids: [payload.audience_id]
```
brennan, line 87: Open question for future iterations; no action item. Should we ever allow customers to specify more than one audience id?

packages/destination-actions/src/destinations/tiktok-audiences/removeUser/generated-types.ts
maryamsharif: yes that is what happened when I stage tested with the change.

packages/destination-actions/src/destinations/tiktok-audiences/functions.ts
maryamsharif: I think if customers want to sync the same engage audience to multiple audiences within their TikTok Audience we could allow them to do so!

packages/destination-actions/src/destinations/tiktok-audiences/functions.ts
maryamsharif: thanks for catching! removed `async`

brennan:
@maryamsharif One last comment - looks like some of the tests in `addUser` and `removeUser` still use the `personas_audience_key` property even though we removed it.

packages/destination-actions/src/destinations/tiktok-audiences/createAudience/__tests__/index.test.ts
marinhero: I'm not sure if we want to do that, this will fall within the user's control and seems very prone to misuse. Have you consulted with Product about this behavior?

maryamsharif: How could it be misused? I don't believe it would have any different consequences than the customer manually creating the curl request. @kdaswani what do you think? I could definitely leave it out if we think it is a bad idea.

kdaswani: Don't think it's a bad idea necessarily, but to @marinhero's comment, I'm worried it may seem a bit hacky and confusing to customers. Let's remove it and rely on customers to create the audience themselves and see what feedback we get during Beta!

packages/destination-actions/src/destinations/tiktok-audiences/functions.ts
kdaswani: Could they use separate Actions for each audience with the current state?

maryamsharif: Yes, in the current state they could create separate actions that are mapped to different audience id's so they could sync to multiple audiences.

marinhero (approved):

PR #1137, opened 2023-03-29
[HGI-431] - Make error code mandatory and introduce APIError class

varadarajan-tw:
This PR does the following changes

**Make code mandatory and set default status as 400 for IntegrationErrors**

IntegrationError doesn't mandate error code or status right now. As a result some of the handled errors from partner built destinations are showing up in sentry. Example: [Heap](https://twilio.slack.com/archives/C04J8QJ1Y59/p1679078393576379). To ensure all that all IntegrationErrors provide a status and error code, this PR defaults the status to 400 and mandates error code to be provided. The following destinations were updated to conform to the new Integration Error - Emarsys, Criteo, FullStory, Heap and livelike-cloud.

**Introduce APIError and API_CALL_FAILED Error Code**

Some destinations capture and re throw Http Errors due to bad error messages from partner API. This error class can be used to identify those class of errors.

## Testing

This updates just the error codes and error type. All unit tests are passing in local.


joe-ayoub-segment (approved):
Looks good to me. 
Tests ran locally. So did yarn validate. 
Defaulting to 400 for IntegrationError seems reasonable. 
Handling and rethrowing the error for a token refresh looks good.
The changes made to existing errors in Partner Integrations look sensible.

packages/core/src/destination-kit/index.ts
```diff
@@ -361,7 +361,13 @@ export class Destination<Settings = JSONObject> {
       return undefined
     }
 
-    return this.authentication.refreshAccessToken(requestClient, { settings, auth: oauthData })
+    return this.authentication.refreshAccessToken(requestClient, { settings, auth: oauthData }).catch((err) => {
+      const message = (err as Error).message ?? 'UNKNOWN'
+      throw new InvalidAuthenticationError(
+        `Failed to refresh access token. Reason:${message}`,
```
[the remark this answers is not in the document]
varadarajan-tw: The other ticket seems to be an error that might occur even before hitting this flow.  This captures only the errors that is thrown from refreshToken method defined in DestinationDefiniton.

packages/destination-actions/src/destinations/emarsys/addToContactList/index.ts
```diff
@@ -68,19 +68,19 @@ const action: ActionDefinition<Settings, Payload> = {
...
             if (body.replyCode === 0) return response
-            else throw new IntegrationError('Something went wrong while adding to contact list')
+            else throw new IntegrationError('Something went wrong while adding to contact list.', 'CONTACT_ADD_FAILED')
           } catch (err) {
-            throw new IntegrationError('Invalid JSON response')
+            throw new IntegrationError('Invalid JSON response', 'CONTACT_ADD_FAILED')
           }
         case 400:
-          throw new IntegrationError('The contact could not be removed from the contact list')
+          throw new IntegrationError('The contact could not be added to the contact list', 'CONTACT_REMOVE_FAILED')
```
rhall-twilio, line 76:
```suggestion
          throw new IntegrationError('The contact could not be added to the contact list', 'CONTACT_ADD_FAILED')
```

packages/destination-actions/src/destinations/emarsys/triggerEvent/index.ts
```diff
@@ -83,19 +84,19 @@ const action: ActionDefinition<Settings, Payload> = {
           try {
             const body = await response.json()
             if (body.replyCode === 0) return response
-            else throw new IntegrationError('Something went wrong while triggering the event')
+            else throw new IntegrationError('Something went wrong while triggering the event', 'TRIGGER_EVENT_FAILED')
           } catch (err) {
-            throw new IntegrationError('Invalid JSON response')
+            throw new IntegrationError('Invalid JSON response', 'TRIGGER_EVENT_FAILED')
```
rhall-twilio, line 89: This one doesn't seem generic. Thinking more about it, the contacts add/remove above are specific as well. 

Do we have guidance on how generic we want these errors to be? Considering at some point they will be populated in confluence and event delivery.

Thinking aloud:
Option 1: GENERIC_API_FAILURE - just tells you the call failed, one day maybe we can promote that `err` object as an argument.

Option 2: TRIGGER_EVENT_FAILED - tells you that out of all the calls that failed, it's the trigger one. `/trigger` is probably not generic across all APIS so we have to consider that each destination will come up with a specific error for each /path that they have?

Option 3: EMARSYS_TRIGGER_EVENT_FAILED - it's the most granular, in case the error message for the trigger call failing needs a different message for EMARSYS destination?


I'm kinda leaning more towards Option 1 provided we forward the `err` object, but wondering how do you feel about it. Do we expect users to use codes that are not part of the enum?

packages/destination-actions/src/destinations/emarsys/addToContactList/index.ts
marinhero: ah! I think Varada is right on this one because he's trying to remove the contact from a list. I understand why you'd like to switch to `CONTACT_ADD_FAILED` since that's the action that's causing the issue.

packages/destination-actions/src/destinations/emarsys/triggerEvent/index.ts
varadarajan-tw: Thanks for the suggestion.  I can add `API_CALL_FAILED` and setup a description. 

On specific vs generic - I think, it will be impossible to cover all possible codes as part of the Enum. The Enum helps to bring consistency in debugging experience for most common errors that occur across destinations. 
We should let some destinations define their own error codes even though they might not have a valid description in contentful and these error codes can give better context than a generic one in certain scenarios. 

We default to a generic Error description from Contentful if we don't find a description for the specific error code.
image

varadarajan-tw: For `API_CALL_FAILED`, we will probably write a similar description to `Error` i guess. Should we just retain the custom error code? At least it will give a better context?

prayansh:
Amazing change! love the standardization of error types and code 😍 ! 

Although since we are changing the response code here, maybe we should limit the change to a single destination first and validate any observability tools / retries and then proceed on to the other destinations

varadarajan-tw:
@prayansh / @SyedWasiHaider—Created two small PRs to clean up Heap and Livelike. 
-  [Livelike](https://github.com/segmentio/action-destinations/pull/1295)
-  [Heap](https://github.com/segmentio/action-destinations/pull/1296)

varadarajan-tw:
https://github.com/segmentio/action-destinations/pull/1309 - PR to introduce new Error class!

varadarajan-tw:
This PR was split into multiple PRs and all of them have been deployed separately. Here is the final [PR](https://github.com/segmentio/action-destinations/pull/1322) to make code and status mandatory.

Closing this PR now.

PR #1152, opened 2023-04-04
Add a new @replace directive

SyedWasiHaider:
Add ability to replace values in fields

## Testing

Tested in stage via webhook action:

Example mapping:

```json
      {
        "fieldKey": "data",
        "value": {
          "userIdReplaced": {
            "@replace": {
              "pattern": "@",
              "replacement": "Weeee",
              "value": {
                "@path": "$.userId"
              }
            }
          },
          "everything": {
            "@path": "$."
          }
        }
      },
```
      
  Example output:
  
```json
  {
  "everything": {
    "channel": "server",
    "context": {
      "library": {
        "name": "unknown",
        "version": "unknown"
      }
    },
    "event": "Example Event",
    "integrations": {},
    "messageId": "api-2NyLfjD8lzVtfkLt2pKsnt8wMU0",
    "projectId": "ah35UczHuHeRK8p3qxd6nE",
    "properties": {},
    "receivedAt": "2023-04-04T18:04:39.926Z",
    "timestamp": "2023-04-04T18:04:39.926Z",
    "type": "track",
    "userId": "wasi@haider@segment.com",
    "version": 2
  },
  "userIdReplaced": "wasiWeeeehaiderWeeeesegment.com"
}
```

### DIFF
--- a/packages/core/src/mapping-kit/__tests__/index.iso.test.ts
+++ b/packages/core/src/mapping-kit/__tests__/index.iso.test.ts
@@ -548,6 +548,89 @@ describe('@template', () => {
   })
 })
 
+describe('@replace', () => {
+  test('replace on empty string', () => {
+    const payload = {
+      a: ''
+    }
+    const output = transform(
+      {
+        '@replace': {
+          pattern: '_',
+          replacement: 'rrrrr',
+          value: { '@path': '$.a' }
+        }
+      },
+      payload
+    )
+    expect(output).toStrictEqual('')
+  })
+  test('replace with empty string', () => {
+    const payload = {
+      a: 'nomore_underscore'
+    }
+    const output = transform(
+      {
+        '@replace': {
+          pattern: '_',
+          replacement: '',
+          value: { '@path': '$.a' }
+        }
+      },
+      payload
+    )
+    expect(output).toStrictEqual('nomoreunderscore')
+  })
+  test('replace with non-empty string', () => {
+    const payload = {
+      a: 'nomore_underscore'
+    }
+    const output = transform(
+      {
+        '@replace': {
+          pattern: '_',
+          replacement: 'weird',
+          value: { '@path': '$.a' }
+        }
+      },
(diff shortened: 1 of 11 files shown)


packages/core/src/mapping-kit/__tests__/index.iso.test.ts
```diff
@@ -548,6 +548,89 @@ describe('@template', () => {
...
+          pattern: '+',
+          replacement: '_',
+          value: { '@path': '$.a' }
+        }
+      },
+      payload
+    )
+    expect(output).toStrictEqual('many_different_things')
+  })
+})
```
LLuque-twilio, line 632: could you add a test where the entire original string pattern is matched and replaced with
- an empty string
- a non empty string
something like
```
const payload = {
      a: 'many+different+things'
    }
    const output = transform(
      {
        '@replace': {
          pattern: 'many+different+things',
          replacement: '',
          value: { '@path': '$.a' }
        }
      },
      payload
    )
    expect(output).toStrictEqual('')
```

packages/core/src/mapping-kit/index.ts
```diff
@@ -100,6 +100,33 @@ registerDirective('@case', (opts, payload) => {
   }
 })
 
+registerDirective('@replace', (opts, payload) => {
+  if (!isObject(opts)) {
```
LLuque-twilio, line 104: Should we change this error message? `@replace` should be an object but it doesn't have an `operator` field does it?

LLuque-twilio (approved):
lgtm - left a few non-blocking comments

packages/core/src/mapping-kit/__tests__/index.iso.test.ts
SyedWasiHaider: Oooh excellent test case!

joe-ayoub-segment:
Hi @SyedWasiHaider FYI this PR has been deployed.

PR #1154, opened 2023-04-05
fixing Amplitude Platform field and added tests - HGI-247 

joe-ayoub-segment:
Fix for https://segment.atlassian.net/browse/HGI-247 

Currently Ampltitude's Platform field mappings are not respected in code when the Library field's value is "analytics.js". The value "Web" is sent. 

This PR changes this behavior so that the value passed with the Platform field is actually sent to Amplitude regardless of whether or not the Library field's value is "analytics.js" or not. However if no value is provided for the Platform field and the Library field's value is "analytics.js", then the platform = "Web" will continue to be sent to Amplitude.  

4 Actions were updated. 2 other Actions didn't have the incorrect behavior so were not altered. 

## Testing

Tested locally using Actions Tester. 
Also added 2 new unit tests. 

Test 1: If Library field = “analytics.js” and we send Platform field = “some_value“, we expect that “some_value“ value will be forwarded to Amplitude as the platform. 

image

Test 2: If Library field = “analytics.js” and we send Platform field = undefined or null, we expect that “Web“ value will be forwarded to Amplitude as the platform. 

image


joe-ayoub-segment:
Closing for now as I'm on PTO - will reopen when back from Recharge.

itsarijitray (approved):

joe-ayoub-segment:
PR deployed

PR #1181, opened 2023-04-17
Apr14 main

KWLandry-acoustic:

Refactored extendheader out, tested keys parsing - all good, 

## Testing


packages/destination-actions/src/destinations/acoustic/Utility/tablemaintutilities.ts
```diff
@@ -1,15 +1,36 @@
...
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+        'User-Agent': 'Segment Action (Acoustic Destination)',
+        Connection: 'keep-alive',
+        'Accept-Encoding': 'gzip, deflate, br',
+        Accept: '*/*'
+      }
+    })
+    auth.accessToken = at.data['access_token'] as unknown as string
+  }
```
joe-ayoub-segment, line 27: Hi @KWLandry-acoustic  cc @varadarajan-tw , 

I don't know if this is a good idea or if it will behave as you expect. 
I think we're supposed to rely on the updateAccessToken function to do this, so coding around it isn't something I'm confident will work. 

Keen to hear your thoughts @varadarajan-tw .

packages/destination-actions/src/destinations/acoustic/index.ts
```diff
@@ -120,26 +119,17 @@ const destination: DestinationDefinition<Settings> = {
             grant_type: 'refresh_token'
           }),
           headers: {
-            'Content-Type': 'application/x-www-form-urlencoded'
+            'Content-Type': 'application/x-www-form-urlencoded',
+            'user-agent': `Segment Action (Acoustic Destination)`,
+            Connection: 'keep-alive',
+            'Accept-Encoding': 'gzip, deflate, br',
+            Accept: '*/*'
```
joe-ayoub-segment, line 126: Hi @KWLandry-acoustic, just checking here, but did you test that this request will work with all of these headers? If I recall correctly when we tested the refreshAccessToken function manually on Friday we only included the Content-Type header. I could be wrong though...

joe-ayoub-segment (changes requested):
Hi @KWLandry-acoustic , 

Couple of comments to respond to. 

Also yarn build is failing. Please see the attached file for the terminal output. Please see the attached notes file for what I see when I run the command: 
[notes.txt](https://github.com/segmentio/action-destinations/files/11259882/notes.txt)

packages/destination-actions/src/destinations/acoustic/Utility/tablemaintutilities.ts
KWLandry-acoustic: Hi Joe,
I'm expecting this 'if' block to not execute 99% of the time with refreshToken/updateAccessToken providing the access token, however, if there is no access token by this point then this gets the access token and we can go on to deliver the data, fyi, all subsequent calls for an access token receive the exact same token for the 4 hours the access token is valid.

joe-ayoub-segment: Hi @KWLandry-acoustic , I think the inclusion of this code might actually prevent the access token from being cashed by Segment. However I can't be sure what the behavior will be. 

@varadarajan-tw we discussed this earlier during our call. Could you comment please?

packages/destination-actions/src/destinations/acoustic/index.ts
KWLandry-acoustic: Hi Joe, yes, these are the headers that are in play for all the auth calls in testing prior and in Postman tests as well,  I do know that 'Connection' and 'User-Agent' are not necessary but helpful,

packages/destination-actions/src/destinations/acoustic/Utility/tablemaintutilities.ts
varadarajan-tw: You are right @joe-ayoub-segment. @KWLandry-acoustic - It should be okay to remove this. We have a retry mechanism in place to generate a new token and retry the event in case the call fails due to invalid access token. You can refer the code [here](https://github.com/segmentio/action-destinations/blob/87659c16d3d1e3946783cc9a56a044088d6483e8/packages/core/src/destination-kit/index.ts#L622)

varadarajan-tw: The retry is based on API returning a 401 error. If the API returns a 401, a new token will be automatically be acquired and the event will be retried.
For Acoustic, API calls are expected to fail with 401 for the first time as only refreshToken is configured in Destination settings. The retry mechanism mentioned above will get a new token by invoking refreshAccessToken defined and the tokens will be cached and reused. The problem with generating a token outside this loop is that the token generated will never get cached.

KWLandry-acoustic: Hi Varadarajan, 
Importantly, this is the **Second** API call, so, there should be an access token gained and cached already, however, if there is no access token then it will execute, but,.... again, at this point there should be an Access Token that was already returned and cached by the refreshToken mechanism with the prior call. 
And fyi will retrieve the exact same token as was generated in the 4 hour window the access token is valid, so, effectively a reference to an Acoustic cached access token.

packages/destination-actions/src/destinations/acoustic/index.ts
KWLandry-acoustic: Hi Joe, resolved the yarn build fail on unknown type, updated code is committed,

joe-ayoub-segment (approved):

PR #1185, opened 2023-04-18
[WIP] - Add synchronization support

varadarajan-tw:
For action-destinations like Blackbaud and Adobe Target Cloud, previous access tokens are invalidated on each refresh. Since access tokens are stored in memory of Centrifuge directors and the tokens are not shared across multiple instances of the Centrifuge server, we’ll run into an endless cycle of refreshing access tokens. If the destination invalidates the previous refresh token, then all events subsequently fail. 
Until we figure out a permanent solution for the problem and to enable the affected destinations, we’ll be using redis as a token store and also as the distributed lock store like some of the classic destinations do.

A more detailed note on the impact of this change, the reasoning behind the change and the test results for review are available in this doc- [Redis as token store and lock store for Action Destinations requiring synchronization](https://docs.google.com/document/d/1nfeFyR9AQ1YJ8csrtdYyGxIIWAOgsPZTzsTbMOWzgSM/edit#)

This PR accompanies this integration PR. The following are the changes made to actions-core as part of this PR:

- onTokenRefresh is now an async method to allow storing of tokens in redis 
- lockStore is now passed as part of onEventOptions. If lockStore is passed, the refreshToken invocation will be synchronized by acquiring a lock. The destinations for which the lockstore must be handed on are controlled by integration Monoservice. 

Only [Blackbaud](https://app.segment.com/partner-portal/integration/actions-blackbaud-raisers-edge-nxt/profile) is meant to use it initially.

## Testing


packages/core/src/destination-kit/index.ts
```diff
@@ -223,6 +225,14 @@ export interface TransactionContext {
   setTransaction: (key: string, value: string) => void
 }
 
+/**
+ * LockStore passes handlers to achieve synchronization for token generartion.
```
brennan, line 229:
```suggestion
 * LockStore passes handlers to achieve synchronization for token generation.
```

varadarajan-tw:
Closing this PR and reopening the fresh one as this has become a bit length.

PR #1197, opened 2023-04-19
[STRATCONN-1816] The Trade Desk CRM Destination

maryamsharif:
NEW Trade Desk Integration 🥳 

**This destination will need to be registered in integrations when deploying.**

[EPIC](https://segment.atlassian.net/browse/STRATCONN-1816)
[SDD](https://docs.google.com/document/d/1bTy9CBVYR1ZTzryzNsXUEnKvFbF_ufbEwNpDqURhEXQ/edit#heading=h.2cn9yahfu9ff)

## Testing
[Testing Doc](https://paper.dropbox.com/doc/The-Trade-Desk-CRM-Testings--B3iO4OfRRreIA112CZ28MQG2Ag-27IJpdqwLqoxoD7g1ITQB)


joe-ayoub-segment:
Hi @maryamsharif , 

I'm just back from PTO and am catching up on emails and PRs. 
I noticed that you asked if I could review this PR. I'm happy to help but am not sure I'm best placed to do the review. Usually I review PRs for new Integrations raised by Partners. I schedule a call with the Partner where they explain the use-cases for the Integration as well as the types of data and identifiers to be collected. We also then review how their API works. 

In this case I don't know anything about The Trade Desk's use-cases, the types of data / identifiers to be collected, and I'm completely unfamiliar with their API. I could spend time getting up to speed on these things but it would likely take me some time. 

Let me ping Doosan to see what he thinks! 
Cheers,
Joe

packages/destination-actions/src/destinations/the-trade-desk-crm/functions.ts
```diff
@@ -0,0 +1,194 @@
...
+        'Content-Type': 'application/json',
+        'TTD-Auth': settings.auth_token
+      }
+    }
+  )
+  let segments = response.data.Segments
+  // pagingToken leads you to the next page
+  let pagingToken = response.data.PagingToken
+  // keep iterating through pages until the last empty page
+  while (segments.length > 0) {
```
nick-Ag, line 67: How does this scale? Could this potentially make tons and tons of calls to get each Segment/Page? Does that pose a timeout issue?

maryamsharif: Yes, we are aware of this risk and plan to ask The Trade Desk if they have any limitations with the amount of Segments they allow customers to have so we can have a better understanding of how many calls we may have to make to get them all. 

cc. @rvadera12

packages/destination-actions/src/destinations/the-trade-desk-crm/functions.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  // More than 1 audience returned matches name
+  if (segmentExists.length > 1) {
+    throw new IntegrationError('Multiple audiences found with the same name', 'INVALID_SETTINGS', 400)
+  }
+
+  if (segmentExists.length == 1) {
+    segmentId = segmentExists[0].CrmDataId
+  }
+  // Audience doesn't exist so we need to create one
```
rhall-twilio, line 100: Can you remind me how we are addressing the race condition here? Maybe adding a comment in the code would be helpful too.

maryamsharif: Because the plan is going to be to do 24 hour syncs for this destination since they require 1500 users per upload, we won't be running into the race condition because batches won't be run in parallel, just once a day! I can add a comment as a reminder!

pooyaj (approved):
Looks great 🎉🎉🎉

rhall-twilio (approved):

PR #1202, opened 2023-04-21
[CCP-202] Populate the liquid templates in destination actions

mattcullenmeyer:
https://segment.atlassian.net/browse/CCP-202

## Testing

Tested locally against “sendSms” action and in staging with feature flag “actions-versions-3.153.1-ccp-202.0” using content sid HXede9d7379aa59807522d7d2f8568b4c1.
1. Works correctly if you provide a valid userId, content sid, and set traitEnrichment to false (do not provide a body)
2. Works correctly if you provide a content sid, set traitEnrichment to true, and provide profile traits (do not provide a body or userId)
3. Ensured passing body field with or without trait enrichment still works (for backward compatibility)
4. Ensured that using a content type other than ‘twilio/text’ (eg ‘twilio/media’) throws an error (content sid HX90d8eabc79f309d09b9fae21f6840652)


packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
```diff
@@ -23,6 +35,17 @@ export class SmsMessageSender extends MessageSender<Payload> {
...
   getBody = async (phone: string) => {
+    if (!this.payload.body && !this.payload.contentSid) {
+      this.logger?.error(
+        `TE Messaging: Unable to process SMS, no body provided and no content sid provided - ${this.settings.spaceId}`
+      )
+      throw new IntegrationError(
+        'Unable to process sms, no body provided and no content sid provided',
+        'Invalid parameters',
+        400
+      )
```
varadarajan-tw, line 46: A better error in this case will be the PayloadValidationErrror. Payload Validation error messages are shown like below in EventDelivery along with the error message.
image


```suggestion
      throw new PayloadValidationError(
        'Unable to process sms, no body provided and no content sid provided'
      )
```

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
```diff
@@ -93,4 +124,42 @@ export class SmsMessageSender extends MessageSender<Payload> {
...
+      this.statsClient?.incr('actions-personas-messaging-twilio.error', 1, this.tags)
+      this.logger?.error(
+        `TE Messaging: SMS failed request to fetch content template from Twilio Content API - ${this.settings.spaceId} - [${error}]`
+      )
+      throw new IntegrationError('Unable to fetch content template', 'Twilio Content API request failure', 500)
+    }
+  }
+
+  private getUnparsedContentBody = (data: ContentTemplateResponse): string => {
+    const type = Object.keys(data.types)[0] // eg 'twilio/text', 'twilio/media', etc
```
varadarajan-tw, line 153: Is it possible for the data.types to not contain any keys? Should a length check be added here or in the getContentTemplate method above?

varadarajan-tw (commented):
Overall the changes look good. I left couple of questions.

The PR changes a required field to optional field which is safe.
It also adds a new non-required field.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
```diff
@@ -8,6 +8,18 @@ import { MessageSender, RequestFn } from '../utils/message-sender'
 
 const Liquid = new LiquidJs()
 
+enum ContentType {
```
varadarajan-tw, line 11: nit: If it is just going to be a single enum item, it could be just a constant rather an enum.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
mattcullenmeyer: I haven't seen a response from the Twilio Content API that hasn't returned a "types" key before, but I agree we should handle this edge case. Thank you for the suggestion

varadarajan-tw (approved):
Thanks for doing the suggested changes 🪄

joe-ayoub-segment:
note: deployed by Sayan 02-May

PR #1208, opened 2023-04-25
[WIP] Add support for array value in @path directive

SyedWasiHaider:
Users are unable to map properties that include a dot character in the key. This PR is an attempt to fix that problem by using an array value for specifying the path. On the app side, we'd need to detect the presence of a dot in the key and then construct/pass an array value.

## Testing

### DIFF
--- a/packages/core/src/mapping-kit/index.ts
+++ b/packages/core/src/mapping-kit/index.ts
@@ -11,6 +11,7 @@ export type InputData = { [key: string]: unknown }
 export type Features = { [key: string]: boolean }
 type Directive = (options: JSONValue, payload: JSONObject) => JSONLike
 type StringDirective = (value: string, payload: JSONObject) => JSONLike
+type StringOrArrayDirective = (value: string | JSONLike[], payload: JSONObject) => JSONLike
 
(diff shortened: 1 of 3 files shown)


packages/core/src/mapping-kit/index.ts
```diff
@@ -183,12 +171,41 @@ registerDirective('@arrayPath', (data, payload) => {
...
+    throw new Error(`${name}: expected string, array or object but got ${realTypeOf(value)}`)
+  }
+
+  const getAsStringPath = (obj: JSONObject, strPath: string) => get(obj, strPath.replace('$.', ''))
+
+  switch (realType) {
+    case 'string':
+      return getAsStringPath(payload, value as string) as JSONLike
+    case 'array':
+      return get(payload, value as string[])
```
SyedWasiHaider, line 186: This is the real change for `@path` while the other parts are more of a refactoring / calling out the supported types more explictly.

SyedWasiHaider:
Re-opened here with a different branch name (because I can't publish canary to npm with this branch name):

https://github.com/segmentio/action-destinations/pull/1282

PR #1214, opened 2023-04-27
CCP-204: remove unlayer apikey and generateEmailHtml

zawadzkip:

- The engage send email action destination will no longer be parsing unlayer designs. Now everything coming into the AD should be in HTML format

## Testing


packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -453,15 +409,20 @@ const action: ActionDefinition<Settings, Payload> = {
       )
       let parsedBodyHtml
 
-      if (payload.bodyUrl && settings.unlayerApiKey) {
+      if (payload.bodyUrl) {
+        if (payload.bodyType === 'design') {
+          tags.push('reason:design_body_type_with_bodyUrl')
+          statsClient?.incr('actions-personas-messaging-sendgrid.bodyUrl_failure', 1, tags)
```
zawadzkip, line 415: do we need to configure anything in datadog for these to appear?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/generated-types.ts
```diff
@@ -1,10 +1,6 @@
 // Generated file. DO NOT MODIFY IT BY HAND.
 
 export interface Settings {
-  /**
-   * The API key for your Unlayer account
-   */
-  unlayerApiKey?: string
```
hocaho, line 7: I remember someone mentioning that removing destination action settings may cause issues (while adding is OK). Unfortunately, I can't recall who mentioned this 🤔 @pmunin can you confirm?

pmunin: since that field is optional - I don't think it should matter. I guess @dlasky would be best person to confirm

zawadzkip: @pmunin @hocaho 

I talked a bit with @dlasky about it, sounds like the real issue is only adding required fields because the existing definitions in the DB wont have it, so it would require a CPS migration first.

Is it worthwhile having the AD throw an error for the design type and logging it as well?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/index.ts
```diff
@@ -10,11 +10,6 @@ const destination: DestinationDefinition<Settings> = {
   authentication: {
     scheme: 'custom',
     fields: {
-      unlayerApiKey: {
-        label: 'Unlayer API Key',
```
zawadzkip, line 14: Since we're removing this, do we think its worthwhile to just fail `design` bodyType requests?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -454,14 +390,19 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          throw new IntegrationError(
```
hocaho, line 397: There are existing journeys that still have bodyType='design' with a valid bodyUrl that are currently broken. eg. project id `tPPVUnfirfSU5UEWREsQj3`, `rggwqjYmmSkBGAbFes1R8e`. This is the query for retrieving destination action config settings: `select * from destination_action_config_settings where project_id = 'tPPVUnfirfSU5UEWREsQj3';`

hocaho (changes requested):

PR #1220, opened 2023-05-02
More path mappings for the email field

slavivanov:
Adding some more path mappings for the email field. 
The event test in the Segment app places the email in the root of the event, which this causes the default test events to not map the email, giving a weird experience for users.

Also, added a test case to the core, because I wasn't sure if nested directives would work. I think it would be useful for it to stay as a reference for other integrations.

## Testing


joe-ayoub-segment (commented):
hi @slavivanov ,
I just reviewed this PR now and I think the way you have your email fields mapped already makes sense.
Our documentation doesn't ask customers to set the email field at the root level of the payload.
The email value is always in the following:

identify() and goup() calls - '$.traits.email'
track() and page() calls - '$.context.traits.email' or '$.properties.email'

The testing tool you use when building and testing your Integration has the email at the root level of the payload, however this is a bug and needs to be fixed!

I hope this makes sense.

Please confirm what you'd like to do.

Kind regards,
Joe

slavivanov:
Hi @joe-ayoub-segment, if the UI won't accept nested mappings, we can close this request.
I'm a bit wary of not using top-level `email` property. 
In the event testing functionality that end-users see when connecting destinations, the sample events have the email as a top-level. See this screenshot: https://share.getcloudapp.com/YEueLzdY

joe-ayoub-segment:
Hi @slavivanov , 
The dummy event we provide in the Actions tester you use when developing places the email address at the root of the payload, which is not something customers ever do. So from that perspective you are safe!

We need to update the code which incorrectly sets the email field at the root in the Actions Tester. I'll schedule time to do this. 

Based on the above, would you like to close this request?

slavivanov:
Sure, closing it.

PR #1223, opened 2023-05-02
resolveWhen bug

silesky:
force reject / do not load destination after 3 seconds -- this is maybe too short? Just an example. 

Assumption is that if a single plugin rejects, we just don't load that plugin.

This is a long term fix 

TODO: 
fix vwo specifically

### DIFF
--- a/packages/browser-destinations/src/runtime/__tests__/resolve-when.test.ts
+++ b/packages/browser-destinations/src/runtime/__tests__/resolve-when.test.ts
@@ -1,0 +1,42 @@
+import { resolveWhen } from '../resolve-when'
+
+// jest.useFakeTimers was hard to use
+const sleep = (time: number): Promise<void> =>
+  new Promise((resolve) => {
+    setTimeout(resolve, time)
+  })
+
+describe(resolveWhen, () => {
+  it('should resolve immediately', async () => {
+    const resolveCb = jest.fn()
+    const errorCb = jest.fn()
+
+    await resolveWhen(() => true, 100)
+      .then(resolveCb)
+      .catch(errorCb)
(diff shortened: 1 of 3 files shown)


packages/browser-destinations/src/runtime/resolve-when.ts
```diff
@@ -1,19 +1,30 @@
...
+import { pTimeout } from './promise-with-timeout'
+
+/**
+ * Will reject if error
+ *
+ * @param condition - condition after which function will resolve
+ * @param checkInterval - how often to check the condition
+ * @param timeout - max timeout
+ */
+export async function resolveWhen(condition: () => boolean, checkInterval: number, timeout = 3000): Promise<void> {
```
silesky, line 10: default timeout guys for force reject?

pooyaj (approved):

packages/browser-destinations/src/runtime/__tests__/resolve-when.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    expect(errorCb).not.toHaveBeenCalled()
+    expect(resolveCb).toHaveBeenCalled()
+  })
+
+  it('should reject after some time', async () => {
+    const errorCb = jest.fn()
+
+    await resolveWhen(() => false, 200).catch(errorCb)
+    expect(errorCb).toHaveBeenCalled()
+  })
```
oscb, line 20: Should we also have a test for the timeout?
callback returns true but after sleeping n + 1 ms
timeout is set to n.

silesky:
Closing this in favor of a simpler PR: https://github.com/segmentio/action-destinations/pull/1224

PR #1224, opened 2023-05-02
Temp fix vwo

silesky:
temporarily ignore `useExistingJquery` setting  because of bug that presented window.VWO from ever being present, causing analytics to completely fail.

loading multiple copies of jquery doesn't usually break things; the other option is to just throw an error for the customers who use this.


oscb (approved):

zikaari (changes requested):
This is not going to fix the SEV. It is no different than what's happening already.

I'd recommend this:

```ts
if (!useExistingJquery) {
  throw new Error('VWO is broken deep inside, and is also breaking everything else')
}
```

Reference -
image

PR #1227, opened 2023-05-02
Fix VWO destination error

zikaari:
This is fix to deal with `j.php` script not always setting the `window.VWO` variable thus leading to `resolveWhen` freeze-up. This patch now checks if the execution of `j.php` immediately set `window.VWO` variable or it didn't. It resolves when it does, and rejects when it doesn't. Either way, preventing the freeze up of entire Analytics SDK.

### DIFF
--- a/packages/browser-destinations/src/destinations/vwo/init-script.ts
+++ b/packages/browser-destinations/src/destinations/vwo/init-script.ts
@@ -8,74 +8,82 @@ export function initScript({
   useExistingJquery = false,
   isSpa = 1
 }) {
-  window._vwo_code =
-    window._vwo_code ||
-    (function () {
-      var account_id = vwoAccountId,
-        settings_tolerance = settingsTolerance,
-        library_tolerance = libraryTolerance,
-        use_existing_jquery = useExistingJquery,
-        is_spa = isSpa,
-        hide_element = 'body',
-        /* DO NOT EDIT BELOW THIS LINE */
-        f = false,
-        d = document,
-        code = {
-          use_existing_jquery: function () {
-            return use_existing_jquery
-          },
-          library_tolerance: function () {
-            return library_tolerance
-          },
-          finish: function () {
-            if (!f) {
-              f = true
-              var a = d.getElementById('_vis_opt_path_hides')
-              if (a) a.parentNode.removeChild(a)
+  return new Promise((resolve, reject) => {
+    if(window._vwo_code) {
+      resolve()
+    }
+    
+    window._vwo_code ||= (function () {
+        var account_id = vwoAccountId,
+          settings_tolerance = settingsTolerance,
+          library_tolerance = libraryTolerance,
+          use_existing_jquery = useExistingJquery,
+          is_spa = isSpa,
+          hide_element = 'body',
+          /* DO NOT EDIT BELOW THIS LINE */
+          f = false,
+          d = document,
+          code = {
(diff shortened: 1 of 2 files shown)


silesky (approved):

packages/browser-destinations/src/destinations/vwo/init-script.ts
```diff
@@ -8,74 +8,78 @@ export function initScript({
...
-          library_tolerance: function () {
-            return library_tolerance
-          },
-          finish: function () {
-            if (!f) {
-              f = true
-              var a = d.getElementById('_vis_opt_path_hides')
-              if (a) a.parentNode.removeChild(a)
+  return new Promise((resolve, reject) => {
+    window._vwo_code ||= (function () {
```
silesky, line 12: just occurred to me that this promise will only resolve if vwo_code doesn't exist. I think we need to resolve immediately in that case.

silesky (changes requested):
looks good, but q

silesky:
cc @prathamesh-88

packages/browser-destinations/src/destinations/vwo/init-script.ts
zikaari: Good catch. Let me fix that

prathamesh-88:
Hi @zikaari @silesky 
Thank you for bringing this issue to our attention and providing the necessary changes. We appreciate your contribution and attention. After testing the suggested changes, we have discovered some malfunctions in the `VWO.event` and `VWO.visitor` functions within the action-tester environment. Our team is actively investigating the root cause and working towards resolving this issue.

Furthermore, we have encountered some challenges when loading the SmartCode (init-script) with Segment, which has been a persistent concern for our team. We have included the necessary fixes to address this issue in this PR. https://github.com/segmentio/action-destinations/pull/1236

silesky (approved):

chrisradek:
@prathamesh-88 
The issues you mentioned around VWO.event and VWO.visitor - are those existing issues present without the changes in this PR? Or were they introduced by this PR? We're trying to understand if we can merge this change to unblock one issue we're seeing while you separately work on fixing the issues you mentioned.

zikaari:
@prathamesh-88 Please refer to [this comment](https://github.com/segmentio/action-destinations/pull/1236#issuecomment-1541080594)

joe-ayoub-segment:
Hi @zikaari - this looks like an old PR. Do you want to proceed with it or should it be closed?
Best regards,
Joe

joe-ayoub-segment:
Hi @zikaari , following up on this PR again. Is it something you'd like to proceed with?
Best regards,
Joe

prathamesh-88:
Hi @joe-ayoub-segment 
The issue mentioned in this PR was resolved in the latest changes. Thus, these changes are no longer required as these changes posed certain functional issues while testing them.

I just noticed that the destination has been moved to private beta from public beta. Can you please look into it and let us know if anything needs to be done on our end?

joe-ayoub-segment:
hi @prathamesh-88 - quick clarification, can I close this PR?

zikaari:
@joe-ayoub-segment since the authors suggest that the changes in this patch might be problematic - we should close it out.

@prathamesh-88 It has been moved to private beta because it is still unstable, the `window.VWO` object never spawns sometimes, and holds up the entire AJS pipeline. We don't have repro steps since it happens at random.

PR #1236, opened 2023-05-05
VWO Web Mode fixes

prathamesh-88:
This PR is to resolve the following issue:
- When VWO SmartCode (init-script) loads with the Segment code, there's a minor delay in loading which might cause white flashes on the page as VWO script (j.php) tries to manipulate the DOM. So we're adding the setting to not load the initScript code through Segment bundle but rather directly add smartcode to their website

## Testing


zikaari:
What is this smart code, can you show an example of it?

joe-ayoub-segment:
hi @pooyaj FYI

prathamesh-88:
@zikaari 
SmartCode is essentially a code snippet(Script tag) that client needs to install on their website to use VWO.

**Sample SmartCode**

```
<!-- Start VWO Async SmartCode -->
<script type='text/javascript' id='vwoCode'>
window._vwo_code=window._vwo_code || (function() {
var account_id=<Account_ID>,
version = 1.5,
settings_tolerance=2000,
library_tolerance=2500,
use_existing_jquery=false,
is_spa=1,
hide_element='body',
hide_element_style = 'opacity:0 !important;filter:alpha(opacity=0) !important;background:none !important',
/* DO NOT EDIT BELOW THIS LINE */
f=false,d=document,vwoCodeEl=d.querySelector('#vwoCode'),code={use_existing_jquery:function(){return use_existing_jquery},library_tolerance:function(){return library_tolerance},hide_element_style:function(){return'{'+hide_element_style+'}'},finish:function(){if(!f){f=true;var e=d.getElementById('_vis_opt_path_hides');if(e)e.parentNode.removeChild(e)}},finished:function(){return f},load:function(e){var t=d.createElement('script');t.fetchPriority='high';t.src=e;t.type='text/javascript';t.onerror=function(){_vwo_code.finish()};d.getElementsByTagName('head')[0].appendChild(t)},getVersion:function(){return version},getMatchedCookies:function(e){var t=[];if(document.cookie){t=document.cookie.match(e)||[]}return t},getCombinationCookie:function(){var e=code.getMatchedCookies(/(?:^|;)\s?(_vis_opt_exp_\d+_combi=[^;$]*)/gi);e=e.map(function(e){try{var t=decodeURIComponent(e);if(!/_vis_opt_exp_\d+_combi=(?:\d+,?)+\s*$/.test(t)){return''}return t}catch(e){return''}});var i=[];e.forEach(function(e){var t=e.match(/([\d,]+)/g);t&&i.push(t.join('-'))});return i.join('|')},init:function(){if(d.URL.indexOf('__vwo_disable__')>-1)return;window.settings_timer=setTimeout(function(){_vwo_code.finish()},settings_tolerance);var e=d.createElement('style'),t=hide_element?hide_element+'{'+hide_element_style+'}':'',i=d.getElementsByTagName('head')[0];e.setAttribute('id','_vis_opt_path_hides');vwoCodeEl&&e.setAttribute('nonce',vwoCodeEl.nonce);e.setAttribute('type','text/css');if(e.styleSheet)e.styleSheet.cssText=t;else e.appendChild(d.createTextNode(t));i.appendChild(e);var n=this.getCombinationCookie();this.load('https://dev.visualwebsiteoptimizer.com/j.php?a='+account_id+'&u='+encodeURIComponent(d.URL)+'&f='+ +is_spa+'&vn='+version+(n?'&c='+n:''));return settings_timer}};window._vwo_settings_timer = code.init();return code;}());
</script>
<!-- End VWO Async SmartCode -->
```
<Account_ID> is the VWO Account ID of the client

We had altered this code a bit to execute it through Segment.

zikaari:
@prathamesh-88 One of Segment's and VWO mutual customer rev.com is affected. When their VWO account id is passed to the `j.php` script, it doesn't return a functioning script back, the one that sets `window.VWO` variable. Please contact them and rectify the issue with `j.php` script and also check if other account id's are affected by a similar issue.

joe-ayoub-segment:
Hi @prathamesh-88 , just checking who's got the next step with this PR? Are you waiting on anyone from Segment/Twilio?

packages/browser-destinations/src/destinations/vwo/index.ts
```diff
@@ -61,17 +61,27 @@ export const destination: BrowserDestinationDefinition<Settings, VWO> = {
...
       type: 'boolean',
       default: false
+    },
+    addSmartcode: {
+      description:
+        'When enabled VWO SmartCode will load parallel to Segment Code, when disabled SmartCode has to be added manually. It is recommended to add SmartCode directly to the webpage code to avoid flicker.',
+      label: 'Add Asynchronous SmartCode',
+      type: 'boolean',
+      default: true,
+      required: true
```
chrisradek, line 71: @pooyaj Do you know if this will add the setting automatically to any/all users of the action destination? Normally I'd think a new required field is a breaking change but this has a default value that maintains existing behavior.

joe-ayoub-segment: Hi @chrisradek , 
Adding a required field, or changing a field to be required, or even adding code in the perform to treat a non required field as required will result in the Integration breaking for customers who already have it set up. This happens regardless of whether or not you have set a field default, such as in your example. The reason is that the field default value won't be saved to the database for customers who are already using the Integration. 

The way to work around this is to make sure your 'new' field is optional, and to have a default applied, as per your example. But you'd also need to apply a default in the perform function. 

Does that make sense to you?

prathamesh-88: Hi @joe-ayoub-segment 
Have made the changes accordingly. Do let me know if we still have any blockers.

joe-ayoub-segment (approved):
LGTM.

packages/browser-destinations/src/destinations/vwo/index.ts
```diff
@@ -61,17 +61,26 @@ export const destination: BrowserDestinationDefinition<Settings, VWO> = {
       label: 'Use Existing JQuery',
       type: 'boolean',
       default: false
+    },
+    addSmartcode: {
```
chrisradek, line 65: SmartCode was always being used before, right?

I wonder if this should be changed to `disableSmartCode` with a default of `false`.

I think as it stands right now based on @joe-ayoub-segment 's comment, existing users would not have a value set for this, which would mean we'd stop adding the `initScript` for those users. (New users would be fine as they'd default to `true`)

joe-ayoub-segment: Hi @chrisradek and @prathamesh-88 m, that sounds like a better approach and hopefully not too much of an ask. @prathamesh-88 are you OK with Chris' proposed approach?

prathamesh-88: Hi @joe-ayoub-segment @chrisradek 

Our intention behind this setting is to give clients the choice to either load the SmartCode through the Segment library or manually install it on their website (recommended way). Although I get the functional benefits of the proposed approach, I feel that if we name the setting "Disable Smartcode" it might cause confusion about the actual usage of the setting causing some clients to think that it affects the manually installed smartcode on their website.

As to @joe-ayoub-segment 's comment, I think that `settings.addSmartcode != false` condition in the `index.ts` will handle the behavior. Please let me know if that's not the case.

joe-ayoub-segment: Happy to go with either approach. Maybe we just make the description of the field a little clearer?
For example: 

When enabled VWO SmartCode will load parallel to Segment Code, when disabled SmartCode has to be added manually. It is recommended to add SmartCode directly to the webpage code to avoid flicker.

=> 

If set to 'true', Segment will load the VWO SmartCode onto the webpage. When set to 'false' you will have to manually add SmartCode to the webpage yourself. This setting defaults to 'true', however we recommend manually adding SmartCode to the webpage to avoid flicker issues.

prathamesh-88: Hi @joe-ayoub-segment 

Thanks for the prompt response. I agree, I have changed the description as suggested.

PR #1237, opened 2023-05-05
VWO Cloud - Fullstack support added

prathamesh-88:
Support added for VWO Fullstack in the Cloud mode destination. 

## Testing


joe-ayoub-segment:
Hi @prathamesh-88 , 
Thanks for the PR, I'll schedule this to be reviewed ASAP. Please note I'm just back from PTO and am catching up, so there may be a slight delay.

packages/destination-actions/src/destinations/vwo/utility.ts
```diff
@@ -1,19 +1,56 @@
 import type { commonPayload, vwoPayload } from './types'
+import * as crypto from 'crypto'
 
-export function formatPayload(name: string, payload: commonPayload, isCustomEvent: boolean, isTrack = false) {
+const VWO_NAMESPACE = '11e13cd7-6c48-53ec-8679-7e9c752273c5'
```
joe-ayoub-segment, line 4: Hi @prathamesh-88 , is this deliberate? 
Is this hard coded value intended to be sent with every payload for all customers?
note: this repo is publicly visible so anyone can see this value.

packages/destination-actions/src/destinations/vwo/index.ts
```diff
@@ -38,6 +38,12 @@ const destination: DestinationDefinition<Settings> = {
...
         required: true
+      },
+      apikey: {
+        label: 'VWO SDK Key',
+        description: 'VWO Fullstack SDK Key. It is mandatory when using the VWO Fullstack suite.',
+        type: 'string',
+        required: false
       }
     },
     testAuthentication: (_request, { settings }) => {
```
joe-ayoub-segment, line 49: Consider updating the testAuthentication function to also check if the apiKey value is correct (when provided).

joe-ayoub-segment (commented):
Hi @prathamesh-88 , 

Couple of observations and questions about this PR. 

1. You are changing some fields to be optional, which is fine. This shouldn't cause any problems. 
2. You are adding a new optional setting field (apikey). You communicated to the customer that this field is required if they are using VWO Fullstack suite.  This too is also fine. 
3. Unit tests are failing when I run them locally. Please see the attached file. 
4. Could you explain a little what this value is for please? VWO_NAMESPACE = '11e13cd7-6c48-53ec-8679-7e9c752273c5'
5. You're encrypting part of the payload, this is OK as long as the size of the value being hashed is small and as long as it's not being called multiple times. We want to ensure that the execution time isn't too long (or timeouts could occur). 
6. Consider updating the testAuthentication function to also check if the api_key value is correct. 
7. I didn't review the code for the formatPayload() function in detail. On the surface it looks OK. If you don't have customers already using the Integration then I think this is safe to deploy, otherwise I'd like to meet with you to get a better understanding of what's happening in this function. Could you clarify if you already have customers using this Integration please?

[failing_tests.txt](https://github.com/segmentio/action-destinations/files/11429951/failing_tests.txt)

packages/destination-actions/src/destinations/vwo/index.ts
prathamesh-88: Hi @joe-ayoub-segment 
The setting name `apiKey` might be misleading. The key is not used for security purposes but to maintain different environments in VWO's FullStack suite. So it doesn't need to be tested and verified. The setting name was set that way to match our internal terminology.

packages/destination-actions/src/destinations/vwo/utility.ts
prathamesh-88: As our Fullstack SDKs containing the algorithm are also publicly hosted. This algorithm is already available to everyone.

joe-ayoub-segment (approved):
Thanks @prathamesh-88 , 
Acknowledged that tests now pass locally as well, thank you. 

LGTM.

joe-ayoub-segment:
hi @prathamesh-88 this change has been deployed. Please let me know if everything looks good!

PR #1238, opened 2023-05-05
update LCNC templates

jinapark202:
This PR updates low-code, no-code templates to use the proper fields and accommodate different default values.

Testing successfully via CLI:
Screenshot 2023-05-05 at 3 20 28 PM

## Testing

### DIFF
--- a/packages/cli/src/commands/scaffold.ts
+++ b/packages/cli/src/commands/scaffold.ts
@@ -122,11 +122,12 @@ export default class Init extends Command {
       const actionsTargetDirectory = `${targetDirectory}/${action.key}`
 
       action.fields.forEach((field: any) => {
-        const hasDirective = field.hasDefault && field.default && field.default.type === 'directive'
-        const hasDefaultValue = field.hasDefault && field.default && field.default.type !== 'directive'
-        field.hasDefaultValue = hasDefaultValue
-        field.isString = field.default?.type === 'string'
-        field.hasDirective = hasDirective
+        const hasDefault = field.hasDefault && field.default
+        if (hasDefault) {
+          field.hasDefaultValue = field.default.type !== 'directive'
+          field.hasDirective = field.default.type === 'directive'
+          field.isString = field.default.type === 'string'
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/scaffold.ts
```diff
@@ -113,14 +113,22 @@ export default class Init extends Command {
...
+    } catch (err: any) {
       this.spinner.fail(`Snapshot test creation failed: ${chalk.red(err.message)}`)
       this.exit()
     }
 
     for (const action of actions) {
       const actionsTargetDirectory = `${targetDirectory}/${action.key}`
 
+      action.fields.forEach((field: any) => {
+        const hasDirective = field.hasDefault && field.default && field.default.type === 'directive'
```
SyedWasiHaider, line 125: Minor but it looks like this line and the next one have some common checks. Maybe we can refactor that out and use that?

packages/cli/templates/actions/empty-action-lowcode/index.ts
```diff
@@ -12,9 +12,25 @@ const action: ActionDefinition<Settings, Payload> = {
      description: '{{description}}',
      type: '{{type}}',
      required: {{required}},
+     {{#hasDefaultValue}}
+     {{#isString}}
      {{#default}}
-     default: {{{value}}}
+     default: "{{value}}",
      {{/default}}
+     {{/isString}}
```
SyedWasiHaider, line 20: Don't know much about this format:

What do the different characters at the beginning mean of the variables mean (`^` and `#` for instance)?

SyedWasiHaider (approved):
minor questions/comments but LGTM!

packages/cli/templates/actions/empty-action-lowcode/index.ts
jinapark202: those variables are used for if-else statements in Mustache.. so this is essentially saying if the default type is a string, add quotes; otherwise, don't include them

PR #1239, opened 2023-05-08
[STRATCONN-2041] Add synchronization support for refresh access token

varadarajan-tw:
For action-destinations like Blackbaud,  previous access tokens are invalidated on each refresh. Since access tokens are stored in memory of Centrifuge director and not shared across multiple service instances, we are running it endless cycle of access token refreshes and also permanent failure of events if refreshToken is invalidated. 

Until we figure out a permanent solution for the problem and to enable the affected destinations, we’ll be using redis as a token store and also as the distributed lock store like some of the classic destinations do.

A more detailed note on the impact of this change, the reasoning behind the change and the test results for review are available in this doc- [Redis as token store and lock store for Action Destinations requiring synchronization](https://docs.google.com/document/d/1nfeFyR9AQ1YJ8csrtdYyGxIIWAOgsPZTzsTbMOWzgSM/edit#)

This PR accompanies this integration PR. The following are the changes made to actions-core as part of this PR:

- onTokenRefresh is now an async method to allow storing of tokens in redis
- passes a synchronization handler to be executed to acquire lock before token refresh. This will be passed in only for destinations that require it. Only [Blackbaud](https://app.segment.com/partner-portal/integration/actions-blackbaud-raisers-edge-nxt/profile) is meant to use it initially. This handler will only acquire the lock and not release the lock. The lock will auto-expire after 15s which is greater than the default timeout for cloudevent calls. The events that reach retry with refreshAccessToken stage will get a RESOURCE_LOCKED_ERROR that will be retried by Centrifuge. Therefore, i have left the release lock portion out of it for simplicity.

## Testing

Testing completed successfully in staging with HubSpot. There is a challenge with testing this change with Blackbaud in staging as we don't have proper test account and also the oauth setup is incomplete. But test results with HubSpot and datadog stats point that this should be safe to move to production.

[Test Results](https://docs.google.com/document/d/1o1YrZW4Dlb9agy_Trup4nZQu2giQFmhvHBjzbS03J_0/edit#
)

### DIFF
--- a/packages/core/src/destination-kit/index.ts
+++ b/packages/core/src/destination-kit/index.ts
@@ -364,8 +364,8 @@ export class Destination<Settings = JSONObject> {
       return undefined
     }
 
-    // Acquire Lock if acquireLock option is passed. Acquire Lock is to required to perform synchronized token refresh
-    // for destinations that invalidate previous tokens
+    // Invoke synchronizeRefreshAccessToken handler if synchronizeRefreshAccessToken option is passed.
+    // This will ensure that there is only one active refresh happening at a time.
     await synchronizeRefreshAccessToken?.()
     return this.authentication.refreshAccessToken(requestClient, { settings, auth: oauthData })
   }
--- a/packages/core/src/destination-kit/index.ts
+++ b/packages/core/src/destination-kit/index.ts
@@ -213,6 +213,8 @@ interface OnEventOptions {
   logger?: Logger
   transactionContext?: TransactionContext
   stateContext?: StateContext
+  /** Handler to perform synchronization. If set, refresh access token action will be synchronized across all events*/
+  synchronizeRefreshAccessToken?: () => Promise<void>
 }
 
 /** Transaction variables and setTransaction method are passed from mono service for few Segment built integrations.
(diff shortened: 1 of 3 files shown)


packages/core/src/destination-kit/index.ts
```diff
@@ -361,6 +364,9 @@ export class Destination<Settings = JSONObject> {
       return undefined
     }
 
+    // Acquire Lock if acquireLock option is passed. Acquire Lock is to required to perform synchronized token refresh
+    // for destinations that invalidate previoius tokens
```
rhall-twilio, line 368:
```suggestion
    // for destinations that invalidate previous tokens
```

packages/core/src/destination-kit/index.ts
```diff
@@ -361,6 +364,9 @@ export class Destination<Settings = JSONObject> {
       return undefined
     }
 
+    // Acquire Lock if acquireLock option is passed. Acquire Lock is to required to perform synchronized token refresh
```
rhall-twilio, line 367: what is acquireLock option? I thought the definition of the synchronizeRefreshAccessToken function was the control.

marinhero (approved):

packages/core/src/destination-kit/index.ts
varadarajan-tw: I was back and forth on the naming and completely missed this comment. Fixed the comment.

rhall-twilio (approved):

SyedWasiHaider (approved):

packages/core/src/destination-kit/index.ts
```diff
@@ -213,6 +213,8 @@ interface OnEventOptions {
   logger?: Logger
   transactionContext?: TransactionContext
   stateContext?: StateContext
+  /** Handler to perform synchronization. If set, refresh access token action will be synchronized across all events*/
```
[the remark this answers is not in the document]
varadarajan-tw: That makes sense. I updated it to `all events across multiple instances of the same destination using the same account for a given source` as only if the account is same, they will use the same key. Thank you!

PR #1240, opened 2023-05-08
Stratconn 2192 - Handle disabled project condition for datafile

hvardhan-unth:
In Optimizely (cloud) action destination, if a project is disabled then the datafile doesn't give error instead it give `undefined` in data. So, this condition has been handled and an error message updated according to that in this PR.
JIRA ticker link:- https://segment.atlassian.net/browse/STRATCONN-2192

## Testing

-  Tested end-to-end using the action-tester(https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)

Screenshot:-
Screenshot 2023-05-08 at 7 09 55 PM
Screenshot 2023-05-08 at 7 10 01 PM


packages/destination-actions/src/destinations/optimizely-feature-experimentation-actions/trackEvent/index.ts
```diff
@@ -108,6 +108,13 @@ const action: ActionDefinition<Settings, Payload> = {
   },
   perform: async (request, { settings, payload }) => {
     const result = await request<ProjectConfig>(settings.dataFileUrl)
+    if (!result.data) {
```
varadarajan-tw, line 111: Are we sure only this error can occur if data is not available? 
Also, could you pls add a test case?

hvardhan-unth: Yes, I have checked in optimizely official documents and tried with datafile url but only in this condition I was getting `undefined` in data.

packages/destination-actions/src/destinations/optimizely-feature-experimentation-actions/trackEvent/mock-dataFile.ts
```diff
@@ -120,3 +120,5 @@ export const dataFile = {
   ],
   revision: '55'
 }
+
+export const dataFileWithDisabledProject = undefined
```
varadarajan-tw, line 124: nit:Do we need an export const here for undefined?

packages/destination-actions/src/destinations/optimizely-feature-experimentation-actions/trackEvent/index.ts
```diff
@@ -108,6 +108,9 @@ const action: ActionDefinition<Settings, Payload> = {
   },
   perform: async (request, { settings, payload }) => {
     const result = await request<ProjectConfig>(settings.dataFileUrl)
+    if (!result.data && typeof result.content === 'string') {
```
varadarajan-tw, line 111: nit: result.content's type is a string. Should we just check if result.content exists?

Also, what should we do if result.content and result.data both are missing?

varadarajan-tw (approved):
Approving as the comments are minor and aren't a blocker for releasing this fix.

packages/destination-actions/src/destinations/optimizely-feature-experimentation-actions/trackEvent/mock-dataFile.ts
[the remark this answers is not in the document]
hvardhan-unth: We can also define it in test file, not mandatory to export it.

packages/destination-actions/src/destinations/optimizely-feature-experimentation-actions/trackEvent/index.ts
hvardhan-unth: I don't think there will be condition where these two will be missing other than if there will be error. But we can throw a customized error message in that condition.

PR #1247, opened 2023-05-11
Add company enrichment to Upollo

cayden-uw:
This PR optionally adds company information enrichment to Identify requests. This includes company name, industry and employee count. 

## Testing

### DIFF
--- a/packages/browser-destinations/src/destinations/upollo/enrichUser/__tests__/index.test.ts
+++ b/packages/browser-destinations/src/destinations/upollo/enrichUser/__tests__/index.test.ts
@@ -1,0 +1,77 @@
+import { Analytics, Context } from '@segment/analytics-next'
+
+import identify from '..'
+import { UpolloClient } from '../../types'
+import { Payload } from '../generated-types'
+
+it('should enrich', async () => {
+  const client = {
+    checkEmail: jest.fn().mockResolvedValue({ company: { name: 'Bar' } })
+  } as any as UpolloClient
+
+  const context = new Context({
+    type: 'identify',
+    event: 'Signed Up'
+  })
+
+  await identify.perform(client as any as UpolloClient, {
+    settings: { apiKey: '123' },
+    analytics: jest.fn() as any as Analytics,
+    context: context,
+    payload: {
+      user_id: 'u1',
+      email: 'foo@bar.com',
+      phone: '+611231234',
+      name: 'Mr Foo',
+      avatar_image_url: 'http://smile',
+      custom_traits: {
+        DOB: '1990-01-01',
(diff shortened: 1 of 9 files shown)


pooyaj (commented):
I'm curios about the use-case for this change. Are we trying to enrich the events that eventually will go to Segment? Looks like we are no longer sending identify calls to Upollo, but instead fetch company info and just enrich the events 🤔

cayden-uw:
Hey @pooyaj, This does still send the event to Upollo it just adds optional enrichment to the Segment event when it does.

cayden-uw:
Any concerns @pooyaj? Happy to grab time to chat if you do

pooyaj:
@cayden-uw Some concern here as mutating the event in a destination could cause unpredictable behavior. If user has multiple destinations, some of them may see the `traits.company` property and some may not, and that behavior is not predictable, as destinations are running asynchronously and not in a predictable order. 

There are other examples of action destinations in the repo that use `context.updateEvent` to mutate the event, but those destinations use `lifeCycleHook` property to make sure they run before regular destinations.

Happy to chat and find an alternative solution if mutating the event is particularly important in this case.

cayden-uw:
Thanks @pooyaj for the thoughtful feedback. It sounds like if I add the lifeCycleHook and set that to enrichment then that would solve the issue of unpredictable behaviour for other destinations? 

If that is the case I will update this PR to that affect now.

cayden-uw:
@pooyaj I updated the lifecycle hook. Let me know if any other changes are needed!

pooyaj:
@cayden-uw sorry for the delay, was getting second opinion on the approach 😄 In this case changing the lifecycleHook of the existing action is not ideal for few reason: 
- It is a breaking change for the destination, as enrichment type actions won't support certain destination features such as destination filters. 
- The enrichment feature will halt the delivery pipeline by making call to the Upollo endpoint and waiting for it to be resolved. 

Is it possible to separate the enrichment logic into its own action that can be optionally turned on by customers? So we will have two actions: 
- `identifyUser` --> which just sends data to Upollo 
- `enrichCompany` --> makes a call to Upollo and enriches the event

cayden-uw:
Thanks for the thoughtful and detailed response @pooyaj. A few things I hadn't thought about when I added this feature originally. 

I have updated the PR to split it into the two actions as suggested

cayden-uw:
@pooyaj let me know if you have any feedback on this or if it is good to go

cayden-uw:
Friendly ping for review

pooyaj (approved):
🚢

cayden-uw:
Thanks @pooyaj!

PR #1250, opened 2023-05-11
Update Loops Action Destination

askkaz:
_Updating the Loops action destination. Adding fields to `createOrUpdateContact` and creating `sendEvent`._

## Testing

- [X] Updated unit tests for `createOrUpdateContact`
- [X] Added unit tests for `sendEvent`


joe-ayoub-segment:
Hi @askkaz , 
Thanks for raising this PR. I'll do a review next week. 
Best regards,
Joe

packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
```diff
@@ -7,6 +7,20 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    createdAt: {
+      label: 'Contact Created Date',
+      description: 'Date the contact was created.',
+      type: 'datetime',
+      format: 'date-time',
+      required: false
+    },
+    customAttributes: {
+      label: 'Custom Contact Attributes',
+      description: 'Attributes maintained by your team.',
```
joe-ayoub-segment, line 19: hi @askkaz , Is this description sufficient for the customer to understand what this field is for?

packages/destination-actions/src/destinations/loops/sendEvent/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      type: 'string',
+      format: 'email',
+      required: false,
+      default: {
+        '@if': {
+          exists: { '@path': '$.traits.email' },
+          then: { '@path': '$.traits.email' },
+          else: { '@path': '$.email' }
+        }
+      }
```
joe-ayoub-segment, line 23: Hi @askkaz , for a track() event you need to check these 2 locations for the email address. See below. 

```suggestion
      default: {
        '@if': {
          exists: { '@path': '$.properties.email' },
          then: { '@path': '$.properties.email' },
          else: { '@path': '$.context.traits.email' }
        }
      }
```

packages/destination-actions/src/destinations/loops/sendEvent/index.ts
```diff
@@ -0,0 +1,55 @@
...
+      required: true,
+      default: {
+        '@path': '$.userId'
+      }
+    }
+  },
+  perform: (request, { payload }) => {
+    return request('https://app.loops.so/api/v1/events/send', {
+      method: 'post',
+      json: {
```
joe-ayoub-segment, line 46: hi @askkaz , do you also want/need to collect event properties as well? Most customers include a properties object in every track() call.

joe-ayoub-segment (changes requested):
Hi @askkaz , 
Please find some feedback on the PR. 
The changes to your createOrUpdateContact Action look safe.
Couple of minor questions for the new Action.

packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
askkaz: @joe-ayoub-segment Our platform allows users to create custom attributes for their list. For example `favoriteColor` is not a default property on our platform but a user can create it for their list. On other platforms (Zapier for example), we use an API endpoint to fetch the available fields for the user when they set up the integration and then we display them. It seems that Segment does not support this, and based on earlier talks with Segment engineers, it was recommended we just dump these attributes into dictionary and then destructure it. I agree its not ideal, but I think its the best we can do with the way that Segment is set up. We will include specific examples of how to use this properly in our documentation.

Open to any suggestions here though :-)

joe-ayoub-segment: Thanks @askkaz this makes sense!

joe-ayoub-segment (approved):
Change looks safe. 
No new required fields. 
Local tests pass. 
yarn validate passes.

joe-ayoub-segment:
Hi @askkaz this PR been deployed. Please test and let me know if everything looks alright!

PR #1252, opened 2023-05-11
Avoiding double hashing when email already comes hashed for Snapchat CAPI and Google Ads Enhanced Conversions

seg-leonelsanches:
Client is having problems with the `email` field, that already comes hashed when it touches the destination. This simply tests if `email` is a SHA256 hash (64 bytes, 0-9 and a-f). If it is, it doesn't hash the `email` a second time. 

## Testing


packages/destination-actions/src/destinations/snap-conversions-api/snap-capi-properties.ts
```diff
@@ -351,7 +351,7 @@ export const formatPayload = (payload: Payload): Object => {
     event_conversion_type: payload?.event_conversion_type,
     event_tag: payload?.event_tag,
     timestamp: Date.parse(payload?.timestamp),
-    hashed_email: hash(payload?.email),
+    hashed_email: new RegExp(/[0-9abcdef]{64}/gi).test(String(payload?.email)) ? payload?.email : hash(payload?.email),
```
nick-Ag, line 354: Could we move the regex into a method to keep the code readable?
```suggestion
    hashed_email: isHashedEmail(payload?.email) ? payload?.email : hash(payload?.email),
```

nick-Ag (approved):
Could you add the method to the two places in Google Enhanced Conversions where you are using the regex as well? Other than that lgtm 👍

PR #1245, opened 2023-05-10
Sev-4480/fix sms subscription status and logging

pmunin:
Fixed behavior of sms sender to support channelType:
- we use the first _subscribed_ externalId even if it's not the first one in the list of supported channelTypes
- we don't throw error on unrecognized subscription status, but just log it, stats it and treat it as Not Sendable.

Logging enriched for different steps so it's always clear what happened on which step and what caused the exit

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [x] Added & updated [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] [Segmenters] Tested in the staging environment

PR #1255, opened 2023-05-12
Sev-4480/fix sms subscription w logging

pmunin:
Continuation of https://github.com/segmentio/action-destinations/pull/1245

Which had to be reverted from production because of Error Code containing the entire message, which increased Datadog tag cardinality

This PR has a minor fix for that error code and only put there the code from Twilio response or if it's not available then statusCode

## Testing


packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
```diff
@@ -145,7 +145,7 @@ export class SmsMessageSender extends MessageSender<Payload> {
       this.tags.push('reason:get_content_template')
       this.statsClient?.incr('actions-personas-messaging-twilio.error', 1, this.tags)
       this.logError(
-        `SMS failed request to fetch content template from Twilio Content API - ${this.settings.spaceId}, ${JSON.stringify(error)})}`
+        `SMS failed request to fetch content template from Twilio Content API - ${this.settings.spaceId} - ${error}, ${JSON.stringify(error)})}`
```
varadarajan-tw, line 148: Is this change intentional? Because JSON.stringify(error) will get you the information needed in the logs. Adding error object might make the long look unreadable.

varadarajan-tw (approved):
Added one question about the change to log statement. Otherwise, the PR LGTM.

packages/destination-actions/src/destinations/engage-messaging-twilio/sendSms/sms-sender.ts
pmunin: Yeah, unfortunately it's not always that easy:
image
It depends on the error I guess, but I added that just in case JSON won't work for that error

varadarajan-tw: Got it! 👍

marinhero:
I'm failing to understand how this is going to fix the issue as you are passing the same parameters or more. For example, I don't think we should log stuff like this `${this.settings.spaceId}` that would effectively create a log aggregate per space/source. You should be able to already filter by that within the Datadog dashboard.

PR #1273, opened 2023-05-18
Iterable destination updates

KakeJopulsky:
Updates for the Iterable destination:
* Updates destination auth call
* Removes preferUserId field from all actions
* Removes dataFields input from items and sets additionalProperties: true in commerce actions  
* Adds descriptions to items inputs in commerce actions
* Adds utils.js file that contains date formatting code
* Updates snapshots

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/iterable/index.ts
+++ b/packages/destination-actions/src/destinations/iterable/index.ts
@@ -42,7 +42,7 @@ const destination: DestinationDefinition<Settings> = {
   presets: [
     {
       name: 'Track Calls',
-      subscribe: 'type = "track" and event != "Order Completed" and event != "Update Cart"',
+      subscribe: 'type = "track" and event != "Order Completed" and event != "Cart Updated"',
       partnerAction: 'trackEvent',
       mapping: defaultValues(trackEvent.fields)
     },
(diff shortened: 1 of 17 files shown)


joe-ayoub-segment:
Thanks @KakeJopulsky , I'll have feedback for you in the next 15 mins

joe-ayoub-segment (changes requested):
Few minor things: 

1. Can we rename the Action to be Upsert User instead of Update User?

2. Can we change the name of the event from "Update Cart" to "Cart Updated" so it matches the Segment spec better?

3. For "Track Purchase" Action, do we need to reference "dollars" in the description for the total field? If Iterable only works in dollars that's good, otherwise I think we should be currency agnostic. 

4. I think we shouldn't hardcode the collection of the phone trait. The customer won't know this is happening, and it's the type of 'magic' we don't want to have in our mappings. I tried a few different options on my local machine and think the best option would be to include a completely separate new field just for phone, and then provide default mappings to both '$.traits.phone' or '$.context.traits.phone'. Keen to hear wha you think?

KakeJopulsky:
Thanks for the feedback, @joe-ayoub-segment! I push a few commits per your suggestions.

joe-ayoub-segment:
Hi @KakeJopulsky .  This PR has been deployed.

PR #1276, opened 2023-05-18
Standalone and independently versioned browser-destinations

zikaari:
This pull requests makes each browser destinations accessible through NPM by making it standalone and independently versioned. This will let customers install the destinations from NPM and have it packaged into their main app bundle through webpack or other bundlers, thus eliminating the reliance on CDN and a lot of concurrent network requests.

This pull request is multi-stage process, each of which is a separate commit so that while reviewing you're not being blasted with all the changes at once thus overwhelming you too much. It is advised that you use the commit filter in to isolate and audit one commit at a time.

Most of the existing functionality and behaviors work with the exception of the following:

- `build-cjs` script has been removed, unless there's an explicit need for it, the output is not warranted
- Manifest generation needs to be sorted for CLI consumption, this can be discussed after on strategies to do it right


NolanChan:
Hi @zikaari , I'm more than happy to move to a convo on slack, but tldr; actions team was hoping to deprecate the cli here in favor of the private action-cli (for security reasons). Would you be able to make these changes in the action-cli repo instead?

silesky:
@zikaari Can we have documentation on the new steps to publish an action destinations to npm?

danieljackins (approved):
Looks good to me - in particular functionality with browser action bundles/versioning still looks good. I think getting sign off from someone on the action destinations team would be good too.

silesky:
I only have one concern -- I saw the name format "@segment/actions-[pluginName]" -- would it make sense to be "@segment/analytics-actions-[pluginName]"   (I don't feel strongly, but it seems like using 'analytics' might be helpful for SEO / discoverability, since this is tightly coupled to analytics.js)

NolanChan (approved):
Thanks for working with us and testing this! Appreciate the meeting walking us through ✅

silesky (approved):

packages/destination-subscriptions/package.json
```diff
@@ -22,16 +22,6 @@
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "types": "dist/cjs/index.d.ts",
-  "exports": {
```
NolanChan, line 25: why is this here?

PR #1277, opened 2023-05-19
Introduce API Error Class

varadarajan-tw:
This PR introduces a new API Error class and new Error Code - API_CALL_FAILED. 

The new error class and code can be used to categorize API errors handled by builders within action code. Builders may decide to capture and rethrow API errors with better contextual error messages or interpret response and decide whether the API call is a failure or a success. **Emarsys** is the only destination that currently seem to need this error class right now. 

The previous implementation of these errors didn't assign an error code or status. Such error are considered as unhandled errors by integrations and show up in Sentry.

Emarsys is a partner build destination and has very low traffic in production. So, there is a chance that we won't see these errors until we see more adoption. These error code changes are safe and seem appropriate. These errors will not be retried.

**Note** 
- We currently don't show error messages in Event delivery page except for **Payload Validation Errors**. The error messages themselves are of no use for Integration Errors or API Error.

## Testing
I don't have an emarsys account. So, i used HubSpot to simulate the new error in staging. This is what they look like in Event delivery. 

image

The message shown in this pane can be customized from contentful
image

### DIFF
--- a/packages/destination-actions/src/destinations/emarsys/addToContactList/index.ts
+++ b/packages/destination-actions/src/destinations/emarsys/addToContactList/index.ts
@@ -1,4 +1,4 @@
-import { IntegrationError, RetryableError } from '@segment/actions-core'
+import { RetryableError, PayloadValidationError, APIError } from '@segment/actions-core'
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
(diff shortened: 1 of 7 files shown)


packages/destination-actions/src/destinations/emarsys/addToContactList/index.ts
```diff
@@ -46,9 +46,9 @@ const action: ActionDefinition<Settings, Payload> = {
     }
   },
   perform: async (request, data) => {
-    if (!data?.payload?.key_field) throw new IntegrationError('Missing key field')
+    if (!data?.payload?.key_field) throw new PayloadValidationError('Missing key field')
```
varadarajan-tw, line 49: This should be a safe change. PayloadValidationError is already used by many destinations in production indicate a required mapping is missing. The destination won't be able to proceed with execution in these cases.

packages/destination-actions/src/destinations/emarsys/removeFromContactList/index.ts
```diff
@@ -46,9 +46,9 @@ const action: ActionDefinition<Settings, Payload> = {
     }
   },
   perform: async (request, data) => {
-    if (!data?.payload?.key_field) throw new IntegrationError('Missing key field')
+    if (!data?.payload?.key_field) throw new PayloadValidationError('Missing key field')
 
-    if (!data?.payload?.key_value) throw new IntegrationError('Missing key value')
+    if (!data?.payload?.key_value) throw new PayloadValidationError('Missing key value')
```
varadarajan-tw, line 51: The above two validations are actually not required. key_field and key_value are required fields and will be validated by the framework itself.

PR #1301, opened 2023-05-25
Initialize new NOOP internal destination

LLuque-twilio:
This PR addresses JIRA [EN-102](https://segment.atlassian.net/browse/EN-102)

In it, we create an empty action-destination for an unnamed internal service.

## Testing

Testing not needed because it is an empty destination used for an internal service.

[EN-102]: https://segment.atlassian.net/browse/EN-102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/destination-actions/src/destinations/noop/noop/index.ts
+++ b/packages/destination-actions/src/destinations/noop/noop/index.ts
@@ -1,0 +1,24 @@
+import { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Noop',
+  description: 'A NOOP Action used for private internal services',
+  fields: {
+    noop: {
+      label: 'NOOP',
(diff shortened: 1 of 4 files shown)


dlasky (approved):

packages/destination-actions/src/destinations/noop/noop/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  description: 'A NOOP Action used for private internal services',
+  fields: {
+    noop: {
+      label: 'NOOP',
+      description: 'A single NOOP field',
+      type: 'string'
+    }
+  },
+  perform: (_, { statsContext }) => {
+    statsContext?.statsClient.incr('actions_noop_perform_hit', 1)
```
LLuque-twilio, line 16: FWIW this perform block should never actually be hit, just adding a stat here so we can monitor it just in case.

PR #1319, opened 2023-06-02
Add type suffixes to nested FullStory cloud mode properties

ndonze-fs:
Before this PR, FullStory custom event and custom user properties were type suffixed as required by FullStory's v1 events and users APIs. However, type suffixing was only applied to root level track properties and identify traits. In practice, mutual Segment and FullStory customers often pass in track properties or identify traits which are complex objects containing nested properties. Since these nested properties were not type suffixed, these track and identify events would fail to be processed by FullStory's server APIs.

This PR:
- Expands type suffixing for FullStory cloud mode properties up to 5 levels deep -- a depth which appears sufficient for the majority of FullStory customer custom event and custom user properties.
- Stops adding an `_obj` type suffix to inferred object properties. This matches type suffixing behavior with FullStory browser API behavior and isn't expected to negatively impact mutual FullStory and Segment customers already using the FullStory Segment cloud mode destination.

## Testing

Unit tests were run, and full integration tests were carried out against FullStory production server APIs.

### DIFF
--- a/packages/destination-actions/src/destinations/fullstory/__tests__/var.test.ts
+++ b/packages/destination-actions/src/destinations/fullstory/__tests__/var.test.ts
@@ -1,22 +1,21 @@
 import { normalizePropertyNames } from '../vars'
 
-const suffixToExampleValuesMap: Record<string, any[]> = {
-  str: ['some string'],
-  bool: [true, false],
+const suffixToExampleValuesMap: Record<string, any> = {
+  str: 'some string',
+  strs: ['string 1', 'string 2'],
+  bool: true,
+  bools: [true, false],
+  real: 1.23,
   // We include an int value with real example values since real will always be preferred when inferring
   // custom var types from values.
-  real: [1, 1.23],
-  int: [1],
-  date: [new Date()],
-  obj: [{}, { secondLevelProp: 'some value' }]
+  reals: [1, 4.56],
+  int: 1,
+  ints: [2, 3],
+  date: new Date(2000, 1),
+  dates: [new Date(2000, 2), new Date(2000, 3)],
+  obj: {},
+  objs: [{}, { nested_string_str: 'nested string' }]
 }
-
-suffixToExampleValuesMap.strs = [suffixToExampleValuesMap.str]
-suffixToExampleValuesMap.bools = [suffixToExampleValuesMap.bool]
-suffixToExampleValuesMap.reals = [suffixToExampleValuesMap.real]
-suffixToExampleValuesMap.ints = [suffixToExampleValuesMap.int]
-suffixToExampleValuesMap.dates = [suffixToExampleValuesMap.date]
-suffixToExampleValuesMap.objs = [suffixToExampleValuesMap.obj]
 
 describe('normalizePropertyNames', () => {
   it('does not add type suffix for undefined values', () => {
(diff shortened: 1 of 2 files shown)


ndonze-fs:
I'm going to close this pull requests as I'd like to consider one additional change to our property name type suffixing logic before we merge and deploy this.

ndonze-fs:
Reopening this PR after adding a related change to stop adding `_obj` type suffixes to inferred object properties.

ndonze-fs:
Reopening this PR after adding a related change to stop adding `_obj` type suffixes to inferred object properties.

joe-ayoub-segment (commented):
Hi @ndonze-fs , 
Thanks for raising this PR. 

Looks like the main reason for this change is to support transforming the names of object properties at up to 5 levels deep. 

The code looks fine to me. 

However our main priority with a change like this is that it doesn't negatively impact customers who are already using this Integration to send data to FullStory. 

Could you explain a little more how this was tested? 
For example it would be great to be able to see an example of a payload coming into Segment, and how that payload gets transformed into a payload which gets sent to Fullstory (both before and after this change). Would that be possible?

Feel free to attach screenshots of any testing. 

Or if you prefer, we could catch up to do some tests. If you'd like to do this please schedule a call with me using this link: http://calendly.com/joe_ayoub/integration-workshop 

Other notes: 
1. There are no new fields being added, or no changes to existing fields. 
2. Tests and yarn validate pass

Kind regards,
Joe

ndonze-fs:
Hi @joe-ayoub-segment,

Before this change, attempting to send identify traits or track properties containing an object with nested properties to FullStory would fail with a 400 Bad Request error:

01_event_tester_400_error_nested_properties

Our FullStory cloud mode destination would translate the Segment identify payload into the following request payload for FullStory's server APIs:

```
{
  "serverSingle_obj": {
    "text": "some string"
  }
}
```

This payload has 2 important characteristics which this PR addresses:
1. `serverSingle_obj` has been suffixed with the `_obj` type suffix, which we've learned is not only unnecessary but also deviates from FullStory's browser API behavior -- setting customers up for an inconsistent experience.
2. More important relative to the 400 Bad Request response, the `text` property has not been suffixed with the `_str` type suffix as required by FullStory server APIs.

As a result, customers have not been able to send objects with nested properties as identify traits or track properties via the FullStory cloud mode destination. All these requests would have returned 400 Bad Request responses unless the customer themself added required type suffixes to nested properties in the original traits or properties sent to Segment. We've verified on our end that mutual customers have not been adding these type suffixes on their own.

This PR:
- Stops adding the unnecessary `_obj` type suffix to object properties containing nested properties
- Adds required type suffix properties to nested object properties up to 5 levels deep

As a result, the previous identify payload would make a request to FullStory's server APIs with the following payload:

```
{
  "serverSingle": {
    "text_str": "some string"
  }
}
```

Notice:
1. The `serverSingle` property has not been type suffixed with `_obj` which is the desired behavior
2. The `text_str` property has received the `_str` type suffix as required by FullStory server APIs

As a result (verified through the local Action Tester and production FullStory server APIs):
- Customers will no longer receive 400 Bad Request responses when including an object with nested properties as identify traits or track properties
- Customers will be able to see these traits or properties within FullStory and filter on them

02_fs_serverSingle_view

03_fs_serverSingle_filter

Local Action Tester with successful response sending identify traits containing an object with nested properties:

04_action_tester_success

While only the identify test appears in this screenshot, an equivalent test was verified with track as well.

joe-ayoub-segment (approved):

ndonze-fs:
@joe-ayoub-segment Testing looks good between our updated cloud mode destination and FullStory. Thank you!

PR #1320, opened 2023-06-05
Stratconn 2625 add error when no users are extracted

hvardhan-unth:
This Pull request is to handle the case when email and/or advertising_id is not in the payload for TikTok audiences,  then don't extract any users and no request is sent however it just return a "successful" response for the events.  So, in this case now an error will thrown.

## Testing


marinhero (approved):
Nice work! Thanks for adding a unit test 🙏

joe-ayoub-segment:
hi @hvardhan-unth your PR has been deployed.

PR #1328, opened 2023-06-07
Koala Server Side Destination

nettofarah:
This is the initial implementation of the Koala server side tracking API.
The main idea is that we'd like to proxy server side `track` and `identify` calls from Segment directly into Koala.

There are still a couple of open questions that I would love to get some clarity on:

### 1. Is there a way that we could only proxy events that did not originate from AJS?

I suspect the answer would be to inspect the `context` object and check the originating library.
We'd like to check for this because the Koala client side integration already listens to all Segment events automatically 

### 2. Is it possible to skip an API call?
We would like to only perform API calls to Koala if certain conditions are met, such as the payload containing a few special keys.

Right now, we do that by returning `undefined` instead of returning a request, but the event tester returns a 404 when we do that. 

## Testing


packages/destination-actions/src/destinations/koala/index.ts
```diff
@@ -0,0 +1,38 @@
...
+    fields: {
+      public_key: {
+        label: 'Public Key',
+        description: 'Your public key',
+        type: 'string',
+        required: true
+      }
+    },
+    testAuthentication: (request, { settings }) => {
+      return request(`${KOALA_PATH}/${settings.public_key}/batch`, { method: 'POST' })
```
mshwery, line 28: @nettofarah rather than a request to /batch – should we hit /metrics or the settings endpoint?

packages/destination-actions/src/destinations/koala/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const KOALA_PATH =
+  process.env.NODE_ENV === 'development'
+    ? 'http://localhost:3001/web/projects'
+    : 'https://api2.getkoala.com/web/projects'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Koala',
+  slug: 'actions-koala',
+  mode: 'cloud',
```
mshwery, line 15: Curious if it's cool to use the same slug as the client-side destination?

packages/destination-actions/src/destinations/koala/track/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  title: 'Track',
+  description: 'Send an Event to Koala.',
+  defaultSubscription: 'type = "track"',
+  fields: {
+    email: {
+      type: 'string',
+      label: 'Email',
+      description: 'The email to associate with the user',
+      required: false,
+      default: { '@path': '$.email' }
```
mshwery, line 16: Is there a top-level Segment event key called `email`?

packages/destination-actions/src/destinations/koala/track/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    message_id: {
+      type: 'string',
+      required: true,
+      description: 'The Segment messageId',
+      label: 'MessageId',
+      default: { '@path': '$.messageId' }
+    }
+  },
+  perform: (request, data) => {
+    const profileId = data.payload.properties?.profile_id
```
mshwery, line 70: Should we add `profile_id` as a field?

packages/destination-actions/src/destinations/koala/track/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    const traits = data.payload.traits ?? {}
+    const email = data.payload.email ?? data.payload.properties?.email ?? traits.email
+    const ip = data.payload.properties?.ip ?? data.payload.device_ip
+
+    if (!profileId && !email) {
+      // Skip call if no identifier is found
+      return
```
mshwery, line 77: Could do `return Promise.resolve()` if it must be a promise?

packages/destination-actions/src/destinations/koala/index.ts
nettofarah: Yeah, I think those would work too. I could also create a new test-auth one.

packages/destination-actions/src/destinations/koala/index.ts
nettofarah: Me too! This was auto-generated by the CLI.

packages/destination-actions/src/destinations/koala/track/index.ts
nettofarah: Yes! It's available when you identify someone with an email.

packages/destination-actions/src/destinations/koala/track/index.ts
nettofarah: You mean as path mapping?

packages/destination-actions/src/destinations/koala/track/index.ts
nettofarah: This is not failing compilation, but I could do it.

packages/destination-actions/src/destinations/koala/track/index.ts
mshwery: Yeah i meant in `fields` above. But it probably doesn't make much difference since you would have to know to pass it in the event properties anyway. Ignore me!

packages/destination-actions/src/destinations/koala/track/index.ts
joe-ayoub-segment: hi @nettofarah and @mshwery - I'm not reviewing this PR, but noticed @mshwery 's question about email being at the root level of a payload. I don't ever recall seeing this in my 4.5 years with Segment. 
@nettofarah, any chance you could explain how this happens in practice please? Is there a doc or API call you can share showing this? 
Reason I want to know is that I get the same question from Partners all the time!

nettofarah: hey, @joe-ayoub-segment.
I can't remember exactly how it gets there, but I believe it being part of the Event spec when I built the first version of Analytics.js 2.0.

All I remember is that it is part of the spec, I suspect some library also sets it when they see an email.

packages/destination-actions/src/destinations/koala/track/index.ts
nettofarah: @pooyaj Is there any special way to handle a "skipped request"?

pooyaj: I think either resolve the promise ( if it is not considered an error ) or could do something like this: 
```
      throw new IntegrationError(
        `ProfileId and email are missing`,
        'Invalid payload',
        400
      )
```
which will surface the error in the delivery tab for the destination.

packages/destination-actions/src/destinations/koala/index.ts
pooyaj: Should use a different slug here as there are some uniqueness guarantees on slugs

packages/destination-actions/src/destinations/koala/track/index.ts
pooyaj: This should be: 
```
      default: {
        '@path': '$.traits.email'
      }
```
I think AJS sets userId at the top level event ( which could be email, or other Ids )

packages/destination-actions/src/destinations/koala/index.ts
nettofarah: what do you suggest @pooyaj ?

packages/destination-actions/src/destinations/koala/track/index.ts
nettofarah: Could we handle it as a condition instead? I would prefer to drop the API call, because it's theoretically not an error

packages/destination-actions/src/destinations/koala/index.ts
pooyaj: `actions-koala-cloud` maybe?

packages/destination-actions/src/destinations/koala/track/index.ts
pooyaj: Yes, in that case doing the `return Promise.resolve()` should be more than enough. It will be considered success from the integrations pipeline.

nettofarah: Awesome. Thank you!

packages/destination-actions/src/destinations/koala/lib/ajs-detection.ts
```diff
@@ -0,0 +1,9 @@
+export interface AnalyticsEventContext {
+  library?: {
+    name?: string
+  }
+}
+
+export function isAJSEvent(eventContext?: AnalyticsEventContext): boolean {
```
pooyaj, line 7: Is the purpose of this method to avoid sending the same data twice ( once via browser and once via server )? If so, this probably won't solve for that, because the data may be coming from AJS, but doesn't necessarily mean it was sent to Koala client-side. User has to connect the Koala Web destination for client-side delivery.

pooyaj (approved):
Looks great 🙌 only one comment around AJS detection

packages/destination-actions/src/destinations/koala/lib/ajs-detection.ts
nettofarah: Yes and no.

We actually have automatic segment detection enabled in the Koala pixel that will piggy back on existing AJS events, even if it wasn't enabled client side via Segment.

joe-ayoub-segment (approved):
@pooyaj reviewed and said that this PR looks good. So I will not review. 
Merging.

PR #1350, opened 2023-06-21
Stratconn 2684 dynamic field to select audience

hvardhan-unth:
TikTok require users to input the audience_id that they want to sync to. In this PR the audience id in mapping is converted to dynamic  field. 

## Testing

Screenshot:- 
Screenshot 2023-06-22 at 12 16 12 AM


packages/destination-actions/src/destinations/tiktok-audiences/addUser/index.ts
```diff
@@ -43,6 +43,21 @@ const action: ActionDefinition<Settings, Payload> = {
           }
         }
       }
+    },
+    audience_id: async (request, { payload }) => {
+      try {
+        const tiktok = new TikTokAudiences(request)
+
+        return tiktok.fetchAudiences(payload.selected_advertiser_id)
```
varadarajan-tw, line 51: For the catch block to work, this statement needs to be awaited.
```suggestion
        return await tiktok.fetchAudiences(payload.selected_advertiser_id)
```

varadarajan-tw: Another option is chaining catch and returning a value.
```suggestion
        return tiktok.fetchAudiences(payload.selected_advertiser_id).catch(err => { 
           return {
          choices: [],
          error: {
            message: JSON.stringify(err),
            code: '500'
          }
        }
        }
        })
```

packages/destination-actions/src/destinations/tiktok-audiences/removeUser/index.ts
```diff
@@ -43,6 +42,21 @@ const action: ActionDefinition<Settings, Payload> = {
...
+        return tiktok.fetchAudiences(payload.selected_advertiser_id)
+      } catch (err) {
+        return {
+          choices: [],
+          error: {
+            message: JSON.stringify(err),
+            code: '500'
+          }
+        }
+      }
```
varadarajan-tw, line 59: Pls read suggestions in addUser action and see if you can apply the same.

packages/destination-actions/src/destinations/tiktok-audiences/api/index.ts
```diff
@@ -41,20 +42,66 @@ export class TikTokAudiences {
     })
   }
 
-  async getAudiences(page_number: number, page_size: number): Promise<GetAudienceAPIResponse> {
```
varadarajan-tw, line 44: was getAudiences get renamed or removed?

varadarajan-tw (commented):
Left few comments. Could you add tests for dynamic fields? [Example](https://github.com/segmentio/action-destinations/blob/70d93d3ccad42ffd8ac5d96b970a0e2d2866a214/packages/destination-actions/src/destinations/hubspot/upsertCustomObjectRecord/__tests__/index.test.ts#L149)

packages/destination-actions/src/destinations/tiktok-audiences/api/index.ts
hvardhan-unth: It is renamed as `fetchAudiences`.

maryamsharif (changes requested):
Can you include tests on the error messages to ensure they appear correctly? Also, I think this may break mappings customers already have. When I go into my staging destinations that I had set up before it now says that I am missing required fields and it is missing the audience_id I had previously inputed. Please do not merge until we ensure that it will not break customers current mappings.

maryamsharif (approved):
Based on the tests in here : https://docs.google.com/document/d/1mxw2sVoNlz2UrL1gZNYM-cvJzqS9EafP7WzFgHJL1t4/edit#heading=h.6qlwdta7t6ue changing the field does not cause any breaking changes.

PR #1355, opened 2023-06-23
Audience settings and Webhook audience as an example

SyedWasiHaider:
This PR adds first class support for audience-specific functionality in actions. Specifically creating/getting audiences are now built-in functions which will be called prior to any calls to `perform` (or `performBatch`). This ensures that implementors don't have to use any workarounds such as adding a "createAudience" action or pollute their `perform`/`performBatch` block with extraneous calls to ensure the external audience exists.

We've chosen webhooks as an initial pilot destination and this destination will also serve as an example implementation for future audience based destinations.

Future work:
* Support for real-time audience (this PR adds the type but at the moment, it doesn't actually do anything).
* Add ability to specify a request per second limit which ultimately would limit the max size of batches that can flow in to `performBatch`

## Testing


packages/core/src/destination-kit/types.ts
```diff
@@ -76,6 +76,7 @@ export interface GlobalSetting {
   default?: string | number | boolean
   properties?: InputField['properties']
   format?: InputField['format']
+  audienceOnly?: boolean
```
SyedWasiHaider, line 79: TODO: need to delete this

packages/core/src/destination-kit/index.ts
```diff
@@ -61,6 +61,32 @@ export interface BaseDefinition {
   presets?: Subscription[]
 }
 
+export type AudienceResult = {
+  externalId: string
+}
+
+export type AudienceMode = { type: 'realtime' } | { type: 'synced'; full_audience_sync: boolean }
```
NolanChan, line 68: woah, a conditional setting 😄 this makes so much sense in typescript haha

packages/destination-actions/src/destinations/webhook-audience/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  authentication: {
+    scheme: 'custom',
+    fields: {
+      sharedSecret: {
+        type: 'string',
+        label: 'Shared Secret',
+        description:
+          'If set, Segment will sign requests with an HMAC in the "X-Signature" request header. The HMAC is a hex-encoded SHA1 hash generated using this shared secret and the request body.'
+      },
+      createAudienceUrl: {
```
NolanChan, line 24: Curious, do we need explicit settings for a `createAudienceUrl` or `getAudienceUrl`? I know that for action-destinations, some (like intercom actions) hardcode the url right into the perform block. Other's (like amplitude) have helper methods (like in `regional-endpoints.ts`).  I know Braze has it as a setting I think... I'm just wondering since we might not want to lock in customers to use one url if they have many

packages/core/src/destination-kit/index.ts
```diff
@@ -61,6 +61,32 @@ export interface BaseDefinition {
...
+  audienceName: string
+}
+
+export type GetAudienceInput<Settings = unknown> = {
+  settings: Settings
+
+  externalId: string
+}
+
+export interface AudienceDestinationSettings<Settings = unknown> {
```
pooyaj, line 82: Wonder if we should tweak the type of ensure either both `createAudience` and `getAudience` are defined, or none of them is defined.

packages/destination-actions/src/destinations/webhook-audience/index.ts
```diff
@@ -0,0 +1,108 @@
+import type { AudienceResult, DestinationDefinition } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+
+import send from '../webhook/send'
+import { IntegrationError } from '@segment/actions-core'
+import { createHmac } from 'crypto'
+
+const externalIdKey = 'externalId'
```
marinhero, line 8: What are these and why are they required?

packages/destination-actions/src/destinations/webhook-audience/index.ts
```diff
@@ -0,0 +1,108 @@
...
+        label: 'Shared Secret',
+        description:
+          'If set, Segment will sign requests with an HMAC in the "X-Signature" request header. The HMAC is a hex-encoded SHA1 hash generated using this shared secret and the request body.'
+      },
+      createAudienceUrl: {
+        type: 'string',
+        label: 'URL to send requests to create an audience',
+        description: `If set, Segment will send a POST request with the ${audienceNameKey} in the JSON to the provided URL to create the audience. The expected JSON response must have "${externalIdKey}".`
+      },
+      getAudienceUrl: {
```
marinhero, line 29: Isn't `getAudience` an operation that we run internally to get the external ID from our own DB?

packages/destination-actions/src/destinations/webhook-audience/index.ts
marinhero, line 24: Not sure if these should be settings instead of part of the `createAudience` method implementation as they rarely change.

packages/core/src/destination-kit/index.ts
marinhero: +1 with @pooyaj for a tighter interface.

marinhero (commented):
Looking good! Left a few comments :)

packages/destination-actions/src/destinations/webhook-audience/index.ts
SyedWasiHaider: The user needs to confirm to some expected structure. Otherwise, we won't know what property to use as the external id from the webhook response.

packages/core/src/destination-kit/index.ts
SyedWasiHaider: good call

packages/destination-actions/src/destinations/webhook-audience/index.ts
SyedWasiHaider: we also make a call to the end partner api to ensure it is a valid external id

packages/core/src/destination-kit/index.ts
SyedWasiHaider: Actually given that both properties are non-nullable, isn't that already true?

pooyaj: Currently they both have to be defined regardless of the mode. Ideally we want to make them both or none ( so that we can define audience destinations that do not implement create/get Audience methods )

packages/destination-actions/src/destinations/webhook-audience/index.ts
pooyaj: Given this is a generic webhook it makes sense not to hardcode the URL

packages/core/src/destination-kit/index.ts
SyedWasiHaider: ah I see what you mean

valerieernst (approved):
:dinostomp:

marinhero (approved):

PR #1358, opened 2023-06-26
[CCP-240] - Update email destination action to accept and call API lookups

ryanrouleau:

This PR adds the ability for API lookups to be called from the sendgrid engage destination action.  API lookups allow customers to make arbitrary 3rd party API calls at the send time of their message and use the response to customize their messages (through liquid syntax).

This PR supports our POC, we will be adding a caching mechanism in an upcoming PR.

More info on API lookups: https://docs.google.com/document/d/1-ietddV4hWdzue_WkPqY-BvIyU024vzsIpOLp2XvckE/edit?usp=sharing

https://segment.atlassian.net/browse/CCP-240

## Testing

- [ x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [x ] [Segmenters] Tested in the staging environment


joe-ayoub-segment:
Thanks for raising this @ryanrouleau - I'll schedule for review.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/utils/api-lookups.ts
```diff
@@ -0,0 +1,154 @@
...
+    logger?.error(
+      `TE Messaging: Email api lookup body parse failure - api lookup id: ${id} - ${settings.spaceId} - [${error}]`
+    )
+    tags.push('reason:parse_apilookup_body')
+    statsClient?.incr('actions-personas-messaging-sendgrid-error', 1, tags)
+    throw new IntegrationError('Unable to parse email api lookup body', 'api lookup body parse failure', 400)
+  }
+
+  try {
+    const res = await request(renderedUrl, {
```
zawadzkip, line 56: https://github.com/segmentio/action-destinations/blob/2abde6b72f8a5b52175389968ce1b8e15e79af3b/packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts#L231

it might be worth adding `skipResponseCloning` to this. We had a sev in the past because certain responses would be too large and cause the requests to silently fail. I would double check to see if the bug that requires this work around was fixed but since we can't except certain response sizes it might be worth adding.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -224,18 +218,45 @@ const attemptEmailDelivery = async (
   }
 
   const bcc = JSON.parse(payload.bcc ?? '[]')
-  const parsedSubject = await parseTemplating(payload.subject, profile, 'Subject', statsClient, tags, settings, logger)
+  const parsedSubject = await parseTemplating(
```
zawadzkip, line 221: might be over kill but do you think it would be worth making `parseTemplating` and `performApiLookups` in a `Promise.all` here?

mbreske: @zawadzkip good call

ryanrouleau: 👍

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
```diff
@@ -458,6 +464,13 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'boolean',
       default: false
     },
+    apiLookups: {
+      label: 'API Lookups',
+      description: 'Any API lookup configs that are needed to send the template',
+      type: 'object',
+      multiple: true,
+      properties: apiLookupActionFields
```
varadarajan-tw, line 472: For my learning - Does this destination get exposed to the customer in the UI?   Because the Action Tester UI is not able to show an array of objects properly and I doubt it will work in the app. I see another field already defined in the destination that is configured similarly.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/previewApiLookup/index.ts
```diff
@@ -0,0 +1,31 @@
...
+      required: false
+    }
+  },
+  perform: async (request, { settings, payload, statsContext, logger }) => {
+    const statsClient = statsContext?.statsClient
+    const tags = statsContext?.tags ?? []
+    const { traits = {}, ...apiLookupConfig } = payload
+    const profile: Profile = { traits: traits as Record<string, string> }
+    const data = await performApiLookup(request, apiLookupConfig, profile, statsClient, tags, settings, logger)
+    return {
```
varadarajan-tw, line 25: I would like to understand this new action a little better. 
How will this previewApiLookUp action be used?  Is the response important here? If so, how is it parsed?

varadarajan-tw (commented):
Hey, I am filling in for Joe this week. The code largely looks good to me. I left a couple of questions for my understanding.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/index.ts
ryanrouleau: No it is not exposed in the UI.  This destination is used internally for engage only

packages/destination-actions/src/destinations/engage-messaging-sendgrid/previewApiLookup/index.ts
ryanrouleau: Yes a little background - we needed a way to call an api lookup (formerly known as data feeds) outside of sending an email, then display the response of it back to the user so they can test/preview their configuration (see attached screenshot of the design for more context).  

We created a new action to do this because we need these arbitrary 3rd party APIs (that customers configure) to be called outside of the internal segment network for security reasons.  Doing this in destination actions is not ideal and is a temporary solution until we either 1) build out our own api lookup/data feed service 2) build on top of the new enrichment platform.

It will only be called through the /test-action endpoint from engage and will be triggered through the UI.  personas-service will be responsible for parsing the response and returning it to the UI.

Screenshot 2023-07-05 at 12 21 11 PM

rhall-twilio (approved):

PR #1365, opened 2023-06-28
Custom validation in Google Ads for email address to allow hashed email addresses.

seg-leonelsanches:
Disabling default AJV default email validation to allow incoming hashed emails.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
@@ -3,6 +3,7 @@ import { ConversionCustomVariable, PartialErrorResponse, QueryResponse } from '.
 import { ModifiedResponse, RequestClient, IntegrationError, PayloadValidationError } from '@segment/actions-core'
 import { StatsContext } from '@segment/actions-core/src/destination-kit'
 import { Features } from '@segment/actions-core/src/mapping-kit'
+import { fullFormats } from 'ajv-formats/dist/formats'
 
 export const API_VERSION = 'v12'
 export const CANARY_API_VERSION = 'v13'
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
```diff
@@ -95,3 +95,14 @@ export function getApiVersion(features?: Features, statsContext?: StatsContext):
 }
 
 export const isHashedEmail = (email: string): boolean => new RegExp(/[0-9abcdef]{64}/gi).test(email)
+export const commonHashedEmailValidation = (email: string): string => {
```
NolanChan, line 98: just curious about what "common" means in the name?

packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
```diff
@@ -95,3 +95,14 @@ export function getApiVersion(features?: Features, statsContext?: StatsContext):
...
+  if (isHashedEmail(email)) {
+    return email
+  }
+
+  if (!/^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/.test(email)) {
```
NolanChan, line 103: also curious how this regex was formed? I'm not too great with regex :D

packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
seg-leonelsanches: @NolanChan I reuse the method in two actions.

packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
NolanChan: I'm personally wary of trying to roll our own email validator. It might not be straightforward, but I'd rather try to lean on AJV or possibly something we're already using to validate emails. I don't have a clear solution in my mind, but I don't think we want to have our own email regex

NolanChan: Maybe you could use this instead! https://github.com/ajv-validator/ajv-formats/blob/master/src/formats.ts#L64-L65

seg-leonelsanches: @NolanChan Thanks for the suggestion. I added the Regex. As the package is a beta, I decided to not add the package right now.

NolanChan: Hey Leonel, I believe it's ok to add the package since we already use the package elsewhere

NolanChan (approved):

PR #1369, opened 2023-06-30
Changed object Id to number

hvardhan-unth:
Change the field type to number for ObjectId from Custom Behavioral Event requests. 
As per the API (/events/v3/send), the object Id in payload must be numeric. So, when customer gives string it throws 400 error. To prevent this error, change the type to number in mapping.

JIRA ticket link:-  https://segment.atlassian.net/browse/STRATCONN-2875

## Testing

Screenshot 2023-06-30 at 6 49 57 PM

### DIFF
--- a/packages/destination-actions/src/destinations/hubspot/sendCustomBehavioralEvent/index.ts
+++ b/packages/destination-actions/src/destinations/hubspot/sendCustomBehavioralEvent/index.ts
@@ -10,7 +10,7 @@ interface CustomBehavioralEvent {
   properties?: { [key: string]: unknown }
   utk?: string
   email?: string
-  objectId?: string
+  objectId?: number
 }
 
 const action: ActionDefinition<Settings, Payload> = {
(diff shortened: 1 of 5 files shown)


packages/destination-actions/src/destinations/hubspot/sendCustomBehavioralEvent/index.ts
```diff
@@ -57,7 +57,7 @@ const action: ActionDefinition<Settings, Payload> = {
       label: 'Object ID',
       description:
         'The ID of the object associated with this event. This can be the HubSpot contact ID, company ID, or ID of any other object. This is required if no email or user token is provided.',
-      type: 'string'
+      type: 'number'
```
varadarajan-tw, line 60: @hvardhan-unth - Could you link the hubspot document page where it says ObjectID has to be a number?

As per hubspot docs, the [Custom Behavioral Endpoint ](https://developers.hubspot.com/docs/api/analytics/events)expects objectId to be a string.
image

hvardhan-unth: As per the response from API call, I get to know that the objectId should be numeric.

Screenshot 2023-07-03 at 2 46 00 PM
Screenshot 2023-07-03 at 2 50 52 PM

varadarajan-tw: Got it! I believe they support both numeric and string values then. And in case if we provide a string value, it must be a number. Do we have high proportion of calls failing with this? How would this affect existing mappings in production?

PR #1370, opened 2023-06-30
Adding AppFit destination

claude:
This adds the AppFit (https://appfit.io) destination and sets up the initial action to track events.

## Testing


varadarajan-tw:
@claude - Thank you for the PR. I'll schedule a review. 

This seems like a new destination. Could you help me with answers to the following questions? This will help us understand and give you feedback appropriate for the destination.
1. What type of customers would want to use this destination?
2. Could you provide top 3 scenarios on how they would want to use it?

claude:
Thanks @varadarajan-tw:

**What type of customers would want to use this destination?**

AppFit is a product of Uptech Studio - a consulting company that helps startups and fast moving companies build launch and refine their products.
At Uptech Studio, we use AppFit as a way for our startup companies to easily look at their metrics on a weekly basis.
Right now, we have an integration with Amplitude, where we are putting all of our partners but we are looking to add Segment as another option for our customers.
Here is a list of all of our customers that we could potentially put on this integration: https://www.uptechstudio.com/clients

**Could you provide top 3 scenarios on how they would want to use it?**
A new startup that is not tracking metrics at all right now:
A startup that is currently not tracking their metrics could sign up for AppFit and use Segment to send their events to AppFit. In this case, they would then install the Segment SDK and set up the the events they wanted to track to AppFit. They would then go into Segment and add a new Destination for AppFit using the API key that we would provide.

An existing startup that is not tracking metrics, but are using Segment for other services like Marketing Automation
In this case, the team would need to sign up for AppFit account, add a data destination in Segment, and copy and paste their API key into Segment. They would then need to go into AppFit to configure the metrics that they wanted to track.

An existing startup that is tracking metrics through another analytics platform via Segment
In this case, the team would need to sign up for AppFit account, add a data destination in Segment, and copy and paste their API key into Segment. They would then need to go into AppFit to make their metrics match the metrics they are tracking elsewhere.

packages/destination-actions/src/destinations/app-fit/track/index.ts
```diff
@@ -0,0 +1,112 @@
...
+    },
+    osName: {
+      label: 'OS Name',
+      description: 'The name of the operating system',
+      type: 'string',
+      required: true,
+      default: {
+        '@path': '$.context.os.name'
+      }
+    },
```
varadarajan-tw, line 78: Do you expect the events to come from Mobile Apps always?

Pls note that deviceId, deviceType and osName are only captured in our Mobile SDKs. 
Refer [context-fields-automatically-collected](https://segment.com/docs/connections/spec/common/#context-fields-automatically-collected)

varadarajan-tw (commented):
Thanks for the explanation! The PR looks good to me. Just left a question for my understanding.

packages/destination-actions/src/destinations/app-fit/track/index.ts
claude: I didn't realize those fields were mobile only, just made them optional. Thanks!

maryamsharif (dismissed):
Thanks for making the suggested changes.

rhall-twilio (approved):

claude:
@rhall-twilio Thanks for approving! What are the next steps to getting this merged?

PR #1378, opened 2023-07-04
Added m3ter destination action

wkalucki:
This adds [m3ter](https://www.m3ter.com/) cloud destination, so that we can support customers that want to submit their usage data to m3ter based on segment events.

The destination uses oath to [authenticate with m3ter](https://www.m3ter.com/docs/guides/authenticating-with-the-platform/service-authentication)

Destination includes the `Submit Measurements` action which translates to [SubmitMeasurements](https://www.m3ter.com/docs/api#tag/Measurements/operation/SubmitMeasurements)  m3ter API operation. It will transform the segment event to m3ter measurement and submit it via REST API. The action supports batching and if possible it should be the prefered way of using the action. By default the action subscribes to [track](https://segment.com/docs/connections/spec/track/) events however its up to the customer to configure different event types if their usage data comes from different events.

### DIFF
--- a/packages/destination-actions/src/destinations/m3ter/auth.ts
+++ b/packages/destination-actions/src/destinations/m3ter/auth.ts
@@ -1,0 +1,20 @@
+import { M3TER_AUTH_API } from './constants'
+import { RequestClient } from '@segment/actions-core'
+
+export interface AccessTokenResponse {
+  token_type: string
+  access_token: string
+  expires_in: number
+}
+
+export default async function getAccessToken(request: RequestClient, accessKeyId: string, apiSecret: string): Promise<string> {
+  const res = await request<AccessTokenResponse>(`${M3TER_AUTH_API}/oauth/token`, {
+    method: 'POST',
+    body: '{"grant_type":"client_credentials"}',
+    headers: {
+      authorization: `Basic ${Buffer.from(`${accessKeyId}:${apiSecret}`).toString('base64')}`,
+      'Content-Type': 'application/json'
+    }
+  })
+  return res.data.access_token
+}
--- a/packages/destination-actions/src/destinations/m3ter/index.ts
+++ b/packages/destination-actions/src/destinations/m3ter/index.ts
@@ -1,0 +1,67 @@
+import type { Settings } from './generated-types'
+
+import submitMeasurements from './submitMeasurements'
+import getAccessToken from './auth'
+import { USER_AGENT_HEADER } from './constants'
+import { defaultValues, DestinationDefinition } from '@segment/actions-core'
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'm3ter',
+  slug: 'actions-m3ter',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'oauth2',
+    fields: {
+      access_key_id: {
+        label: 'm3ter Access Key Id',
+        description: 'Your service user Access Key Id. You can generate the service user and its Access Key Id in your m3ter console under "Settings" -> "Access" -> "Service Users" -> "Create Service User".' +
+          ' Step by step guide can be found in [m3ter Docs](https://www.m3ter.com/docs/guides/authenticating-with-the-platform/service-authentication#generating-an-api-key-and-secret-for-a-service-user)',
+        type: 'string',
+        required: true
+      },
+      api_secret: {
+        label: 'm3ter Api Secret',
+        description: 'Your service user Api Secret. You can generate the service user and its Api Secret in your m3ter console under "Settings" -> "Access" -> "Service Users" -> "Create Service User".' +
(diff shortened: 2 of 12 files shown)


martinconolly:
Hi @varadarajan-tw

_Who are the users the destination intended for? Why?_

m3ter is a usage-based billing and analytics platform; we aim to make the initial integration and ongoing process of ingesting usage data as quick and easy as possible for our customers. This destination is therefore intended for current and future m3ter customers.


_Could you explain top 3 scenarios on how you expect a user to leverage the destination in Segment?_

Our customers may already have their usage events in Segment, in which case the scenario is that they would use the destination to send those events to m3ter. Alternatively, they may use Segment for other purposes, and the minimal effort to integrate with m3ter via the destination would encourage them to expand their use of Segment to include those usage events.


@wkalucki and I are happy to hop on a call to discuss further if it would be useful.

packages/destination-actions/src/destinations/m3ter/submitMeasurements/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      type: 'object',
+      description: 'Numeric cost values for measurements associated with costs',
+      required: false
+    },
+    income: {
+      label: 'Income',
+      type: 'object',
+      description: 'Numeric income values for measurements associated with income',
+      required: false
+    }
```
tcgilbert, line 88:
```suggestion
    }, 
enable_batching: {
      type: 'boolean',
      label: 'Batch Data to M3ter',
      description:
        'When enabled, the action will use the M3ter batch API. Which we recommend for...',
      required: true,
      default: true
  },
```

wkalucki: Thanks, @tcgilbert added the field

packages/destination-actions/src/destinations/m3ter/auth.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface AccessTokenResponse {
+  token_type: string
+  access_token: string
+  expires_in: number
+}
+
+export default async function getAccessToken(request: RequestClient, accessKeyId: string, apiSecret: string): Promise<string> {
+  const res = await request<AccessTokenResponse>(`${M3TER_AUTH_API}/oauth/token`, {
+    method: 'POST',
+    body: '{"grant_type":"client_credentials"}',
```
rhall-twilio, line 13: @joe-ayoub-segment we don't have full support for client_credentials flow for partner builds. I've requested it to be added to the Q4 pipeline, but we will need to hardcode this destination slug to hide the OAuth login button before it goes GA (if before Q4)

rhall-twilio (approved):

packages/destination-actions/src/destinations/m3ter/auth.ts
wkalucki: Hi @rhall-twilio @joe-ayoub-segment 
What does it mean for us?

If it's any help, there are other destinations that use it

https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/1plusx-asset-api/index.ts#L44

https://github.com/segmentio/action-destinations/blob/dee4caf80bb5d1c298a3343fb5fef16578a1dd32/packages/destination-actions/src/destinations/criteo-audiences/criteo-audiences.ts#L55

https://github.com/segmentio/action-destinations/blob/dee4caf80bb5d1c298a3343fb5fef16578a1dd32/packages/destination-actions/src/destinations/salesforce-marketing-cloud/index.ts#L57

packages/destination-actions/src/destinations/m3ter/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'oauth2',
+    fields: {
+      access_key_id: {
+        label: 'm3ter Access Key Id',
+        description: 'Your service user Access Key Id. You can generate the service user and its Access Key Id in your m3ter console under "Settings" -> "Access" -> "Service Users" -> "Create Service User".' +
+          ' Step by step guide can be found in [m3ter Docs](https://www.m3ter.com/docs/guides/authenticating-with-the-platform/service-authentication#generating-an-api-key-and-secret-for-a-service-user)',
+        type: 'password',
```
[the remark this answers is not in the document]
wkalucki: made it regular string

wkalucki:
Hi @rhall-twilio , thanks for your review and approval. What are the next steps to merge this and get it deployed to private beta?

packages/destination-actions/src/destinations/m3ter/auth.ts
rhall-twilio: it means that Segment needs to do a one-line code in internal repositories so that we don't show a Login button for the user in the settings of this destination.

rhall-twilio (approved):

PR #1384, opened 2023-07-06
add configurable default request timeout

peterdemartini:
In order to support bulk batches, we need to be able to extend the default timeout for the requests. This is more of a proposal for the change and I am open to changes/suggestions.

## Testing

Tested as part of: https://github.com/segmentio/integrations/pull/2674

### DIFF
--- a/packages/core/src/create-request-client.ts
+++ b/packages/core/src/create-request-client.ts
@@ -1,7 +1,7 @@
 import addBasicAuthHeader from './middleware/before-request/add-basic-auth-header'
 import prepareHeaders from './middleware/after-response/prepare-headers'
 import prepareResponse from './middleware/after-response/prepare-response'
-import createInstance, { AllRequestOptions, RequestOptions } from './request-client'
+import createInstance, { AllRequestOptions, RequestOptions, DEFAULT_REQUEST_TIMEOUT } from './request-client'
 import type { ModifiedResponse } from './types'
 
 export interface ResponseError extends Error {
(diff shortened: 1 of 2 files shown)


packages/core/src/create-request-client.ts
```diff
@@ -1,15 +1,15 @@
...
+import createInstance, { AllRequestOptions, RequestOptions, DEFAULT_REQUEST_TIMEOUT } from './request-client'
 import type { ModifiedResponse } from './types'
 
 export interface ResponseError extends Error {
   status?: number
 }
 
 const baseClient = createInstance({
-  timeout: 10000,
+  timeout: DEFAULT_REQUEST_TIMEOUT,
```
SyedWasiHaider, line 12: Curious, how does this timeout works with existing 4s timeout we have in integrations/centrifuge? 

Guess I'm wondering if setting this value to anything more than 4s has any affect.

peterdemartini: We can, and do, increase it beyond 4s in Centrifuge/Integrations Consumer for certain destinations/customers. To be correct though, this _should_ be dynamic based on time remaining from the deadline time passed from Centrifuge. With that said, there are multiple layers of protections across the different services. Having a timeout larger here may not be a big deal, it could lead to potentially duplicates (even though that may be unavoidable)

SyedWasiHaider (approved):

dlasky:
would we ever consider passing this as part of the event spec? it seems like that would match up with the flow control config a bit better than global tuning?

peterdemartini:
> would we ever consider passing this as part of the event spec? it seems like that would match up with the flow control config a bit better than global tuning?

If this could use the `deadline` from Integration Spec, then this wouldn't be as much as an issue. However, it is not necessarily bad to have multiple service layers of protection that operate independent of each other.

PR #1387, opened 2023-07-07
[TikTok Audiences] fix normalization of emails

maryamsharif:
Upon investigation we found that our email normalization is removing all of the periods in an email instead of just the ones before the @ symbol. For example we would take we would take this `local-part.123+suffix @domain.com` and normalize it to `local-part123@domaincom` but it should be `local-part123@domain.com`

## Testing


rvadera12 (approved):
Looks Great! Can we add a test for the email hashing.

PR #1394, opened 2023-07-12
Updating GA4 docs to make clear what values can be used as Data Steam Type parameter.

seg-leonelsanches:
The docs are not very clear about Data Steam Type parameter regarding:

- What is the default value ("Web");
- What are the possible values.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
+++ b/packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
@@ -356,6 +356,6 @@ export const data_stream_type: InputField = {
   type: 'string',
   choices: [DataStreamType.Web, DataStreamType.MobileApp],
   description:
-    'The type of data stream this data belongs in. This can either be a web stream or a mobile app stream (iOS or Android).',
+    'The type of data stream this data belongs in. This can either be a web stream or a mobile app stream (iOS or Android). Possible values: "Web" (default) and "Mobile App".',
   default: DataStreamType.Web
 }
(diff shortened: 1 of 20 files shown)


packages/destination-actions/src/destinations/google-analytics-4/ga4-properties.ts
```diff
@@ -356,6 +356,6 @@ export const data_stream_type: InputField = {
   type: 'string',
   choices: [DataStreamType.Web, DataStreamType.MobileApp],
   description:
-    'The type of data stream this data belongs in. This can either be a web stream or a mobile app stream (iOS or Android).',
+    'The type of data stream this data belongs in. This can either be a web stream or a mobile app stream (iOS or Android). Possible values: "Web" (default) and "Mobile App".',
   default: DataStreamType.Web
```
marinhero, line 360: Hey @seg-leonelsanches. The default value will be visible when a mapping is about to be created. Do you think that's not illustrative enough? Also, would we update the destination docs instead of the destination definition?

seg-leonelsanches: @marinhero It is not. The ideal would be a validation between this field and whether the instrumentation id is provided, or whether the firebase id is provided, but for now this validation message is good enough. 

Another motivation is that clients don't know what values to add. I have to read the code to find out.

seg-leonelsanches: @marinhero Regarding your second question, this section is used by the official docs. there's an include that points to this definition in the other repo.

marinhero (approved):

PR #1399, opened 2023-07-14
Add Attio destination action

nickcampbell18:
Hi team 👋 

We'd like to contribute a new destination for [Attio](https://attio.com/). Attio is a flexible CRM platform, and we'd like to allow our customers to be able to listen on Segment events to send data into Attio.

This destination offers 3 actions:

  1. `identifyUser`, which we recommend runs on an `identify` event and creates a Person + User using the Attio API
  2. `groupWorkspace`, which we recommend runs on a `group` event and creates a Company + Workspace using the Attio API
  3. `assertRecord`, which is a generic version of both of these that can be customised to create arbitrary types of record and run on any event requested

At the moment, we believe this solves the main problems our customers have been asking for, but hopefully we can add more in future if demand changes. Really the only main API call this makes is an "assert", which behaves like an upsert and either creates or updates a record based on matching attributes.

Authentication-wise, we'd like Segment to please integrate our OAuth 2.0 flow to allow customers to easily setup from inside their dashboard. Our docs are here: https://attio.com/developers/authentication#generating-access-tokens, I can provide access tokens privately, you can reach me at nick.campbell@attio.com

## Testing

There's unit & mocked HTTP tests for each action, but we've also been testing this locally using the tester against our production API so we're fairly confident it works. Segmenters - I can happily grant you access tokens for an Attio demo environment if you'd like to test there also.


packages/destination-actions/src/destinations/attio/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  slug: 'actions-attio',
+  mode: 'cloud',
+
+  authentication: {
+    scheme: 'oauth-managed',
+    fields: {},
+    testAuthentication: (request) => request('https://api.attio.com/v1/token')
+  },
+
+  extendRequest({ auth }) {
```
rhall-twilio, line 19: does Attio support the refresh token flow? If not, for how long are the access tokens good for?

nickcampbell18: Nope, we don't use refresh tokens, and access tokens have no expiry.

sayan-das-in (approved):

sayan-das-in:
Hi @nickcampbell18, I have approved the PR. In the mean time I wanted to ask you a couple of things for record keeping since this is a new destination.

1. Who are the customers this destination intended for and why?
2. Could you give examples of use cases for the destination?

nickcampbell18:
Attio is a CRM product so the main use-case our customers have is for passing data into Attio when their customers engage with their product. Initially, the two big uses we anticipate are:

1. Making a copy of relevant user data when somebody signs into their product (`identify`).
2. Linking those users to company records after a `group` event

A typical Attio customer is a technology-savvy startup who is looking to use a combination of SaaS products to keep track of their data effectively. They likely already use Segment and are familiar with the process of sending data to and from it, so a Segment integration will help them funnel their data into Attio so they can do further reporting, analysis and support for their customers. The reason we're adding this now is that a number of customers have been asking for this functionality.

An example use-case:

* I use Segment to track users on my app
* I use Attio to keep track of all the times I've contacted those users (e.g. billing, support) 
* With this integration, new users in Segment will get added to my Attio workspace so I can support them as well

You can find more information about Attio on our homepage: https://attio.com/

sayan-das-in:
Thank you for the quick response.

I noticed that you have not returned the response of the `request` calls in `perform` method of the supported actions (`identifyUser`, `groupWorkspace`, `assertRecord`). This can potentially lead to issues in our event tester where user may not be able to see the exact cause of the error.

Could you please raise another PR to quickly address this issue?
[Here's an example](https://github.com/segmentio/action-destinations/blob/dc657cc8c827282d97bfc047230b0528d5c5c047/packages/destination-actions/src/destinations/braze/trackEvent/index.ts#L86) on how destinations usually do this.

nickcampbell18:
Sure, opened https://github.com/segmentio/action-destinations/pull/1428

PR #1400, opened 2023-07-14
[SDK-3485] braze subscription groups

davidbielik:
Adds a new Action to Set Subscription Groups.

Questions for @pooyaj 
- [x] is this the right approach from Segment perspective? yes
- [x] should the be a `track` property instead, or is the implementation generic? doesn't matter, it's generic
- [x] how should the tests be written
- [ ] how to fix the existing typescript errors
- [x] how to update snapshot tests? `-u` flag in the test command

The expected use is a customer calling:
```
analytics.identify("id", {
  braze_subscription_groups: [{
          "subscription_group_id": "5ertykiuyfjyttgkf",
          "subscription_group_state": "unsubscribed"
      }, {
          "subscription_group_id": "ytghkuguymjghb",
          "subscription_group_state": "unsubscribed"
    }]
})
```

We would loop through each group and call the `addToSubscriptionGroup`/`removeFromSubscriptionGroup` methods.

## Testing

image

image


davidbielik:
@spvismaya @wesleyorbin FYI

packages/browser-destinations/destinations/braze/src/__tests__/updateUserProfile.test.ts
```diff
@@ -142,4 +144,52 @@ describe('updateUserProfile', () => {
...
+          braze_subscription_groups: braze_subscription_groups,
+          custom_attributes: { foo: 'bar' }
+        }
+      })
+    )
+
+    expect(destination.actions.updateUserProfile.perform).toHaveBeenCalledWith(
+      expect.objectContaining({
+        instance: expect.objectContaining({
+          changeUser: expect.any(Function)
```
davidbielik, line 183: @pooyaj question for your team - this doesn't actually test that the method was called, i think it just checks that the Braze SDK loaded has that method in its JSON. 

Is there a way to check that a given braze sdk method was called? I'm not sure how valuable this test assertion is.

joe-ayoub-segment:
hi @pooyaj - I'm just back from leave and am catching up on everything. Will you be completing this PR review?

davidbielik:
Tested it via the actions tester and looks good on my end!

joe-ayoub-segment:
Hi @pooyaj , 
I just had a call with @davidbielik where we discussed options for testing this Integration. 
Would it be possible to deploy this so that it's only available for David's Workspace?
David's Workspace slug is braze-sdks 

Thanks,
Joe

packages/browser-destinations/destinations/braze/src/updateUserProfile/index.ts
```diff
@@ -154,7 +180,7 @@ const action: BrowserActionDefinition<Settings, BrazeDestinationClient, Payload>
     }
 
     // TODO - addAlias / addToCustomAttributeArray?
-    if (payload.external_id !== undefined) {
+    if (payload.external_id && typeof payload.external_id === 'string') {
```
joe-ayoub-segment, line 183: This change isn't needed.

packages/browser-destinations/destinations/braze/src/updateUserProfile/index.ts
```diff
@@ -180,8 +206,8 @@ const action: BrowserActionDefinition<Settings, BrazeDestinationClient, Payload>
     }
 
     // Adding `firstName` and `lastName` here as these fields are mapped using cammel_case.
-    const reservedFields = [...Object.keys(action.fields), 'firstName', 'lastName']
-    if (payload.custom_attributes !== undefined) {
+    const reservedFields = [...Object.keys(action.fields), 'firstName', 'lastName', 'braze_subscription_groups']
+    if (payload.custom_attributes && typeof payload.custom_attributes === 'object') {
```
joe-ayoub-segment, line 210: The change on this line isn't necessary

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
Hi @pooyaj , 
I just had a call with @davidbielik where we discussed options for testing this Integration. 
Would it be possible to deploy this so that it's only available for David's Workspace?
David's Workspace slug is braze-sdks 

Also @davidbielik , can you please attach some testing screenshots? Other than that the PR looks good to me! 

Thanks,
Joe

davidbielik:
Braze /v3/data payload shows correct custom attributes (excluding sub groups) and correct "events" to add subscription groups (excluding the "bad" one)
image

image

joe-ayoub-segment (approved):
Looks good, thanks for adding the test payloads.

joe-ayoub-segment:
Hi @pooyaj can this go out with the next web deploy?

packages/browser-destinations/destinations/braze/src/__tests__/updateUserProfile.test.ts
pooyaj: @davidbielik That's right, the test is not valuable because it just tests the perform method, rather than what the braze SDK is being called with. Not a blocker, but we can improve the tests using examples from other destinations: https://github.com/segmentio/action-destinations/blob/main/packages/browser-destinations/destinations/hubspot-web/src/trackPageView/__tests__/index.test.ts

davidbielik:
Hi @joe-ayoub-segment @pooyaj any update on this one?

joe-ayoub-segment:
Hi @davidbielik this is ready to deploy. We're in a deploy freeze at the moment. 
@pooyaj , when can we schedule a Web deploy? Would we do it Monday 28th?

PR #1420, opened 2023-07-18
[TTD] Batch Settings

maryamsharif:
This PR addresses [STRATCONN-2944](https://segment.atlassian.net/browse/STRATCONN-2944) to add an error if min batch size in not met and to add a `batch_size` mapping to configure a max batch size.

## Testing

[STRATCONN-2944]: https://segment.atlassian.net/browse/STRATCONN-2944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/destination-actions/src/destinations/the-trade-desk-crm/functions.ts
```diff
@@ -35,6 +35,12 @@ export interface CREATE_API_RESPONSE {
 }
 
 export async function processPayload(request: RequestClient, settings: Settings, payloads: Payload[]) {
+  const TTD_MIN_RECORD_COUNT = 1500
+  if (payloads.length < TTD_MIN_RECORD_COUNT) {
+    throw new PayloadValidationError(
```
varadarajan-tw, line 40: For my understanding: How are the retries handled for batches? Because, if we error out with 400 here, wouldn't the entire batch be discarded? Or does bulk-batching pipeline reconstruct a batch even for 400s?

rvadera12: The entire batch will be discarded, however Trade Desk is supporting daily full audience syncs. For example, if we receive a payload of under the TTD_MIN_RECROD_COUNT then the event will be retried and discarded. However the events in reality will get retried the next day since we have full audience syncs, and if it is above the TTD_MIN_RECORD_COUNT then we will send it to Trade Desk. For further context, Trade Desk only supports uploads above 1500 users so even if we sent this downstream, Trade Desk will fail to upload any records.

PR #1424, opened 2023-07-20
Update Wisepops script URL

clementai:
We changed our script URL, the old one has a redirection but it will be better to skip it.

## Testing

It should work has before, this is just a URL change.

### DIFF
--- a/packages/browser-destinations/destinations/wisepops/src/index.ts
+++ b/packages/browser-destinations/destinations/wisepops/src/index.ts
@@ -63,7 +63,7 @@ export const destination: BrowserDestinationDefinition<Settings, Wisepops> = {
   settings: {
     websiteId: {
       description:
-        "The identifier of your Wisepops' website. You can find it in [your setup code on Wisepops](https://app.wisepops.com/f/settings/websites).",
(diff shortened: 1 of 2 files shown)


packages/browser-destinations/destinations/wisepops/src/index.ts
```diff
@@ -63,24 +63,23 @@ export const destination: BrowserDestinationDefinition<Settings, Wisepops> = {
...
   },
 
   initialize: async ({ settings }, deps) => {
-    window.WisePopsObject = 'wisepops'
     window.wisepops =
       window.wisepops ||
-      function (...arg) {
-        ;(window.wisepops.q = window.wisepops.q || []).push(arg)
+      function () {
+        (window.wisepops.q = window.wisepops.q || []).push(arguments)
```
pooyaj, line 77: Looks like this change is causing linting error: 
```
Error:   77:60  error    Use the rest parameters instead of 'arguments'  prefer-rest-params
```

pooyaj (changes requested):
- Please fix the linting issue 
- Also seeing CI tests failing (https://github.com/segmentio/action-destinations/actions/runs/5612748874/job/15208537397?pr=1424), but some of the failures are strange. Will re-run after linting issues are fixed

pooyaj (approved):

PR #1428, opened 2023-07-21
Add return values for API calls in Attio actions

nickcampbell18:
This is a follow-up to https://github.com/segmentio/action-destinations/pull/1399#issuecomment-1645403215

Note that the `groupWorkspace` and `identifyUser` commands make _two_ API calls. The first one should basically always succeed, the second is more likely to fail depending on object configuration on the Attio side, so I took the decision to return that second result.

@sayan-das-in is it possible to return multiple values like `[person, user]` or `[company, workspace]`? If so, I can swap.


sayan-das-in:
@nickcampbell18 thank you for making the changes. This looks good!

sayan-das-in (approved):

PR #1431, opened 2023-07-21
Fullstory V2 API Actions

mattgrosvenor-fs:
### Adding two new Actions to allow users to migrate from our V1 to V2 API
 - Track Events V2
 - Identify User V2
 
 New actions are being created over changing existing actions to allow users to make the change when they want instead of being automatically switched.
 
### Updated `onDelete` to use our new V2 API

This isn't really a callable action so little harm in automatically switching this call to V2.

## Testing

Created unit tests for new functions/actions being added.
Ran everything through the local server and verified it came through correctly.

### DIFF
--- a/packages/destination-actions/src/destinations/fullstory/__tests__/fullstory.test.ts
+++ b/packages/destination-actions/src/destinations/fullstory/__tests__/fullstory.test.ts
@@ -209,4 +209,93 @@ describe('FullStory', () => {
       })
     })
   })
+
+  describe('trackEventV2', () => {
+    it('makes expected request with default mappings', async () => {
+      nock(baseUrl).post(`/v2beta/events?${integrationSourceQueryParam}`).reply(200)
+      const eventName = 'test-event'
+
+      const sessionId = '12345:678'
+
+      const properties = {
+        'first-property': 'first-value',
+        second_property: 'second_value',
+        thirdProperty: 'thirdValue',
+        useRecentSession: true,
+        sessionUrl: `session/url/${encodeURIComponent(sessionId)}`
+      }
+
+      const timestamp = new Date(Date.UTC(2022, 1, 2, 3, 4, 5)).toISOString()
+
+      const event = createTestEvent({
+        type: 'track',
+        userId,
+        event: eventName,
+        timestamp,
+        properties
+      })
+
+      const [response] = await testDestination.testAction('trackEventV2', {
+        settings,
+        event,
+        // Default mappings defined under fields in ../trackEventV2/index.ts
+        useDefaultMappings: true,
+        mapping: {
+          useRecentSession: {
+            '@path': '$.properties.useRecentSession'
+          },
+          sessionUrl: {
+            '@path': '$.properties.sessionUrl'
+          }
+        }
+      })
+
+      expect(response.status).toBe(200)
+      expect(JSON.parse(response.options.body as string)).toEqual({
+        name: eventName,
+        properties: {
+          firstproperty: 'first-value',
+          second_property: 'second_value',
+          thirdProperty: 'thirdValue',
+          useRecentSession: true,
+          sessionUrl: `session/url/${encodeURIComponent(sessionId)}`
+        },
+        user: {
+          uid: userId
+        },
+        timestamp,
+        session: {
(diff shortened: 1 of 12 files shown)


marinhero:
Hi Matt! this looks good to me, just a quick question. Is V2 ready on your end? because deploying this will start to render this to our mutual customers.

joshrynneFS:
Hi @marinhero - jumping in while Matt is on vacation. Yes, our v2 APIs are ready and have been in-use by customers for a while now. We should be okay to deploy this change especially since the v2 actions are not enabled by default.

marinhero (approved):

marinhero:
All right, marking this as ready for release. The next release train will launch on Tuesday August 2nd. Thank you for your help and patience. Let me know if you have any questions. Best.

joe-ayoub-segment:
Hi @joshrynneFS and @mattgrosvenor-fs , 

I'm just back from 4 weeks leave. My colleague @marinhero was kindly helping out with some PR reviews while I was out. 
As you can imagine I've a few things to catch up on, including this PR.

I'm going to ask my team if there's a standard way for us to deploy the code in such a way so that the new Actions are hidden from customers until we've had the opportunity to test them properly. 

Because of this I'm not sure that this PR will be deployed tomorrow. I'd also like to have the opportunity to review the code myself (unless @marinhero tells me that he's 100% happy with it). 

Does the above sound OK to you @joshrynneFS and @mattgrosvenor-fs ?

Kind regards,
Joe

joshrynneFS:
Welcome back @joe-ayoub-segment! When you say, "in such a way so that the new Actions are hidden from customers until we've had the opportunity to test them properly" does this mean more than just making them not installed by default? We were envisioning that customers would still be able to find them via the "New Mapping" button in our Cloud Mode destination.

It's okay if the changes don't make it into this week's deploy so that you perform a review. Do you think we could aim for next week's deploy?

Thanks,
Josh

joe-ayoub-segment:
Hi Josh, 

By "in such a way so that the new Actions are hidden from customers until we've had the opportunity to test them properly" I meant that it would be desirable to be able to make these 2 new Actions available to just you and I in Production, and maybe one or two select customers. We could then ensure (via testing) that the new Actions work exactly as intended before other customers start to use them. Once customers start using an Action it makes it more complicated to change the code in those Actions. Does that make sense?

Cheers,
Joe

joshrynneFS:
That does make sense, thanks. I didn't realize that was an option, but if there is a way to do this, we would love the opportunity to test it in a more controlled way.

joshrynneFS:
Hi Joe,

The second option seems okay with us. What kind of preparation would you need from us (sample data, environment setup, etc.) before the screenshare?

-Josh

joe-ayoub-segment:
I'd need to deploy your code to Staging, then schedule a call with you. 

Do you have credentials which we can use when sending data from Segment to Fullstory (for the new Action)?
We can hand craft some events for testing purposes, so as long as you are clear on the type of events customers will forward to the new Actions then we should be good.

joshrynneFS:
Great. You'll just need an API key for one of our sandboxes which we can provide on the call.

joe-ayoub-segment:
Deployed to Staging and did some testing with Nate and Josh. 

V1 tests
image
image

V2 tests
image
image

Josh and Nate confirmed the updates were made on Fullstory.

joe-ayoub-segment (approved):
Did testing in Staging for both legacy and new V2 Actions.

joe-ayoub-segment:
Hi @mattgrosvenor-fs , 
The flakey test has been fixed, however you'll need to get the latest code from main before the CI tests will pass for this PR. 
Are you OK doing this please?

Best regards,
Joe

mattgrosvenor-fs:
Hi @joe-ayoub-segment ,
I updated our fork's main with latest segment main and rebased this branch on that. Hopefully tests pass now.

Best,
Matt

joe-ayoub-segment:
Hi @mattgrosvenor-fs , this PR has been deployed. Please check and let me know if everything looks good or not!

mattgrosvenor-fs:
@joe-ayoub-segment , thank you! Every looks good.

PR #1440, opened 2023-07-25
Add support for self hosted Adobe Library

marinhero:
This PR will allow users to bring their own version of AT.JS. This branch is not supposed to be merged against `main` since it will be released as a pinned version to a given source.

## Testing

Testing completed successfully via replit. Ask via DM for it if you are interested.

### DIFF
--- a/packages/browser-destinations/destinations/adobe-target/src/upsertProfile/index.ts
+++ b/packages/browser-destinations/destinations/adobe-target/src/upsertProfile/index.ts
@@ -33,7 +33,7 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
   },
   perform: (Adobe, event) => {
     if (!Object.prototype.hasOwnProperty.call(window, 'adobe')) {
-      return Promise.reject()
+      return Promise.resolve()
     }
 
     /*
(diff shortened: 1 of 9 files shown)


packages/browser-destinations/destinations/adobe-target/src/upsertProfile/index.ts
```diff
@@ -32,6 +32,10 @@ const action: BrowserActionDefinition<Settings, Adobe, Payload> = {
     }
   },
   perform: (Adobe, event) => {
+    if (!Object.prototype.hasOwnProperty.call(window, 'adobe')) {
+      return Promise.reject()
```
pooyaj, line 36: This might not be an issue, but I'm not sure what will be the downstream effect of this in AJS, may worth testing this path to see if it causes exceptions in the console or retries. If it does cause exceptions, may be better to just returning undefined or doing a `Promise.resolve()`.

pooyaj (approved):

packages/browser-destinations/destinations/adobe-target/src/upsertProfile/index.ts
marinhero: How can you check for retries? I had a `console.log` here before and verified that it only went inside this once. But if there's no issue with having `Promise.resolve()` I'll do just that.

PR #1441, opened 2023-07-25
[external] add new launchdarkly-audiences destination

k3llymariee:
This PR adds a new destination - `launchdarkly-audiences` - in order to allow LaunchDarkly customers the ability to send attribution events to the LaunchDarkly API. 

- We are using a `CUSTOM` authentication using `apiKey` - customers will need to provide a valid service token provided in the LaunchDarkly UI. 
- We have only one action - `syncAudience` - which will subscribe to both `track` and `identify` segment events. This action makes a POST request to the LaunchDarkly API sync segment endpoint to add or remove emails from a LaunchDarkly segment list. 
- We expect the following fields to be included in the payload:
    - `context.personas.computation_key` which will be used as the name and ID of the LaunchDarkly segment
    - `context.personas.computation_class` which will be needed to determine proper handling of the audience
    - `.context.traits.email` OR `.traits.email` - user identifier
    - `.properties` OR `.traits`
- We support both single and batched events

## Testing


packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  description: 'Sync Engage Audiences to LaunchDarkly segments',
+  defaultSubscription: 'type = "identify" or type = "track"',
+  fields: {
+    custom_audience_name: {
+      label: 'Audience Key',
+      description: 'Segment Audience key to which user identifier should be added or removed',
+      type: 'hidden',
+      required: true,
+      default: {
+        '@path': '$.context.personas.computation_key'
```
marinhero, line 17: Hi! Marín here from the integrations team. I'm going to be reviewing your PR. To start, I'd like to know where did you learn about the `personas` values that you are using here? also, does the `computation_key` matches anything in your DB? For example, the ID of an audience on that end.

ldhenry: Hi @marinhero,

We learned about the `persons` values from an initial sync call with @joe-ayoub-segment. The intention is for this destination to be used for the sole purpose of syncing Segment Audiences with [LaunchDarkly segments](https://docs.launchdarkly.com/home/contexts/segments). Joe mentioned that `context.persona.computation_key` will be the snake-cased version of the Audience name. We are using this (along with a unique prefix) as the Segment key on our side.

packages/destination-actions/src/destinations/launchdarkly-audiences/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      // The sdk/goals/{clientID} endpoint returns a 200 if the client ID is valid and a 404 otherwise.
+      return Promise.all([
+        request(`${CONSTANTS.LD_CLIENT_SDK_BASE_URL}/sdk/goals/${settings.clientId}`, { method: 'head' }),
+        request(`${CONSTANTS.LD_API_BASE_URL}/versions`, {
+          method: 'GET',
+          headers: {
+            Authorization: `${settings.apiKey}`
+          }
+        })
+      ])
```
joe-ayoub-segment, line 43: hi @varadarajan-tw , I think returning a Promise.all will work, but would value another opinion.

packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/index.ts
```diff
@@ -0,0 +1,74 @@
...
+        '@path': '$.context.personas.computation_class'
+      }
+    },
+    context_kind: {
+      label: 'Context kind',
+      description:
+        "The event's context kind. To learn more about context kinds and where you can find a list of context kinds LaunchDarkly has observed, read [Context kinds](https://docs.launchdarkly.com/home/contexts/context-kinds).",
+      type: 'string',
+      required: true,
+      default: 'user'
```
joe-ayoub-segment, line 35: What other options should be available here? 
If there are a limited number then we turn this into an Enum, like this: 

      choices: [
        { label: 'User', value: 'user' },
        { label: 'Account', value: 'account' }
      ], 
      default: 'user'
      
 in the example above, field validation will fail if the field doesn't contain either 'user' or 'account'.

packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/index.ts
```diff
@@ -0,0 +1,74 @@
...
+      description: 'The unique LaunchDarkly context key. In most cases the Segment `userId` should be used.',
+      type: 'string',
+      required: true,
+      default: {
+        '@if': {
+          exists: { '@path': '$.userId' },
+          then: { '@path': '$.userId' },
+          else: { '@path': '$.email' }
+        }
+      }
```
joe-ayoub-segment, line 48: The email address will never be at the root level of a Segment payload. I know in the testing tool it is shown at the root, but that's a bug that needs to be fixed!! 

If email is included in an identify() call it will likely be located at traits.email. 
If email is included in a track() call it will likely be located either at properties.email or context.traits.email. 

However, why not fall back to the anonymousId instead? A Segment payload will always contain either a userId or an anonymousId. It is never guaranteed to contain an email address. 

So we could this this: 

      default: {
        '@if': {
          exists: { '@path': '$.userId' },
          then: { '@path': '$.userId' },
          else: { '@path': '$.anonymousId' }
        }
      }

packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/custom-audience-operations.ts
```diff
@@ -0,0 +1,119 @@
...
+  for (const batch of audienceBatches) {
+    if (batch.batch.length === 0) {
+      continue
+    }
+    const ldRequest = request(CONSTANTS.LD_API_BASE_URL + CONSTANTS.LD_API_CUSTOM_AUDIENCE_ENDPOINT, {
+      method: 'POST',
+      json: batch
+    })
+    promises.push(ldRequest)
+  }
```
joe-ayoub-segment, line 114: hi @k3llymariee , 

Looks like you are preparing a different Request/fetch for each individual Audience. 
Segment's maximum batch size is 1000 events, so in theory if the customer had created 1000 different Audiences then the batch could contain 1000 events each for a different Audience. This would mean that your Promise.all would need to fire 1000 requests, which would probably time out. 

In practice this is unlikely to happen, but it is a technical limitation of this approach.  

The only options we have available are: 
1. Disable batching, so that each request is sent individually
2. Keep batching enabled, but if a customer notices an issue we can ask them to disable batching for their individual Audience (which they can do via the UI). 

Keen to hear your thoughts.

joe-ayoub-segment (changes requested):
Hi @k3llymariee , cc @varadarajan-tw  and @pooyaj , 

Thanks again for submitting this PR. I think it looks really good! 

I left a few questions inline for you. I've cc'd my colleague Varada to ask his opinion on one of the questions (re: testAuthentication). 

My colleague Pooya (cc'd) has built out a capability which makes it easier to create Audience Destinations. I'm wondering if this new LaunchDarkly Audiences Destination could be a good candidate to leverage this new capability?  I'm going to discuss this with Pooya offline and will get back to you. Either way we won't require you to make any code changes - Pooya and I would apply any chances using the same logic you provided in this PR. 

Best regards,
Joe

packages/destination-actions/src/destinations/launchdarkly-audiences/index.ts
varadarajan-tw: Yeah, this will work.

packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/index.ts
ldhenry: Our customers can create arbitrary context kinds, so unfortunately we can't provide a list of options.

packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/index.ts
ldhenry: That sounds like a good idea to me.

packages/destination-actions/src/destinations/launchdarkly-audiences/syncAudience/custom-audience-operations.ts
ldhenry: If you look up in `parseCustomAudienceBatches` above, we are actually just creating separate batches for each unique Audience key that we receive in the events payload array. My assumption is that the Audience key is always the same for every event in the payload array so we we would just have one batch under normal conditions. Is this assumption wrong?

joe-ayoub-segment: Hi @ldhenry , 
The array of events which are sent into the `perform` function could contain events from multiple Audiences. 
It's likely to be the case a large proportion of the time. 

Is your endpoint able to accept a single batch API call containing updates for multiple Audiences? If so, it might be best to do it that way?

ldhenry: Sorry our endpoint can't handle a batch containing multiple audiences today.

joe-ayoub-segment:
hi @ldhenry , we can go ahead with it, and monitor. If batching causes any issues we can disable it. OK with you?

ldhenry:
> hi @ldhenry , we can go ahead with it, and monitor. If batching causes any issues we can disable it. OK with you?

That sounds good to me. We are just planning to use it internal for the time being so we can keep a close eye on it.

joe-ayoub-segment:
hi @k3llymariee , Mike and @ldhenry , 

I spent some time figuring out if there would be value in refactoring the code so that it works with the new 'Native Audiences' capability, and in the end I didn't believe that there would be sufficient benefit. In fact there would be a couple of added drawbacks. For now I'd recommend that we keep the code you've written. 

However I did add a couple of changes. I tried to submit a PR to your repo but didn't have the permissions, so instead you can download the full code [here](https://drive.google.com/drive/folders/1uyv5AVFkPS5KBrlSwvEdcBw-e7BQH1gl?usp=drive_link).  

I've shared the code folder with Mike and Henry but I didn't have your email address @k3llymariee . 

The changes I've added make the Destination a little more flexible and correct an issue with the mapping for the context_key field. 

The main change is that the customer now has more options for how to map a value to the context_key. 

Please pull down the latest code from the main Segment action-destinations branch before you add the folder. 

I recorded a [short video](https://drive.google.com/file/d/1_ZsZcktVCWPN9TDiHKZfIhdHkyiM3dod/view?usp=sharing) explaining the changes.

Best regards,
Joe

ldhenry:
@joe-ayoub-segment,

I just pushed up the changes you emailed and they work great! Thanks for helping - I think it's an improvement overall.

I tested it out locally and can confirm that a Segment was created in LaunchDarkly.

CleanShot 2023-08-16 at 3 14 45

joe-ayoub-segment (approved):

PR #1462, opened 2023-08-02
Extra context fields requested by client in Mixpanel

seg-leonelsanches:
Client requested optional mappings for context fields to be sent to Mixpanel (all supported events).

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
@@ -325,7 +325,11 @@ export const eventProperties: Record<string, InputField> = {
     description: 'The Event Original Name, if applicable',
     required: false,
     default: {
-      '@path': '$.name'
+      '@if': {
+        exists: { '@path': '$.event' },
+        then: { '@path': '$.event' },
+        else: { '@path': '$.name' }
+      }
     }
   },
   event_properties: {
--- a/packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
@@ -283,6 +283,51 @@ export const eventProperties: Record<string, InputField> = {
       '@path': '$.context.userAgent'
     }
   },
+  advertising_id: {
+    label: 'Advertising ID',
+    type: 'string',
+    description: 'Advertising ID',
+    required: false,
+    default: {
+      '@path': '$.context.device.advertisingId'
+    }
+  },
+  ad_tracking_enabled: {
+    label: 'Ad Tracking Enabled',
+    type: 'string',
+    description: 'Ad Tracking Enabled (true or false)',
+    required: false,
+    default: {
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
```diff
@@ -283,6 +283,51 @@ export const eventProperties: Record<string, InputField> = {
...
+      '@path': '$.context.timezone'
+    }
+  },
+  app_platform: {
+    label: 'App Platform',
+    type: 'string',
+    description: 'The App Platform, if applicable',
+    required: false,
+    default: {
+      '@path': '$.context.timezone'
```
shouri-thangellapalli, line 319: $.context.app.platform?

seg-leonelsanches: Fixed.

marinhero (commented):
This looks good to me. Letting @joe-ayoub-segment have the final say.

packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
```diff
@@ -283,6 +283,51 @@ export const eventProperties: Record<string, InputField> = {
...
+    description: 'The App Platform, if applicable',
+    required: false,
+    default: {
+      '@path': '$.context.app.platform'
+    }
+  },
+  name: {
+    label: 'Event Original Name',
+    type: 'string',
+    description: 'The Event Original Name, if applicable',
```
joe-ayoub-segment, line 325: Hi @seg-leonelsanches , 

screen() and page() events might contain a value at the `$.name` path. They will never contain any value at the `$.event` path. 

track() events never contain a value at the `$.name` path, but always contain a value at the `$.event` path. 

So I'm not sure if your description for this field is accurate, or if you are configuring it to collect the information you think will be at that path. 

Maybe the description just needs to read "The name of the page or screen, if applicable", or something like that?

Happy to discuss if you can explain a little more about what the customer wants to collect with this field.

joe-ayoub-segment (changes requested):
Hi @seg-leonelsanches , 
Just one question about a field description. 
Other than that it looks fine to me. 

Note that this change will affect all customers who configure a new instance of a Mixpanel Destination going forward. 

Customers with Mixpanel already configured as a Destination will need to open the Destination configuration, ensure that there are paths configured for the new fields, then save the Destination configuration. 

Regards.
Joe

packages/destination-actions/src/destinations/mixpanel/mixpanel-properties.ts
seg-leonelsanches: Hi @joe-ayoub-segment, 

In fact this is something that the client asked to send downstream. They want to pass two names: one for the mapped event, another one for the event original name. I believe this can be used also for page() and screen(), and I can improve the description.

joe-ayoub-segment:
Approved - will merge prior to deploy. scheduled for 15-August-23

PR #1466, opened 2023-08-03
[Twilio] Add support for self hosted AT.JS in Adobe Target - Web

marinhero:
The [first version](https://github.com/segmentio/action-destinations/pull/1440) of this helped us discover that due to a race condition the personalization component of AT breaks when at.js loads before Segment's a.js. Said first version, also helped us create an asset pipeline that now we can leverage to feed Twilio's own version of at.js into our destination.

This PR helps us do just that, point the dependency loader towards the static URL that the Twilio team has confirmed will never change and load at.js from there. That will mitigate any race condition and restore the usability of the personalization component.

This branch won't be merged to master. Instead it will be pinned to Twilio's Sources.

## Testing


nick-Ag (approved):
🚀

PR #1496, opened 2023-08-14
Added condition for empty string in mapping

hvardhan-unth:
When using FB Conversions API Actions, if we REMOVE the Email and Phone from the mappings, we pass the same hashed value to FB's side everytime. It seems it sends the “empty“ hashed value.  So, in this PR a condition has been added to check the empty string.
  
## Testing

SCREENSHOT:- 

Screenshot 2023-08-11 at 2 50 52 PM

Screenshot 2023-08-11 at 2 48 54 PM


packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/user-data.test.ts
```diff
@@ -1,4 +1,14 @@
...
+
+interface HashedData {
+  em: string
+  ph: string
+  db: string
+  ge: string
+  zp: string
+  ct: string
+  st: string
+}
```
varadarajan-tw, line 11: FWIW, it is not necessary to define type for every object in typescript. You can still access objects like in javascript. If you want to enforce type validation in more than one place, then it would make sense definining something like this. You can potentially use Record<string,string> here as well.

varadarajan-tw (approved):
LGTM! ✨

PR #1506, opened 2023-08-16
Conman 627

vikramkumar19:
This PR adds SendOptOut( Send to all statuses expect for Unsubscribed) functionality to engage destinations.

## Testing

Testing completed in Stage.

image

image


pmunin:
I do have some questions about this, but not sure if I'm able to finish review by EOD

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -96,6 +97,8 @@ export abstract class MessageSendPerformer<
 
   static readonly nonSendableStatuses = ['unsubscribed', 'did not subscribed', 'false'] // do we need that??
   static readonly sendableStatuses = ['subscribed', 'true']
+  static readonly nonSendableStatusesOptOut = ['unsubscribed', 'false'] // do we need that??
+  static readonly sendableStatusesOptOut = ['subscribed', 'true', 'did not subscribed', 'null']
```
pmunin, line 102: this does not make sense how `did not subscribe` is sendable status for optOut? Changing for optout model introduces confusion. I suggest to remove that and change the logic of methods that use it

```suggestion
```

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -119,6 +122,20 @@ export abstract class MessageSendPerformer<
...
+    const staticMems = this.getStaticMembersOfThisClass()
+    const subStatus = extId.subscriptionStatus?.toString()?.toLowerCase()
+    if (subStatus === null || subStatus === '') return true
+    if (!subStatus) return false // falsy status is valid and considered to be Not Subscribed, so return false
+    // if subStatus is not in any of the lists of valid statuses, then return true
+    if (staticMems.sendableStatusesOptOut.includes(subStatus)) return true
+    if (staticMems.nonSendableStatusesOptOut.includes(subStatus)) return false
+    return undefined //Invalid subscriptionStatus
+  }
+
```
pmunin, line 138: Instead of creating this separate method. I suggest to change the existing method in a very simple way:

```
  isExternalIdSubscribed(extId: ExtId<TPayload>): boolean | undefined {
    const staticMems = this.getStaticMembersOfThisClass()
    const subStatus = extId.subscriptionStatus?.toString()?.toLowerCase()

    + const isOptOutModel = this.payload.sendBasedOnOptOut

    + if(isOptOutModel) 
    + {
    +   if (!subStatus) return true //falsy status considered to be subscribed for opt-out model
    +   // allow sending messages to all subscription state apart from unsubscribed
    +   // so if subStatus is not in any of the lists of unsubscribed, return true, otherwise false
    +   return (!staticMems.nonSendableStatuses.includes(subStatus)) 
    + }
    if (!subStatus) return false // falsy status is valid and considered to be Not Subscribed, so return false
    // if subStatus is not in any of the lists of valid statuses, then return true
    if (staticMems.sendableStatuses.includes(subStatus)) return true
    if (staticMems.nonSendableStatuses.includes(subStatus)) return false
    return undefined //Invalid subscriptionStatus
  }

```

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -131,7 +148,12 @@ export abstract class MessageSendPerformer<
...
     const supportedExtIdsWithSub = this.payload.externalIds
       ?.filter((extId) => this.isSupportedExternalId(extId))
-      .map((extId) => ({ extId, isSubscribed: this.isExternalIdSubscribed(extId) }))
+      .map((extId) => ({
+        extId,
+        isSubscribed: this.payload.sendBasedOnOptOut
+          ? this.isExternalIdSubscribedOptOutModel(extId)
+          : this.isExternalIdSubscribed(extId)
+      }))
     if (!supportedExtIdsWithSub || !supportedExtIdsWithSub.length)
```
pmunin, line 157: according to suggested logic above I suggest to revert this change

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
```diff
@@ -37,8 +37,14 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload>
...
-    const isSubscribed = super.isExternalIdSubscribed(extId)
+    const send_opt_out = this.payload.sendBasedOnOptOut !== undefined && this.payload.sendBasedOnOptOut
+    let isSubscribed
+    if (send_opt_out) {
+      this.currentOperation?.logs.push('OptOut subscription')
+      isSubscribed = super.isExternalIdSubscribedOptOutModel(extId)
+    } else {
+      isSubscribed = super.isExternalIdSubscribed(extId)
+    }
     if (!isSubscribed) return isSubscribed // may be undefined => as invalid
```
pmunin, line 48: Having those suggestions in the MessageSendPerformer - you don't need that change here. If you do need that extra log message here - you should put it in the super.isExternalIdSubscribed in MessageSendPerformer

```suggestion
```

pmunin (changes requested):
plz see suggestions

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
pmunin: not sure why it captured line 48 in that suggestion above, so make sure you don't apply it as it

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -10,14 +10,26 @@ import { track } from './track'
...
 export enum SendabilityStatus {
+  /**
+   * No. of externalIds for which the message send is triggered based on the customer chosen send type (bypass, opt-out, opt-in)
+   */
   ShouldSend = 'should_send',
   /**
-   * No externalIds that supported by this Channel exist in the payload
+   * No. of externalIds for which the message send is not triggered based on the customer chosen send type (bypass, opt-out, opt-in)
+   */
+  ShouldNotSend = 'should_not_send',
```
pmunin, line 20: Seems like you are replacing `NotSubscribed` with `ShouldNotSend` - which does not look right, as the purpose of this status is to explain the reason why it's not being sent (if it's not ShouldSend) just by looking at it. I suggest to revert that replacement

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -145,28 +187,33 @@ export abstract class MessageSendPerformer<
...
 
-    const subscribedExtIds = supportedExtIdsWithSub.filter((e) => e.isSubscribed === true).map((e) => e.extId)
+    const shouldSendExtIds = supportedExtIdsBasedOnSendType.filter((e) => e.isSubscribed === true).map((e) => e.extId)
 
-    if (!subscribedExtIds.length) {
+    if (!shouldSendExtIds.length) {
       return invalidSubStatuses.length == 0
         ? {
-            sendabilityStatus: SendabilityStatus.NotSubscribed
+            sendabilityStatus: SendabilityStatus.ShouldNotSend
```
pmunin, line 216: See the comment above. I suggest to revert it

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -145,28 +187,33 @@ export abstract class MessageSendPerformer<
       return { sendabilityStatus: SendabilityStatus.SendDisabled }
     }
 
-    // list of extenalIds that are supported by this Channel, if none - exit
-    const supportedExtIdsWithSub = this.payload.externalIds
+    this.logPssSubscriptionStatesForExternalIds()
+
+    // list of extenalIds that are supported by this Channel, if none - exit based on customer chosen send type (bypass, opt-out, opt-in)
+    const supportedExtIdsBasedOnSendType = this.payload.externalIds
```
pmunin, line 193: this name is confusing, what is send type? it looks like the behavior is still defined based on the subscription status.  So I suggest to keep the old name `supportedExtIdsWithSub`

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -145,28 +187,33 @@ export abstract class MessageSendPerformer<
...
         sendabilityStatus: SendabilityStatus.NoSupportedExternalIds
       }
 
-    const invalidSubStatuses = supportedExtIdsWithSub.filter((e) => e.isSubscribed === undefined).map((e) => e.extId)
+    const invalidSubStatuses = supportedExtIdsBasedOnSendType
+      .filter((e) => e.isSubscribed === undefined)
+      .map((e) => e.extId)
 
-    const subscribedExtIds = supportedExtIdsWithSub.filter((e) => e.isSubscribed === true).map((e) => e.extId)
+    const shouldSendExtIds = supportedExtIdsBasedOnSendType.filter((e) => e.isSubscribed === true).map((e) => e.extId)
```
pmunin, line 211: same here

pmunin (changes requested):

joe-ayoub-segment:
hi @vikramkumar19 and @pmunin please let me know when the PR is ready for me to review / deploy.

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
pmunin: > The main idea behind this PR

Yep, I was reviewing it based on this exact understanding, I read the jira ticket.

pmunin: if the status is `did not subscribe` it should be noSendableStatus, same as of optIn model

also make sure you read the other comments I posted below.

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
VenkataNeeharikaKondipati: Based on customer chosen send type i.e bypass, opt-out / opt-in we decide if we 'should send' or 'should not send'. 
The subscription state of the user could be unsubscribed or did-not-subscribe but we still send the email if the customer has chosen 'bypass'. In this case, it confusing to have it 'NotSubscribed' as the reason here

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -106,14 +119,16 @@ export abstract class MessageSendPerformer<
     return this.constructor as ThisStaticClass
   }
 
-  /**
-   * check if extId is (un)subscribed (returns true|false) or if subscription status is invalid (returns undefined)
-   */
+  // /**
+  //  * check if extId is (un)subscribed (returns true|false) or if subscription status is invalid (returns undefined)
+  //  */
```
pmunin, line 124: I guess you're unintentionally commented it
```suggestion
  /**
   * check if extId is (un)subscribed (returns true|false) or if subscription status is invalid (returns undefined)
   */
```

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -94,8 +107,8 @@ export abstract class MessageSendPerformer<
    */
   abstract isSupportedExternalId(externalId: NonNullable<TPayload['externalIds']>[number]): boolean
 
-  static readonly nonSendableStatuses = ['unsubscribed', 'did not subscribed', 'false'] // do we need that??
-  static readonly sendableStatuses = ['subscribed', 'true']
+  static readonly nonSendableStatuses = ['false']
+  static readonly sendableStatuses = ['true']
```
pmunin, line 111: should we just in case also leave `unsubscribed` and `subscribed` in those arrays or you're absolutely sure those value will never come here? because if they will - nothing will be sent.  keep in mind test message scenario (destination-actions-service)

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -106,14 +119,16 @@ export abstract class MessageSendPerformer<
...
+  //  * check if extId is (un)subscribed (returns true|false) or if subscription status is invalid (returns undefined)
+  //  */
   isExternalIdSubscribed(extId: ExtId<TPayload>): boolean | undefined {
     const staticMems = this.getStaticMembersOfThisClass()
     const subStatus = extId.subscriptionStatus?.toString()?.toLowerCase()
-    if (!subStatus) return false // falsy status is valid and considered to be Not Subscribed, so return false
-    // if subStatus is not in any of the lists of valid statuses, then return true
+
+    // isOptOutModel if true means we can target to any statuses apart from unsubscribed
+    const isOptOutModel = this.payload.sendBasedOnOptOut ? true : false
```
pmunin, line 130: what about that `if(subStatus=='null')? subStatus = null`?

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -128,20 +143,45 @@ export abstract class MessageSendPerformer<
...
+        case 'true':
+          this.currentOperation?.tags.push('subscribed:')
+          break
+        // No. of externalIds that have an explicit status: "did-not-subscribe" or does not have an entry in PSS (i.e PSS returned 'isSubscribed: null')
+        case '':
+          this.currentOperation?.tags.push('did_not_subscribe:')
+          break
+        // Not a valid subscription status
+        default:
+          this.currentOperation?.tags.push('unknown_subscription_state:')
```
pmunin, line 170: You are not passing value to those tags. `tags` is the array. In your code you will just have a lot of `{tagName}:` items here with empty value. that does not look right. consider using some aggregate or `Array.reduce`

pmunin (changes requested):

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
vikramkumar19: As we walked through the scenarios this should not happen.

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
vikramkumar19: I am mostly sure about this. However, for keeping this more flexible shall add `unsubscribed ` and `subscribed ` back

pmunin: yeah I'd add it back just to be safe

pmunin (approved):
Thanks for all the changes

joe-ayoub-segment:
hi @vikramkumar19 the change looks fine to me - however I'm only looking for basic syntax issues, and other potential issues which might cause problems (such as adding a new required field for example). 

Usually we don't allow PRs containing changes to more than one Destination at a time, however in this case it makes sense. 

I see you did some testing. Did you do some regression testing as well?

My main question from the PR relates to how you are accessing the new field. 

Remember that for customers who already have the Integration configured prior to your deploy, `payload.sendBasedOnOptOut` will always return `undefined`. So you need to write your code to handle the 3 states for this optional boolean flag. 

If you can confirm that this is what you've done then I'm good to approve!

vikramkumar19:
> hi @vikramkumar19 the change looks fine to me - however I'm only looking for basic syntax issues, and other potential issues which might cause problems (such as adding a new required field for example).
> 
> Usually we don't allow PRs containing changes to more than one Destination at a time, however in this case it makes sense.
> 
> I see you did some testing. Did you do some regression testing as well?
> 
> My main question from the PR relates to how you are accessing the new field.
> 
> Remember that for customers who already have the Integration configured prior to your deploy, `payload.sendBasedOnOptOut` will always return `undefined`. So you need to write your code to handle the 3 states for this optional boolean flag.
> 
> If you can confirm that this is what you've done then I'm good to approve!

Hey @joe-ayoub-segment !
All these are good questions.
> I see you did some testing. Did you do some regression testing as well?
 Tested these changes in stage and all looked good. 
I have added couple more test cases here for all three states true, false and undefined.
Agreed on the backward compatibility we have added few fields in past and had success going forward.

pmunin:
@joe-ayoub-segment 

> for customers who already have the Integration configured prior to your deploy, `payload.sendBasedOnOptOut` will always return undefined

Wouldn't the old integrations fallback to default=false instead of undefined according to [the mappings](https://github.com/segmentio/action-destinations/pull/1506/files#diff-24cb4244068776ed36ee2fe097e6f31bc60868dc2acf9251e00d094cac45839bR126)? the code will handle both undefined and false equally, but i'm just currious

VenkataNeeharikaKondipati (approved):
LGTM! Thanks for all the additional stats

joe-ayoub-segment:
> 

Hi @pmunin , no, pre-existing Integrations won't default to false as there won't be an entry in the control plane DB for that instance of the Destination. This is one of the gotchas I need to look out for when approving PRs. 

The background on this is as follows: 

1. When an instance of an Integration is created by a customer the mappings for each field for that specific instance are saved to the database. So for that particular instance of the Destination, the database only contains mapping details for fields which existing when the instance of the Destination was created

2. When we deploy a new version of the Integration which has added fields, the new field mapping data in not present for pre existing instances. So the new field value will always result in undefined for those old instances.

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @vikramkumar19 this PR has been deployed.

PR #1507, opened 2023-08-17
[STRATCONN-2850] Return transformed payload from Segment Profile Destination

varadarajan-tw:
[RFC](https://docs.google.com/document/d/19uKcgtbNE9TJZHUYIPWqE9i0QwunkYPpBIDqP8vMeHw/edit) - Refer Solution 3.

[JIRA EPIC](https://segment.atlassian.net/browse/STRATCONN-2850)

This PR updates Segment Profiles to return the transformed payload and not make the API call. The transformed payload will no longer be sent from action-destination and instead will be delivered from centrifuge. To aid in returning transformed payload, this PR also adds a `data` node to Result object returned by action. This will be referenced in this [Integrations PR](https://github.com/segmentio/integrations/pull/2700)
Other Related PR
- [integrations-go](https://github.com/segmentio/integrations-go/pull/380)

There will be one followup PR to perform similar updates to `Segment Connections` destinations

## Testing

Stage test results can be found here for [Segment Destination](https://docs.google.com/document/d/19uKcgtbNE9TJZHUYIPWqE9i0QwunkYPpBIDqP8vMeHw/edit#heading=h.8ogs60px2e7z) and [Segment Profile](https://docs.google.com/document/d/19uKcgtbNE9TJZHUYIPWqE9i0QwunkYPpBIDqP8vMeHw/edit#heading=h.6zo61ihebjxu).

Changes to int monoservice response as a result of this change

**Flag Off**
Exchange is recorded as destination works as is.

image

**Flag On**
Transformed payload is returned, but no exchange is recorded as destination doesn't deliver to TAPI endpoint.

image


packages/core/src/create-test-integration.ts
```diff
@@ -73,6 +74,38 @@ class TestDestination<T, AudienceSettings = any> extends Destination<T, Audience
...
+      logger,
+      transactionContext,
+      stateContext
+    })
+
+    return responses
+  }
+
+  /** Same as testAction but returns responses and results */
+  async testAction2(
```
varadarajan-tw, line 94: I could use a better naming suggestion for this method 😅

packages/core/src/destination-kit/action.ts
```diff
@@ -155,7 +155,7 @@ export class Action<Settings, Payload extends JSONLikeObject, AudienceSettings =
 
     // Construct the request client and perform the action
     const output = await this.performRequest(this.definition.perform, dataBundle)
-    results.push({ output: output as JSONObject })
+    results.push({ data: output as JSONObject, output: 'Action Executed' })
```
varadarajan-tw, line 158: It's difficult to filter out data that we need to send back from a destination in integrations. So, I am adding a data node and setting the output there. Open to other suggestions as well.

varadarajan-tw: Results will be parsed in integrations like [here](https://github.com/segmentio/integrations/pull/2700/files#diff-6db3e5ddc729acbbb968d5dd771dfa6d542245e980e9b9d12bdb2417d2e05163R416)

marinhero (approved):
Hey Varada, I have a few questions regarding this PR forgive me if I'm making you backtrack:

- You mention a problem with Transformations but there's no clear explanation of why it's an issue and what those are.

- Why are we handling batches in the regular perform blocks?

- What does the current state look like and how will it change once your changes are implemented?

Thanks!

varadarajan-tw:
Thanks for the questions @marinhero 
- **You mention a problem with Transformations but there's no clear explanation of why it's an issue and what those are.**
Updated the [doc](https://docs.google.com/document/d/19uKcgtbNE9TJZHUYIPWqE9i0QwunkYPpBIDqP8vMeHw/edit#heading=h.cd1refkvt8jb) to give more context regarding transformation problem. TLDR; transformations can only happen at int-monoservice in the current pipeline but int-monoservice cannot communicate with any internal endpoints. So, we retain transformations in int monoservice but deliver the transformed payload from Centrifuge to the internal endpoint.

- **Why are we handling batches in the regular perform blocks?**
We are not handling batching in `perform` block. I'll add a comment to clarify 😅 . The transformed payload is returned in a format compatible to delivered via `https://api.segment.io/batch` endpoint and hence it is wrapped within a batch array. There is no solid reason for doing this. Repeater destination uses the batch endpoint even for delivering a single payload and I decided to adopt the same approach just for convenience of code in [int-go](https://github.com/segmentio/integrations-go/blob/v3/trackingapi/trackingapi.go).

- **What does the current state look like and how will it change once your changes are implemented?**
Updated the snapshot of how this change affects the int monoservice response in description. Apologies for not doing it earlier.

dlasky (approved):
lgtm - i don't _think_ that results array was ever actually used so its cool that you are doing somethign with it.

joe-ayoub-segment:
hi @varadarajan-tw this PR has been deployed.

PR #1522, opened 2023-08-24
[STRATCONN] updated setConfigurationField from set to config

AnkitSegment:

Worked on[ STRATCONN-3049](https://segment.atlassian.net/browse/STRATCONN-3049?atlOrigin=eyJpIjoiY2I2MzNmZjgzNTE2NDEyZmE0YzY1ZjMwMGYxNDIxNjMiLCJwIjoiaiJ9) 

In this build removed configuration field from index file and moved to setConfigurationField actions as gtag config tags run only once.

Screenshot 2023-08-24 at 10 09 00 AM

Screenshot 2023-08-24 at 10 08 40 AM

## Testing

Tested successfully on Local and stage server


packages/browser-destinations/destinations/google-analytics-4-web/src/setConfigurationFields/index.ts
jsabor, line 1: Line 106 send_page_view: true - Can the boolean value pull to what is set within the Segment Settings toggle like it did in the past?  We had it right before this commit change: https://github.com/segmentio/action-destinations/commit/7cea9239fce357be9a94703e6b46a4c3571688e2  For context, what was creating the dupe page error was a culmination of items, but the main issue was the gtag('event') and that error is corrected with these changes in this commit.

For line 106, we should add pageView setting, back to this file: packages/browser-destinations/destinations/google-analytics-4-web/src/generated-types.ts
Then update line 106 to read:
`send_page_view: settings.pageView`

jsabor (commented):
1 item on the send_page_view property.

varadarajan-tw (approved):

packages/browser-destinations/destinations/google-analytics-4-web/src/index.ts
```diff
@@ -134,29 +134,23 @@ export const destination: BrowserDestinationDefinition<Settings, Function> = {
         'If your CMP loads asynchronously, it might not always run before the Google tag. To handle such situations, specify a millisecond value to control how long to wait before the consent state update is sent. Please input the wait_for_update in milliseconds.',
       label: 'Wait Time to Update Consent State',
       type: 'number'
+    },
+    pageView: {
+      description: 'Set to false to prevent the default snippet from sending page views. Enabled by default.',
+      label: 'Page Views',
+      type: 'boolean',
+      default: true
```
varadarajan-tw, line 142: Sorry. I missed this. This value will only be applied to new subscriptions that'll be created. For existing instances, the default value won't apply, and it will be always `false`. We'll have to push-defaults but considering that GA4 is very popular, it will be a tedious task to update all instances with the new default value. So, to ensure backward compatibility,  suggest checking for undefined in perform block and assume `true` if so or else use whatever the value of `pageView` is.

varadarajan-tw (changes requested):

packages/browser-destinations/destinations/google-analytics-4-web/src/index.ts
jsabor: On index.ts it may not matter because the pageview will come from the setConfigurationFields and the ability to send a pageView was removed from index.ts on line 159.  
On gtag load (line 148), you can set consent defaults to 'denied' from the settings (line 155) and override the value on the analytics.page() or some track() event that maps to setConfigurationFields

packages/browser-destinations/destinations/google-analytics-4-web/src/setConfigurationFields/index.ts
jsabor, line 1: Thought about this overnight, we are missing the payload value for userId.  After line 117, need to add:
`if (payload.user_id){
config.user_id = payload.user_id
}`
We need the ability to send and null userId with the config command

jsabor (commented):
Added comments to index.ts about consent - I think we are good here.
Added one parameter add for user_id in setConfigurationFields.
Thank you.

varadarajan-tw (approved):

packages/browser-destinations/destinations/google-analytics-4-web/src/index.ts
jshwhitlow: I'd like to mention as an end-user that this change wasn't communicated by Segment to us in advance and we relied on gtag to do the pageView sending. (We had disabled page views within all our GA4 Web properties to avoid duplicate page views from both gtag and analytics.page call) We finally noticed something was wrong due to reporting being really low at the end of September but lost half a month's worth of page views.

In the future, better communication is needed on critical changes like this one.

ryanaip: I agree.  This was a major miss, and we also lost out on several weeks worth of page views.

ryanaip: @jshwhitlow would you mind sharing how you resolved your situation?  Are you manually sending page_view events through a different Segment mapping or manually with native gtag JS?

rvadera12: @jshwhitlow @ryanaip 

Thank you for bringing this to our attention. We acknowledge the issue and understand the impact it had on your reporting due to the uncommunicated change regarding page view sending through gtag. We're committed to resolving this and ensuring that such misunderstandings do not recur in the future.

To fix the issue, please ensure you have the "Set Configuration Fields" mapping configured and enabled in your mappings. This mapping is crucial for data collection by the gtag.js SDK. If you disabled or removed this mapping to resolve page_view duplication issues, please re-add and enable it immediately to ensure data is collected accurately.

If you need to send a page_view with your Configuration Mapping, toggle the "Send Page View" option on in the Segment settings. This step is essential to prevent any further data loss and ensure proper page_view tracking

ryanaip: @rvadera12 what do we do if we would like to send data both from our Identify and our Page calls? For example, Identify might have user properties like "company" and Page might have page properties like "category" (GA4 content_group).

If we enable "Send Page View", my understanding is that GA4 would then send duplicate page views, one for each of those two methods?  Our desired behavior would be for Identify to not send pageviews and Page to send pageviews.  Is that currently possible?

PR #1533, opened 2023-08-29
Add ABsmartly first actions implementation

marcio-absmartly:
This PR adds an destination action integration with ABsmartly.
This initial implementation converts track, page and screen calls and sends the events as goals to the ABsmartly collector.
It may also allow sending custom exposure events through Segment.

This is a refactoring of our destination function at https://github.com/absmartly/segment to fit the actions framework (currently deployed).

## Testing


joe-ayoub-segment:
Thank you @marcio-absmartly - I'll schedule for review!

packages/destination-actions/src/destinations/absmartly/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    mapping: defaultValues(trackExposure.fields),
+    type: 'automatic'
+  }
+]
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'ABsmartly (Actions)',
+  slug: 'actions-absmartly',
+  mode: 'cloud',
+
```
joe-ayoub-segment, line 56: Let's put a description in here. 
For example: 
description: 'Send analytics events and user profile data to events to ABsmartly...'

packages/destination-actions/src/destinations/absmartly/index.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+  },
+
+  /*
+  Not implemented in this release, but will be in the future.
+  Pending questions:
+    - Can we specify mappings / UI for this call as well? we need to map the userId/anonymousId to values on our platform, like we do for the other calls.
+    */
+  // onDelete: async (request, { settings, payload }) => {
+  // },
```
joe-ayoub-segment, line 105: OnDelete just passes the userId and nothing else. Customers generally aren't so concerned about anonymous data.

packages/destination-actions/src/destinations/absmartly/goal.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+  },
+  achievedAt: {
+    label: 'Goal Achievement Time',
+    type: 'datetime',
+    required: true,
+    description:
+      'Exact timestamp when the goal was achieved (measured by the client clock). Must be an ISO 8601 date-time string, or a Unix timestamp (milliseconds) number',
+    default: {
+      '@path': '$.originalTimestamp'
```
joe-ayoub-segment, line 53:
```suggestion
      '@path': '$.timestamp'
```

packages/destination-actions/src/destinations/absmartly/goal.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export const defaultGoalFields: Record<string, InputField> = {
+  units: {
+    label: 'Units',
+    type: 'object',
+    required: true,
+    description:
+      'The units of the goal to track. Mapping of unit name to source property in the event payload. Create Units in the Settings -> Units section of the ABsmartly Web Console',
+    defaultObjectUI: 'keyvalue:only',
+    default: {
```
joe-ayoub-segment, line 28: Do you need to specify if userId or anonymousId is required?

packages/destination-actions/src/destinations/absmartly/event.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export const defaultEventFields: Record<string, InputField> = {
+  publishedAt: {
+    label: 'Event Sent Time',
+    type: 'datetime',
+    required: true,
+    description:
+      'Exact timestamp when the event was sent (measured by the client clock). Must be an ISO 8601 date-time string, or a Unix timestamp (milliseconds) number',
+    default: {
+      '@path': '$.sentAt'
```
joe-ayoub-segment, line 30: Not sure why you want to collect the sentAt timestamp, but of course you can if you like! 

I'll explain what the 4 different Segment timestamps are: 

**originalTimestamp** - this is the time on the client when the event was generated. Note the time can be inacurate if the client's clock is incorrect. 
**sentAt** - this is the time on the client when the event was actually sent from the client. Note the time can be inacurate if the client's clock is incorrect. 
**receivedAt** - this is the time on Segment's servers when the event was received by Segment's servers. We guarantee that the time is correct. 
**timestamp** - If Segment detect that there's a difference between the **sentAt** and **receivedAt** timestamps, we infer that the Client's clock is off by that amount of time. We then correct the **originalTimestamp** value by that difference to arrive at the **timestamp** value.

packages/destination-actions/src/destinations/absmartly/event.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  },
+  agent: {
+    label: 'Agent',
+    type: 'string',
+    required: false,
+    description: 'Optional agent identifier that originated the event. Used to identify which SDK generated the event.',
+    default: 'segment'
```
joe-ayoub-segment, line 38:
```suggestion
    default: {'@path': '$.context.library.name'}
```

This will pull the actual name of the Segment library which was used to send the event. e.g. analytics.js for Web

packages/destination-actions/src/destinations/absmartly/event.ts
```diff
@@ -0,0 +1,74 @@
...
+  },
+  application: {
+    label: 'Application',
+    type: 'string',
+    required: false,
+    description:
+      'Optional application name that originated this event. Must exist if not empty. Create Applications in the Settings -> Applications section of the ABsmartly Web Console',
+    default: ''
```
joe-ayoub-segment, line 46: If you are not setting a default then just remove this line.

packages/destination-actions/src/destinations/absmartly/exposure.ts
```diff
@@ -0,0 +1,86 @@
...
+    return false
+  }
+
+  const attributes = exposure['attributes']
+  if (attributes != null && !Array.isArray(attributes)) {
+    return false
+  }
+
+  return true
+}
```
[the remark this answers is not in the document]
joe-ayoub-segment: To help me understand, could you paste in a JSON object for a valid Exposure payload please?

packages/destination-actions/src/destinations/absmartly/exposure.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  return sendEvent(
+    request,
+    settings,
+    {
+      ...exposure,
+      publishedAt: unixTimestampOf(payload.publishedAt)
+    },
+    payload.agent,
+    payload.application
```
joe-ayoub-segment, line 84: Hi @marcio-absmartly you don't seem to be sending the other fields from defaultEventFields here. Is that intentional?

Which user identifier are you sending?

joe-ayoub-segment (changes requested):
Hi @marcio-absmartly , 

Thanks for raising this PR - shall we schedule a call to review together. I've a few questions which would be best discussed in person. Hopefully we can get it all fixed up and approved during the call. 

You can schedule a call with me here http://calendly.com/joe_ayoub/integration-workshop

Best regards,
Joe

packages/destination-actions/src/destinations/absmartly/index.ts
marcio-absmartly: Sure, but we still the mapping, just like in other events. We need to know what corresponds to Segment's userId on our platform.

packages/destination-actions/src/destinations/absmartly/goal.ts
marcio-absmartly: Yes, we should specify that at least one is required. How can we do this?

packages/destination-actions/src/destinations/absmartly/event.ts
marcio-absmartly: It's because we do the same calculation that you do to get the final timestamp. So if we send the originals we get a little extra information and we just treat Segment as another SDK.

packages/destination-actions/src/destinations/absmartly/exposure.ts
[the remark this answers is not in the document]
marcio-absmartly: The idea is that we don't want the user to mess with this data in Segment - we want the payload generated by the SDK to be an atomic piece of data that gets passed directly here - because it's subject to change with different versions of the SDK, etc...

This code is just a safety measure to loosely validate that that is the case.

What this code is trying to do is saying that, for example, we cannot have any goals in this payload. For that to be true either there is no "goals" field, or the field is an empty array. Depending on the SDK, language, and Json encoding library used, sometimes they omit the field, sometimes the field is encoded as null, or sometimes it is an empty array, all valid cases of having no goals.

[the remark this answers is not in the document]
marcio-absmartly: Here is an example payload from our Javascript SDK's tests.

```
{
  publishedAt: 1611141535829,
  units: publishUnits,
  hashed: true,
  exposures: [
    {
      id: 1,
      name: "exp_test_ab",
      unit: "session_id",
      exposedAt: 1611141535729,
      variant: 1,
      assigned: true,
      eligible: true,
      overridden: false,
      fullOn: false,
      custom: false,
      audienceMismatch: false,
    },
  ],
  attributes: [
    {
      name: "attr1",
      setAt: 1611141535729,
      value: "value1",
    },
    {
      name: "attr2",
      setAt: 1611141535729,
      value: "value2",
    },
    {
      name: "attr3",
      setAt: 1611141535729,
      value: 3,
    },
  ],
},
```

packages/destination-actions/src/destinations/absmartly/exposure.ts
marcio-absmartly: Yes, so for these, we are sending the "atomic" payload directly from our SDK in that "exposure" property, which is already correctly formatted by definition.

Our exposure events are constantly evolving, so I thought it better to pass the payload directly and verbatim, instead of having us update the integration, and customers manage this mapping with each change to the exposure Schema.

This usage is not your typical Segment usage. We don't expect many customers to use this, and currently only one customer is going down this route. This is basically to support using Segment to publish our exposures, instead of sending them directly to us, and also send "dumber" 'Experiment Viewed' events to Segment to process by other destinations.

I am expecting this to also be useful if/when we build an ABsmartly Source, but must admit haven't done any research on that yet.

packages/destination-actions/src/destinations/absmartly/index.ts
joe-ayoub-segment: Hi @marcio-absmartly , 

Here's an example delete function from another Integration: 

  onDelete: async (request, { payload }) => {
    return request(`https://api.hilohq.com/v1/events/delete`, {
      method: 'POST',
      json: {
        event: {
          contact_id: payload.userId
        }
      }
    })
  }

Does that make sense?

packages/destination-actions/src/destinations/absmartly/index.ts
```diff
@@ -0,0 +1,116 @@
...
+  mode: 'cloud',
+  description:
+    'ABsmartly is an online experimentation service. Send events to ABsmartly to track experiments and goals.',
+
+  authentication: {
+    scheme: 'custom',
+    fields: {
+      collectorEndpoint: {
+        label: 'Collector Endpoint',
+        description: 'ABsmartly Collector endpoint, for example: https://you-subdomain.absmartly.io/v1',
```
joe-ayoub-segment, line 64:
```suggestion
        description: 'ABsmartly Collector endpoint, for example: https://your-subdomain.absmartly.io/v1. Contact ABsmartly Support if you don\'t know your Collector Endpoint',
```

packages/destination-actions/src/destinations/absmartly/index.ts
```diff
@@ -0,0 +1,116 @@
...
+        label: 'API Key',
+        description:
+          'ABsmartly SDK API Key. Create SDK Api Keys in the Settings > API Keys section of the ABsmartly Web Console',
+        type: 'string',
+        required: true
+      },
+      environment: {
+        label: 'Environment',
+        description:
+          'Environment name. Create Environments in the Settings > Environments section of the ABsmartly Web Console',
```
joe-ayoub-segment, line 79:
```suggestion
          'Environment name. Create Environments in the Settings > Environments section of the ABsmartly Web Console'. Environment name needs to match what\'s in the console. ,
```

packages/destination-actions/src/destinations/absmartly/event.ts
```diff
@@ -0,0 +1,73 @@
...
+    default: {
+      '@path': '$.sentAt'
+    }
+  },
+  agent: {
+    label: 'Agent',
+    type: 'string',
+    required: false,
+    description: 'Optional agent identifier that originated the event. Used to identify which SDK generated the event.',
+    default: {'@path': '$.context.library.name'}
```
joe-ayoub-segment, line 38:
```suggestion
        default: {
      '@if': {
        exists: { '@path': '$.context.library.name' },
        then: { '@path': '$.context.library.name' },
        else: 'segment'
      }
    }
```

packages/destination-actions/src/destinations/absmartly/trackGoal/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { defaultGoalFields, sendGoal } from '../goal'
+
+const fields = { ...defaultGoalFields }
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Track Goal',
+  description: 'Send a goal event to ABsmartly',
+  fields: fields,
```
joe-ayoub-segment, line 11: Add this line: 

  defaultSubscription: 'type = "track" and event != "Experiment Viewed"',

packages/destination-actions/src/destinations/absmartly/trackExposure/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { defaultExposureFields, sendExposure } from '../exposure'
+
+const fields = { ...defaultExposureFields }
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Track Exposure',
+  description: 'Send an experiment exposure event to ABsmartly',
+  fields: fields,
```
joe-ayoub-segment, line 11: Please add this line: 

defaultSubscription: 'type = "track" and event = "Experiment Viewed"',

joe-ayoub-segment (changes requested):

joe-ayoub-segment (approved):

PR #1537, opened 2023-08-29
Install necessary dep yet exclude  other optional deps

zikaari:
One of the optional deps is referenced by a post-install script, but we must also ignore other optional deps to preserve the integrity of the test


zikaari:
No longer applicable

PR #1555, opened 2023-09-05
STRATCONN-3095: Develop new Google Campaign Manager Action Destination

rhall-twilio:
This PR adds a new Browser Destination: Google Tag for Campaign Manager
Implementation spec: https://docs.google.com/document/d/1bj8CQcT030xHAtg3LtYE83AJEKybYgWYzSMWR4F5Iw8/edit

## Testing

Integration testing handled by @jsabor by virtue of deploying the destination in a local e-commerce environment, firing the analytics event, and confirming that the payloads matched our [product spec](https://docs.google.com/document/d/1bj8CQcT030xHAtg3LtYE83AJEKybYgWYzSMWR4F5Iw8/edit#heading=h.q6tba6dykg9y). We cannot test the E2E flow given that we don't have an active Ad campaign so will defer to customer for final validation.


jsabor (approved):
Looks good and passed testing, GTG!

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
```diff
@@ -0,0 +1,71 @@
+import type { Settings } from './generated-types'
+import type { BrowserDestinationDefinition } from '@segment/browser-destination-runtime/types'
+import { browserDestination } from '@segment/browser-destination-runtime/shim'
+
```
marinhero, line 4: Nit: keep your imports together.
```suggestion
```

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+import counterActivity from './counterActivity'
+
```
marinhero, line 6: Nit: keep your imports together.
```suggestion
```

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+declare global {
+  interface Window {
+    gtag: Function
+    dataLayer: any
+  }
+}
+
+export const destination: BrowserDestinationDefinition<Settings, Function> = {
+  name: 'Google Tag for Campaign Manager',
+  slug: 'actions-google-campaign-manager',
```
marinhero, line 18: confirmed match 👍🏽

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  mode: 'device',
+
+  settings: {
+    advertiserId: {
+      description:
+        'In Campaign Manager, go to Floodlight -> Configuration, and Advertiser ID is located under the Configuration heading.',
+      label: 'Advertiser ID',
+      type: 'string',
+      required: true,
+      default: 'DC-'
```
marinhero, line 28: Why do you need a default value that's not actionable? `DC-` points to nothing and can produce broken configuration easier than making the setting mandatory. What do you think?

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
```diff
@@ -0,0 +1,71 @@
...
+      window.dataLayer.push(arguments)
+    }
+
+    window.gtag('set', 'allow_ad_personalization_signals', settings.allowAdPersonalizationSignals)
+    window.gtag('js', new Date())
+    window.gtag('config', settings.advertiserId, {
+      conversion_linker: settings.conversionLinker
+    })
+    const script = `https://www.googletagmanager.com/gtag/js?id=${settings.advertiserId}`
+    await deps.loadScript(script)
```
marinhero, line 61: Consider adding a validation for correct loading:

```
    await deps.resolveWhen(() => Object.prototype.hasOwnProperty.call(window, 'theObjectYouLoadedWithTheScriptAbove'), 100)
```

marinhero: I see it might be an issue because you initialize `gtag` here but I recommend looking for something that indicates that the script loaded properly

packages/browser-destinations/destinations/google-campaign-manager/src/salesActivity/index.ts
```diff
@@ -0,0 +1,98 @@
+import type { BrowserActionDefinition } from '@segment/browser-destination-runtime/types'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const action: BrowserActionDefinition<Settings, Function, Payload> = {
+  title: 'Sales Activity',
+  description: 'Record monetary data for conversions, such as cost and the number of items sold',
```
marinhero, line 7:
```suggestion
  description: 'Record monetary data for conversions, such as cost and the number of items sold.',
```

packages/browser-destinations/destinations/google-campaign-manager/src/counterActivity/index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+// TODO: reasonable default paths
+
+const action: BrowserActionDefinition<Settings, Function, Payload> = {
+  title: 'Counter Activity',
+  description: 'Record non-monetary conversion data such as unique users, conversions, and session length.',
+  defaultSubscription: 'type = "track"',
+  platform: 'web',
+  fields: {
+    activityGroupTagString: {
```
marinhero, line 13: Consider grouping shared fields between actions into a single file and import them to each action. Will make our lives easier in the long run.

marinhero (commented):
Leaving a few comments for your consideration. Let me know what you think and I'll take another look.

packages/browser-destinations/destinations/google-campaign-manager/README.md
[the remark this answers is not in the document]
rhall-twilio: I didn't even see that. I think it was added by the scaffolder.

packages/browser-destinations/destinations/google-campaign-manager/src/counterActivity/index.ts
rhall-twilio: we'd be trading readability for consistency on fields that just don't change that much

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
rhall-twilio: this is to indicate the user that they should paste the ID that starts with `DC-`. Otherwise, I risk them putting just the number and then things will break. Ideally I would use the `placeholder` property but it's not available in this object for some reason.

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
rhall-twilio: This is following the same pattern as GA4: https://github.com/segmentio/action-destinations/blob/07b542f85c09b92e0a4f9202732455a18b27293e/packages/browser-destinations/destinations/google-analytics-4-web/src/index.ts#L146

Do we need to update them too?

Also a lot of other browser destinations:
image

packages/browser-destinations/destinations/google-campaign-manager/src/types.ts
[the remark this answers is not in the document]
rhall-twilio: This is wrecking havoc on the testcases mocking. I will timebox 1 hour to try to fix.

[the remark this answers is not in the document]
rhall-twilio: that was not too bad actually.

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
jsabor: The DC- is the prefix from Google, and when a user grabs their Advertiser ID, it does not include the DC- prefix, it is only a string of numbers.  Doubtful if this will break the configuration.  Most users of Campaign Manager would be saavy enough to include and understand the format.

marinhero: Ah, that's unfortunate. If the value aims to act as a placeholder then proceed.

packages/browser-destinations/destinations/google-campaign-manager/src/types.ts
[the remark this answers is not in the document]
rhall-twilio: actually,  I don't know. this broke GA4: 
image

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
marinhero: Volume != correctness 😄 I'd add it as it makes the behavior safer but it's ultimately up to you. Perhaps, it's not as important as I think.

packages/browser-destinations/destinations/google-campaign-manager/src/counterActivity/index.ts
marinhero: same as above, up to you.

marinhero (approved):
Just nits left. Thanks for addressing them all!

packages/browser-destinations/destinations/google-campaign-manager/src/index.ts
rhall-twilio: Let me continue as-is and will open a JIRA ticket to audit the browser destinations as I think there's more than two that should make this change.

jsabor: There is redundancy here calling gtag twice if they have GA4, but it wouldn't break anything, it is slightly wonky.  There is the potential that if they don't have the 'set' commands in parity, one would override the other; however, that would be an edge case.

Ideally in perfect world, we would have a destination for Google Tag where you could load gtag call the set and config commands for all Google Ids (GA4, Google Ads, Search Ads, Campaign Manager), and then the individual G destinations would handle gtag event calls... but we are not there and for now, this is the best option.

rhall-twilio: JIRA ticket for investigating later: https://segment.atlassian.net/browse/STRATCONN-3146

packages/browser-destinations/destinations/google-campaign-manager/src/counterActivity/index.ts
rhall-twilio: I'd say this is okay. I see it making sense for something like GA4 where they have thousands of actions.

packages/browser-destinations/destinations/google-campaign-manager/src/types.ts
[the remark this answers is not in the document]
rhall-twilio: reverted and filed: https://segment.atlassian.net/browse/STRATCONN-3147 for this sprint

PR #1567, opened 2023-09-08
Update Attio integration instructions, drop required `name` property

nickcampbell18:
We've done a bit of beta testing and found a few minor issues we'd like to fix:

* The actions tester presents the properties object in the opposite way to production, so I've swapped the descriptive text around to make that clearer:

Screenshot 2023-09-08 at 16 58 42
Screenshot 2023-09-08 at 16 58 49

* When we originally wrote this destination, we intended for `name` to be a required property, but this has changed in our platform; so I'm removing it here from the payload and setting the default value with a different name to show it's just an example with no specific meaning.

### DIFF
--- a/packages/destination-actions/src/destinations/attio/assertRecord/generated-types.ts
+++ b/packages/destination-actions/src/destinations/attio/assertRecord/generated-types.ts
@@ -10,7 +10,7 @@ export interface Payload {
    */
   matching_attribute: string
   /**
-   * Attributes to either set or update on the Attio Record. The values on the left should be Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs. The Matching Attribute must be included for assertion to work.
+   * Attributes to either set or update on the Attio Record. The values on the left should be Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs, for example: traits.name → name. The Matching Attribute must be included for assertion to work.
    */
   attributes?: {
     [k: string]: unknown
--- a/packages/destination-actions/src/destinations/attio/assertRecord/generated-types.ts
+++ b/packages/destination-actions/src/destinations/attio/assertRecord/generated-types.ts
@@ -10,10 +10,9 @@ export interface Payload {
    */
   matching_attribute: string
   /**
-   * Attributes to either set or update on the Attio Record. The keys on the left should be Attio Attribute IDs or Slugs, and the values on the right are Segment attributes or custom text. The Matching Attribute must be included for assertion to work.
+   * Attributes to either set or update on the Attio Record. The values on the left should be Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs. The Matching Attribute must be included for assertion to work.
    */
   attributes?: {
-    name?: string
     [k: string]: unknown
   }
 }
--- a/packages/destination-actions/src/destinations/attio/identifyUser/index.ts
+++ b/packages/destination-actions/src/destinations/attio/identifyUser/index.ts
@@ -24,21 +24,21 @@ const user_attributes: InputField = {
   label: 'Additional User attributes',
   description:
     'Additional attributes to either set or update on the Attio User Record. The values on the left should be ' +
-    'Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs.',
+    'Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs. ' +
+    'For example: traits.name → name',
   defaultObjectUI: 'keyvalue:only',
   additionalProperties: true,
   properties: {},
-  default: {
-    an_example_attribute: { '@path': '$.traits.name' }
-  }
+  default: {}
 }
 
 const person_attributes: InputField = {
   type: 'object',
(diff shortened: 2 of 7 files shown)


nickcampbell18:
One more question for the team please, while I'm here - I got this error while testing in production:

> request to https://api.attio.com/v2/objects/people/records/simple?matching_attribute=email_addresses failed, reason: The https request exceeded the deadline of 3.5s

Is this deadline applied to all network requests from an Actions destination? We might have to do some work to bring that time confidently under 3.5 seconds if so...

packages/destination-actions/src/destinations/attio/assertRecord/generated-types.ts
```diff
@@ -10,10 +10,9 @@ export interface Payload {
    */
   matching_attribute: string
   /**
-   * Attributes to either set or update on the Attio Record. The keys on the left should be Attio Attribute IDs or Slugs, and the values on the right are Segment attributes or custom text. The Matching Attribute must be included for assertion to work.
+   * Attributes to either set or update on the Attio Record. The values on the left should be Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs. The Matching Attribute must be included for assertion to work.
```
joe-ayoub-segment, line 13: Hi @nickcampbell18 , 
How about providing a simple example payload in the description?

packages/destination-actions/src/destinations/attio/identifyUser/index.ts
```diff
@@ -23,18 +23,13 @@ const user_attributes: InputField = {
...
-  properties: {
-    name: {
-      label: 'Name',
-      type: 'string'
-    }
-  },
+  properties: {},
   default: {
-    name: { '@path': '$.traits.name' }
+    an_example_attribute: { '@path': '$.traits.name' }
```
joe-ayoub-segment, line 32: Hi @nickcampbell18 , leaving this default as 'an_example_attribute' will likely result in lots of customers sending in an attribute called 'an_example_attribute' ;). Most customers won't know or think to change this mapping. 

It probably makes more sense to leave it as 'name', but that's up to you.

joe-ayoub-segment (changes requested):
Hi @nickcampbell18 , 
Thanks for raising the PR. Just a couple of minor questions. Let me know how you'd like to respond to them and I'll approve and merge the PR. 
Thanks,
Joe

joe-ayoub-segment:
> 

Hi @nickcampbell18 , in Production you do have more than 3.5 seconds. 
If a Production invocation does time out it will be reported in the delivery dashboard for that Destination. i.e the customer will be able to see it. Is that something you've noticed at all?

Do you know why the request is taking 3.5 seconds? Are you doing multiple requests serially or in parallel (I've not looked into the code)?

packages/destination-actions/src/destinations/attio/identifyUser/index.ts
nickcampbell18: I've removed the default and added a description instead, thanks for the tip 🙏

nickcampbell18:
Thanks for the clarification!

> Do you know why the request is taking 3.5 seconds? Are you doing multiple requests serially or in parallel (I've not looked into the code)?

In a nutshell it does two chained HTTP requests where the second depends on the result of the first. Performance is a tracked concern for the API that Segment is hitting, and we do occasionally have P99 tail latencies in the region of 10s due to a data modelling issue inside Attio itself. Depending on the popularity of the integration, we might need to find a way to improve the performance on the Attio side, probably queueing the writes (this would also allow us to support batching from the Segment side) - if so we'll submit a PR to update that implementation later.

> Is that something you've noticed at all?

In this case the error was reported to us by a customer, although they were using the "Test Mapping" button I believe, this wasn't a production event travelling through Segment normally.

joe-ayoub-segment:
Hi @nickcampbell18 ,
I believe that a 10s response time will result in a timeout on Segment's side. I don't think it will be retried either due to the lack of a response code (such as a 429). 

The only way I can think of making it work would be super hackey, and I wouldn't endorse it. You could wait for 3 or 4 seconds and then throw a RetryableError from the Segment perform() function. Segment will then retry the entire event later. You'd need to make sure that your first API call is idempotent though. Overall not a great solution!  

Having an ingestion queue would be a good idea. I've seen other Segment Partners do that successfully. 

Happy to discuss if you like. 
Cheers,
Joe

nickcampbell18:
Sounds good - thanks @joe-ayoub-segment. We might do that, but it would be a follow-up PR - this one is all good from our side 🙏

joe-ayoub-segment:
hi @nickcampbell18 , CI checks are breaking. Any chance you could fix please?

nickcampbell18:
Sorry about that @joe-ayoub-segment - I was having trouble running tests because pulling dependencies was breaking because of this private dependency, not sure if anybody else has hit this problem but I thought I'd flag it!

https://github.com/segmentio/action-destinations/blob/main/packages/cli-internal/package.json#L82

joe-ayoub-segment:
hi @nickcampbell18 this PR has been deployed.

nickcampbell18:
Thank you! 🙏

PR #1491, opened 2023-08-11
TTD - Controlled Batching

sayan-das-in:
This PR allows The Trade Desk destination to send events to a special S3 bucket for additional processing.

## Testing

Tested successfully on local and staging.
Testing document is available [here](https://docs.google.com/document/d/1cyPtGNdiJdPJp7Yd_BQM_Kh6yKQOKFiKvIYa7owZaoM/edit).

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [x] [Segmenters] Tested in the staging environment

PR #1569, opened 2023-09-11
LiveRamp Controlled Batching

sayan-das-in:
This PR adds Controlled Batching support to LiveRamp destination.
This is similar to Controlled Batching support in TTD (https://github.com/segmentio/action-destinations/pull/1491)

Note: This is one of the many PRs for LiveRamp. There will be a follow-up PR to refactor LiveRamp and The TradeDesk code and to include additional test cases to increase test coverage.

## Testing
Testing completed successfully in local instance of integrations monoservice with stage AWS account.
Test document is available here: https://docs.google.com/document/d/1Vc8h1chLLGPAzFZI6dEKr3I88aH_QiX14oJIMdb-fF4/edit#heading=h.aytgmb2bjig3

### DIFF
--- a/packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
+++ b/packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
@@ -6,7 +6,7 @@ import { LIVERAMP_MIN_RECORD_COUNT, LIVERAMP_LEGACY_FLOW_FLAG_NAME } from '../pr
 
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import type { ProcessDataInput } from '../operations'
+import type { RawData, ExecuteInputRaw, ProcessDataInput } from '../operations'
 
 const action: ActionDefinition<Settings, Payload> = {
   title: 'Audience Entered (S3)',
@@ -84,28 +84,22 @@ const action: ActionDefinition<Settings, Payload> = {
       unsafe_hidden: true,
       required: false,
       default: 170000
-    },
-    _segment_audience_computation_id: {
-      label: 'Segment Audience Compute ID',
-      description:
-        'The Audience Compute ID. This is used to identify audience chunks from a sync during aggregation. Note: This field is not visible on the UI.',
-      type: 'string',
-      unsafe_hidden: true,
-      default: { '@path': '$.context.personas.computation_id' }
     }
   },
-  perform: async (request, { payload, features }) => {
+  perform: async (request, { payload, features, rawData }: ExecuteInputRaw<Settings, Payload, RawData>) => {
     return processData({
       request,
       payloads: [payload],
-      features
+      features,
+      rawData: rawData ? [rawData] : []
     })
   },
-  performBatch: (request, { payload, features }) => {
+  performBatch: (request, { payload, features, rawData }: ExecuteInputRaw<Settings, Payload[], RawData[]>) => {
     return processData({
       request,
       payloads: payload,
-      features
+      features,
+      rawData
     })
   }
 }
@@ -131,7 +125,7 @@ async function processData(input: ProcessDataInput<Payload>) {
     // AWS FLOW
     // -----------
     return sendEventToAWS(input.request, {
-      audienceComputeId: input.payloads[0]._segment_audience_computation_id,
+      audienceComputeId: input.rawData?.[0].context?.personas?.computation_id,
       uploadType: 's3',
       filename,
       fileContents,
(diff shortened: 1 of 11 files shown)


packages/destination-actions/src/destinations/liveramp-audiences/operations.ts
```diff
@@ -1,11 +1,20 @@
...
+  payloads: T[]
+  features?: Record<string, boolean>
+}
 
 /*
 Generates the LiveRamp ingestion file. Expected format:
 liveramp_audience_key[1],identifier_data[0..n]
 */
-function generateFile(payloads: Payload[]) {
+function generateFile(payloads: sftpPayload[] | sftpPayload[]) {
```
varadarajan-tw, line 17: The types here seem to be redundant.

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
```diff
@@ -80,29 +84,65 @@ const action: ActionDefinition<Settings, Payload> = {
       unsafe_hidden: true,
       required: false,
       default: 170000
+    },
+    _segment_audience_computation_id: {
```
rhall-twilio, line 88: Will this break other users if a malicious attacker changes this? This can be intercepted in the GraphQL request before it's sent out.

packages/destination-actions/src/destinations/liveramp-audiences/operations.ts
rhall-twilio: they are as they're both sftp. one should be s3.

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
sayan-das-in: The mapping is used to get the Compute ID, this is then used as the directory where all chunks are stored. So in theory if this value is somehow altered, it will break the aggregator. Should we access `raw_data` directly in that case ?

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredSftp/index.ts
```diff
@@ -78,30 +82,66 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    //------------
+    // LEGACY FLOW
+    // -----------
+    const sftpClient = new ClientSFTP()
+    return uploadSFTP(sftpClient, input.payloads[0], filename, fileContents)
+  } else {
+    //------------
+    // AWS FLOW
+    // -----------
+    return sendEventToAWS(input.request, {
```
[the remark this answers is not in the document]
sayan-das-in: The payload goes to a Lambda function which aggregates all audience chunks back into one file. We store metadata in another which contents the actual destination the user intended to upload. Another Lambda function performs the final upload to S3 or SFTP server. We are re-using a part of the TTD pipeline.

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
rhall-twilio: I think so. We need to remain cognizant that unsafe_hidden doesn't mean that the user cannot change it, just that they cannot see it.

If it breaks only for this user is fine, but if they can use this to attack and overwrite another user it's a problem.

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredSftp/index.ts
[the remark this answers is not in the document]
rhall-twilio: so to confirm that other Lambda function will know how to write both to S3 and SFTP protocols?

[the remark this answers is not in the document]
sayan-das-in: Yes, the second Lambda function will have the context and will perform the actual upload to S3 and SFTP.

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
sayan-das-in: I have removed the hidden field and changed the logic to use raw data.

packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredSftp/index.ts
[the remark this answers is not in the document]
rhall-twilio: hey Sayan, can you point me to this Lambda function that uploads to SFTP? It seems to not be working.

PR #1581, opened 2023-09-15
[Algolia] feat: added addToCart and filterClicked events and made existing events more flexible

wwalser:
I've added two new event types to the Algolia Insights destination:
- [filterClicked event](https://www.algolia.com/doc/rest-api/insights/#method-param-filters)
- [addToCart event](https://www.algolia.com/doc/rest-api/insights/#method-param-eventname)

In addition to these two new events, I've extended the other events in the Algolia Insights destination to forward properties which are present on the event but not explicitly mapped. This is to allow us (Algolia) to expand the Insights API without requiring us to update this destination in lock-step with each udpate. Before this change, each change that we make to our API would need to be manually added to this destination. This additional flexibility enables customers to use any fields we add to the API in the future even if we haven't updated the Action Destination to explicitly map those fields. 

I've taken care to ensure this change is backward compatible so that it doesn't break any existing installs of the Algolia Insights destination.

## Testing
I've added tests for the two new events.
I've manually tested those events.
I've manually tested the new flexibility layer ensure it doesn't break existing installations. 

### DIFF
--- a/packages/browser-destinations/destinations/cdpresolution/src/generated-types.ts
+++ b/packages/browser-destinations/destinations/cdpresolution/src/generated-types.ts
@@ -2,11 +2,11 @@
 
 export interface Settings {
   /**
+   * Client identifier provided by CDP Resolution [Hashed Account ID]
+   */
+  clientIdentifier: string
+  /**
    * Identity resolution endpoint. [CDP Resolution documentation](https://www.cdpresolution.com/docs/)
    */
   endpoint: string
-  /**
-   * Client identifier provided by CDP Resolution [CDP Resultion Account](https://www.cdpresolution.com/accountsettings)
-   */
-  clientIdentifier: string
 }
--- a/packages/destination-actions/src/destinations/algolia-insights/__tests__/__snapshots__/snapshot.test.ts.snap
+++ b/packages/destination-actions/src/destinations/algolia-insights/__tests__/__snapshots__/snapshot.test.ts.snap
@@ -10,12 +10,8 @@ Object {
       "objectIDs": Array [
         "U[ABpE$k",
       ],
-      "products": Array [
-        Object {
-          "product_id": "U[ABpE$k",
-        },
-      ],
       "queryID": "U[ABpE$k",
+      "testType": "U[ABpE$k",
       "timestamp": null,
       "userToken": "U[ABpE$k",
     },
@@ -33,12 +29,42 @@ Object {
       "objectIDs": Array [
         "U[ABpE$k",
       ],
-      "products": Array [
-        Object {
-          "product_id": "U[ABpE$k",
-        },
-      ],
       "userToken": "U[ABpE$k",
+    },
+  ],
+}
+`;
+
+exports[`Testing snapshot for actions-algolia-insights destination: productAddedEvents action - all fields 1`] = `
+Object {
+  "events": Array [
+    Object {
+      "eventName": "Add to cart",
+      "eventType": "conversion",
+      "index": "g)$f*TeM",
+      "objectIDs": Array [
+        "g)$f*TeM",
+      ],
+      "queryID": "g)$f*TeM",
+      "testType": "g)$f*TeM",
+      "timestamp": null,
+      "userToken": "g)$f*TeM",
+    },
+  ],
+}
+`;
+
+exports[`Testing snapshot for actions-algolia-insights destination: productAddedEvents action - required fields 1`] = `
+Object {
+  "events": Array [
+    Object {
+      "eventName": "Add to cart",
+      "eventType": "conversion",
+      "index": "g)$f*TeM",
+      "objectIDs": Array [
+        "g)$f*TeM",
+      ],
+      "userToken": "g)$f*TeM",
     },
   ],
 }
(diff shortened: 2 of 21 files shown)


joe-ayoub-segment:
Hey @wwalser thanks for raising the PR. Hope everything is going well with you!
I'll schedule for review. 
Joe

packages/destination-actions/src/destinations/algolia-insights/filterClickedEvents/index.ts
```diff
@@ -0,0 +1,99 @@
...
+      label: 'Filters',
+      description:
+        'Populates the filters field in the Algolia Insights API, a list of up to 10 facet filters. Field should be an array of strings with format ${attribute}:${value}.',
+      type: 'string',
+      multiple: true,
+      required: true,
+      default: {
+        '@path': '$.properties.filters'
+      }
+    },
```
joe-ayoub-segment, line 21: You could strongly type this, as follows: 

```
     filters: {
      label: 'Filters',
      description:
        'Populates the filters field in the Algolia Insights API, a list of up to 10 facet filters. Field should be an array of strings with format ${attribute}:${value}.',
      type: 'object',
      multiple: true,
      required: true,
      defaultObjectUI: 'keyvalue',
      additionalProperties:false,
      properties: {
         attribute: {
            label: 'Filter name',
            description: 'The name of the Filter',
            type: 'string',
            required: true
         },
         value: {
            label: 'Filter value',
            description: 'The value of the Filter',
            type: 'string',
            required: true
         }
      },
      default: {
        '@arrayPath': [
          '$.properties.filters',
          {
            attribute: {
              '@path': '$.attribute'
            },
            value: {
              '@path': '$.value'
            }
          }
        ]
      }
    }, 
```

It will render something like this: 

image

In the real UI the "Add Key/Value" button won't be showing (which is good).

joe-ayoub-segment:
hi @wwalser , also just noticed that there's a minor CI issue. You can run `yarn validate` to check it. 

image

packages/destination-actions/src/destinations/algolia-insights/filterClickedEvents/index.ts
wwalser: It's useful to know how to do this in case I need to do it in other areas but in this specific case I think it'd be detrimental to Algolia's customers ability to integrate these events.

The "${attribute}:${value}" format is how our [customers pass filters to us](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/in-depth/filters-and-facetfilters/). It will often be the case that they already have it in this format when they need to emit the event. Requiring them to separate the two would require many to write their own tokenization. It's just splitting on `:`, but it's still a step that many don't have to take at the moment.

wwalser:
The `yarn validate` issue should be resolved.

packages/destination-actions/src/destinations/algolia-insights/filterClickedEvents/index.ts
joe-ayoub-segment: Hi @wwalser , 
I'm happy to go with whichever approach you like - but just want to point out that based on the Segment ecommerce spec we usually see customers  passing filtered data in this format: 

https://segment.com/docs/connections/spec/ecommerce/v2/#product-list-filtered

image

Let me know which you'd like to go for and I'll get this deployed. 
Cheers,
Joe

wwalser: mm, good point. I have tried to match the ecommerce spec where reasonable. I'll integrate this.

wwalser:
Integrated the strong typing on filter attributes and values.

wwalser:
Did some more manual testing and I think this is ready to go.

wwalser:
squashed and rebased on main.

joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @wwalser , 
PR approved - will deploy tomorrow. 

Minor comment: 
Segment does have a recommended event name for when a user adds items to a cart and for when a user applies a filter. They are "[Product Added](https://segment.com/docs/connections/spec/ecommerce/v2/#product-added)" and "[Product List Filtered](https://segment.com/docs/connections/spec/ecommerce/v2/#product-list-filtered)" respectively. This might not make sense for Algolia though, but just wanted to point it out in case you decide you want to change the event names for the default mappings. 

Cheers,
Joe

wwalser:
Thanks for the note! When I originally wrote the destination I tried to stick as closely as possible to the v2 ecommerce spec but basically forgot about that goal this time around.

I've renamed everything to have the correct name based on that spec.

I'm ready to land this when you are.

----

While I've got you here, I have a question:
A customer has noticed that one of the fields in this integration takes a single item (a string) when it should accept an array of strings. The `productViewedEvents` action is the one in question, it should accept multiple objectIds for when a customer is viewing a list of items. Based on the contributor notes I feel like changing that field defintion now will break anyone else who is already using the destination successfully, is that accurate? 

If there is no upgrade path for changing the existing field, I may add another action for the plural case. Either way it'll be in a separate PR from this one.

joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @wwalser , tests were breaking and I didn't have access to your branch, so I raised a PR in a local branch https://github.com/segmentio/action-destinations/pull/1615.

PR #1589, opened 2023-09-19
Update Attio Actions destination to use IDs as unique lookup field

nickcampbell18:
Hi team 👋 ,

This is a relatively small change based on some feedback with our early beta testers. We've changed the data model slightly in Attio, so that with the suggested "identify User" and "group Workspace" mappings you need to specify an additional `user_id` or `workspace_id` attribute (that is likely part of your Segment event already).

While it's technically a breaking change, none of our existing testers are actually sending production data through this Action yet, so we thought it a good time to make the change, and we'll coordinate with them as it's released.

## Testing

I've updated the unit tests and run through some manual end-to-end tests locally, this should still work great like before.

### DIFF
--- a/packages/destination-actions/src/destinations/attio/groupWorkspace/generated-types.ts
+++ b/packages/destination-actions/src/destinations/attio/groupWorkspace/generated-types.ts
@@ -6,9 +6,9 @@ export interface Payload {
    */
   domain: string
   /**
-   * The name of the Workspace
+   * The ID of the Workspace
    */
-  name: string
+  workspace_id: string
   /**
    * Additional attributes to either set or update on the Attio Company Record. The values on the left should be Segment attributes or custom text, and the values on the right are Attio Attribute IDs or Slugs. For example: traits.name → name
    */
--- a/packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
+++ b/packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
@@ -27,9 +27,9 @@ const workspace_id: InputField = {
   required: true,
   default: {
     '@if': {
-      exists: { '@path': '$.context.groupId' },
-      then: { '@path': '$.context.groupId' },
-      else: { '@path': '$.id' }
+      exists: { '@path': '$.groupId' },
+      then: { '@path': '$.groupId' },
+      else: { '@path': '$.context.group_id' }
     }
   }
 }
--- a/packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
+++ b/packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
@@ -19,17 +19,17 @@ const domain: InputField = {
   }
 }
 
-const name: InputField = {
+const workspace_id: InputField = {
   type: 'string',
-  label: 'Name',
-  description: 'The name of the Workspace',
+  label: 'ID',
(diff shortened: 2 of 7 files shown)


nickcampbell18:
Hi @joe-ayoub-segment, can I provide any additional context to help get this over the line? Thanks!

joe-ayoub-segment:
Hi @nickcampbell18 , thanks for raising this PR. 
I usually review code for all new Integrations but was on leave when your initial PR to add a new Integration was raised. 

Given that you are open to making a breaking change - there's a few things in the Integration which I have questions about. Could we have a quick sync to run through them please?

http://calendly.com/joe_ayoub/integration-workshop 

Best regards,
Joe

packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
```diff
@@ -19,17 +19,17 @@ const domain: InputField = {
...
   format: 'text',
   required: true,
   default: {
     '@if': {
-      exists: { '@path': '$.traits.name' },
-      then: { '@path': '$.traits.name' },
-      else: { '@path': '$.name' }
+      exists: { '@path': '$.context.groupId' },
+      then: { '@path': '$.context.groupId' },
+      else: { '@path': '$.id' }
```
joe-ayoub-segment, line 32:
```suggestion
      exists: { '@path': '$.groupId' },
      then: { '@path': '$.groupId' },
      else: { '@path': '$.context.group_id' }
```

packages/destination-actions/src/destinations/attio/groupWorkspace/generated-types.ts
```diff
@@ -6,9 +6,9 @@ export interface Payload {
    */
   domain: string
   /**
-   * The name of the Workspace
+   * The ID of the Workspace
    */
-  name: string
+  workspace_id: string
```
joe-ayoub-segment, line 11: Note: adding a new required field will break the Action for any customer who is already using it. 
Discussed with Nick and he is aware of this - no customer using this Action currently.

packages/destination-actions/src/destinations/attio/identifyUser/generated-types.ts
```diff
@@ -5,6 +5,10 @@ export interface Payload {
    * The email address of the person to link the user to
    */
   email_address: string
+  /**
+   * The ID of the User
+   */
+  user_id: string
```
joe-ayoub-segment, line 11: Another required field. This will break the experience for existing customers.

packages/destination-actions/src/destinations/attio/identifyUser/index.ts
```diff
@@ -19,6 +19,21 @@ const email_address: InputField = {
...
+  description: 'The ID of the User',
+  format: 'text',
+  required: true,
+  default: {
+    '@if': {
+      exists: { '@path': '$.userId' },
+      then: { '@path': '$.userId' },
+      else: { '@path': '$.anonymousId' }
+    }
+  }
```
joe-ayoub-segment, line 34:
```suggestion
  default: { '@path': '$.userId' }
```

joe-ayoub-segment (changes requested):

joe-ayoub-segment (approved):

PR #1605, opened 2023-09-22
STRATCONN-3184 - fix for iterable date issue

joe-ayoub-segment:
As per https://segment.atlassian.net/browse/STRATCONN-3184

Before a payload is sent to Iterable, the object is parsed for strings that might be dates. If there's a match then the date string is converted into a date string format that Iterable expects. 

However this parsing is converting non date strings into Iterable date strings. 

For example the following string "1234" gets converted into "1234-01-01 00:00:00 +00:00"

Proposed fix for this is to edit the regex responsible for evaluating strings as dates so that it only works if the string starts with a YYYY-MM-DD format.  

## Testing

Tested with local testing tool. 
Also wrote a unit test.

### DIFF
--- a/packages/browser-destinations/destinations/cdpresolution/src/generated-types.ts
+++ b/packages/browser-destinations/destinations/cdpresolution/src/generated-types.ts
@@ -2,11 +2,11 @@
 
 export interface Settings {
   /**
-   * Client identifier provided by CDP Resolution [Hashed Account ID]
-   */
-  clientIdentifier: string
-  /**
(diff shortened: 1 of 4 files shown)


KakeJopulsky:
Successfully tested locally and through unit tests.

rhall-twilio (approved):

joe-ayoub-segment:
Approved for deploy by Partner

joe-ayoub-segment:
Hi @KakeJopulsky this PR has been deployed.

PR #1613, opened 2023-09-26
[Trade Desk] Fix Auth Error for createAudience

maryamsharif:
The auth field was originally only defined in the `extendRequest` function but when createAudience is called from ADS the `extendRequest` does not actually input the values to these requests resulting in audiences not being created and NULL being saved as the external_id. This PR fixes this issue.

Also removing the dropdown for region because the UI does not support the drop-downs correctly causing it to look like the field is filled but it is really not saved in the DB.

## Testing

Before
Screenshot 2023-09-25 at 6 05 06 PM

After
Screenshot 2023-09-25 at 6 05 27 PM

### DIFF
--- a/packages/browser-destinations/destinations/cdpresolution/src/generated-types.ts
+++ b/packages/browser-destinations/destinations/cdpresolution/src/generated-types.ts
@@ -2,11 +2,11 @@
 
 export interface Settings {
   /**
-   * Client identifier provided by CDP Resolution [Hashed Account ID]
-   */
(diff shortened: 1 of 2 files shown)


rvadera12 (approved):

marinhero:
Ah, yes. This would do that since it's not actually an auth field. I'm sorry I misled you. Good catch.

joe-ayoub-segment:
note: this was deployed 26-Sep-23

PR #1650, opened 2023-10-12
Removes the concurrency=1 setting from the top-level build commands

chrisradek:
This PR removes the concurrency=1 from the root-level build scripts.

## Context on changes
Simply removing concurrency=1 from the scripts caused builds to fail - this repeatedly manifested as `browser-destination-runtime` failing to build due to missing types on `actions-core`.

Long story short, the issue was that nearly every package's build command also included a clean command which ended up removing their dependencies before rebuilding them - causing chaos when more than 1 package was being built at a time.

## Background on the build steps

For those interested...

Most of our packages were using `tsc -b tsconfig.build.json` for building, and the tsconfig.build.json includes references to in-repo packages they depend on. You can learn more about build mode for TypeScript here:
https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript
> Running tsc --build (tsc -b for short) will do the following:
> - Find all referenced projects
> - Detect if they are up-to-date
> - Build out-of-date projects in the correct order

This is really useful when working on a single package, as you can run build within it and any in-repo dependencies will be built if needed.

However when combined with the `--clean` flag, it will also remove the build artifacts of its dependencies.

`yarn clean`, which contained `tsc -b tsconfig.build.json --clean` for most packages, was the first step ran as many package's `build` script. This is how we ended up running into issues when 2+ packages that share a dependency were being built at the same time - one's clean step could kick in while the other was trying to build.


package.json
```diff
@@ -18,8 +18,8 @@
     "cli-internal": "yarn workspace @segment/actions-cli-internal",
     "core": "yarn workspace @segment/actions-core",
     "bootstrap": "lerna bootstrap",
-    "build": "./bin/run generate:types && lerna run build  --concurrency 1 --stream --ignore @segment/actions-cli-internal && yarn browser build-web",
-    "build:browser-destinations": "yarn lerna run build --concurrency 1 --scope=@segment/destinations-manifest --include-dependencies --stream && yarn browser build-web",
+    "build": "./bin/run generate:types && lerna run build --stream --ignore @segment/actions-cli-internal && yarn browser build-web",
```
dlasky, line 21: 🔥

packages/actions-shared/package.json
```diff
@@ -23,7 +23,7 @@
     "registry": "https://registry.npmjs.org"
   },
   "scripts": {
-    "build": "yarn clean && yarn tsc -b tsconfig.build.json",
+    "build": "yarn tsc -p tsconfig.build.json",
```
dlasky, line 26: curious if this actually makes a difference?

dlasky (approved):
thanks for digging in here!!!

packages/actions-shared/package.json
chrisradek: I actually want to try turning from `-p` back to `-b` for the build step - I think that's still useful functionality to preserve if we can.

dlasky (approved):

packages/actions-shared/package.json
chrisradek: It works locally putting it back, fingers crossed CI doesn't complain! (It shouldn't, really just the clean was causing problems)

silesky: I kinda like this change, as tsc -p is simpler to reason sboyt! Lerna is already dependency-aware -- so using lerna to then trigger another dependency aware build makes my head hurt

silesky: I kind of prefer that -p since it's easier to reason sbout, as lerna is already dependency aware... so by using lerna with tsc -b and parallization, we are simultaneously running two graph build systems that have the potential to trip over eachother (since lerna also runs all sub packages AFAIK)

silesky (approved):

PR #1651, opened 2023-10-13
Improve CI speed, add nx

silesky:
image

What does this do?
- breaks up CI into more steps, reducing total CI time.
- adds nx so we can have CI / local build caching (have not added it to CI)
- replace the slow generate type check on every build with a CI check for ensuring that types are up-to-date.
- QoL updates like adding titles to the test runs

Limitations:
- This does not enable distributed build caching in CI -- something @dlasky was going to look at
- This does not speed up a full test run on CI without caching -- seems to be some hard resource constraints with jest.
- No lazy linting at all, since Monorepo does not have individual lint commands for individual projects.
- No lazy testing for destinations  (e.g. running tests on only the specific destinations that changed),  since monorepo does not currently have individual test commands for individual destinations.

image


dlasky (approved):

NolanChan (approved):

NolanChan:
oh interesting, we could make testing really fast by lazy loading testing once we develop that :)

PR #1638, opened 2023-10-09
Action Destination for Apollo.io

karunkumarjha:
Implemented a plugin for Apollo in action destinations. It ingests data from track API and send it to a webhook.

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

PR #1672, opened 2023-10-19
New Destination - Apollo.io 

joe-ayoub-segment:
New Cloud Mode Destination. 
Original [PR raised here ](https://github.com/segmentio/action-destinations/pull/1638) - with review and feedback. Raised this new PR to fix some tests. 

## Testing

tested locally


packages/destination-actions/src/destinations/apolloio/track/index.ts
```diff
@@ -0,0 +1,164 @@
...
+          description: 'The search query of the page where the event occurred.',
+          label: 'Search'
+        }
+      }
+    }
+  },
+  perform: (request, data) => {
+    return request(baseURL, {
+      method: 'post',
+      json: [data.payload]
```
varadarajan-tw, line 153: Why is the payload array here but single event in `performBatch`?  Maybe the two needs to be swapped?

packages/destination-actions/src/destinations/apolloio/track/index.ts
```diff
@@ -0,0 +1,164 @@
...
+      default: { '@path': '$.event' }
+    },
+    properties: {
+      type: 'object',
+      label: 'Properties',
+      description: 'Properties to associate with the event',
+      default: { '@path': '$.properties' }
+    },
+    timestamp: {
+      type: 'string',
```
varadarajan-tw, line 31: Why not use datetime?

packages/destination-actions/src/destinations/apolloio/index.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { Settings } from './generated-types'
+import { IntegrationError, defaultValues } from '@segment/actions-core'
+
+import track from './track'
+export const baseURL = process?.env?.ACTIONS_APOLLOIO_BASE_URL_SECRET
+  ? `${process.env.ACTIONS_APOLLOIO_BASE_URL_SECRET}`
+  : 'https://apollo.io/'
+export const authURL = process?.env?.ACTIONS_APOLLOIO_AUTH_URL_SECRET
+  ? `${process.env.ACTIONS_APOLLOIO_AUTH_URL_SECRET}`
+  : 'https://apollo.io/'
```
varadarajan-tw, line 11: nit: This is a lot more succinct. [Null coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)
```suggestion
export const authURL = process?.env?.ACTIONS_APOLLOIO_AUTH_URL_SECRET ?? 'https://apollo.io/'
```

varadarajan-tw (approved):
LGTM. Left few comments.

packages/destination-actions/src/destinations/apolloio/track/index.ts
joe-ayoub-segment: I think it's already correct - the Partner wants to send an array regardless. And data.payload is already an array in the performBatch function.

joe-ayoub-segment:
note: PR deployed

PR #1677, opened 2023-10-23
Conman 703/adding field

JusteenRandev:
## Description

- This adds an optional field to the send email for the `engage-messaging-sendgrid` destination
- New field is ip_pool_name
- Allows sender to choose which ip pool they are sending their email from (Specifies which ip address they are using)
- If no ip pool name is given the default behaviour continues
- No pool name or a valid pool name MUST be given. If an invalid pool name is entered the email will not send

## Testing

Additional manual testing was done to check the email gets sent from the correct IP address when the pool name is given.
See below screenshots from the sendgrid account (Repeated for each ip pool):
image
IPPoolDemo
Untitled
IPPoolNameNone


packages/destination-actions/src/destinations/engage/sendgrid/__tests__/send-email.test.ts
```diff
@@ -1495,6 +1496,57 @@ describe.each([
...
+        }),
+        settings,
+        mapping: getDefaultMapping({ ipPool: 'testPoolName' })
+      })
+
+      expect(responses.length).toBeGreaterThan(0)
+      expect(sendGridRequest.isDone()).toEqual(true)
+    })
+
+    it('sends the email to ip pool when name is not specified', async () => {
```
vikramkumar19, line 1530: Can we add a test case when ipPool === `''`. To make sure we capture all scenario's of ipPools `undefined` or `''`

vikramkumar19 (commented):
Overall LGTM!Can we add few tests around all the possible values of ipPool. Also can you check if `null` is referred as `undefined`

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
```diff
@@ -193,6 +193,16 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload>
       mailContent = mailContentSubscriptionHonored
       this.statsClient?.incr('request.dont_pass_subscription', 1)
     }
+    // Check if ip pool name is provided and sends the email with the ip pool name if it is
+    if (this.payload.ipPool) {
+      mailContent = {
+        ...mailContent,
+        ip_pool_name: this.payload.ipPool
```
sumitbhanwala, line 200: @JusteenRandev I am pretty sure that you have used the right field name as documentation suggests  `ipPoolName` I also remember running into the discrepancy into the documentation vs what works in send grid API

sumitbhanwala: Also what Vikram said if this.payload.ipPool resolves to an empty string ''. Will it evaluate to true.
If yes than we shouldn't be going inside the if block

JusteenRandev: It works when I use that field when manually testing and I see it in the documentation here: https://docs.sendgrid.com/api-reference/mail-send/mail-send

sumitbhanwala: Yes that's what I meant, If we use there client they do it camelCase
ipPoolName: 'transactional email'.
Probably the client will convert before sending to server.

vikramkumar19 (approved):
LGTM!

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
JusteenRandev: If it resolves to an empty string it evaluates to false. I tested this manually and added it as a test case. The ip_pool_name field doesn't get set. The mailContent seems to be getting sent as a json object in a post request [here](https://github.com/segmentio/action-destinations/pull/1677/files#diff-0b5e29bcd732565f68cff10ca6765608743a263d3d4131894734d79f044f9272L201) so I followed the documentation for the curl command for the field. It's seems to work when testing manually so I'm not sure what else to check that it is the right field name.

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/actionDefinition.ts
```diff
@@ -119,6 +119,12 @@ export const actionDefinition: ActionDefinition<Settings, Payload> = {
       type: 'boolean',
       default: false
     },
+    ipPool: {
+      label: 'IP Pool',
+      description: 'Send email with an ip pool',
+      type: 'string',
+      default: ''
```
NolanChan, line 126: Just wondering if there's an explicit reason to assign an empty string default? It shouldn't affect anything either way, so no need to change it

NolanChan (approved):

sumitbhanwala (approved):
LGTM

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/actionDefinition.ts
JusteenRandev: There isn't any explicit reason.

joe-ayoub-segment:
hi @JusteenRandev this PR has been deployed.

PR #1683, opened 2023-10-24
APP-86828 Pendo browser destination

wordythebyrd:

Pendo would like to release a 1:1 Segment data => Pendo data implementation and possibly add more customizations or mappings for the user later.  I removed the settings for `accountId`, `parentAccountId`, and `setVisitorIdOnLoad`.  We want the Pendo visitorId to always be the Segment userId or Segment anonymous id with a Pendo prefix applied to it.  Same with Segment groupId to Pendo accountId, then we forward the user traits and group traits to Pendo as well.

## Testing

### DIFF
--- a/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
+++ b/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
@@ -11,59 +11,49 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
   fields: {
     visitorId: {
       label: 'Visitor ID',
-      description: 'Pendo Visitor ID. Defaults to Segment userId',
+      description: 'Pendo Visitor ID. Maps to Segment userId',
       type: 'string',
       required: true,
       default: {
         '@path': '$.userId'
-      }
+      },
+      readOnly: true
     },
     accountId: {
       label: 'Account ID',
-      description: 'Pendo Account ID. This overrides the Pendo Account ID setting',
+      description: 'Pendo Account ID',
       type: 'string',
       required: true,
-      default: { '@path': '$.groupId' }
+      default: { '@path': '$.groupId' },
+      readOnly: true
     },
     accountData: {
       label: 'Account Metadata',
       description: 'Additional Account data to send',
       type: 'object',
-      required: false
-    },
-    parentAccountId: {
-      label: 'Parent Account ID',
-      description:
-        'Pendo Parent Account ID. This overrides the Pendo Parent Account ID setting. Note: Contact Pendo to request enablement of Parent Account feature.',
-      type: 'string',
-      required: false
-    },
-    parentAccountData: {
-      label: 'Parent Account Metadata',
-      description:
-        'Additional Parent Account data to send. Note: Contact Pendo to request enablement of Parent Account feature.',
-      type: 'object',
-      required: false
+      required: false,
+      default: { '@path': '$.traits' },
+      readOnly: true
     }
   },
   perform: (pendo, event) => {
     const payload: PendoOptions = {
       visitor: {
         id: event.payload.visitorId
-      }
-    }
-    if (event.payload.accountId) {
-      payload.account = {
+      },
+      account: {
         ...event.payload.accountData,
         id: event.payload.accountId
       }
     }
-    if (event.payload.parentAccountId) {
-      payload.parentAccount = {
-        ...event.payload.parentAccountData,
-        id: event.payload.parentAccountId
-      }
+
+    // If parentAccount exists in group traits, lift it out of account properties into options properties
+    // https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/pendo/lib/index.js#L136
+    if (payload.account?.parentAccount) {
+      payload.parentAccount = payload.account.parentAccount
+      delete payload.account.parentAccount
     }
+
     pendo.identify(payload)
   }
 }
--- a/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
+++ b/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
@@ -1,7 +1,7 @@
 import type { BrowserActionDefinition } from '@segment/browser-destination-runtime/types'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import type { PendoSDK, identifyPayload } from '../types'
+import type { PendoSDK, PendoOptions } from '../types'
 
 const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
   title: 'Send Group Event',
@@ -47,21 +47,21 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
     }
   },
   perform: (pendo, event) => {
-    const payload: identifyPayload = {
+    const payload: PendoOptions = {
       visitor: {
         id: event.payload.visitorId
       }
     }
-    if (event.payload.accountId || event.settings.accountId) {
+    if (event.payload.accountId) {
       payload.account = {
-        id: event.payload.accountId ?? (event.settings.accountId as string),
-        ...event.payload.accountData
+        ...event.payload.accountData,
+        id: event.payload.accountId
       }
     }
-    if (event.payload.parentAccountId || event.settings.parentAccountId) {
+    if (event.payload.parentAccountId) {
       payload.parentAccount = {
-        id: (event.payload.parentAccountId as string) ?? (event.settings.parentAccountId as string),
-        ...event.payload.parentAccountData
+        ...event.payload.parentAccountData,
+        id: event.payload.parentAccountId
       }
     }
     pendo.identify(payload)
--- a/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
+++ b/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
@@ -34,6 +34,22 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
       required: false,
       default: { '@path': '$.traits' },
       readOnly: true
+    },
+    parentAccountData: {
+      label: 'Parent Account Metadata',
+      description:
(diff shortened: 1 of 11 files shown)


joe-ayoub-segment:
Thanks for the PR @wordythebyrd! I'll schedule for review.

packages/browser-destinations/destinations/pendo-web-actions/src/index.ts
```diff
@@ -28,44 +29,17 @@ export const destination: BrowserDestinationDefinition<Settings, PendoSDK> = {
...
     region: {
       label: 'Region',
       type: 'string',
       description: "The Pendo Region you'd like to send data to",
       required: true,
       default: 'io',
       choices: [
-        { value: 'io', label: 'io' },
-        { value: 'eu', label: 'eu' }
+        { value: 'io', label: 'app.pendo.io' },
```
wpj, line 39: What do you think about having these values be the full agent CDN URL instead of just the region identifer? That way, we wouldn't need to construct the CDN URL at runtime.

packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
```diff
@@ -47,21 +47,21 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
     }
   },
   perform: (pendo, event) => {
-    const payload: identifyPayload = {
```
wpj, line 50: I think we may want to construct a payload that's identical to the `initialize` payload so that Segment traits/Pendo metadata stays consistent between those two.

packages/browser-destinations/destinations/pendo-web-actions/src/identify/index.ts
```diff
@@ -61,22 +61,22 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
     }
   },
   perform: (pendo, event) => {
-    const payload: identifyPayload = {
```
wpj, line 64: I think we may want to construct a payload that's identical to the `initialize` payload so that Segment traits/Pendo metadata stays consistent between those two.

packages/browser-destinations/destinations/pendo-web-actions/src/index.ts
wordythebyrd: Agreed, this has been updated.

packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
wordythebyrd: I've tried to update the group, identify, and track actions to reflect this.  Will need to confirm with @joe-ayoub-segment if setting the fields to readOnly accomplishes what we need here.

packages/browser-destinations/destinations/pendo-web-actions/src/identify/index.ts
wordythebyrd: Addressed in previous comment.

packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
```diff
@@ -11,59 +11,49 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
...
-        id: (event.payload.parentAccountId as string) ?? (event.settings.parentAccountId as string),
-        ...event.payload.parentAccountData
-      }
+
+    // If parentAccount exists in group traits, lift it out of account properties into options properties
+    // https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/pendo/lib/index.js#L136
+    if (payload.account?.parentAccount) {
+      payload.parentAccount = payload.account.parentAccount
+      delete payload.account.parentAccount
     }
```
joe-ayoub-segment, line 55: One of the goals of the platform we're using here (which we call 'Actions') is that we want to allow customers to remap fields from a Segment payload to fields in Pendo in the Segment UI. This is to minimise changes a customer's Developer would need to make to their analytics tracking code on their website. 

So for example, we should let the customer decide where in the Segment payload they want to provide the parentAccount info. Here's 2 examples: 

traits: {
  parentAccount: 'parentAccountValue'
}

or equally: 

traits: {
  pendoParentAccount: 'parentAccountValue'
}

If we go with the approach in your code the customer will lose the ability to remap their Segment payloads to the Pendo payload (at least for the parentAccount). 

For this reason I'd recommend we provide a specific field to capture the mapping for parentAccount. 

Keen to hear your thoughts on this.

joe-ayoub-segment (commented):
hi @wordythebyrd, 

Thanks for suggesting these changes. 
I understand why you have recommended them - however making the majority of the fields 'readOnly' will prevents the customer from being able to remap data before sending it to Pendo. 

For example the customer might send the following traits object in their Segment payload: 

traits: {
  first_name: 'Billy', 
  last_name: 'Bob',
  is_recent_purchaser: true, 
  age: 23,
  gender: 'Male'
}

But they may only want to send some of those traits on to Pendo, or they might want to rename some of those traits before forwarding them to Pendo. For example they might only want to send the following visitorData: 

visitor: {
    fname: 'Billy', 
    lname: 'Bob'
}

I was hoping to find a way to retain the customer's ability to remap fields, while still being able to initialize the Pendo JS library correctly.

packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
wordythebyrd: I think the `parentAccount` property may be the use-case where we could allow the customer to map a trait to a specific field since we do need to lift it up a level in the payload, but doing so will require I remove it from the `pendo.initialize` call.

wordythebyrd:
> hi @wordythebyrd,
> 
> Thanks for suggesting these changes. I understand why you have recommended them - however making the majority of the fields 'readOnly' will prevents the customer from being able to remap data before sending it to Pendo.
> 
> For example the customer might send the following traits object in their Segment payload:
> 
> traits: { first_name: 'Billy', last_name: 'Bob', is_recent_purchaser: true, age: 23, gender: 'Male' }
> 
> But they may only want to send some of those traits on to Pendo, or they might want to rename some of those traits before forwarding them to Pendo. For example they might only want to send the following visitorData:
> 
> visitor: { fname: 'Billy', lname: 'Bob' }
> 
> I was hoping to find a way to retain the customer's ability to remap fields, while still being able to initialize the Pendo JS library correctly.

Pendo clients are expecting the 1:1 relationship with Segment users / user traits / groups / group traits and this is what has been communicated to me from Pendo Product and Eng leadership, while I understand the value of the field mappings, we should wait and collect feedback from our customers to determine if there is a need for this.

In addition, there are risks associated with initializing Pendo without the user data and relying on the identify calls to supply it, since that's where the mappings occur.  Primarily that changes to metadata won't be recognized until the customer's `Segment.identify()` call is made, which could result in gaps of unknown duration where guide's are not correctly retargeted in Pendo.  It also anecdotally appears to open up more ways for the customer to inadvertently misconfigure the setup.

Ultimately, this will get Pendo customers on the actively maintained Analytics 2.0 platform, open the integration to Pendo customers in all our regions (US/EU/JP/US-Restricted), and add support for Pendo customer's using Pendo's CNAME feature to customize their domain, so this is still very high value and exciting to launch.

joe-ayoub-segment:
hi @wordythebyrd if you're happy with the changes let's get them deployed so we can test them. OK with you?

wordythebyrd:
> hi @wordythebyrd if you're happy with the changes let's get them deployed so we can test them. OK with you?

Yes, once merged I will dev test at the direct URL you showed me during our call.

joe-ayoub-segment:
> > hi @wordythebyrd if you're happy with the changes let's get them deployed so we can test them. OK with you?
> 
> Yes, once merged I will dev test at the direct URL you showed me during our call.

Thanks @wordythebyrd . 
I just noticed that CI failed. A test needs to be updated. I'd do it if I had access to your branch! 
Any chance you could take a look please?

image

wordythebyrd:
> > > hi @wordythebyrd if you're happy with the changes let's get them deployed so we can test them. OK with you?
> > 
> > 
> > Yes, once merged I will dev test at the direct URL you showed me during our call.
> 
> Thanks @wordythebyrd . I just noticed that CI failed. A test needs to be updated. I'd do it if I had access to your branch! Any chance you could take a look please?
> 
> image

Yes, I believe I pushed up a fix on Friday.

joe-ayoub-segment:
Thanks @wordythebyrd - CI reran and passed. I'll deploy today.

joe-ayoub-segment:
hi @wordythebyrd - quick correction - the deploy won't happen today as we're in a deploy freeze. The earliest will be this Wednesday, but more likely next Monday/Tuesday.

wordythebyrd:
> hi @wordythebyrd - quick correction - the deploy won't happen today as we're in a deploy freeze. The earliest will be this Wednesday, but more likely next Monday/Tuesday.

@joe-ayoub-segment  Ah, we were hoping to demo this internally on Wednesday. (~3pm)  Do you think we could make that work?  I assume its code freeze for end of quarter?

joe-ayoub-segment:
> > hi @wordythebyrd - quick correction - the deploy won't happen today as we're in a deploy freeze. The earliest will be this Wednesday, but more likely next Monday/Tuesday.
> 
> @joe-ayoub-segment Ah, we were hoping to demo this internally on Wednesday. (~3pm) Do you think we could make that work? I assume its code freeze for end of quarter?

Hi @wordythebyrd - I'm sorry, but unfortunately I can't deploy it this week due to a company wide deploy freeze (due to an event taking place in London).

wordythebyrd:
Hey @joe-ayoub-segment, hope the event went well!  Are you able to merge?

joe-ayoub-segment:
Hi @wordythebyrd the deploy has been delated again (it's a company wide thing). We're planning to deploy across tomorrow and Thursday at this point. Apologies for the delays.

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @wordythebyrd this PR has been deployed. 

Would you like to schedule some time to run through some test scenarios? 

http://calendly.com/joe_ayoub/integration-workshop

Best regards,
Joe

PR #1693, opened 2023-10-27
Update Attio action to link User<>Workspace and append values

nickcampbell18:
Thanks Segment team for all of your help so far. We've got a few customers using our action now, the biggest being Attio itself. Overall we're quite happy with the stability of this feature. There are two minor changes in this PR that we hope will complete our journey.

  1. **Update the handler for the `group` events so that it actually links Users and Workspaces together**. We always intended to do this but the designs of these objects were still in flux, it's ready now. With this change when calling the `groupWorkspace` action now (and having opted-in on the field) it will find an existing User and link it to the Workspace. This change is non-breaking (I believe!)

  2. **Pass an extra parameter in the API call to append values**. A subtle one this, the Attio API was overwriting one-to-many relationship attributes whenever updated, so you could only ever set a single User for a Workspace at a time. Very annoying. Our API has already been updated to support this new flag so it's safe to roll out at your leisure.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/attio/api/index.ts
+++ b/packages/destination-actions/src/destinations/attio/api/index.ts
@@ -30,7 +30,8 @@ export class AttioClient {
   }
 
   /**
-   * Either create or update a Record in the Attio system.
+   * Either create or update a Record in the Attio system. Multi-select attribute values
+   * are always appended, never replaced.
    *
    * @param matching_attribute The Attribute to match the Record on (e.g. an email address)
    * @param object The Attio Object (id / api_slug) that this Record should belong to (e.g. "people")
@@ -49,7 +50,7 @@ export class AttioClient {
     requestOptions?: Partial<RequestOptions>
   }): Promise<ModifiedResponse<AssertResponse>> {
     return await this.request(
-      `${this.api_url}/v2/objects/${object}/records/simple?matching_attribute=${matching_attribute}`,
+      `${this.api_url}/v2/objects/${object}/records/simple?matching_attribute=${matching_attribute}&append_to_existing_values=true`,
       {
         method: 'put',
         json: { data: { values } },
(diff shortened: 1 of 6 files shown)


packages/destination-actions/src/destinations/attio/api/index.ts
```diff
@@ -49,7 +50,7 @@ export class AttioClient {
     requestOptions?: Partial<RequestOptions>
   }): Promise<ModifiedResponse<AssertResponse>> {
     return await this.request(
-      `${this.api_url}/v2/objects/${object}/records/simple?matching_attribute=${matching_attribute}`,
+      `${this.api_url}/v2/objects/${object}/records/simple?matching_attribute=${matching_attribute}&append_to_existing_values=true`,
```
mulholo, line 53: @nickcampbell18, do we expect this behaviour change to affect existing users in any serious way?

joe-ayoub-segment (approved):

joe-ayoub-segment:
Thanks for the PR @nickcampbell18 - we'll deploy next Tuesday. Just need to check the CI checks pass.

packages/destination-actions/src/destinations/attio/api/index.ts
nickcampbell18: Only if they were depending on some overwrite behaviour... AFAIK none of our existing customers would be expecting that, since we're pushing them towards using the default configurations and those would normally be additive.

We'll reach out nonetheless when this is released to let them know :)

nickcampbell18:
Thanks @joe-ayoub-segment 🙏

joe-ayoub-segment:
Hi @nickcampbell18 this PR has been deployed. Can you confirm that you are happy with the change please?

nickcampbell18:
Thanks @joe-ayoub-segment - can confirm, this looks great and seems to be working well with our test customer. I've also just submitted some catalog information so that we can get to Public Beta hopefully soon!

PR #1708, opened 2023-11-03
Add additional field for updating only existing users if toggled to true 

mishranik:
_Added additional boolean field support. If it is toggled to true only existing users will. be updated ._

1. If the field is mapped to True, no new users sent from Segment will be created inside MoEngage. Only existing users will be updated.

2. if the field is mapped to False, new users sent from Segment will be created inside MoEngage. Existing users will be updated as well.

field_name: update_existing_only
## Testing


joe-ayoub-segment:
Thanks for the PR @mishranik , I'll schedule for review!

joe-ayoub-segment:
Hi @mishranik , 
What's the default behavior if this field resolves to 'undefined'? Will it cause an error, or will it default to true or false?

The reason I ask is that when we add a new field to an Integration, customers who already have the Integration configured won't see that new field. The field will then resolve to undefined. 

Can you clarify please?

mishranik:
@joe-ayoub-segment Hi 
 thanks for reviewing the PR.  I had one doubt regarding the new field.
Since I have added the default boolean value as false, I was expecting it will apply to all existing customers. Do you think it will work? Else let me know how to proceed here while adding new field.

joe-ayoub-segment:
> I had one doubt regarding the new field

Hi @mishranik , the field will have 3 possible states, depending on when the customer has configured the Integration: 

1. Customers who configured the Integration before we deploy this PR: field value will be 'undefined'
2 and 3: Customers who configure the Integration after we deploy this PR: field value will be either true or false

Do you need to update your code to handle scenario 1 above?

mishranik:
@joe-ayoub-segment Yes, I will have to possibly handle the 1st case, can you please share any reference, what would be the best way to handle this or any possible case, while introducing a new field. Thanks

joe-ayoub-segment:
Hi @mishranik , I just noticed that your PR just defines the update_existing_only field, but doesn't actually use it anywhere. 

There's 2 things you need to do: 

1. I assume you need to update the event payload you are sending to MoEngage to include info from the new update_existing_only field? I'm not sure where you will add this info in the payload though - I assume that you should know this? 

When you are adding the value you can check for undefined and correct as you need. 

    const event = {
      type: payload.type,
      user_id: payload.userId,
      traits: payload.traits,
      context: {
        app: { version: payload.app_version },
        os: { name: payload.os_name },
        library: { version: payload.library_version }
      },
      anonymous_id: payload.anonymousId,
      timestamp: payload.timestamp
    }

3. You will need to update the generated-types.ts file to include a reference to your new update_existing_only field. Don't edit this field manually. Instead run this command from the command line: 

./bin/run generate:types 

Best regards,
Joe

mishranik:
hi @joe-ayoub-segment I have added handling for the undefined case. As well as generated file for types as suggested by you.  Can you please check now. 

Also, I'm expecting i'm reading the update_existing_only field correctly with @path variable.

joe-ayoub-segment:
Hi @mishranik , it's getting there! 

It all looks good apart from the @path value. 

If you want the new field to take its value from the Segment event payload, then you will probably want the code to look like this: 

 ```
 update_existing_only: {
      label: 'Update existing users only',
      type: 'boolean',
      description: 'Setting this to true will not create new users in MoEngage. Only existing users will be updated',
      default: {
        '@path': '$.properties.update_existing_only'
      }
    }
```

However if you just want to field to default to true or false you can do this. The customer can always override it anyway.  

```
    update_existing_only: {
      label: 'Update existing users only',
      type: 'boolean',
      description: 'Setting this to true will not create new users in MoEngage. Only existing users will be updated',
      default: true
    }
```

OR 

```
    update_existing_only: {
      label: 'Update existing users only',
      type: 'boolean',
      description: 'Setting this to true will not create new users in MoEngage. Only existing users will be updated',
      default: false
    }
```

mishranik:
Hi @joe-ayoub-segment thanks for the explanation, i have updated the code where customer can choose if they want to have Update existing users only set as a true or false with false being default.

mishranik:
@joe-ayoub-segment  any idea by when we can merge this?

joe-ayoub-segment:
Looks good to me @mishranik - I'll rerun CI and see if it passes.

mishranik:
not sure what I can do to validate Synk steps @joe-ayoub-segment  rest of them seems to have passed

joe-ayoub-segment:
> not sure what I can do to validate Synk steps @joe-ayoub-segment rest of them seems to have passed

The sync check will always fail for externally raised PRs. So we're all good! I'll merge it before deploying it Wednesday next week :)

mishranik:
Hi @joe-ayoub-segment  I hope we can get this merged and release today.

mishranik:
hey @joe-ayoub-segment sorry for bumping this again, just wanted to check if we are good with deployment today?

joe-ayoub-segment:
Hi @mishranik yes I'm preparing the deploy now.

joe-ayoub-segment:
Hi @mishranik this PR has been deployed. Can you confirm that you are happy with it please?

mishranik:
Hi @joe-ayoub-segment regional endpoint mentioned in this PR is not visible need your confirmation if that is deployed or not yet.

PR #1709, opened 2023-11-03
Add Bucket Web Actions destination

Munter:
This adds a browser actions destination for the [Bucket tracking SDK](https://github.com/bucketco/bucket-tracking-sdk).

Bucket is targeting B2B SAAS companies with a specific focus on product management and engineering, and thus is only interested in tracking `identify`, `group` and `event` calls.

Further, the Bucket SDK will establish a live server push connection between the users bowser and the Bucket backend, which can trigger a satisfaction and feedback UI based on the users triggered events and the backend service deeming feedback prompting to be relevant.

Bucket already has a [direct cloud destination](https://segment.com/docs/connections/destinations/catalog/bucket/), but we'd like to ease the installation of the browser SDK further.

More information about Bucket:
- [Bucket website](https://bucket.co/)
- [Live Satisfaction](https://bucket.co/blog/introducing-live-satisfaction)

## Testing


joe-ayoub-segment:
Please note that we've paused deploys for this week due to an internal issue.

joe-ayoub-segment:
hi @Munter just following up on this to see if you are still interested in getting this PR deployed?

Munter:
@joe-ayoub-segment I am indeed interested in getting this deployed

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
Hi @Munter - my apologies it looks like I mustn't have actually submitted my review - it was pending for a long time.

packages/browser-destinations/destinations/bucket/src/group/index.ts
```diff
@@ -0,0 +1,61 @@
...
+          type: 'string',
+          required: false,
+          label: 'Name',
+          description: 'The company name',
+          default: {
+            '@path': '$.traits.name'
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
Munter: I think I'll fall back to just mapping `$.traits` directly without attempting to promote certain fields

Munter:
@joe-ayoub-segment No worries about the delay. It just so happens I was unable to work the past week anyway.

I've pushed changes that should adapt to all your review comments

packages/browser-destinations/destinations/bucket/src/group/index.ts
```diff
@@ -0,0 +1,48 @@
+import type { BrowserActionDefinition } from '@segment/browser-destination-runtime/types'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import type { Bucket } from '../types'
+
+const action: BrowserActionDefinition<Settings, Bucket, Payload> = {
+  title: 'Identify Company',
+  description: 'Creates or updates a Company in Bucket and associates the user with it',
+  platform: 'web',
+  defaultSubscription: 'type = "group" and userId != null',
```
joe-ayoub-segment, line 10:
```suggestion
  defaultSubscription: 'type = "group"',
```

packages/browser-destinations/destinations/bucket/src/group/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      default: {
+        '@path': '$.traits'
+      }
+    }
+  },
+  perform: (bucket, { payload }) => {
+    // Ensure we never call Bucket.company() without a user ID
+    if (payload.userId) {
+      void bucket.company(payload.groupId, payload.traits, payload.userId)
+    }
```
joe-ayoub-segment, line 44:
```suggestion
      void bucket.company(payload.groupId, payload.traits, payload.userId)
```

packages/browser-destinations/destinations/bucket/src/identifyUser/generated-types.ts
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * The User's full name
+     */
+    name?: string
+    /**
+     * The User's e-mail address
+     */
+    email?: string
+    [k: string]: unknown
+  }
```
joe-ayoub-segment, line 21: Hi @Munter can you run this command to update the generated-types.ts file please?

./bin/run generate:types

The traits field definition needs to be updated to match what you specified in code.

packages/browser-destinations/destinations/bucket/src/trackEvent/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      default: {
+        '@path': '$.properties'
+      }
+    }
+  },
+  perform: (bucket, { payload }) => {
+    // Ensure we never call Bucket.track() without a user ID
+    if (payload.userId) {
+      void bucket.track(payload.name, payload.properties, payload.userId)
+    }
```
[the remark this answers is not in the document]
joe-ayoub-segment: Map a Segment track() event to Bucket, so there's no need to check for the userId.

packages/browser-destinations/destinations/bucket/src/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    {
+      name: 'Identify User',
+      subscribe: 'type = "identify"',
+      partnerAction: 'identifyUser',
+      mapping: defaultValues(identifyUser.fields),
+      type: 'automatic'
+    },
+    {
+      name: 'Group',
+      subscribe: 'type = "group" and userId != null',
```
joe-ayoub-segment, line 33:
```suggestion
      subscribe: 'type = "group"',
```

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
hi @Munter - 
There's a few very minor but important things which need changing. I've included suggestions for them. 
If you can get this back to me today or tomorrow I can deploy this week - otherwise the next deploy is next week. 

Best regards,
Joe

packages/browser-destinations/destinations/bucket/src/trackEvent/index.ts
```diff
@@ -0,0 +1,48 @@
+import type { BrowserActionDefinition } from '@segment/browser-destination-runtime/types'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import type { Bucket } from '../types'
+
+const action: BrowserActionDefinition<Settings, Bucket, Payload> = {
+  title: 'Track Event',
+  description: 'Map a Segment track() event to Bucket',
+  platform: 'web',
+  defaultSubscription: 'type = "track" and userId != null',
```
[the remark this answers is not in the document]
Munter: I couldn't actually find a way to verify this behavior with the actions tester. In the actions tester, if I just subscribe to `type = "track"` and I have `userId: null`, it says the event matches:
image

If I use the actions tester UI to add the `userId != null` condition, that same event no longer matches, which is what I intended:
image

I don't recall it being a requirement that I have previously called `analytis.identify()` before calling `analytics.track()`, so I don't see how a `userId` can be guaranteed on a track call.

packages/browser-destinations/destinations/bucket/src/group/index.ts
Munter: I just ran a live test where I called `analytics.load('...'); analytics.group('test-group-id')` resulting in a group request to segment with the following payload:

```json
{
  "timestamp": "2023-11-27T21:02:49.076Z",
  "anonymousId": "9471e694-5e3e-4e8f-b401-60e3ecc6465c",
  "type": "group",
  "traits": {},
  "groupId": "peter-testing",
  "context": {
  },
  "messageId": "ajs-next-c42d4b094f81e9e15f8f286bf4ca0704",
  "writeKey": "...",
  "userId": null,
}
```

It's fully possible to create a group call with a `null` `userId`, and Bucket can't do anything with them, so this is the only way I could see I could not subscribe to them.

Is there something I'm missing?

packages/browser-destinations/destinations/bucket/src/trackEvent/index.ts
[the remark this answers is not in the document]
Munter: I also ran a live test for calling `analytics.track()` as an anonymous user with the same result as in https://github.com/segmentio/action-destinations/pull/1709#discussion_r1406749877 - where I get `userId: null`

Bucket is not interested in anonymous user events, so this seems to be the way to filter them out

packages/browser-destinations/destinations/bucket/src/group/index.ts
Munter: I haven't figured out how to run the tests in a way where I'm actually running the full Bucket destination and call the full lifecycle of the analytics client to perform the events that I need to verify in my destination. It's all mocked and called directly, so I needed this switch in the code in order to test correct behaviour when `userId: null`.

I'd love to get some guidance on how to test in a way so the subscription FQL is taken into account and `perform()` is never called if the subscription doesn't match.

Munter:
Hi @joe-ayoub-segment 

I've generates the types now.

Some of your comments don't seem to match with the behaviour I'm observing in the production version of analytics.js when I'm testing it locally. I'm unable to verify that `userId` is actually a required property and a guaranteed string for the `analytics.group()` and `analytics.track()` calls, as I'm observing the opposite.

I'd love some help to modify my tests to actually run a full analytics instance with all the lifecycle methods called directly on it, so I can verify correct behaviour based on the analytics client evaluating the subscription FQL per event. Currently I'm using a test pattern I found in other browser destinations, which is the only thing I could get to work. But this requires double book keeping and an extra guard in the actions `perform()` methods

joe-ayoub-segment:
Hi @Munter thanks for regenerating the types file. That was the most important thing. 

My apologies I should have been more specific: If you specifically want to prevent a null valued userId, you can do it like this in the userId field definition (note the allowNull:false attribute).  

     userId: {
      type: 'string',
      allowNull: false,
      required: true,
      description: 'Unique identifier for the user',
      label: 'User ID',
      default: {
        '@path': '$.userId'
      }
    }

Doing this will mean that only events containing a userId (which is not null) will be passed into the perform() function. All other events will be dropped. 
When you do this it will be safe to remove all the other userId != null checks.

Munter:
> Doing this will mean that only events containing a userId (which is not null) will be passed into the perform() function. All other events will be dropped.

I'll make that change immediately. I'll see if that changes anything in the perform function itself, since I still haven't found a way to test everything through an analytics instance, but instead am calling it directly

Munter:
@joe-ayoub-segment I've pushed to updates that mode the `userId` requirement from the raw FQL into the field definition with `allowNull: false`.

I think the only thing left to address is the extra `payload.userId` checks in the `trackEvent` and `group` actions. I don't have a good way of removing them unless I'm introduced to a better way of testing through a fully initialized analytics instance with the bucket web destination loaded. I haven't been able to create that in tests yet without a lot o stuff breaking

joe-ayoub-segment (approved):

Munter:
The CI did another build and failed again. None of the errors look like something I could have introduced:

```
[3/5] Fetching packages...
error Command failed.
Exit code: 128
Command: git
Arguments: ls-remote --tags --heads ssh://git@github.com/segmentio/control-plane-service-js-client.git
Directory: /home/runner/work/action-destinations/action-destinations
Output:
fatal: could not read Username for 'https://github.com/': terminal prompts disabled
error https://registry.yarnpkg.com/@segment/remote-service-client/-/remote-service-client-2.4.1.tgz: Request failed "404 Not Found"
```

```
`snyk` requires an authenticated account. Please run `snyk auth` and try again.
```

I am unsure if there is anything I can do right now to move this PR along

joe-ayoub-segment:
hi @Munter 
You can ignore the snyk failure. However we need to get the other checks to pass. 

I'll rerun the CI checks now.

joe-ayoub-segment:
hmm they all failed really quickly. 
@Munter could you please get all the latest files from the main branch please?

PR #1710, opened 2023-11-05
Yahoo: removed Identifier setting and mapping

longstoryshort:
Remover Identifier audience setting and action mapping. We inherit identifiers from the Configurable ID sync: IDs mapped in the Configurable ID sync setting will be sent to the destination.


joe-ayoub-segment:
Hi @longstoryshort this PR has been deployed.

joe-ayoub-segment:
DB migration still to happen. Will notify you .

PR #1711, opened 2023-11-06
Complete Trackey integration

clickoutio:
Hi, last time you merged my pull request but there were some changes left.
I have moved each action to its seperate file and fixed some issue about the connection with our backend.

## Testing
I have included unit tests for each of the separate integration

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

Hope everything is ok

Best regards,
Javier

PR #1720, opened 2023-11-14
Changes to trackey added by Partner

joe-ayoub-segment:
Trackey is not yet in use by customers. 

Partner has refactored out the Actions into separate folders so that tests could be created. 

The original PR is [here](https://github.com/segmentio/action-destinations/pull/1711), however I created this PR to ensure we keep the Action folder names the same so that we can avoid having to do a DB migration. 

## Testing

Purpose of this PR was to refactor and add tests.


sayan-das-in (approved):

joe-ayoub-segment:
Hi @clickoutio this PR has been deployed. Can you confirm that it looks good please?

PR #1728, opened 2023-11-16
Algolia Insightes - adding 2 new fields

joe-ayoub-segment:
Update allows customers to select the name of the event and type of event to send to Algolia for each Action. 

Default values will remain in place. 

## Testing

Tested with Actions Tester locally

### DIFF
--- a/packages/destination-actions/src/destinations/algolia-insights/__tests__/__snapshots__/snapshot.test.ts.snap
+++ b/packages/destination-actions/src/destinations/algolia-insights/__tests__/__snapshots__/snapshot.test.ts.snap
@@ -4,8 +4,8 @@ exports[`Testing snapshot for actions-algolia-insights destination: conversionEv
 Object {
   "events": Array [
     Object {
-      "eventName": "Conversion Event",
-      "eventType": "conversion",
+      "eventName": "U[ABpE$k",
+      "eventType": "view",
       "index": "U[ABpE$k",
       "objectIDs": Array [
         "U[ABpE$k",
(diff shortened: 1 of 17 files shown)


joe-ayoub-segment:
hi @wwalser , 

Following up on my email from a week or so ago: 

image

I raised a PR on the Algolia Integration - and would like to check with you if it makes sense to merge and deploy it. 

Here's a link to a [video](https://drive.google.com/file/d/1HibmaiIc9oqtEmzhmHE7C5VyJ_6IaPWw/view?usp=sharing) explaining the change. 

I'm unsure if this change is a good idea or not - it was requested by a customer, but it looks like you were pretty deliberate when you coded the integration so that only specified values would be sent for Event Name and Event Type. 

Keen to hear what you think...

wwalser:
Hi Joe, I'm away from work for some life events for a few more weeks. I should be able to look at this once I return.

joe-ayoub-segment:
Hi @wwalser @JasonBerry this PR has been deployed.

wwalser:
Sry I was away on pat leave while this was being considered.

I don't think it's a huge deal given the small number of users, but I believe all existing users of this integration now have this error state after this PR landed:
Screenshot 2024-01-10 at 1 20 16 PM

It seem that adding a new, `required` field causes the destination to be placed in an error state even if that field was added with a `default` value. That default value is not added to existing configurations. 

This could be unintended at the Action Destinations platform level?

joe-ayoub-segment:
Hi @wwalser thanks for pointing this out. This shouldn't have happened. We'll deploy a fix today.

wwalser:
To clarify, do you plan to roll back this change or fix this at the platform level?

joe-ayoub-segment:
Hi @wwalser I updated the code to make the fields optional - this was the original intent. 
A default value gets applied if the field if no value is passed into the field. 

The code was deployed yesterday so the issue should now be resolved. 

I would like to reach out to customers about this though. Would you be able to share any details of who contacted you about this issue please (perhaps over email)?

PR #1729, opened 2023-11-16
Snap - adding default paths for click id and cookie

joe-ayoub-segment:
Adding default paths for the cookie and click id fields. 
These will pick up values from client side events fired from a browser. 

## Testing
- Tested in the staging environment and local tester tool 

image


varadarajan-tw (approved):

joe-ayoub-segment:
Note: PR deployed.

PR #1731, opened 2023-11-17
Updated Criteo API version (2023-01 -> 2023-10)

MarcDillar:
## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/criteo-audiences/criteo-audiences.ts
+++ b/packages/destination-actions/src/destinations/criteo-audiences/criteo-audiences.ts
@@ -2,7 +2,7 @@ import { createHash } from 'crypto'
 import { IntegrationError, RetryableError } from '@segment/actions-core'
 import type { RequestClient } from '@segment/actions-core'
 
-const BASE_API_URL = 'https://api.criteo.com/2023-01'
+const BASE_API_URL = 'https://api.criteo.com/2023-10'
 
 export const hash = (value: string | undefined): string | undefined => {
   if (value === undefined) return
@@ -23,7 +23,7 @@ class CriteoAPIError extends IntegrationError {
 
 export type Operation = {
   operation_type: string
-  audience_id: string
+  contactlist_id: string
   user_list: string[]
 }
 
@@ -78,15 +78,15 @@ export const criteoAuthenticate = async (
   return credentials
 }
 
-export const patchAudience = async (
+export const patchContactList = async (
   request: RequestClient,
   operation: Operation,
   credentials: ClientCredentials
 ): Promise<Response> => {
-  if (isNaN(+operation.audience_id))
-    throw new IntegrationError(`The Audience ID should be a number (${operation.audience_id})`, 'Invalid input', 400)
+  if (isNaN(+operation.contactlist_id))
+    throw new IntegrationError(`The Audience Segment ID should be a number (${operation.contactlist_id})`, 'Invalid input', 400)
 
-  const endpoint = `${BASE_API_URL}/audiences/${operation.audience_id}/contactlist`
+  const endpoint = `${BASE_API_URL}/marketing-solutions/audience-segments/${operation.contactlist_id}/contact-list`
   const headers = await getRequestHeaders(request, credentials)
   const payload = {
     data: {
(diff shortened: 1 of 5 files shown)


joe-ayoub-segment:
Thanks for the PR @MarcDillar - I'll schedule for review.

joe-ayoub-segment:
hi @MarcDillar thanks again for raising this PR. It looks good from my perspective. 

The main risk with a change like this is that we break data collection for customers who are already using the Integration. 

We don't have a capbility to deploy code dark, and it's not possible to do side by side testing either. So we're going to have to rely on some other mechanisms.  

I'm going to run an internal test to run some live traffic through it to see if any internal errors generated are the same as with the previous code. 

We could deploy the code to an internal Staging environment, and if you were free to jump on a call we could connect the Integration and sync an Audience to Criteo (which you could then verify looked good). Would you be open to that?

Best regards,
Joe

joe-ayoub-segment:
note: meeting with Marc Dillar Monday

joe-ayoub-segment:
Thanks for meeting earlier today @MarcDillar . 
Shall we schedule another call to test again later this week?
Here's my Calendly link: http://calendly.com/joe_ayoub/integration-workshop

joe-ayoub-segment:
Hi @MarcDillar - looks like the CI checks are failing. 
image

joe-ayoub-segment:
You can run the tests with this command: 

yarn jest --testPathPattern='./packages/destination-actions/src/destinations/criteo-audiences'

MarcDillar:
Hi @joe-ayoub-segment 

Thank you, the unit tests should now work correctly:
image

joe-ayoub-segment:
hi @MarcDillar  - in the end I decided to not deploy this PR until I am fully able to test it end to end locally without any potential issues. I raised a bug on our testing tool as I suspect there is an issue with it. Once this is done I'll retest and deploy. It will likely be in Jan 2024. 

Do we have enough time for this, or will the old API version stop working soon?

MarcDillar:
Hi @joe-ayoub-segment, no worries, the endpoints that we had to replace are present until version 2023-07, which is supported until July 2024 so we still have a bit of time to make that change.

joe-ayoub-segment:
Hi @MarcDillar , 

Happy New Year!
I'd like to get this PR deployed as soon as possible. I'm going to try doing a retest now - but just noticed that the linter is complaining about some 'var' definitions. I tried to fix myself but I don't have permission to push to your repo.  

image

MarcDillar:
Hi @joe-ayoub-segment 

Happy new year to you too! I just pushed a commit that should solve the linting issue.

joe-ayoub-segment (changes requested):

packages/destination-actions/src/destinations/criteo-audiences/criteo-audiences.ts
```diff
@@ -105,104 +105,124 @@ export const patchAudience = async (
...
+          advertiser_id
+        ]
+      }
+    }
+  }
 
-  const body = await response.json()
+  let continue_search = true
+  let offset = 0
+  do {
```
joe-ayoub-segment, line 134: Finally discovered the bug which was causing requests to hang. The response sizes from the request to get the contactlist ID are too large. Approx 30kb. So Segment is hanging.

We can override this limit by adding the 'skipResponseCloning: true' attribute to the Request.

Also, we don't need to await response.json(). Instead we can do as below:

  interface AudienceSegment {
    attributes: {
      [key: string]: unknown
    }
    id: string
    type: string
  }

  interface ApiResponse {
    data: AudienceSegment[]
    meta: {
      totalItems: number
    }
  }

  let body: ApiResponse

  do {
    const endpoint = `${BASE_API_URL}/marketing-solutions/audience-segments/search?limit=${LIMIT}&offset=${offset}`

    const response = await request(endpoint, {
      method: 'POST',
      skipResponseCloning: true,
      headers: headers,
      json: payload
    })

    body = response.data as ApiResponse

    if (response.status !== 200)
      // Centrifuge will automatically retry the batch if there's
      // an issue fetching the Advertiser's audiences from Criteo.
      throw new RetryableError("Error while fetching the Advertiser's audiences")

    // If the contact list is found, return the corresponding ID
    for (const contactlist of body.data) {
      if (contactlist.attributes.name === audience_segment_name) return contactlist.id
    }

    // Else, continue searching
    offset += LIMIT
    continue_search = body.meta.totalItems > offset
  } while (continue_search)
}

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
hi @MarcDillar I did some more testing and found the issue which was causing the code to periodically hang.
I'm glad we didn't deploy it because it would have caused an incident :). 

Please see my comment above which should resolve the issue. Can you please update the PR accordingly?

While testing the PR I reviewed the code in more detail and noticed that there are a 3 potential failure points with the Integration code. 

1. We are fetching the contactlist ID with every invocation, which is inefficient. We do have a mechanism to cache these responses, but it will involve a more detailed change. 
2. The access_token is requested with each invocation, which is also inefficient. We do have a solution where we can cache the access_token as well, but that too will require a more in-depth change to the Integration code. 
3. Finally, I believe that the using a do..while loop to fetch the contactlist ID is inefficient. Each response can be large, and multiple requests may need to be made before the correct contactlist ID is retrieved. Would it not be possible to update the Criteo API to accept a contactlist Name and return a contactlist ID in a single request? I think we discussed this previously and you said that it's not possible yet. 

I look forward to hearing back from you. 
Kind regards,
Joe

MarcDillar:
Hi @joe-ayoub-segment thank you for your review, I just made a commit following your recommendations.

joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @MarcDillar this PR has been deployed. 

Could you please confirm that the change looks good from your side please? Is there any dashboard data you can check to verify this?

Best regards,
Joe

PR #1739, opened 2023-11-23
STRATCONN-3283 Added default value for cookiePath

AnkitSegment:
Added Validation for cookiePath, Now no cookie_path key is passed unless user indicates to override a default value. 

Screenshot 2023-12-14 at 1 49 39 PM

## Testing


jsabor:
@AnkitSegment I don't think this is intended behavior.  In the same vein as all config parameters, there are default values that do not have to be configured.  The `cookie_path` will default to `/` if not indicated.  Also what I'm seeing in your screenshot is there are two session ids , see cookie `_ga_{{G-MEASUREMENT_ID}}` , this will break the session in GA4 and I have seen this is with several clients, although, I thought this was from their application of a.js not being executed from the top level domain.

PR #1749, opened 2023-11-30
Remove import of lodash

LLuque-twilio:
This import was causing certain web destinations to crash due to overriding their local `_` variable

## Testing

Testing not needed. Implemented an equal check without the use of Lodash.

### DIFF
--- a/packages/core/src/mapping-kit/value-keys.ts
+++ b/packages/core/src/mapping-kit/value-keys.ts
@@ -1,8 +1,8 @@
-// import { isDirective } from './is-directive'
-// eslint-disable-next-line lodash/import-scope
-import _ from 'lodash'
-
 type ValueType = 'enrichment' | 'function' | 'literal' | 'variable'
+
+function isObject(value: any): value is object {
+  return value !== null && typeof value === 'object'
+}
 
 export interface DirectiveMetadata {
   _metadata?: {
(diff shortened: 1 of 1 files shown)


chrisradek (approved):

packages/core/src/mapping-kit/value-keys.ts
```diff
@@ -1,9 +1,9 @@
-// import { isDirective } from './is-directive'
-// eslint-disable-next-line lodash/import-scope
-import _ from 'lodash'
-
 type ValueType = 'enrichment' | 'function' | 'literal' | 'variable'
 
+function isObject(value: any): value is object {
```
mericsson, line 3: It looks like there are other `isObject` in this code-base. Should we re-use those?

mericsson: I guess this one is different? https://github.com/segmentio/action-destinations/blob/c9dad648d1f5fb1bca1ab98e00858c24fb45291e/packages/core/src/real-type-of.ts#L25

LLuque-twilio: I saw that. But that one is specifically for an object `{}` and does not include arrays `[]` which we need

dlasky (approved):

packages/core/src/mapping-kit/value-keys.ts
mericsson: Should we add a unit test to confirm that expectation and make it clear?

LLuque-twilio: I guess another way would just be re-using both the already-existing internal methods `isArray` and `isObject` like `isObject(value) || isArray(value)`. Don't have a preference on this.

PR #1751, opened 2023-12-01
fix(google-analytics-4-web): scope user updates to the relevant measurement ID

backjo:

This PR fixes what I believe to be a bug in the GA4 Web Destination. The GA4 Web Destination in Segment requires configuration of a Measurement ID, implying that it is scoped to a single measurement. Currently, however, changes to the user_id and user_properties are 'leaking' into other measurements that are not being managed by Segment. Per the [gTag docs](https://developers.google.com/tag-platform/gtagjs/reference#parameter_scope), the 'set' operation sets for all measurements. To properly scope the user_id and user_properties, this PR updates the problematic usage of 'set' to use 'config' and passed in the relevant measurement ID to ensure that it is scoped. 

## Testing

### DIFF
--- a/packages/browser-destinations/destinations/google-analytics-4-web/src/addPaymentInfo/index.ts
+++ b/packages/browser-destinations/destinations/google-analytics-4-web/src/addPaymentInfo/index.ts
@@ -32,8 +32,8 @@ const action: BrowserActionDefinition<Settings, Function, Payload> = {
     user_properties: user_properties,
     params: params
   },
-  perform: (gtag, { payload }) => {
-    updateUser(payload.user_id, payload.user_properties, gtag)
+  perform: (gtag, { payload, settings }) => {
+    updateUser(payload.user_id, payload.user_properties, gtag, settings.measurementID)
     gtag('event', 'add_payment_info', {
       currency: payload.currency,
       value: payload.value,
--- a/packages/browser-destinations/destinations/google-analytics-4-web/src/addToCart/index.ts
+++ b/packages/browser-destinations/destinations/google-analytics-4-web/src/addToCart/index.ts
@@ -21,8 +21,8 @@ const action: BrowserActionDefinition<Settings, Function, Payload> = {
     user_properties: user_properties,
     params: params
   },
-  perform: (gtag, { payload }) => {
-    updateUser(payload.user_id, payload.user_properties, gtag)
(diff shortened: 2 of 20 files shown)


dossy:
> The GA4 Web Destination in Segment requires configuration of a Measurement ID, implying that it is scoped to a single measurement.

It needs the Measurement ID because the destination is loading Gtag.js for the measurement ID on behalf of the user when the Destination is added.  I didn't think that necessarily implied that events would be scoped to that single measurement ID, but I can definitely see that as a reasonable interpretation.  I think the documentation needs to be clear about this and state explicitly what the user can expect.

I think a setting that controls whether properties should use `gtag('set', ...)` or `gtag('config', 'TAG_ID', ...)` would help here, 
 as it's useful to be able to do both, depending on the user's requirements.

backjo:
> > The GA4 Web Destination in Segment requires configuration of a Measurement ID, implying that it is scoped to a single measurement.
> 
> It needs the Measurement ID because the destination is loading Gtag.js for the measurement ID on behalf of the user when the Destination is added. I didn't think that necessarily implied that events would be scoped to that single measurement ID, but I can definitely see that as a reasonable interpretation. I think the documentation needs to be clear about this and state explicitly what the user can expect.
> 
> I think a setting that controls whether properties should use `gtag('set', ...)` or `gtag('config', 'TAG_ID', ...)` would help here, as it's useful to be able to do either, depending on the user's requirements.

Agreed on a setting make sense here! I imagine that even if this wasn't the intended original behavior, some folks are probably relying on it in production for their other tags!

Would it make sense for me to propose that in this PR? Happy to do it - but given the amount of time the PR has been open before this chat, I'd want to make sure that there was a chance of these changes being reviewed and being able to land :)

dossy:
> [...] given the amount of time the PR has been open before this chat, I'd want to make sure that there was a chance of these changes being reviewed and being able to land :)

I don't know how the Segment team decides what contributions they will accept and what contributions they let sit, so it's up to you whether you want to invest the time.

I just submitted PR #1806 which fixes the "Page Views" setting and make it actually work, and while I have a work-around in place as I need that functionality right now, I have no idea if and when it'll get merged for others to benefit, but I at least submitted the one-byte change :)

backjo:
fixed by #1817

PR #1803, opened 2024-01-12
[STRATCONN] [GA4] Added send_to key in all events

AnkitSegment:
In this ticket added send_to key in all events of GA4. Now events are routed to specific group.

JIRA TICKET: https://segment.atlassian.net/browse/STRATCONN-3325

Screenshot 2024-01-08 at 10 28 20 PM

## Testing


varadarajan-tw (commented):
LGTM! Should the tests be updated for this change?

AnkitSegment:
> LGTM! Should the tests be updated for this change?

yes, this changes are deployed on stage and tested.

dossy:
Before this PR, events are sent to the `default` group which is the behavior when `send_to` is not specified.

https://developers.google.com/tag-platform/gtagjs/routing#default_group

This PR breaks this, and makes it impossible for someone using the GA4 Web destination to do so, as `send_to: settings.measurementID` is now hard-coded into the event params.

It would be better if someone who wants to use an explicit `send_to` event parameter for a mapped event to specify it in their `payload.params` and not force this on everyone.

jsabor:
Yes, the config does set the measurement id for all events; however, if there are multiple configs, or (1) config with multiple tag ids, if you don't limit the event with the `send_to` parameter, the event will broadcast to all tag ids.  Specifically, there is a issue where some GA4 event would broadcast to GAds since they share similar events, and it would cause (and has caused) a conflict.  If you look at how Google handles this in GTM, all events sent via a GTM GA4 Event tag include a send_to param so it sends specifically to the GA4 measurement id indicated in the tag.

Backwards compatibility will be updated, as well as documentation.

packages/browser-destinations/destinations/google-analytics-4-web/src/customEvent/index.ts
```diff
@@ -41,11 +41,11 @@ const action: BrowserActionDefinition<Settings, Function, Payload> = {
     user_properties: user_properties,
     params: params
   },
-  perform: (gtag, { payload }) => {
+  perform: (gtag, { payload, settings }) => {
     updateUser(payload.user_id, payload.user_properties, gtag)
     const event_name = normalizeEventName(payload.name, payload.lowercase)
 
-    gtag('event', event_name, payload.params)
+    gtag('event', event_name, { params: payload.params, send_to: settings.measurementID })
```
dossy, line 48: I think this needs to be:

```
gtag('event', event_name, {
  send_to: settings.measurementID,
  ...payload.params
})
```

dossy (changes requested):

dossy:
> Backwards compatibility will be updated, as well as documentation.

Okay.  This is definitely going to be a big gotcha for anyone who's currently using the GA4 Web destination.  Is there a way to version the destinations and pin your Segment configuration to a particular version?

packages/browser-destinations/destinations/google-analytics-4-web/src/customEvent/index.ts
dossy: @varadarajan-tw what do you think?

jsabor:
The changes do support backwards capability.  If you want to run the event like it has in the past, it is equivalent to `send_to: default`  If the the user wants to add guardrails to the event, the user has to make the selection of `Yes` and that gives the output of `send_to: G-xxxx`

jsabor:
Backwards capability is applied to the update.  If you have a mapping created, there will be no selection and it will output `send_to: default` .  If you create a new mapping, it will pre-select yes, but the user can flip to no.

packages/browser-destinations/destinations/google-analytics-4-web/src/ga4-properties.ts
```diff
@@ -367,3 +367,9 @@ export const items_multi_products: InputField = {
     ]
   }
 }
+export const send_to: InputField = {
+  label: 'Send To',
+  type: 'boolean',
+  default: true,
+  description: 'If a send_to parameter is not set, events are routed to the default target group.'
```
dossy, line 374: This is not true.  If the `send_to` parameter is not set, events are routed to `settings.measurementID`, not the `default` target group:

```
send_to: payload.send_to == undefined || payload.send_to == true ? settings.measurementID : 'default',
```

In order to not break backwards-compatibility, if `send_to` is not set, events SHOULD be routed to the `default` target group, which is what the current behavior is before this change.  But, this PR does not implement this, as of this moment.

dossy: @jsabor FYI.

jsabor: The condition line was changed yesterday, the latest commit has the changes.   

The default to `true` will happen for net new mappings.  Previous mappings will not have a value mapped and result in `send_to:default`

PR #1806, opened 2024-01-13
Fix GA4 Web destination's "Page Views" setting

dossy:
The Segment GA4 Web destination is a device-mode destination that loads the Google Gtag.js, which [sends page views by default][1].  In order to stop it from sending page views, one must do:

```javascript
  gtag('config', 'TAG_ID', {
    send_page_view: false
  });
```

[1]: https://developers.google.com/analytics/devguides/collection/ga4/views?client_type=gtag#default-behavior

The Segment GA4 Web destination implementation will set `config.send_page_view = true` when `settings.pageView == true`, but when `settings.pageView == false`, then the `if (settings.pageView)` means `config.send_page_view` will never be set to `false`.

Changing the guard to `if (!settings.pageView)` fixes this, so that `config.send_page_view` will be set to `false` if `settings.pageView` is false, which is the desired behavior.


brennan:
This issue was addressed in https://github.com/segmentio/action-destinations/pull/1922

PR #1820, opened 2024-01-22
[stratconn-2993] [Braze] Adds email as a user identification field

nick-Ag:
This PR makes a minor update to the Braze destination. An email field is added and it is added to the list of acceptable IDs users can pass for identification. 

Ticket: https://segment.atlassian.net/browse/STRATCONN-2993

## Testing

Tested successfully in staging. Passing the `email` field correctly, Braze accepts it as the only user ID field required.
Screenshot 2024-01-22 at 3 43 25 PM


packages/destination-actions/src/destinations/braze/trackPurchase/__tests__/__snapshots__/snapshot.test.ts.snap
```diff
@@ -49,6 +50,7 @@ Object {
       "app_id": "my-app-id",
       "braze_id": undefined,
       "currency": "USD",
+      "email": undefined,
```
marinhero, line 53: Why do you add the key at all if it will be undefined? can the API handle it gracefully or will it create new users given that you'll be adding a new key per every existing record?

Before:

User = { brazeId: 42 }

After:

User = { brazeId: 42, email: undefined }

marinhero (commented):
I have a question that I'm hoping you can clarify.

packages/destination-actions/src/destinations/braze/trackPurchase/__tests__/__snapshots__/snapshot.test.ts.snap
nick-Ag: Hey, I just tested it out and I think this is something to do with the snapshot generation. I have a default mapping for `email`, but in the corresponding unit test there is nothing at that mapping so it evaluates to `undefined`. Actions will filter that out when it sends requests. I just tested that out in staging and verified we aren't sending `email: undefined` downstream to Braze.

marinhero: Okay. Thanks for checking. Is there anywhere in the code where you could save this piece of knowledge in a comment?

nick-Ag: Ok added that comment in the unit test file 👍

marinhero (approved):

joe-ayoub-segment:
Hi @nick-Ag this PR has been deployed. I did not push it yet as the diff looked suspicious. Please see my Slack messages

PR #1830, opened 2024-01-24
Fix SMS/MMS and email with <nil> traits to display default

miguelpdiaz8:
Fixes Broadcasts returning NIL for merge tags in SMS/MMS messages when there is no value for the trait. In order to compensate for the compute service sending “<nil>” for non existent profile traits, we have to exclude that value from profile traits, replacing it with null before parsing liquid js.

## Testing

- [x ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality

### DIFF
--- a/packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
+++ b/packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
@@ -90,15 +90,14 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload> 
     },
     contentType: string
   ) {
-    for (const trait of Object.keys(liquidData.profile.traits || {})) {
-      if (
-        liquidData.profile.traits &&
-        (liquidData.profile.traits[trait] === '<nil>' || liquidData.profile.traits[trait].trim() === '')
-      ) {
-        liquidData.profile.traits[trait] = ''
+    const profileCopy = { ...liquidData.profile }
+    for (const trait of Object.keys(profileCopy.traits || {})) {
+      if (profileCopy.traits && (profileCopy.traits[trait] === '<nil>' || profileCopy.traits[trait].trim() === '')) {
+        profileCopy.traits[trait] = ''
       }
     }
-    const parsedContent = content == null ? content : await Liquid.parseAndRender(content, liquidData)
+    const parsedContent =
+      content == null ? content : await Liquid.parseAndRender(content, { ...liquidData, profile: profileCopy })
     this.logOnError(() => 'Content type: ' + contentType)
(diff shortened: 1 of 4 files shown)


alfrimpong:
Can you please deploy and test this on staging?
Also, lets test prod to make sure no weird behaviors were passed through by this PR's predecessor

miguelpdiaz8:
> Can you please deploy and test this on staging? Also, lets test prod to make sure no weird behaviors were passed through by this PR's predecessor

I already tested Prod and it is behaving the same way than before the previous fix, it only works different if the content is '<nil>' where we return '<nil>' directly but that is fine since '<nil>' content should return '<nil>' anyway since it does not have traits

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
```diff
@@ -90,10 +90,15 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload>
...
-    const parsedContent =
-      content == null || content === '<nil>' || content.trim() === ''
-        ? content
-        : await Liquid.parseAndRender(content, liquidData)
+    for (const trait of Object.keys(liquidData.profile.traits || {})) {
+      if (
+        liquidData.profile.traits &&
+        (liquidData.profile.traits[trait] === '<nil>' || liquidData.profile.traits[trait].trim() === '')
+      ) {
+        liquidData.profile.traits[trait] = ''
```
alfrimpong, line 98: Can we keep this function pure? I believe this line and the one in TwilioMessageSender:43 both mutate the original object. We should avoid that to keep the functions pure

joe-ayoub-segment:
hi @alfrimpong and @miguelpdiaz8 - is this PR ready for me to review and then deploy?

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
miguelpdiaz8: Done

miguelpdiaz8:
> hi @alfrimpong and @miguelpdiaz8 - is this PR ready for me to review and then deploy?

Hi @joe-ayoub-segment yes this is ready for review and then deploy, thanks

alfrimpong (approved):
Looks good to me. Please ensure this is tested on stage.

miguelpdiaz8:
> Hi @miguelpdiaz8 I'm happy to deploy once testing on Stage has been confirmed!

@joe-ayoub-segment you can deploy now, thank you!

joe-ayoub-segment (approved):

PR #1842, opened 2024-01-30
Add Window object change unit test

danieljackins:
This PR adds a unit test that loads a web destination that doesn't alter the Window object and then checks for Window changes to confirm that our actions core isn't making unintended changes to Window

## Testing

### DIFF
--- a/packages/browser-destinations/destinations/__tests__/indext.test.ts
+++ b/packages/browser-destinations/destinations/__tests__/indext.test.ts
@@ -1,0 +1,29 @@
+import { Analytics, Context } from '@segment/analytics-next'
+import segmentUtilitiesDestination from '../segment-utilities-web'
+
+it('window object shouldnt be changed by actions core', async () => {
+  const windowBefore = Object.keys(window)
+
+  // load a plugin that doesn't alter window object
+  const [plugin] = await segmentUtilitiesDestination({
+    throttleWindow: 3000,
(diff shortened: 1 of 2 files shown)


packages/browser-destinations/destinations/__tests__/indext.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    ]
+  })
+
+  await plugin.load(Context.system(), {} as Analytics)
+
+  const windowAfter = Object.keys(window)
+  const diff = windowAfter.filter((element) => !windowBefore.includes(element))
+
+  // window object shouldn't change as long as actions-core isn't changing it
+  expect(diff.length).toBe(0)
```
silesky, line 28: Would it be better to use an intersection so we know what keys actually changed?

i.e 
```ts
expect(diff.sort()).toEqual(windowBefore.sort())
```

silesky (approved):

packages/browser-destinations/destinations/__tests__/indext.test.ts
```diff
@@ -0,0 +1,28 @@
+import { Analytics, Context } from '@segment/analytics-next'
```
silesky, line 1: This file is named indext.test -- spelling

silesky (changes requested):
file misspelled

silesky (approved):

PR #1863, opened 2024-02-07
DV360 - Support no-auth endpoint for legacy destinations

marinhero:
Legacy destinations don't have an oauth object until customers authenticate into the new destination. However, the bulkUploader API does not require an OAuth object to function; it can continue to relay data by verifying ownership through the externalId. Only legacy destinations will have an externalId and no auth object. While we preserve the security benefits of OAuth, we also ensure that current users can maintain uninterrupted data flow without needing to take any action.

## Testing


joe-ayoub-segment:
hi @marinhero PR deployed!

PR #1878, opened 2024-02-13
feat(Jimo/Actions): sendTrackEvent

imsamdez:
Adding an action for track events

## Testing

Nothing too hard or specific, we basically subscribe to track events and call a method of our SDK.
That said, I saw that we were not able to access the `receivedAt` property of the event payload. Not sure why because in the testing tool, we do see the mapping happening, yet, the field is undefined.

### DIFF
--- a/packages/browser-destinations/destinations/jimo/src/sendTrackEvent/index.ts
+++ b/packages/browser-destinations/destinations/jimo/src/sendTrackEvent/index.ts
@@ -1,0 +1,96 @@
+import type { BrowserActionDefinition } from '@segment/browser-destination-runtime/types'
+import { JimoSDK } from 'src/types'
+import type { Settings } from '../generated-types'
+import { Payload } from './generated-types'
+
+const action: BrowserActionDefinition<Settings, JimoSDK, Payload> = {
+  title: 'Send Track Event',
+  description: 'Submit an event to Jimo',
+  defaultSubscription: 'type = "track"',
+  platform: 'web',
+  fields: {
+    messageId: {
+      description: 'The internal id of the message.',
+      label: 'Message Id',
+      type: 'string',
+      required: true,
+      default: {
+        '@path': '$.messageId'
+      }
+    },
+    timestamp: {
+      description: 'The timestamp of the event.',
+      label: 'Timestamp',
+      type: 'string',
+      required: true,
+      default: {
+        '@path': '$.timestamp'
+      }
+    },
+    receivedAt: {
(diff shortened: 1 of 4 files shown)


joe-ayoub-segment:
Thanks for raising the PR @imsamdez . I'll schedule for review.

packages/browser-destinations/destinations/jimo/src/sendTrackEvent/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    },
+    receivedAt: {
+      description: 'The reception date of the event.',
+      label: 'Received at',
+      type: 'string',
+      required: true,
+      default: {
+        '@path': '$.receivedAt'
+      }
+    },
```
joe-ayoub-segment, line 38: I would remove this field as it will always be undefined. 

Values at the $.receivedAt position in the payload are only added when the payload reaches Segment's servers. In your case you are transforming the event in the browser and then calling the Jimo API. So the payload you will be transforming will not contain anything at this location.

packages/browser-destinations/destinations/jimo/src/sendTrackEvent/index.ts
```diff
@@ -0,0 +1,97 @@
...
+    const {
+      event_name,
+      userId,
+      anonymousId,
+      timestamp,
+      /* receivedAt, (for some reason, receivedAt is undefined) */
+      messageId,
+      properties
+    } = payload
+    const receivedAt = timestamp
```
joe-ayoub-segment, line 86: receivedAt will always be undefined for a web based Integration.

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
Other than the receivedAt issue the PR looks good @imsamdez !

joe-ayoub-segment:
Hi @imsamdez this PR has been deployed. Please check and let me know if there are any issues!

imsamdez:
Hey @joe-ayoub-segment,

Quick question, is it normal that the new action isn't mapped automatically for existing project.
For example, i've a project that only have the "Send user data" action mapped. But if i go to New mapping, i'm able to add the new Track action.
Also, after adding it, it seems to work well.
00a5d5ec04a995078531f4196248e1b7
67c87de2cb6880a0b8acb2986476d4bc

imsamdez:
Hey @joe-ayoub-segment. 

Alright, so we need to communicate to our customers each time we had a new action so that they can manually add it to their existing integration?

PR #1884, opened 2024-02-16
Revamp DV360 refresh_token flow

marinhero:
The refresh token belongs to every user that authenticates with Segment. We keep it and should use it every time we refresh the auth_token. The arguments that belong to Segment as part of this flow are `client_id` and `client_secret` and should be pulled from the env. 

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/display-video-360/shared.ts
+++ b/packages/destination-actions/src/destinations/display-video-360/shared.ts
@@ -1,7 +1,6 @@
 import { IntegrationError, RequestClient, StatsContext } from '@segment/actions-core'
 import { OAUTH_URL, USER_UPLOAD_ENDPOINT } from './constants'
 import type { RefreshTokenResponse } from './types'
-import type { OAuth2ClientCredentials } from '@segment/actions-core/destination-kit/oauth2'
 
 import {
   UserIdType,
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/display-video-360/shared.ts
```diff
@@ -15,35 +14,42 @@ import { ListOperation, UpdateHandlerPayload, UserOperation } from './types'
 import type { AudienceSettings, Settings } from './generated-types'
 import { GetAudienceInput } from '@segment/actions-core/destination-kit/execute'
 
-type SettingsWithOauth = Settings & { oauth: OAuth2ClientCredentials }
+type SettingsWithOauth = Settings & { oauth: DV360AuthCredentials }
+type DV360AuthCredentials = { refresh_token: string; access_token: string; client_id: string; client_secret: string }
```
nick-Ag, line 18: I wonder if we should mark `refresh_token` as optional, and throw an error if we try to refresh the access token without it. I suggest this because, if we run into that scenario where Google does not return the refresh token to us and we delete the existing one, it would be easier to debug with a custom error message.

nick-Ag (approved):

packages/destination-actions/src/destinations/display-video-360/shared.ts
marinhero: This is a good idea. Let me add it.

joe-ayoub-segment:
hi @marinhero and @nick-Ag looks like CI failing for this PR even though it's labeled for release. Can someone take a look please?

joe-ayoub-segment:
Hi @marinhero and @nick-Ag - I can't deploy this today as there are failing tests. I had a quick look but it wasn't clear to me what the intended logic should be so I wasn't sure if the test code or actual underlying logic needed to change.

marinhero:
Looks like I missed to update a unit test. I'll release this by running an out of schedule deploy.

joe-ayoub-segment:
Hi @marinhero PR deployed

PR #1887, opened 2024-02-16
[LinkedIn Conversions] Update Conversion Rules

nick-Ag:
This PR provides new logic for the `onMappingSave` hook which will determine whether the rule should be updated, rather than having an entirely new rule created. 

Related PRs:
https://github.com/segmentio/integrations/pull/2817
https://github.com/segmentio/app/pull/19605

## Testing

Tested successfully in staging.

https://github.com/segmentio/action-destinations/assets/27820201/5040db30-d42e-40fa-bdd5-064571d941f9


packages/destination-actions/src/destinations/linkedin-conversions/types.ts
```diff
@@ -100,6 +100,9 @@ export interface ConversionRuleCreationResponse {
   type: string
 }
 
+/** This request returns 204 no content */
+export interface ConversionRuleUpdateResponse {}
```
marinhero, line 104: This is an interesting use case of an `interface`. We typically use them to declare a contract that needs to be fulfilled, right? In your case, the contract is empty and using a `type` could work the same way and provide a "final" state as they are immutable. However, all of this file is using interfaces, therefore, I'd vouch for the use of `interface` even if the use case will let you get away with it.

packages/destination-actions/src/destinations/linkedin-conversions/api/api.test.ts
marinhero, line 1: Nice tests!

marinhero (approved):

jinapark202 (approved):

PR #1913, opened 2024-03-06
[Snap v3 CAPI] Additional SnapV3 connector tweaks

bordoley:
* Update logic for selecting the app or pixel id based upon the event_conversion_type
* Trim leading/trailing whitespace from string values.
* Ensure num_items is an integer value

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/utils.ts
+++ b/packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/utils.ts
@@ -77,5 +77,7 @@ export const splitListValueToArray = (input: string): readonly string[] | undefi
   return result.length > 0 ? result : undefined
 }
 
-export const emptyStringToUndefined = (v: string | undefined): string | undefined =>
-  (v ?? '').length > 0 ? v : undefined
+export const emptyStringToUndefined = (v: string | undefined): string | undefined => {
+  const trimmed = v?.trim()
+  return (trimmed ?? '').length > 0 ? trimmed : undefined
+}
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/utils.ts
```diff
@@ -77,5 +77,17 @@ export const splitListValueToArray = (input: string): readonly string[] | undefi
...
-  (v ?? '').length > 0 ? v : undefined
+export const emptyStringToUndefined = (v: string | undefined): string | undefined => {
+  const trimmed = v?.trim()
+  return (trimmed ?? '').length > 0 ? trimmed : undefined
+}
+
+export const parseNumberSafe = (v: string | number | undefined): number | undefined => {
+  if (Number.isSafeInteger(v)) {
+    return v as number
+  } else if (v != null) {
```
joe-ayoub-segment, line 88: should this be a check for undefined instead of null?

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
hi @bordoley just one minor comment to look at/clarify.

packages/destination-actions/src/destinations/snap-conversions-api/reportConversionEvent/utils.ts
bordoley: Weak equality in javascript allows you to use `!= null` to check both the undefined and null object case. It's a trick I learned at FB that allows you be a bit more safe in codebases that don't explicitly ban the use of the null object via linters.

joe-ayoub-segment:
hi @bordoley PR deployed

PR #1920, opened 2024-03-11
Additional responsys validation

joe-ayoub-segment:
Additional validation for Responsys

## Testing

Tested locally


joe-ayoub-segment:
PR deployed

PR #1924, opened 2024-03-12
Add mantle action destination

jnormore:
Adds the [Mantle](https://heymantle.com) destination with `identify` and `pushEvent` actions.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mantle/identify/index.ts
+++ b/packages/destination-actions/src/destinations/mantle/identify/index.ts
@@ -15,32 +15,40 @@ const action: ActionDefinition<Settings, Payload> = {
       description:
         'The unique identifier for the Shopify shop. This is used to associate the customer with a Shopify shop in Mantle',
       type: 'string',
-      required: true
+      required: true,
+      default: { '@path': '$.traits.shopId' }
     },
     myshopifyDomain: {
       label: 'Shopify Shop Domain',
       description:
         'The unique .myshopify.com domain of the Shopify shop. This is used to associate the customer with a Shopify shop in Mantle',
       type: 'string',
-      required: true
+      required: true,
+      default: { '@path': '$.traits.shopifyDomain' }
     },
     name: {
       label: 'Name',
       description: 'The name of the customer / shop',
       type: 'string',
-      required: false
+      required: false,
+      default: { '@path': '$.traits.name' }
     },
     email: {
       label: 'Email',
       description: 'The email of the customer',
       type: 'string',
-      required: false
+      required: false,
+      default: { '@path': '$.traits.email' }
     },
     customFields: {
       label: 'Custom Fields',
       description: 'The custom fields of the customer / shop',
       type: 'object',
-      required: false
+      required: false,
+      defaultObjectUI: 'keyvalue',
+      default: {
+        '@path': '$.traits'
+      }
     }
   },
   perform: (request, data) => {
--- a/packages/destination-actions/src/destinations/mantle/identify/index.ts
+++ b/packages/destination-actions/src/destinations/mantle/identify/index.ts
@@ -1,0 +1,61 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+import { API_URL } from '../config'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Identify',
+  description:
+    'Identify (create or update) a customer for your app in Mantle, including any additional information about the customer',
+  defaultSubscription: 'type = "identify"',
+  fields: {
+    platformId: {
+      label: 'Shopify Shop ID',
+      description:
+        'The unique identifier for the Shopify shop. This is used to associate the customer with a Shopify shop in Mantle',
+      type: 'string',
+      required: true
+    },
+    myshopifyDomain: {
+      label: 'Shopify Shop Domain',
+      description:
+        'The unique .myshopify.com domain of the Shopify shop. This is used to associate the customer with a Shopify shop in Mantle',
+      type: 'string',
+      required: true
+    },
+    name: {
+      label: 'Name',
+      description: 'The name of the customer / shop',
+      type: 'string',
+      required: false
+    },
+    email: {
+      label: 'Email',
+      description: 'The email of the customer',
+      type: 'string',
+      required: false
+    },
+    customFields: {
+      label: 'Custom Fields',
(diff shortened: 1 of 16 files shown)


packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Push Event',
+  description: 'Push event to your app in Mantle',
+  defaultSubscription: 'type = "track"',
+  fields: {
+    eventName: {
+      label: 'Event Name',
+      description: 'The name of the event you want to push to Mantle',
+      type: 'string',
+      required: true
```
joe-ayoub-segment, line 16: consider adding a default mapping here. e.g. 

`default: {'@path': '$.event'}`

packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    eventId: {
+      label: 'Event ID',
+      description: 'The unique identifier for the event. This is used to deduplicate events in Mantle',
+      type: 'string',
+      required: false
```
joe-ayoub-segment, line 22: Recommend this default: 

default: {'@path': '$.messageId'}

packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    customerId: {
+      label: 'Customer ID',
+      description:
+        'The unique identifier for the customer. This is used to associate the event with a customer in Mantle. It can be the internal customer ID, API token, Shopify shop ID, or Shopify shop domain',
+      type: 'string',
+      required: true
```
joe-ayoub-segment, line 29: Recommend adding a default mapping here as well. 
However this should be done carefully. 

There are 2 main identifiers you could map to. 
userId is an identifier which should be available when the end user is known to the customer. 
anonymousId is an anonymous identifier - basically a 'cookie' value which gets included in all events generated by client side Segment libraries (web, iOS, Android, etc). 

If your platform supports only 'known' users, then I'd recommend defaulting to userId: 

default: {'@path': '$.userId'}

If your platform supports only anonymous users, when I'd recommend defaulting to anonymousId: 

default: {'@path': '$.anonymousId'}

However if your platform supports both known and anonymous users then you can either use the userId and fall back to the anonymousId, or you could set up 2 different fields to collect each of these values. 

Here's how you can fall back from userId to anonymousId 

default: {
        '@if': {
          exists: { '@path': '$.userId' },
          then: { '@path': '$.userId' },
          else: { '@path': '$.anonymousId' }
        }
}

packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    properties: {
+      label: 'Event Properties',
+      description: 'The properties of the event. This is the extra data you want to attach to the event',
+      type: 'object',
+      required: false
```
joe-ayoub-segment, line 35: recommend adding default mapping: 

```
default: {
  '@path': '$.properties'
}
```

packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    timestamp: {
+      label: 'Event timestamp',
+      description: 'The timestamp of the event, defaults to the current time',
+      type: 'datetime',
+      required: false
```
joe-ayoub-segment, line 41: default should be: 

```
default: {
  '@path': '$.timestamp'
}
```

joe-ayoub-segment, line 41: you can make timestamp required. It will always be there.

packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
joe-ayoub-segment, line 22: You can make messageId required. It will always be there.

packages/destination-actions/src/destinations/mantle/pushEvent/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    })
+  },
+  performBatch: (request, data) => {
+    const events = data.payload.map((payload) => ({
+      event_name: payload.eventName,
+      ...(payload.eventId ? { event_id: payload.eventId } : {}),
+      customerId: payload.customerId,
+      properties: payload.properties || {},
+      ...(payload.timestamp ? { timestamp: payload.timestamp } : {})
+    }))
```
joe-ayoub-segment, line 64: Generally this looks fine. Just wondering why you would default to adding {} if there is no timestamp or event_id? Shouldn't these default to undefined or '' at most?

packages/destination-actions/src/destinations/mantle/identify/index.ts
```diff
@@ -0,0 +1,61 @@
...
+      label: 'Email',
+      description: 'The email of the customer',
+      type: 'string',
+      required: false
+    },
+    customFields: {
+      label: 'Custom Fields',
+      description: 'The custom fields of the customer / shop',
+      type: 'object',
+      required: false
```
joe-ayoub-segment, line 43: This should probably map to the following by default: 

`default: {'@path': '$.traits'} `

packages/destination-actions/src/destinations/mantle/identify/index.ts
```diff
@@ -0,0 +1,61 @@
...
+      type: 'string',
+      required: true
+    },
+    name: {
+      label: 'Name',
+      description: 'The name of the customer / shop',
+      type: 'string',
+      required: false
+    },
+    email: {
```
joe-ayoub-segment, line 33: This should probably map to the following by default: 

`default: {'@path': '$.traits.email'} `

joe-ayoub-segment (changes requested):

joe-ayoub-segment (approved):

joe-ayoub-segment:
Thanks @jnormore 
I merged and then made a couple of changes to some of the default paths. 
I think these changes should make it explicitly clear that the customer needs to carefully decide what to send as the customer_id, shop_id and shopify_domain. 

You can always update these / change the default with another PR if you disagree. 

Regards,
Joe

PR #1929, opened 2024-03-13
[Snap V3 CAPI] use pixel id for offline events

bordoley:
Update the Snap V3 connector to use the pixel id for offline events, and not fallback to the app id.

## Testing

During production testing, we found one case of AppID being used for OFFLINE conversion events. In this case pixel ID should actually be used.


joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @bordoley - PR deployed. I'll flip the flags shortly and ping you when done.

PR #1931, opened 2024-03-14
timeout error 429 and stringify

mayur-pitale:

Add a timeout error 0f 429 to avoid Race condition when Upsert List & Send Custom Traits/Audience get triggered at the same time
Add option on the "Send Custom Traits" and "Send Audience" actions to stringify all values if checked

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/responsys/sendAudience/index.ts
+++ b/packages/destination-actions/src/destinations/responsys/sendAudience/index.ts
@@ -75,6 +75,15 @@ const action: ActionDefinition<Settings, Payload> = {
       default: { '@path': '$.context.personas.computation_class' },
       choices: [{ label: 'Audience', value: 'audience' }]
     },
+    stringify_all_data: {
+      label: 'Convert All Data to String',
+      description: `If checked, all data is converted to String before sending to Responsys;
+                    otherwise, data is sent as is and must match to the data type in Responsys.
+                    eg. Value 1 will be sent as 1 if unchecked and as "1" if checked.`,
+      type: 'boolean',
+      default: false,
+      required: false
+    },
     enable_batching: enable_batching,
     batch_size: batch_size,
     timestamp: {
--- a/packages/destination-actions/src/destinations/responsys/sendCustomTraits/index.ts
+++ b/packages/destination-actions/src/destinations/responsys/sendCustomTraits/index.ts
@@ -37,6 +37,15 @@ const action: ActionDefinition<Settings, Payload> = {
         CUSTOMER_ID_: { '@path': '$.userId' }
       }
     },
+    stringify_all_data: {
+      label: 'Convert All Data to String',
+      description: `If checked, all data is converted to String before sending to Responsys;
+                    otherwise, data is sent as is and must match to the data type in Responsys.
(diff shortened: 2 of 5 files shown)


packages/destination-actions/src/destinations/responsys/sendAudience/index.ts
```diff
@@ -75,6 +75,15 @@ const action: ActionDefinition<Settings, Payload> = {
       default: { '@path': '$.context.personas.computation_class' },
       choices: [{ label: 'Audience', value: 'audience' }]
     },
+    stringify_all_data: {
+      label: 'Convert All Data to String',
+      description: `If checked, all data is converted to String before sending to Responsys;
+                    otherwise, data is sent as is and must match to the data type in Responsys.
+                    eg. Value 1 will be sent as 1 if unchecked and as "1" if checked.`,
```
joe-ayoub-segment, line 82:
```suggestion
      description: `If checked, all data will be converted to strings before being sent to Responsys. For example the value 1 will be sent as "1"`,
```

packages/destination-actions/src/destinations/responsys/sendCustomTraits/index.ts
```diff
@@ -37,6 +37,15 @@ const action: ActionDefinition<Settings, Payload> = {
         CUSTOMER_ID_: { '@path': '$.userId' }
       }
     },
+    stringify_all_data: {
+      label: 'Convert All Data to String',
+      description: `If checked, all data is converted to String before sending to Responsys;
+                    otherwise, data is sent as is and must match to the data type in Responsys.
+                    eg. Value 1 will be sent as 1 if unchecked and as "1" if checked.`,
```
joe-ayoub-segment, line 44: description: `If checked, all data will be converted to strings before being sent to Responsys. For example the value 1 will be sent as "1"`,

packages/destination-actions/src/destinations/responsys/utils.ts
```diff
@@ -80,11 +80,15 @@ export const sendCustomTraits = async (
...
     return userDataFieldNames.map((fieldName) => {
-      return (userData as Record<string, string>) && fieldName in (userData as Record<string, string>)
-        ? (userData as Record<string, string>)[fieldName]
-        : ''
+      if ((userData as Record<string, string>) && fieldName in (userData as Record<string, string>)) {
+        const value = (userData as Record<string, string>)[fieldName]
+        return payload[0].stringify_all_data && typeof value !== 'string' ? JSON.stringify(value) : value
+      } else {
+        return ''
+      }
```
joe-ayoub-segment, line 91: Hi @mayur-pitale it's not a good idea to simply check the payload[0] and apply the setting to all other items in the array. Each payload item might have a different value for stringify_all_data, so we need to handle that scenario.

joe-ayoub-segment: An alternative would be to make a global setting called stringify_all_data instead of having the setting at an Action level.

joe-ayoub-segment (changes requested):

joe-ayoub-segment:
hi @mayur-pitale this PR has been deployed.

PR #1956, opened 2024-03-27
[STRATCONN-3673] - Fetch all commits in publish workflow

varadarajan-tw:
Publish workflow failed during last week's deploy with the following error. This was because the publish workflow doesn't clone all the commits and hence it wasn't able to process commits beyond the current HEAD commit. Since we are generating changelog as part of the pulbish workflow, we fetch all commits instead of just the current HEAD. This PR updates the checkout action to fetch all [commits](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4).

image

## Testing

Testing completed successfully in local.

Shallow cloned the github repo with `git clone --branch=main git@github.com:segmentio/action-destinations.git --depth 1` 

image

Ran `git rev-list --count HEAD~ --grep "Publish" --since="00:00"` - This resulted in an error

image

On cloning the entire repo, the error went away

image

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,10 +45,6 @@ jobs:
       - name: Build
         run: NODE_ENV=production yarn build
 
-      - name: Fetch Latest Tags
-        run: |
-          git fetch --tags
-
       - name: Set NPM Token
         run: |
           npm set '//registry.npmjs.org/:_authToken' ${{ secrets.NPM_PUBLISH_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,6 +25,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           token: ${{ secrets.GH_PAT }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           token: ${{ secrets.GH_PAT }}
+          fetch-depth: 0 # Required as we compute the version based on the number of commits since the last tag
```
varadarajan-tw, line 28: The fetch-depth input controls how many commits Git will fetch when running the workflow. By setting fetch-depth to 0, the action fetches all history for all branches and tags, which is necessary here as the version is computed based on the number of commits since the last tag.

.github/workflows/publish.yml
```diff
@@ -44,10 +45,6 @@ jobs:
       - name: Build
         run: NODE_ENV=production yarn build
 
-      - name: Fetch Latest Tags
```
varadarajan-tw, line 47: Removing fetch latest tags step as fetch-depth:0 change to checkout step would pull fetch all tags.

PR #1959, opened 2024-04-01
Adds merge directive

valerieernst:

Adds `@merge` directive. Looks like some form of this used to exist, but it is no longer implemented in the code. This PRs adds the directive from scratch and cleans up references to the old `options.merge` in the readme.

## Testing

### DIFF
--- a/packages/core/src/mapping-kit/README.md
+++ b/packages/core/src/mapping-kit/README.md
@@ -209,78 +209,6 @@ suite][schema.test.js] is a good source-of-truth for current implementation beha
 
 [schema.test.js]: https://github.com/segmentio/fab-5-engine/blob/master/packages/destination-actions/src/lib/mapping-kit/__tests__
 
-## Options
-
-Options can be passed to the `transform()` function as the third parameter:
-
-```js
-const output = transform(mapping, input, options)
-```
-
-Available options:
-
-```js
-{
-  merge: true // default false
-}
-```
-
-### merge
-
-If true, `merge` will cause the mapped value to be merged onto the input payload. This is useful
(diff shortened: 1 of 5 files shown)


packages/core/src/mapping-kit/README.md
```diff
@@ -209,78 +209,6 @@ suite][schema.test.js] is a good source-of-truth for current implementation beha
 
 [schema.test.js]: https://github.com/segmentio/fab-5-engine/blob/master/packages/destination-actions/src/lib/mapping-kit/__tests__
 
-## Options
-
-Options can be passed to the `transform()` function as the third parameter:
```
valerieernst, line 214: `transform` doesn't actually accept a 3rd param 😬

marinhero (approved):

marinhero (dismissed):
Looks like CI is failing. Let's address those and then I can give another 👁️

LLuque-twilio (approved):
lgtm. 

Building off of someone elses comment - maybe we can include an optional string input  param to the `merge()` method such as `left` or `right` that handles which objects key should "win" in the merge (looks like the default is `right`).

We do something similar with the `case()` directive allowing us to switch between lowercase/uppercase - see [here](https://github.com/segmentio/action-destinations/pull/1959/files#diff-d9563d5c7d4ab9ddfcd19dfb24309188891a94c3bfbb144e466cc595bb37f231L91)

dlasky:
maybe optionally and slightly less confusing if there are two inputs the user could switch the order of the inputs to get the desired behavior without having to necessarily explicitly understand how our code works?

dlasky (approved):
test failure looks unrelated -- timezones :-/

PR #1963, opened 2024-04-02
[STRATCONN-3673] - Fix release workflow

varadarajan-tw:
This PR attempts to fix the release workflow that is currently failing due to [permission issue](https://github.com/segmentio/action-destinations/actions/runs/8523646318/job/23346389646). 

image

The reason for permission issue is we use GH_PAT token in publish workflow which has permissions only to work with control-plane-js-client repo. Since action-destinations is open source, the workflow is able to clone the actions repo. 

This PR makes couple of changes to fix:
- To avoid using GH PAT token with write access to Github repo,  I have moved the release tag computation to the`postversion.sh` script. This script will be executed via postversion hook - only for main branch.
- Extract the release steps to separate job that depends on the build-and-publish job to ensure we don't use GH_PAT token. GH_PAT token contains permissions only to pull data from ctl-plane-js-client.

## Testing

Testing completed successfully with staging branch. No packages were published because I tested by commenting out the `publish` step.

[Success Run](https://github.com/segmentio/action-destinations/actions/runs/8525607521/job/23352968214)

image

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,7 +25,6 @@ jobs:
       - uses: actions/checkout@v3
         with:
           token: ${{ secrets.GH_PAT }}
-          fetch-depth: 0 # Required as we compute the version based on the number of commits since the last tag
 
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v3
@@ -45,6 +44,10 @@ jobs:
       - name: Build
         run: NODE_ENV=production yarn build
 
+      - name: Fetch Latest Tags
+        run: |
+          git fetch --tags
+
       - name: Set NPM Token
         run: |
           npm set '//registry.npmjs.org/:_authToken' ${{ secrets.NPM_PUBLISH_TOKEN }}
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -45,6 +44,10 @@ jobs:
       - name: Build
         run: NODE_ENV=production yarn build
 
+      - name: Fetch Latest Tags
```
varadarajan-tw, line 47: Reverting my previous PR commit - https://github.com/segmentio/action-destinations/pull/1956

marinhero (approved):
LGTM. Question though:

> To avoid using GH PAT token with write access to Github repo, I have moved the release tag computation to thepostversion.sh script. This script will be executed via postversion hook - only for main branch.

The reason we aren't creating a token to write on this repo is because it's open source? or why don't replace the token you were using with limited permissions to one that has the privileges you require.

varadarajan-tw:
That is a good question. I thought about it. The reason I don't want to use PAT token is every write action performed will get tagged to the user associated with the PAT token irrespective of whoever triggered the release flow.  This is not good for auditing. Also, it is not possible to restrict PAT token access to just updating tags.

The github-action(bot) by default has action to write to the repo but since `main` is a protected branch and only members of our github org have access to push to `main` directly,  `github-action(bot)` will not able to push tags. PAT tokens are the only way to write to main.

Overall, I felt adding release tag after lerna completes its `Publish` script makes the overall process easier, and we don't have to go through this entire access ordeal.

PR #1974, opened 2024-04-05
[KAMELEOON] added new events for tracking

akalyuzhnyi:
The [Kameleoon](https://www.kameleoon.com/) cloud destination.

## Testing

-  renamed logEvent => track
-  added page event
-  added group event
-  added identify event


packages/destination-actions/src/destinations/kameleoon/index.ts
```diff
@@ -75,7 +79,10 @@ const destination: DestinationDefinition<Settings> = {
   },
   presets,
   actions: {
-    logEvent
+    track,
```
joe-ayoub-segment, line 82: Can we keep the logEvent action name please (instead of track). 

Our deploy process is additive only, so I'd have to write a script to delete the logEvent config from the Production database - which is something I'd prefer not to do. 

So the ask here is as follows: 
1. It's OK to add the page, group and identify Actions
2. please keep the logEvent Action in the folder called logEvent. Please keep the logEvent Action registered on this line, and do not register a new Action called track. 
3. You can provide a new Label and Description for the logEvent Action. This is what the user sees in the UI anyway.

joe-ayoub-segment, line 82: Can we keep the logEvent action name please (instead of track). 

Our deploy process is additive only, so I'd have to write a script to delete the logEvent config from the Production database - which is something I'd prefer not to do. 

So the ask here is as follows: 
1. It's OK to add the page, group and identify Actions
2. please keep the logEvent Action in the folder called logEvent. Please keep the logEvent Action registered on this line, and do not register a new Action called track. 
3. You can provide a new Label and Description for the logEvent Action. This is what the user sees in the UI anyway. i.e you can call it 'Track Event'.

packages/destination-actions/src/destinations/kameleoon/track/index.ts
```diff
@@ -6,13 +6,25 @@ import omit from 'lodash/omit'
...
+    userId: {
+      type: 'string',
+      description: 'The ID associated with the user',
+      label: 'User ID',
+      default: { '@path': '$.userId' }
+    },
     event: {
       type: 'string',
-      required: false,
+      required: true,
```
joe-ayoub-segment, line 27: Making the event field required if OK if you are handling only track() events. 
However I'd recommend removing the default mapping which points to $.name below, as this is only used in page() events, and page() events don't always have a value in the $.name location of the payload. 

So change the default mapping to this: 

`default: { '@path': '$.event' },`

packages/destination-actions/src/destinations/kameleoon/identify/index.ts
```diff
@@ -0,0 +1,79 @@
...
+      type: 'string',
+      description: 'The ID associated with the user',
+      label: 'User ID',
+      default: { '@path': '$.userId' }
+    },
+    properties: {
+      type: 'object',
+      required: false,
+      description: 'Traits to send with the event',
+      label: 'user traits',
```
joe-ayoub-segment, line 30:
```suggestion
      label: 'User Traits',
```

packages/destination-actions/src/destinations/kameleoon/group/index.ts
```diff
@@ -0,0 +1,85 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import omit from 'lodash/omit'
+
+import { BASE_URL } from '../properties'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Group',
+  description: 'Group user in Kameleoon',
```
joe-ayoub-segment, line 10: Can we add a more explicit description here please?

packages/destination-actions/src/destinations/kameleoon/group/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      description: 'The group id',
+      label: 'Group ID',
+      required: true,
+      default: { '@path': '$.groupId' }
+    },
+    properties: {
+      type: 'object',
+      required: false,
+      description: 'Traits to send with the event',
+      label: 'user traits',
```
joe-ayoub-segment, line 36: Should this say 'User Traits' or something which aligns more with what a group is?
Please also Capital Case the label. 

```suggestion
      label: 'User Traits',
```

joe-ayoub-segment:
Hi @akalyuzhnyi thanks for raising the PR. 
I've provided some feedback. Main thing is that I'd prefer that you not rename the logEvent Action to track. You can still update the label and description for the Action though, so in the UI it can say 'Track Event'. 

Does that make sense?

packages/destination-actions/src/destinations/kameleoon/index.ts
akalyuzhnyi: Hello! Ok, I will add back logEvent

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @akalyuzhnyi PR deployed. Please confirm that the change looks good!

joe-ayoub-segment:
FYI I had to update some breaking tests.

PR #1987, opened 2024-04-10
Dynamic Yield: turning Destination into Audience Destination

joe-ayoub-segment:
Turning DY Destination into an Audience Destination

## Testing

None done yet. Do not deploy


packages/destination-actions/src/destinations/dynamic-yield-audiences/syncAudience/index.ts
```diff
@@ -42,57 +66,90 @@ const action: ActionDefinition<Settings, Payload> = {
...
-    const audienceID = payload.segment_audience_id
-    const audienceValue = d?.rawData?.properties?.[audienceName] ?? d?.rawData?.traits?.[audienceName]
+    const { audienceSettings, payload, settings } = data  
+    const audienceName = audienceSettings?.audience_name ?? payload.segment_audience_key
+    
+    const audienceValue = payload.traits_or_props[payload.segment_audience_key]
+
+    let primaryIdentifier
+
+    switch(settings.identifier_type){
```
djluis, line 110: If I'm understanding correctly, this works for the primary identifiers, but later we need to add some dynamic code to fetch a custom identifier, like the example below:
```
default: {
   primaryIdentifier = payload[settings.identifier_type] ?? undefined
}
```

djluis (approved):
Looks good for the initial mapping with the primary identifiers, later we must include the custom "dynamic" identifiers. 👍 thanks again for the support @joe-ayoub-segment

packages/destination-actions/src/destinations/dynamic-yield-audiences/syncAudience/index.ts
joe-ayoub-segment: > If I'm understanding correctly, this works for the primary identifiers, but later we need to add some dynamic code to fetch a custom identifier, like the example below:
> 
> ```
> default: {
>    primaryIdentifier = payload[settings.identifier_type] ?? undefined
> }
> ```

hi @djluis I decided to go ahead and hard code the 3 most common identifiers into a dropdown list, and to make email the default. If you are able to expose an API we can use to populate a dynamic list then we can make use of it.

Currently, the code only adds the identifier for the selected identifier_type into the payload which gets sent to DY. We could change this to include all identifiers we have, if you like. Though you'll have to show me how to structure that information in the payload.

packages/destination-actions/src/destinations/dynamic-yield-audiences/syncAudience/index.ts
```diff
@@ -42,57 +76,90 @@ const action: ActionDefinition<Settings, Payload> = {
...
+            section_id : settings.sectionId,
+            identifier_type: settings.identifier_type,
+            accessKey: settings.accessKey
+          }
+        },
+        user_profiles : [ {  
+          user_identities : [ {   
+            type : settings.identifier_type,
+            encoding : "sha-256",
+            value : hashAndEncode(primaryIdentifier)
```
djluis, line 155: @joe-ayoub-segment We need to discuss this on next meeting ⚠️

djluis (changes requested):

joe-ayoub-segment:
PR deployed

PR #1990, opened 2024-04-15
feat(screeb): identify on init

samber:
At Screeb, we need to identify users when the tag is initialized. If we do it later, then the tag will create an identify for a short period (between `init()` and `identify()`). Customers are disappointed by this behavior, so we inject the segment userId.

AnonymousId is ignored since Screeb does not have such a concept.

### DIFF
--- a/packages/browser-destinations/destinations/screeb/src/index.ts
+++ b/packages/browser-destinations/destinations/screeb/src/index.ts
@@ -1,6 +1,7 @@
 import type { Settings } from './generated-types'
 import type { BrowserDestinationDefinition } from '@segment/browser-destination-runtime/types'
 import { browserDestination } from '@segment/browser-destination-runtime/shim'
+import { ID } from '@segment/analytics-next'
 import { Screeb } from './types'
 import { defaultValues } from '@segment/actions-core'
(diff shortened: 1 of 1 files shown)


packages/browser-destinations/destinations/screeb/src/index.ts
```diff
@@ -71,7 +72,12 @@ export const destination: BrowserDestinationDefinition<Settings, Screeb> = {
     await deps.loadScript('https://t.screeb.app/tag.js')
     await deps.resolveWhen(() => window.$screeb !== preloadFunction, 500)
 
-    window.$screeb('init', settings.websiteId)
+    let visitorId: ID = null
```
joe-ayoub-segment, line 75: hi @samber - not sure if the ID type needs to be imported. 
Can you just do this please?

let visitorId: string | null | undefined = null

joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @samber - this PR has been deployed. Please confirm that it looks correct to you!

samber:
@joe-ayoub-segment i just tested and it works very well, thanks ;)

PR #2002, opened 2024-04-19
Adding 'Segment' to TT Pixel change

joe-ayoub-segment:
Minor change to TT Pixel Destination

## Testing

Tested with Jae from TikTok.


joe-ayoub-segment:
PR deployed

PR #2003, opened 2024-04-19
TikTok Conversions API migration

joe-ayoub-segment:
Migrating TikTok Conversions to a new API. 

## Testing

This has been tested by TikTok as a Sandbox Destination which is already live on Production but in Private Beta mode.


sayan-das-in (approved):
Changes looks good however I am not able to verify the correctness of the API change as TikTok is blocked in India.

joe-ayoub-segment:
note: PR deployed on Tuesday this week

PR #2011, opened 2024-04-26
Add external_id field to trackEvent Action

tcgilbert:
This update adds the ability to map a Segment field to Klaviyo's external_id [in their create event api](https://developers.klaviyo.com/en/reference/create_event). 

Screenshot 2024-04-26 at 1 57 28 PM

This comes at the request of the Klaviyo team: 

> The only other request I have Thomas is that in track/identify mappings for Klaviyo Actions, I’d love to have the ability to add “external_id” in as a profile property.  This would help on our side of attributing events in the case where if we wanted to map the segment user_id to an external_id on our end, we could track events without a hard email / phone in the payload (like track calls). 

## Testing

Tested locally: 
Screenshot 2024-04-26 at 1 48 40 PM

### DIFF
--- a/packages/destination-actions/src/destinations/klaviyo/trackEvent/index.ts
+++ b/packages/destination-actions/src/destinations/klaviyo/trackEvent/index.ts
@@ -23,6 +23,10 @@ const action: ActionDefinition<Settings, Payload> = {
           label: 'Phone Number',
           type: 'string'
         },
+        external_id: {
+          label: 'External ID',
+          type: 'string'
+        },
         other_properties: {
           label: 'Other Properties',
           type: 'object'
--- a/packages/destination-actions/src/destinations/klaviyo/__tests__/__snapshots__/snapshot.test.ts.snap
+++ b/packages/destination-actions/src/destinations/klaviyo/__tests__/__snapshots__/snapshot.test.ts.snap
@@ -28,6 +28,7 @@ Object {
         "data": Object {
           "attributes": Object {
             "email": "rafebezuv@micfeoz.zm",
+            "external_id": "PE*zlOgIPA]mVozMLBaL",
             "other_properties": Object {
               "testType": "PE*zlOgIPA]mVozMLBaL",
             },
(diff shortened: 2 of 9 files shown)


packages/destination-actions/src/destinations/klaviyo/trackEvent/index.ts
```diff
@@ -23,6 +23,10 @@ const action: ActionDefinition<Settings, Payload> = {
           label: 'Phone Number',
           type: 'string'
         },
+        external_id: {
+          label: 'External ID',
+          type: 'string'
+        },
```
joe-ayoub-segment, line 29: Seems like a straightforward change. Just a few questions: 

1) Does adding support for external_id break with any conventions in other Actions? e.g. does another Action also support external_id, but maybe have it named or handled differently?

2) Should there be default mappings for the external_id (and maybe also phone_number and email ) field?

3) Has this been tested properly?

tcgilbert: > Does adding support for external_id break with any conventions in other Actions? e.g. does another Action also support external_id, but maybe have it named or handled differently?

No. All the other actions actually have an external_id field, so if anything this makes the actions more congruent. 

> Should there be default mappings for the external_id (and maybe also phone_number and email ) field?

Can you add default mappings to properties within a object type? If so, I think this would make sense

> Has this been end to end tested some how? Has it been confirmed that the external_id gets to Klaviyo correctly?

I've been sending events from Actions tester to Klaviyo, and can verify that the external_id is arriving within klaviyo. If there is any other way to test, happy to do so as well. Here is one for "Order Completed", external_id was empty previsouly

Screenshot 2024-04-29 at 9 12 17 AM

tcgilbert:
[waiting for clarity on payload validation from Klaviyo](https://twilio-external.slack.com/archives/C0183R05G8K/p1714397253518169?thread_ts=1708987275.420609&cid=C0183R05G8K)

tcgilbert:
closing as never got confirmation from Klaviyo

PR #2018, opened 2024-04-30
Patch for UI issue with Klaviyo Subscribe

tcgilbert:
When attempting to add the subscribeProfile action in my workspace I get a graphql error. 

Screenshot 2024-04-30 at 3 01 06 PM

This appears to be because the type for custom_source field is a string, and the current default value is set to a number `-59`. I previously thought klaviyo required that this be a number ([based off partner source codes](https://help.klaviyo.com/hc/en-us/articles/1260804673530#h_01HDKHG9AM4BSSM009BM6XBF1H)) but after testing you can set this to a string `"-59"` and the result in Klaviyo is the same. 

Note that there is no issue with adding the new unsibscribeProfile action, so I'm fairly certain this will fix the issue.  

## Testing

Sent events to Klaviyo's subscribe endopint using `custom_source: "-59"` and this successfully adds the right $source param on profile
Screenshot 2024-04-30 at 3 34 11 PM

### DIFF
--- a/packages/destination-actions/src/destinations/klaviyo/functions.ts
+++ b/packages/destination-actions/src/destinations/klaviyo/functions.ts
@@ -227,7 +227,7 @@ export function formatSubscribeRequestBody(
     }
   }
 
-  subData.data.attributes.custom_source = custom_source || -59
+  subData.data.attributes.custom_source = custom_source || '-59'
 
   if (list_id) {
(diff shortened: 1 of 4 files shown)


joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @tcgilbert deploy complete

PR #2024, opened 2024-05-03
channels-1094

cogwizzle:
refactor: Refactoring the name for dataFeedCache to be
engageDestinationCache. This is because this cache is no longer just intended to be used for DataFeed cache, but is now intended to be used as a more generic cache for the Engage Destination.

## Testing
The changes should be tested in the [PR in integrations](https://github.com/segmentio/integrations/pull/2863).

This change was tested in conjunction with the [Channels-1094 branch in integrations](https://github.com/segmentio/integrations/pull/2863/files). We saw 2000 duplicates prevented during our [load test](https://segment.datadoghq.com/notebook/9113137/channels-1094?range=3540000&tpl_var_correlation_id=ssd_2jkl2hgnb9jny4bwg5jpzgnqj2m&start=1722012720000&live=false). Additionally we had 624 duplicates that still showed up in Splunk. In prior load test of the same size this number was close to 3000.


packages/core/src/create-test-integration.ts
```diff
@@ -36,15 +36,17 @@ interface InputData<Settings> {
    */
   useDefaultMappings?: boolean
   auth?: AuthTokens
-  /**
-   * The features available in the request based on the customer's sourceID;
-   * `features`, `stats`, `logger`, `dataFeedCache`, and `transactionContext` and `stateContext` are for internal Twilio/Segment use only.
-   */
+  /** Engage internal use only. DO NOT USE. */
```
cogwizzle, line 39: Broke this comment apart so that we could get language server support for using each property in the shape of this object.

packages/core/src/destination-kit/types.ts
```diff
@@ -39,15 +41,17 @@ export interface ExecuteInput<
   page?: string
   /** The data needed in OAuth requests */
   readonly auth?: AuthTokens
-  /**
-   * The features available in the request based on the customer's sourceID;
-   * `features`,`stats`, `logger` , `transactionContext` and `stateContext` are for internal Twilio/Segment use only.
-   */
+  /** Engage internal use only. DO NOT USE. */
```
cogwizzle, line 44: Broke this comment apart so that we could get language server support for using each property in the shape of this object.

packages/core/src/destination-kit/index.ts
```diff
@@ -364,11 +367,11 @@ export interface Logger {
...
 
-export interface DataFeedCache {
-  setRequestResponse(requestId: string, response: string, expiryInSeconds: number): Promise<void>
-  getRequestResponse(requestId: string): Promise<string | null>
-  maxResponseSizeBytes: number
-  maxExpirySeconds: number
+export interface EngageDestinationCache {
+  getByKey: (key: string) => Promise<string | null>
+  readonly maxExpirySeconds: number
+  readonly maxResponseSizeBytes: number
```
pmunin, line 373: "response" still specific to datafeed case, should be maxValueSizeBytes or something as mentioned in https://github.com/segmentio/integrations/pull/2863/files#diff-e241cf2449a32291aa0ff2fa6959ea3e83c8cf8cc3d061588037e535cdf2c07a

Also it's strange that we dup the code of validating that size & expiry [in this repo](https://github.com/segmentio/action-destinations/blob/d52ba57aaad1f6339bc0c3e9af691e9886e134b8/packages/destination-actions/src/destinations/engage/utils/apiLookups.ts#L165-L183) as well as in Integrations (validateSet). 
Maybe instead we just try to cache.setByKey and then handle error that it throws

packages/core/src/destination-kit/index.ts
```diff
@@ -401,8 +404,8 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
       this.settingsSchema = fieldsToJsonSchema(this.authentication.fields)
     }
 
-    for (const action of Object.keys(destination.actions)) {
-      this.partnerAction(action, destination.actions[action])
+    for (const [name, action] of Object.entries(destination.actions)) {
+      this.partnerAction(name, action)
```
pmunin, line 408: :nice:

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -91,9 +91,86 @@ export abstract class MessageSendPerformer<
...
+   */
+  @track()
+  async readCache(messageId?: string, recipientId?: string) {
+    if (!messageId || !recipientId || !this.engageDestinationCache) {
+      return
+    }
+    return this.engageDestinationCache.getByKey(messageId + recipientId)
+  }
+
+  @track()
```
pmunin, line 135: I don't think you need this tracked, when the upstream EngageActionPerformer.perform is tracked already

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -91,9 +91,86 @@ export abstract class MessageSendPerformer<
...
+   */
+  @track()
+  async cacheResponse(response: Response, messageId?: string, recipientId?: string) {
+    if (!messageId || !recipientId || !this.engageDestinationCache) {
+      return
+    }
+    // Check if the response is successful.
+    if (response.ok) {
+      const body = await response.text()
+      await this.engageDestinationCache.setByKey(messageId + recipientId, body)
```
pmunin, line 112: we don't want to store entire body in cache - that may be costly and not risk free from performance and redis-faultiness standpoint.
We just need response.status (integer).

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
```diff
@@ -91,9 +91,86 @@ export abstract class MessageSendPerformer<
...
+   * @returns The response from the cache.
+   */
+  @track()
+  async cacheResponse(response: Response, messageId?: string, recipientId?: string) {
+    if (!messageId || !recipientId || !this.engageDestinationCache) {
+      return
+    }
+    // Check if the response is successful.
+    if (response.ok) {
+      const body = await response.text()
```
pmunin, line 111: I might be wrong but reading response.text() here will close the stream and will fail next time response body is tried to be read (during actual logic of processing response), so you might want to check it and if it's the case you may use response.clone() or pass here the actual value to be cached, instead of response object

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
```diff
@@ -361,7 +361,7 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload>
           this.tags,
           this.settings,
           this.logger.loggerClient,
-          this.dataFeedCache
+          this.engageDestinationCache
```
pmunin, line 364: hmm, out of your scope, but just for record: strange to see this in sendgrid code, I'd expect that to happen in base class and be used by both Twilio and Sendgrid.

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
pmunin, line 1: I'd do it easier and in EngageActionPerformer.ts/perform method:
```
    const op = this.currentOperation as OperationContext
    op?.onFinally.push(() => {
      const error = op.error as TrackedError
      if(!error) this.cache.setByKey(messageId, stringifySuccess())
      else if (error && !isErrorRetryable(error)) this.cache.setByKey(messageId, stringifyError(error))
    })

    try {
      const cache = parseCache(this.cache.getByKey(messageId))
      if(isCacheSuccess(cache)) return cache.result
      else if(isCachedError) throw cache.error
    }
    catch()
    {
        //log problem with redis
        // in that case we proceed as if nothing is cached
    }
    ... 
    beforePerform()
    return doPerform()
```

pmunin (changes requested):

packages/core/src/destination-kit/index.ts
cogwizzle: Great call. Will rename this.

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
cogwizzle: Since I'm swapping to use the status code I don't think we need to worry about this.

packages/destination-actions/src/destinations/engage/sendgrid/sendEmail/SendEmailPerformer.ts
cogwizzle: I agree.

packages/destination-actions/src/destinations/engage/utils/MessageSendPerformer.ts
cogwizzle: Can you help me understand why this belongs in EngageActionPerformer rather than in the MessageSendPerformer? I'm certain you have a reason, but I'm not 100% certain the express purpose of each class.

pmunin: EngageActionPerformer is a base class for any engage DA, which contains base logic for logging, stats etc. The only derived class right now is MessageSendPerformer, which is specialized on message sending workflow, i.e. extracting template, personalizing it for profile and then sending it to all recipients.

It's arguable, but I'd suggest for our case it is more practical to dedup on `DA.perform` level rather than on per recipient level - because:
 1) most of actions have single recipient (recently we allowed having multiple recipients per email as a workaround for some subscription related sev, but iirc our strategy is to have a single recipient per `DA.perform`)
2) it is safe to assume that if it message succeeded or failed for one of the recipients - retryability of it will be applicable to the other recipients of that message, because based on what I saw in datadog in 99% of the cases failure happens before hitting TW/SG api call.

Therefore since we're talking about caching of entire DA.perform outcome (and not per each recipient within each perform) for preventing dup calls of DA.perform - that logic belongs to EngageActionPerformer.

Hope it makes sense?

cogwizzle: I would like to keep the current implementation for a few reasons.

I tried to implement the suggested pattern you outlined above. From an outsiders perspective the code was much harder for me to interpret and follow due to the unconventional coding style associated with the suggested change. I really struggled due to the unusual types associated with the onFinally code. I don't think casting the type is a reasonable solution either, because I am not familiar enough with the code base to know what the given type is supposed to be.

The existing implementation handles both the multi-recipient and single-recipient case. If we do support the multi-recipient case shouldn't we build a caching implementation that works this way too? I have concerns about potential side effects if we don't use a compound key such as messageId and recipientId we may return an inappropriate cached value which may cause messages to fail to be sent due to cache collisions. In a sense my fear is that we go from creating duplicates to dropping messages in a way that would not be very transparent to trace through.

If you'd like to discuss this further I recommend we pair on this solution to get the most equitable solution between our two proposed implementations. I'm sure together we can come up with a reasonable solution if you decide it is worth going down this road.

joe-ayoub-segment:
hi @cogwizzle @pmunin please let me know when you are done with this so I can review prior to deploy

pmunin (dismissed):

joe-ayoub-segment:
hi @cogwizzle - there's a bunch of failing CI checks. Can you address these before I review please?

cogwizzle:
> hi @cogwizzle - there's a bunch of failing CI checks. Can you address these before I review please?

Fixed the test and pipeline.

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -91,9 +91,104 @@ export abstract class MessageSendPerformer<
...
+   * @param recipientId The recipientId of the message.
+   * @returns The response from the cache.
+   */
+  @track()
+  async putCache(value: any, messageId?: string, recipientId?: string) {
+    if (!messageId || !recipientId || !this.engageDestinationCache) {
+      return
+    }
+    const status = value
+    await this.engageDestinationCache.setByKey(messageId + recipientId.toLowerCase(), JSON.stringify(status))
```
alfrimpong, line 110: Lets keep an eye out for event loop blocking code like stringify and parse. Can we pull out the relevant `value` in the subclasses and pass it down to this? That way we can avoid the stringify

cogwizzle: Ok something that makes this extremely difficult is the way that the surrounding code is written. It effectively could be type `any` rather than a response type. Fixing the types would be difficult.

I can give my best attempt to address this problem. I see the issue with parse, and stringify that you're talking about and I was not aware that it was going to be such a performance issue.

cogwizzle: So I reworked the code in a way that I created a simple serialization pattern that can be used to communicate all of the data that we need without using JSON.stringify or JSON.parse.

alfrimpong: nice! lmk when its time for a re-review

joe-ayoub-segment:
hi @cogwizzle - converting to draft until this is ready for review.

pmunin:
Closing towards https://github.com/segmentio/action-destinations/pull/2325

PR #2054, opened 2024-05-23
Fix customer.io timestamp issue

nunofgs:
PR #1800 introduced a small regression. If users set `convert_timestamp` to false, then the `timestamp` field would not be converted which is actually not support by the Customer.io API. The API only accepts unix timestamps on the root `timestamp` field.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/__tests__/trackEvent.test.ts
+++ b/packages/destination-actions/src/destinations/customerio/__tests__/trackEvent.test.ts
@@ -172,7 +172,7 @@ describe('CustomerIO', () => {
             ...attributes,
             anonymous_id: event.anonymousId
           },
-          timestamp,
+          timestamp: dayjs.utc(timestamp).unix(),
           type: 'person'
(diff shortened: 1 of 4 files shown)


joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @nunofgs PR deployed. Please check it all looks good!

nunofgs:
Hi Joe,

Our backend fix was deployed yesterday. To recap:

   - Our backend fix made it so an invalid timestamp (ie: not a Unix
   timestamp) is replaced with time.now() instead. This matches the behavior
   we have in the v1 api.
   - The PR you deployed on action-destinations makes it so ISO-8601 dates
   are converted into timestamps so the customer data is kept instead of
   assuming time.now()


Also, we'll soon be submitting one more PR with some changes to defaults if
that's OK. No functionality changes, just better defaults. I've CCed the
team members who are taking care of that.

Thanks!
--
Nuno Sousa

PR #2057, opened 2024-05-24
Implement batch support for Attio Actions

nickcampbell18:
This pull request adds batch support for the Attio Actions destination. We're looking to make this change because our customers are seeing high failure rates (rate limit exceeded, timeouts and 5xx errors) with the single-event mechanisms.

Our secret batching API supports up to 10,000 Segment events in a single batch that will allow us to process these asynchronously, in order, in our own infrastructure in a more controlled and reliable way.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/attio/common-fields.ts
+++ b/packages/destination-actions/src/destinations/attio/common-fields.ts
@@ -4,11 +4,11 @@ import { Settings } from './generated-types'
 export const commonFields: ActionDefinition<Settings>['fields'] = {
   enable_batching: {
     label: 'Send data to Attio in batches',
-    description: 'Send data to Attio in batches for much better performance.',
+    description:
+      'Send batches of events to Attio, for improved performance, however invalid events are silently dropped.',
     type: 'boolean',
     required: false,
-    unsafe_hidden: true,
-    default: true
+    default: false
   },
 
   batch_size: {
--- a/packages/destination-actions/src/destinations/attio/common-fields.ts
+++ b/packages/destination-actions/src/destinations/attio/common-fields.ts
@@ -24,7 +24,7 @@ export const commonFields: ActionDefinition<Settings>['fields'] = {
     label: 'Received at',
     description: 'When the event was received.',
     type: 'datetime',
-    required: true,
+    required: false,
     default: {
       '@path': '$.receivedAt'
     }
--- a/packages/destination-actions/src/destinations/attio/common-fields.ts
+++ b/packages/destination-actions/src/destinations/attio/common-fields.ts
@@ -1,0 +1,32 @@
+import { ActionDefinition } from '@segment/actions-core'
+import { Settings } from './generated-types'
+
+export const commonFields: ActionDefinition<Settings>['fields'] = {
+  enable_batching: {
+    label: 'Send data to Attio in batches',
+    description: 'Send data to Attio in batches for much better performance.',
+    type: 'boolean',
+    required: false,
+    unsafe_hidden: true,
+    default: true
+  },
+
+  batch_size: {
+    label: 'Batch Size',
+    description: 'Max batch size to send to Attio (limit is 10,000)',
+    type: 'number',
+    required: false,
+    unsafe_hidden: true,
+    default: 1_000
+  },
+
+  received_at: {
+    label: 'Received at',
+    description: 'When the event was received.',
+    type: 'datetime',
+    required: true,
+    default: {
+      '@path': '$.receivedAt'
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


packages/destination-actions/src/destinations/attio/common-fields.ts
```diff
@@ -0,0 +1,32 @@
...
+    required: false,
+    unsafe_hidden: true,
+    default: 1_000
+  },
+
+  received_at: {
+    label: 'Received at',
+    description: 'When the event was received.',
+    type: 'datetime',
+    required: true,
```
joe-ayoub-segment, line 27: hi @nickcampbell18 can you make this field optional please?
Its value will be undefined for all instances of the Integration which have already been configured prior to the deploy happening. So you'll need to handle that in the perform() and performBatch() code too.

packages/destination-actions/src/destinations/attio/identifyUser/index.ts
```diff
@@ -85,6 +87,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          person: {
+            object: 'people',
+            mode: 'create-or-update',
+            matching_attribute: 'email_addresses',
+            multiselect_values: 'append',
+            values: {
+              email_addresses: item.email_address,
+              ...(item.person_attributes ?? {})
+            },
+            received_at: item.received_at.toString()
```
joe-ayoub-segment, line 115: received_at needs to be handled defensively as it will be undefined for customers who have the Integration already set up.

packages/destination-actions/src/destinations/attio/identifyUser/index.ts
```diff
@@ -85,6 +87,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          }
+        },
+        received_at: item.received_at.toString()
```
joe-ayoub-segment, line 118: received_at needs to be handled defensively as it will be undefined for customers who have the Integration already set up.

packages/destination-actions/src/destinations/attio/identifyUser/generated-types.ts
```diff
@@ -21,4 +21,16 @@ export interface Payload {
...
+   */
+  enable_batching?: boolean
+  /**
+   * Max batch size to send to Attio (limit is 10,000)
+   */
+  batch_size?: number
+  /**
+   * When the event was received.
+   */
+  received_at: string | number
```
joe-ayoub-segment, line 35: Needs to be optional.

packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
```diff
@@ -101,6 +103,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          company: {
+            object: 'companies',
+            mode: 'create-or-update',
+            matching_attribute: 'domains',
+            multiselect_values: 'append',
+            values: {
+              domains: item.domain,
+              ...(item.company_attributes ?? {})
+            },
+            received_at: item.received_at.toString()
```
joe-ayoub-segment, line 131: received_at needs to be handled defensively as it will be undefined for customers who have the Integration already set up.

packages/destination-actions/src/destinations/attio/groupWorkspace/index.ts
```diff
@@ -101,6 +103,37 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          }
+        },
+        received_at: item.received_at.toString()
```
joe-ayoub-segment, line 134: received_at needs to be handled defensively as it will be undefined for customers who have the Integration already set up.

joe-ayoub-segment (changes requested):
Hi @nickcampbell18 thanks for raising this PR. 
Generally looks good. 
1 major but hopefully simple to fix issue. These 3 new fields you are adding will resolve to undefined for all customers' instances which have already been set up prior to the deploy. So that needs to be handled in code defensively. 

Also note that this means that batching will be disabled for these customers' instances. To enable batching the customer would need to edit the 'Action' , and save the Action. However because the enable_batching field is hidden it might not be possible. 

What's your intention here?

nickcampbell18:
Thanks for the feedback! I've addressed both of those points.

What do you think of the wording here on the second commit? At the moment Attio won't tell the user if their mapping is causing validation failures on our side (although you can still use the event tester I guess to see if it would generally work). I can send a PR to the [docs page](https://github.com/segmentio/segment-docs/blob/develop/src/connections/destinations/catalog/actions-attio/index.md) as well, to clarify if that's easier...

joe-ayoub-segment (approved):

joe-ayoub-segment:
Looks good. I might reword the batch label and description after I merge the code.

joe-ayoub-segment:
hi @nickcampbell18 PR deployed. Please check it all looks good!

PR #2064, opened 2024-05-30
[pendo-web-actions] add setting to exclude certain segment users

wordythebyrd:
Allows end-users to set a specific user trait in Segment to control whether Pendo should initialize and receive events for that particular user.

The end-user defines an exclusion user trait in the settings. This trait acts as a flag to determine if a user should be excluded from Pendo tracking.

This feature provides flexibility to end-users by allowing them to control Pendo tracking on a per-user basis. They can set the designated user trait in Segment to true or 'true' for specific users they want to exclude from Pendo tracking. This can be useful in scenarios where certain users, such as internal team members or test accounts, should not be tracked by Pendo.

### DIFF
--- a/packages/browser-destinations/destinations/pendo-web-actions/src/generated-types.ts
+++ b/packages/browser-destinations/destinations/pendo-web-actions/src/generated-types.ts
@@ -14,6 +14,10 @@ export interface Settings {
    */
   cnameContentHost?: string
   /**
+   * Specify a Segment user trait, which, if exists and is set to true, will prevent Pendo initialization and actions for the Segment user.
+   */
+  exclusionUserTrait?: string
+  /**
    * Override sending Segment's user traits on load. This will prevent Pendo from initializing with the user traits from Segment (analytics.user().traits()). Allowing you to adjust the mapping of visitor metadata in Segment's identify event.
    */
   disableUserTraitsOnLoad?: boolean
--- a/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
+++ b/packages/browser-destinations/destinations/pendo-web-actions/src/group/index.ts
@@ -55,6 +55,10 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
     }
   },
   perform: (pendo, { mapping, payload }) => {
+    if (!pendo) {
+      return
+    }
+
     // remove parentAccountData field data from the accountData if the paths overlap
 
     type pathMapping = {
--- a/packages/browser-destinations/destinations/pendo-web-actions/src/identify/index.ts
+++ b/packages/browser-destinations/destinations/pendo-web-actions/src/identify/index.ts
@@ -30,6 +30,10 @@ const action: BrowserActionDefinition<Settings, PendoSDK, Payload> = {
     }
   },
   perform: (pendo, event) => {
+    if (!pendo) {
+      return
+    }
+
     const payload: PendoOptions = {
       visitor: {
         ...event.payload.visitorData,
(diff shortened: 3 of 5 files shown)


joe-ayoub-segment (changes requested):
hi @wordythebyrd thanks for raising this PR. 

It seems that you would like a way to prevent certain users from having data sent by Segment to Pendo. 

I think that the approach taken here will partially work. I'll try and explain below: 

1. Anonymous user visits website
2. Segment JS loads and creates an anonymous user profile in the browser (with an anonymousId and no traits)
3. The user clicks on things, potentially sending data to Segment and Pendo
4. The exclusionUserTrait trait gets set on the user in the browser, which tells Segment to not send any more events to Pendo. 
5. No more events get sent to Pendo from the web page. 

As you can see from step 3 above events will continue to be sent to Pendo until the exclusionUserTrait value is set in the browser. 

Is this as you expect it to work?

--------------------

This is how I would recommend doing this: 

1. Let the end user specify a boolean querystring value which then gets set as a cookie. 
2. e.g. preventPendo=true
3. On Initialize, check for the querystring value. If it is there, save the value to a cookie. Then check if the cookie exists. If it does exist, prevent the Pendo library from being intialized with Segment

We wouldn't even need any new fields to be added to the Setting or into the Action code. We'd just need to modify the initialize() function. 

We can then update the docs to explain how to use it. 

Does that sound like an approach you'd be willing to go with?

joe-ayoub-segment:
You could use some code like this: 

`import { UniversalStorage } from '@segment/analytics-next'`

Then in the initialize() function: 

   ```
const storage = analytics.storage as UniversalStorage<Record<string, string>> 

   const urlParams = new URLSearchParams(window.location.search)
   const qStringValue: string | null = urlParams.get('preventpendo') || null

   if (qStringValue === 'true' || qStringValue === 'false') {
      storage.set('preventPendoCookie', qStringValue)
   }
   
   const shouldPrevent: 'true' | 'false' | null = storage.get('preventPendoCookie')
```

   // you can then use the shouldPrevent to cancel the initialization of the library.

wordythebyrd:
Thanks @joe-ayoub-segment, I'm discussing the options with my team.

Is it fair to summarize the issue with my approach as "Blocking certain users from having data sent by Segment to Pendo using a Segment user trait will not prevent anonymous user data from being sent to Pendo"

wordythebyrd:
@joe-ayoub-segment, as long as the client sent a Segment identify call immediately with the exclusionUserTrait, it would work right?  I see in the docs that Segment does allow traits to be sent for anonymous users in an identify call. (source: https://segment.com/docs/connections/spec/identify/#anonymous-id)

joe-ayoub-segment:
hi @wordythebyrd, I would summarize the issue as follows: 

"Attempting to block certain users from having data sent by Segment to Pendo will work only some of the time, as the approach relies on the user having the 'blocker trait' value set before any Segment events get triggered". 

It's still likely that a track() event would be triggered from the web page before the identify() event which sets the trait.  

Does that make sense?

That's why I am suggesting we look at an alternate more watertight approach. 

Also - this all depends on why this feature is aimed at. i.e do you want this feature just for your own internal testing, or is it likely that customers will use this featute? If it's the latter, I think we should come up with a solution that guarantees no customer data getting sent. 

Happy to discuss on a call. Here's my Calendly. https://calendly.com/joe_ayoub/
Note that I'm on PTO from Wednesday 12 June. Back Thursday 13 June. 

How urgent is this feature?

Thanks,
Joe

wordythebyrd:
@joe-ayoub-segment Turns out, after reviewing what the client was trying to do, I was able to resolve this without needing a change to our destination.  If your curious, they were conditionally reseting Segment and stubbing out the identify, group, track, and page actions if the user was a super admin, I was able to advise them to also call `pendo.teardown()` which stops all pendo event collection, guides, and essentially makes it so the agent isn’t running.

joe-ayoub-segment:
Thanks @wordythebyrd - sounds good!

PR #2065, opened 2024-05-31
[DC-789] adds subscription sync modes support in destination-kit

konoufo:
This PR adds Sync Modes support to destination-kit as part of the Mappings 2.0 project (See relevant section in the [blueprint](https://docs.google.com/document/d/1h-T40iySIxC9Zm6HVEiM-c0otTVmBUWn7dSamyFebrQ/edit#heading=h.ez9jlsj1msqz)) .

- [x] Adds ability to define sync modes setting for any action
- [x] Injects selected `syncMode` in perform block.
- [x] Adds ability to define `SyncModeCondition` (will allow conditionally displaying fields in app)

## Testing

### DIFF
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -12,7 +12,7 @@ import type {
   ExecuteInput,
   Result,
   SyncMode,
-  SyncModes
+  SyncModeDefinition
 } from './types'
 import { syncModeTypes } from './types'
 import { NormalizedOptions } from '../request-client'
@@ -103,8 +103,8 @@ export interface ActionDefinition<
     >
   }
 
-  /** The sync modes setting definition. This enables subscription sync mode selection when subscribing to this action. */
-  syncModes?: SyncModes
+  /** The sync mode setting definition. This enables subscription sync mode selection when subscribing to this action. */
+  syncMode?: SyncModeDefinition
 }
 
 export const hookTypeStrings = ['onMappingSave', 'retlOnMappingSave'] as const
@@ -280,7 +280,7 @@ export class Action<Settings, Payload extends JSONLikeObject, AudienceSettings =
       }
     }
 
-    const syncMode = this.definition.syncModes ? bundle.mapping?.['__segment_internal_sync_mode'] : undefined
+    const syncMode = this.definition.syncMode ? bundle.mapping?.['__segment_internal_sync_mode'] : undefined
 
     // Construct the data bundle to send to an action
     const dataBundle = {
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -262,6 +262,11 @@ export class Action<Settings, Payload extends JSONLikeObject, AudienceSettings =
     // Remove empty values (`null`, `undefined`, `''`) when not explicitly accepted
     payload = removeEmptyValues(payload, this.schema, true) as Payload
 
+    // Remove internal hidden field
+    if (bundle.mapping && '__segment_internal_sync_mode' in bundle.mapping) {
+      delete payload['__segment_internal_sync_mode']
+    }
+
     // Validate the resolved payload against the schema
     if (this.schema) {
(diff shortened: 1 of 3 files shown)


packages/core/src/destination-kit/types.ts
```diff
@@ -191,18 +193,35 @@ export interface InputField extends InputFieldJSONSchema {
...
+  type?: 'field' // optional for backwards compatibility
   value: Omit<FieldValue, 'Directive'> | Array<Omit<FieldValue, 'Directive'>> | undefined
 }
 
+/**
+ * A single condition defining whether a field should be shown based on the current sync mode.
+ * operator: The operator to use when comparing the sync mode
+ * value: The value to compare against, if undefined, we will match based on whether a sync mode is set or not
+ */
+export interface SyncModeCondition extends BaseCondition {
```
nick-Ag, line 218: I'm wondering how this will be used, is there some example code of a destination that implements `SyncModeCondition`? Will it be something like 'if condition.type == 'syncMode' && condition.value === "add"' -> do not show some field that is irrelevant to add?

packages/core/src/destination-kit/types.ts
```diff
@@ -191,18 +193,35 @@ export interface InputField extends InputFieldJSONSchema {
...
+ * A single condition defining whether a field should be shown based on the value of the field specified by `fieldKey`.
  * fieldKey: The field key in the fields object to look at
  * operator: The operator to use when comparing the field value
  * value: The value we expect that field to have, if undefined, we will match based on whether the field contains a value or not
  */
-export interface Condition {
+export interface FieldCondition extends BaseCondition {
   fieldKey: string
-  operator: 'is' | 'is_not'
+  type?: 'field' // optional for backwards compatibility
```
nick-Ag, line 209: I think it'd be worth it to make `type` required now since there are only a few places using it that you'd have to update

packages/core/src/destination-kit/types.ts
konoufo: > I'm wondering how this will be used, is there some example code of a destination that implements `SyncModeCondition`? Will it be something like 'if condition.type == 'syncMode' && condition.value === "add"' -> do not show some field that is irrelevant to add?

Good question! There is no example code as of now. This is a completely new thing. And it's going to be used exactly as you described to hide fields in app e.g Action editor. Any concerns?

packages/core/src/destination-kit/types.ts
konoufo: Oh interesting. I found that I'd need to make changes in app and this repo. And that's just what I could find quickly. Is it possible to break compatibility anywhere else? Can you help round up all the places this could break?

packages/core/src/destination-kit/types.ts
konoufo: Also it's good to point out that the type itself might not be used directly outside this module so maybe an export is not justified?

packages/core/src/destination-kit/types.ts
konoufo: Example of code that seems like it would break if `type` was required: https://github.com/segmentio/app/blob/fce0074dd2ffe440d67d7fbd895173b7b41b15b1/packages/app/client/containers/DestinationActions/components/MappingInput/use-dependent-fields.test.tsx#L41-L50

pooyaj (dismissed):
Looks great 🙌 couple of comments otherwise lgtm

packages/core/src/destination-kit/action.ts
```diff
@@ -264,6 +280,8 @@ export class Action<Settings, Payload extends JSONLikeObject, AudienceSettings =
       }
     }
 
+    const syncMode = this.definition.syncMode ? bundle.mapping?.['__segment_internal_sync_mode'] : undefined
```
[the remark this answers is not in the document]
konoufo: Yes it would show up. Ok let's get rid of it.

pooyaj (approved):

maryamsharif (approved):

PR #2067, opened 2024-05-31
Fix to show request headers in Action's Tester UI

tcgilbert:
Currently in the Actions Tester, the headers in the displayed request object are always empty:

image

This is because, when converting the request and response made by the action into JSON ([return res.status(200).json(debug)](https://github.com/segmentio/action-destinations/blob/main/packages/cli/src/lib/server.ts#L313)) an empty object will be returned for the request headers:

image

> Side Note: I can't seem to figure out why the headers within the response obj are able to be parsed, as they appear to be the same type, though perhaps there is a difference under the hood

This fix converts the request headers into a standard JavaScript object in order that they can be presented in the Actions Tester UI. I elected to redact certain headers that my contain sensitive creds, not sure if this would be an exhaustive list, so if you think we should simply remove the headers entirely that is also an option. 

## Testing

tested locally: 

Screenshot 2024-05-31 at 11 33 45 AM


nick-Ag (approved):
Thanks for noticing this and addressing it! 🔥

joe-ayoub-segment:
hi @tcgilbert - deploy complete -  Please confirm the change looks good.

PR #2071, opened 2024-06-04
Tiktok Pixel & Web Conversions Destination for additional PII support

jae-rhee-tiktok:
Adding new mappings for all presets to include default mappings for additional PII info:
- first name
- last name
- city
- state
- country
- zipcode


joe-ayoub-segment:
Hi @jae-rhee-tiktok .  This PR has been deployed. Please check that it works as expected :)

PR #2077, opened 2024-06-06
Bryan/audience to list

bwebs:
Adding new audienceToList functionality for Engage Audience to Courier Lists

## Testing


packages/destination-actions/src/destinations/courier/audienceToList/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    return processPayload(request, payload, settings)
+  },
+
+  performBatch: (request, { payload, settings }) => {
+    const promises = payload.map((event) => {
+        return processPayload(request, event, settings)
+    })
+    
+    return Promise.all(promises)
+  }
```
joe-ayoub-segment, line 112: Looks like this just splits out a batch of events into individual API calls. 
I suggest we remove the performBatch() implementation as you've defaulted the enable_batching field to false and made it hidden anyway (so performBatch will never be called anyway).

packages/destination-actions/src/destinations/courier/audienceToList/index.ts
```diff
@@ -0,0 +1,135 @@
...
+    },
+    enable_batching: {
+      type: 'boolean',
+      label: 'Batch events',
+      description:
+        'When enabled, the action will batch events before sending them to Courier. In most cases, batching should be enabled.',
+      required: false,
+      default: false,
+      unsafe_hidden: true,
+    },
```
joe-ayoub-segment, line 89: If you don't intend to handle batches of events, then just delete this field.

packages/destination-actions/src/destinations/courier/audienceToList/index.ts
```diff
@@ -0,0 +1,135 @@
+import { PayloadValidationError, type ActionDefinition, type ModifiedResponse, type RequestOptions } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { getDomain } from '..'
+
+export const CONSTANTS = {
+
+  LD_API_CUSTOM_AUDIENCE_ENDPOINT: '/segment-targets/segment-audiences',
```
joe-ayoub-segment, line 8: Let's remove this constant as it is not used.

joe-ayoub-segment (changes requested):
hi @bwebs thanks for raising this PR. 

Looks like it's a copy of the LaunchDarkly implementation - which is fine. 

I left some comments inline. 

Some additional questions: 

Looks like you are sending the 'segment_audience_id' to your platform as the 'list_id'. Please note that the 'segment_audience_id' looks like a GUID and will probably not be something the customer will recognise on your platform. If your platform is able to accept a more friendly name, you could also send the 'segment_audience_key' value. This is a snake cased version of the Audience name the customer typed into Segment when they created the Audience. I would only send this in addition to the segment_audience_id and not as a replacement, as it would give you something nicer to display to the end user. Does that make sense?

Also - we do have a slightly more robust and flexible way to build Audience Destination now, which involved a little more code. However the advantage is that Segment can make an API call to your platform so you can initialize the Audience before Segment starts sending data to populate the Audience. Please let me know if this is something you would be interested in.

packages/destination-actions/src/destinations/courier/audienceToList/index.ts
bwebs: 👍 removed performBatch until we support a match endpoint

packages/destination-actions/src/destinations/courier/audienceToList/index.ts
bwebs: 👍 removed

joe-ayoub-segment (dismissed):

PR #2084, opened 2024-06-10
disabling iam until it is fixed

joe-ayoub-segment:
Kafka IAM auth isn't working with MSK. Because of this we're disabling it until a fix is implemented. 

## Testing

Tested locally.


joe-ayoub-segment:
hi @varadarajan-tw - looks like the labeller didn't add any labels. There's no error though...
It was a 1 line change in a list of choices in a field which was changed. Any thoughts on why there are no labels?

varadarajan-tw:
Hey @joe-ayoub-segment , the labeler added labels but it seems like you removed it accidentally or there was some lag when you did the update.
image

seg-leonelsanches (approved):

joe-ayoub-segment:
Note: PR deployed

PR #2085, opened 2024-06-10
[MAIN] [STRATCONN] Added conditional fields on Intercom

AnkitSegment:
JIRA TICKETS: https://segment.atlassian.net/browse/STRATCONN-3852

https://github.com/segmentio/action-destinations/assets/139338151/43c0627d-4632-4f87-aaa2-6690f738e6a4

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
+++ b/packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
@@ -25,6 +25,15 @@ const action: ActionDefinition<Settings, Payload> = {
       label: 'External ID',
       default: {
         '@path': '$.userId'
+      },
+      depends_on: {
+        conditions: [
+          {
+            fieldKey: 'role',
+            operator: 'is',
+            value: 'user'
+          }
+        ]
       }
     },
     email: {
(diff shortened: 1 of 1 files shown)


packages/destination-actions/src/destinations/intercom/identifyContact/index.ts
```diff
@@ -25,6 +25,15 @@ const action: ActionDefinition<Settings, Payload> = {
       label: 'External ID',
       default: {
         '@path': '$.userId'
+      },
```
nick-Ag, line 28: The description for this field says 'External IDs cannot be set if the role is `lead`' but the condition defined here will only show the external ID field if role is lead. Shouldn't we hide this external ID field if role is user instead?

AnkitSegment: But for email description says, "Email is required if the role is `user` and External ID is blank."

nick-Ag: hmm - I don't think that would be relevant to `external_id`? I think the condition here should be `fieldKey: 'role', operator: 'is_not', value: 'lead'` based on how I'm reading the description

nick-Ag (approved):

joe-ayoub-segment:
hi @AnkitSegment PR deployed

PR #2099, opened 2024-06-14
Required Fields V2 - Conditionally Required Fields

nick-Ag:
This PR revamps the `required` property for action input fields such that a destination builder can define conditionally required fields. These builder defined conditional requirements are then compiled into a JSONSchema4 object, which is used to validate each mapping before the `perform` block is triggered.

**SDD: [Link](https://docs.google.com/document/d/12ouu-xmUfYhfXmyesC8aNPvKKMwhIuM2OpQAj7x_5Lw/edit?tab=t.0#heading=h.3gvugq439bq1)**

### Conditional Field Example:
For example, the Salesforce Lead `last_name` field is only required when a user is creating a lead. We can now encode that in the definition of the field directly like this:
```
    last_name: {
      label: 'Last Name',
      description: "The lead's last name. **This is required to create a lead.**",
      type: 'string',
      ...
      required: {
        conditions: [
          {
            fieldKey: 'operation',
            operator: 'is',
            value: 'create'
          }
        ]
      },
```
This builder defined condition will then be compiled into a JSONSchema like this:
```
{
  '$schema': 'http://json-schema.org/schema#',
  type: 'object',
  additionalProperties: false,
  properties: {
    operation: {
      title: 'Operation',
	...
    },
	...
    last_name: {
      title: 'Last Name',
	...
    },
	...
  },
  required: [ 'operation' ],
  if: {
    properties: {"operation": {"const": "create"}}
  },
  then: { "required": ["last_name"] }
}

```
And eventually will validate Payloads such as these:
`{ operation: 'create', first_name: 'nick' } // invalid - missing last_name`
`{ operation: 'create', last_name: 'aguilar' } // valid - required last_name is included`
`{ operation: 'update', first_name: 'nick' } // valid - last_name is not required`

## Reviewing this PR:
~2800 lines added:
* 1400 - schema-validation.test.ts.snap - this documents the JSON schema generated for each unit test. This is automated and viewing each of these generated schemas isn't important
* 1100 - schema-validation.test.ts - New unit tests which run various combinations of different conditions a builder can define
* 250 - fields-to-jsonschema.ts - The JSON Schema generation logic

## Testing
Deployed successfully to staging. I have updated the Salesforce Lead action's `last_name` field with a conditional requirement.

As of now these conditions are not implemented in the UI, and I am only expecting new AJV error messages when the conditional requirement is not met.

The conditional requirement is enforced when `operation: create` but `last_name` is missing:
Screenshot 2024-12-02 at 10 19 17 AM

When `operation: create` and `last_name` is defined, the condition is met and event delivery succeeds:
Screenshot 2024-12-02 at 10 21 05 AM

When `operation: update` the condition is not enforced and `last_name` can be undefined:
Screenshot 2024-12-02 at 10 26 43 AM

## Merge & Deploy Timing
This PR will be a part of a series of PRs. It should not be merged or deployed before these other PRs are ready.
@joe-ayoub-segment - I will handle the final deploy of this one
 
`app`: https://github.com/segmentio/app/pull/22618
`action-cli`: https://github.com/segmentio/action-cli/pull/196 💬


joe-ayoub-segment:
hi @nick-Ag - I know this is still in draft, so just out of curiosity what will this do?

```
  groupConditions: {
    anyOf: ['external_id', 'user_alias', 'email', 'braze_id']
  },
```

Will it hide or show fields in the UI depending on which fields have a mapping populated?
Or will it perform a runtime validation which will prevent the perform() function from being invoked if none of the fields have a value?

Keen to understand the intent here as I think the second approach would be really valuable.

nick-Ag:
> hi @nick-Ag - I know this is still in draft, so just out of curiosity what will this do?
> 
> ```
>   groupConditions: {
>     anyOf: ['external_id', 'user_alias', 'email', 'braze_id']
>   },
> ```
> 
> Will it hide or show fields in the UI depending on which fields have a mapping populated? Or will it perform a runtime validation which will prevent the perform() function from being invoked if none of the fields have a value?
> 
> Keen to understand the intent here as I think the second approach would be really valuable.

Planning on the second approach primarily. I also want the validation to appear in the UX so that users can't save mappings where the group condition has not been met. Looking forward to showing more soon!

packages/destination-actions/src/destinations/salesforce/lead/index.ts
```diff
@@ -53,6 +53,15 @@ const action: ActionDefinition<Settings, Payload> = {
           else: { '@path': '$.properties.last_name' }
         }
       },
+      required: {
```
maryamsharif, line 56: Would this not break current implementation because you are making this field required in certain instances where it wasn't before?

joe-ayoub-segment:
hi @nick-Ag I can't wait to start using this new feature :). 
Can you tell me what happens when null, '' and undefined values are passed in to a string field which is conditionally required? Also what if the field is defined as AllowNull: true?

joe-ayoub-segment:
Hi @nick-Ag , qq: 
- does/should the generated-types.ts file change if conditions are used?
- Can you show examples how to use multiple conditions, with all/any (assuming those use-cases are supported)?

nick-Ag:
> Hi @nick-Ag , qq:
> 
> * does/should the generated-types.ts file change if conditions are used?
> * Can you show examples how to use multiple conditions, with all/any (assuming those use-cases are supported)?

* The generated-types file won't change, fields which are conditionally required will be optional (ex: `last_name?: string`). [See here](https://github.com/segmentio/action-destinations/pull/2099/files#diff-c8bb6098d0d776248dd55165e9a6081ecb18d68b3599dc27cb8a4ac20591f788R351)
* Those cases are supported, see the [unit test here](https://github.com/segmentio/action-destinations/pull/2099/files#diff-a1b513c8f4198f826e3c54fd6ee5e3c7ac79826d5305ec5bdc7a8ea54967d0b1R310) for an example. I've tried to test every possible use-case I could think of in that test file

maryamsharif (approved):

nick-Ag:
> hi @nick-Ag I can't wait to start using this new feature :). Can you tell me what happens when null, '' and undefined values are passed in to a string field which is conditionally required? Also what if the field is defined as AllowNull: true?

I'm going to make a small update to the PR to handle `null` values as well, thanks for pointing out this case

packages/core/src/__tests__/schema-validation.test.ts
```diff
@@ -61,6 +62,1170 @@ const schema = fieldsToJsonSchema({
...
+      isValid = validateSchema(b_not_required_mappings[0], schema, { throwIfInvalid: false })
+      expect(isValid).toBe(true)
+
+      isValid = validateSchema(b_not_required_mappings[1], schema, { throwIfInvalid: false })
+      expect(isValid).toBe(true)
+    })
+  })
+
+  // TODO: support sync mode conditions
+  describe.skip('should validate based on sync mode value', () => {})
```
varadarajan-tw, line 1226: Is this something you would be tackling in a separate PR?

packages/core/src/__tests__/schema-validation.test.ts
```diff
@@ -61,6 +62,1170 @@ const schema = fieldsToJsonSchema({
...
+        type: 'string',
+        label: 'b',
+        description: 'b',
+        required: {
+          conditions: [{ fieldKey: 'a', operator: 'is', value: undefined }]
+        }
+      }
+
+      const schema = fieldsToJsonSchema(mockActionFields)
+      console.log('schema', JSON.stringify(schema, null, 2))
```
varadarajan-tw, line 1198: nit:should this be removed?

packages/core/src/__tests__/schema-validation.test.ts
```diff
@@ -61,6 +62,1170 @@ const schema = fieldsToJsonSchema({
...
+      expect(schema).toMatchSnapshot()
+
+      const b_required_mappings = [{ a: null }, { a: null, b: 'b_value' }, {}, { b: 'b_value' }]
```
varadarajan-tw, line 1201: Are we treating null as undefined here? Is there a reason why you decided to do so? 
How would we differentiate between absence of a field and a field with null value?

varadarajan-tw (commented):
Fantastic work! Great tests! 
Left just one question about nulls.

packages/core/src/__tests__/schema-validation.test.ts
nick-Ag: Yes going to follow-up for this 👍

packages/core/src/__tests__/schema-validation.test.ts
```diff
@@ -61,6 +62,1170 @@ const schema = fieldsToJsonSchema({
...
+        required: true,
+        description: 'a'
+      }
+
+      mockActionFields['b'] = {
+        label: 'b',
+        type: 'string',
+        description: 'b',
+        required: {
+          conditions: [{ fieldKey: 'a', operator: 'is', value: 'a_value' }]
```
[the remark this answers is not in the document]
nick-Ag: Not quite, the unit test was testing if b isn't required, not whether the `is_not` operator works correctly. I've merged this unit test with the previous one since they were sharing the same schema and hopefully the test name makes more sense

joe-ayoub-segment:
hi @nick-Ag. Couple of questions about how this code will behave: 

1) what will happen in the following scenario? I assume the UI will not fail validation. However when the event gets sent for processing it will fail? Does this mean that validation on the front end needs to be slightly different to validation that happens when an event is being processed?

```
{
   ... 
   label: 'last_name'
   conditions: [
          {
            fieldKey: 'operation',
            operator: 'is',
            value: 'create'
          }
    ]
 },
{
   ... 
   label: 'operation'
   default: {
      '@path': '$.propertries.operation'
   }
 }
```

2. what will the error messages be when validation fails? How useful will they be? Are there any examples you can share? The reason I ask is that JSON schema error messages aren't always that helpful for the end user.

varadarajan-tw (dismissed):
Thanks for answering my questions! Nice work! ✨ This will be very useful feature saving us some lines of code and bring in some consistency!

nick-Ag:
> hi @nick-Ag. Couple of questions about how this code will behave:
> 
> 1. what will happen in the following scenario? I assume the UI will not fail validation. However when the event gets sent for processing it will fail? Does this mean that validation on the front end needs to be slightly different to validation that happens when an event is being processed?
> 
> ```
> {
>    ... 
>    label: 'last_name'
>    conditions: [
>           {
>             fieldKey: 'operation',
>             operator: 'is',
>             value: 'create'
>           }
>     ]
>  },
> {
>    ... 
>    label: 'operation'
>    default: {
>       '@path': '$.propertries.operation'
>    }
>  }
> ```
> 
> 2. what will the error messages be when validation fails? How useful will they be? Are there any examples you can share? The reason I ask is that JSON schema error messages aren't always that helpful for the end user.

1. Really good question here. In this case the UI will consider that a valid mapping and pass validation. However if there is no value at `$.properties.operation` for the event then the validation will fail at the actions level. The validation between the UI and event delivery is only different in the sense that UI validation checks only the users mapping (before path directives are applied) and the event delivery validation will validate against the final payload (path directives applied to the event). 
2. Here's an example of an error message in the case that last_name is missing. 
```
[
	{
		"statusCode": 500,
		"message": "The root value is missing the required field 'last_name'. The root value must match \"then\" schema.",
		"stack": [
			"AggregateAjvError: The root value is missing the required field 'last_name'. The root value must match \"then\" schema.",
...
```

nick-Ag:
> Looks great 🙌 This is not part of the scope? right?
> 
> ```
>     address: {
>       label: 'Address',
>       type: 'object',
>       properties: {
>         city: {
>           label: 'city',
>           type: 'string',
>           required: [Condition Here]
>         },
> ```
> 
> wondering if we handle it gracefully if someone puts a condition on the required field ( ideal if we can prevent it via typescript )

Fantastic question. Based on a unit test I just wrote that case is not handled correctly. Going to circle back and address it before merging. Will consider both preventing it entirely vs. implementing it correctly

joe-ayoub-segment:
> > Looks great 🙌 This is not part of the scope? right?
> > ```
> >     address: {
> >       label: 'Address',
> >       type: 'object',
> >       properties: {
> >         city: {
> >           label: 'city',
> >           type: 'string',
> >           required: [Condition Here]
> >         },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > wondering if we handle it gracefully if someone puts a condition on the required field ( ideal if we can prevent it via typescript )
> 
> Fantastic question. Based on a unit test I just wrote that case is not handled correctly. Going to circle back and address it before merging. Will consider both preventing it entirely vs. implementing it correctly

Nice question @pooyaj . My vote is to implement it if possible (if @nick-Ag can do it!). The reason is that Object fields are a neat way to squeeze a lot of fields into a small amount of page real estate. and they also allow us to locate related fields close together. It would be a missed opportunity if 'required' didn't work on these sub fields.

nick-Ag:
> > > Looks great 🙌 This is not part of the scope? right?
> > > ```
> > >     address: {
> > >       label: 'Address',
> > >       type: 'object',
> > >       properties: {
> > >         city: {
> > >           label: 'city',
> > >           type: 'string',
> > >           required: [Condition Here]
> > >         },
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > wondering if we handle it gracefully if someone puts a condition on the required field ( ideal if we can prevent it via typescript )
> > 
> > 
> > Fantastic question. Based on a unit test I just wrote that case is not handled correctly. Going to circle back and address it before merging. Will consider both preventing it entirely vs. implementing it correctly
> 
> Nice question @pooyaj . My vote is to implement it if possible (if @nick-Ag can do it!). The reason is that Object fields are a neat way to squeeze a lot of fields into a small amount of page real estate. and they also allow us to locate related fields close together. It would be a missed opportunity if 'required' didn't work on these sub fields.

Have implemented that use case & updated the README with some documentation about this feature. This one is ready to go out. Looking for final review @joe-ayoub-segment , @varadarajan-tw , @pooyaj

README.md
```diff
@@ -358,6 +358,72 @@ const destination = {
...
+            match: 'all',
+            conditions: [
+              {
+                fieldKey: 'operation',
+                operator: 'is',
+                value: 'create'
+              }
+            ]
+          }
+        },
```
joe-ayoub-segment, line 389: hi @nick-Ag . 
Can we include an example with dot notation for a reference to an object.property field?

joe-ayoub-segment (approved):
Reviewed the functionality with @nick-Ag . I'm not the best person to review the code though, so please have someone who understands this part of the codebase do a review.

PR #2119, opened 2024-06-27
[DC-792] injects matchingKey in the perform block

konoufo:
This PR adds the ability to access user-selected record matching field in the perform block. The `matchingKey` is the field key for one of the action fields of category `identifier`. The record matching field is meant to be selected via a dropdown in the mapping UI for RETL (upcoming work). [See the relevant section of the blueprint for more context.](https://docs.google.com/document/d/1h-T40iySIxC9Zm6HVEiM-c0otTVmBUWn7dSamyFebrQ/edit#heading=h.2vwlxsapfz26)

## Testing

### DIFF
--- a/packages/core/src/__tests__/destination-kit.test.ts
+++ b/packages/core/src/__tests__/destination-kit.test.ts
@@ -145,6 +145,33 @@ const destinationWithSyncMode: DestinationDefinition<JSONObject> = {
   }
 }
 
+const destinationWithIdentifier: DestinationDefinition<JSONObject> = {
+  name: 'Actions Google Analytics 4',
+  mode: 'cloud',
+  actions: {
+    customEvent: {
+      title: 'Send a Custom Event',
+      description: 'Send events to a custom event in API',
+      defaultSubscription: 'type = "track"',
+      fields: {
+        userId: {
+          label: 'User ID',
+          description: 'The user ID',
+          type: 'string',
+          required: true,
+          category: 'identifier'
+        }
+      },
+      perform: (_request, { matchingKey }) => {
+        return ['this is a test', matchingKey]
+      },
+      performBatch: (_request, { matchingKey }) => {
+        return ['this is a test', matchingKey]
+      }
+    }
+  }
+}
+
 const destinationWithDynamicFields: DestinationDefinition<JSONObject> = {
   name: 'Actions Dynamic Fields',
   mode: 'cloud',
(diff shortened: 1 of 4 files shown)


packages/core/src/__tests__/destination-kit.test.ts
```diff
@@ -145,6 +145,33 @@ const destinationWithSyncMode: DestinationDefinition<JSONObject> = {
...
+      title: 'Send a Custom Event',
+      description: 'Send events to a custom event in API',
+      defaultSubscription: 'type = "track"',
+      fields: {
+        userId: {
+          label: 'User ID',
+          description: 'The user ID',
+          type: 'string',
+          required: true,
+          category: 'identifier'
```
nick-Ag, line 162: Still trying to understand the multiple identifiers case, suppose we had an additional `identifier` field 'email' here. Then we would automatically push a 'Record Matching' dropdown field that users would pick userID or email from, and depending on that the `matchingKey` would pull the correct value from the payload for the perform block? Is that a good understanding?

konoufo: > Still trying to understand the multiple identifiers case, suppose we had an additional `identifier` field 'email' here. Then we would automatically push a 'Record Matching' dropdown field that users would pick userID or email from, and depending on that the `matchingKey` would pull the correct value from the payload for the perform block? Is that a good understanding?
@nick-Ag Exactly 💯 . So in the UI, the selected field out of all the `identifier` fields defined would have been stored in `mapping['__segment_internal_matching_key']`, which is exactly what `matchingKey` is set to.

nick-Ag (approved):

packages/core/src/destination-kit/action.ts
```diff
@@ -197,6 +197,13 @@ const isSyncMode = (value: unknown): value is SyncMode => {
   return syncModeTypes.find((validValue) => value === validValue) !== undefined
 }
 
+const INTERNAL_HIDDEN_FIELDS = ['__segment_internal_sync_mode', '__segment_internal_matching_key']
```
pooyaj, line 200: Nice! more tidy than before 👏

packages/core/src/destination-kit/action.ts
```diff
@@ -297,6 +302,10 @@ export class Action<Settings, Payload extends JSONLikeObject, AudienceSettings =
 
     const syncMode = this.definition.syncMode ? bundle.mapping?.['__segment_internal_sync_mode'] : undefined
 
+    const matchingKey = Object.values(this.definition.fields).find((field) => field.category === 'identifier')
```
pooyaj, line 305: Safe to remove this check to simplify? Can we just make an attempt to pass `bundle.mapping?.['__segment_internal_matching_key']` ... I think will be undefined if the destination does not support record matching anyways.

pooyaj (approved):
Small nit otherwise looks great 🙌

packages/core/src/destination-kit/action.ts
konoufo: > Safe to remove this check to simplify? Can we just make an attempt to pass `bundle.mapping?.['__segment_internal_matching_key']` ... I think will be undefined if the destination does not support record matching anyways.

Yeah I guess I was trying to be defensive lol. By the way, do you think there should be validation to make sure that the value in `bundle.mapping?.['__segment_internal_matching_key']` is one of the `identifier` fields? 
I'm on the fence. On one hand the UI is unlikely to cause invalid values but on another hand, things happen e.g someone can save a bad mapping via PAPI?

pooyaj (approved):

packages/core/src/destination-kit/action.ts
pooyaj: I considered removing the condition to simplify future changes. If we decide to change the field category used as the indicator for matching keys, fewer parts of the codebase will need to be updated.

nick-Ag (approved):

packages/core/src/destination-kit/action.ts
pooyaj: Also no need for validation since even if someone creates a `__segment_internal_matching_key` with bad value it won't cause any significant issues in the perform block

joe-ayoub-segment:
hi @konoufo PR deployed

PR #2121, opened 2024-06-28
[STRATCONN-3905] [OrderComplete] Add features to bring parity with klaviyo classic

varadarajan-tw:
[JIRA](https://segment.atlassian.net/browse/STRATCONN-3905)

Klaviyo classic destination makes some additional formatting before submitting Order Complete events to klaviyo API. This PR aims to port most of the formatting to Klaviyo Actions. Key changes are
- Adds sub-fields to `properties` and `products` object fields. None of the new fields are required and hence change should be backward compatible. In case customers had defined some other properties, they'll be retained as `additionalProperties:true` attribute has been added.
- Formats property names sent to destination to title case by default. [cloud classic destination](https://github.com/segmentio/integrations/blob/c1042ca507274d7a5ccdd4ca8b06bf68b8158642/integrations/klaviyo/lib/mapper.js#L138), [device classic destination](https://github.com/segmentio/analytics.js-integrations/blob/e09a39466bd1288fe496ead63acebb7ee5510d10/integrations/klaviyo/lib/index.js#L268) has inconsistent formatting for same fields. So, went with the latest [Klaviyo docs](https://developers.klaviyo.com/en/docs/guide_to_integrating_a_platform_without_a_pre_built_klaviyo_integration#placed-order)
- Sets `value` to product price instead of order value in `Products Ordered` event.  
- Fixes default subscription for trackEvent and orderCompleted actions
- Adds two new mappings to allow overriding Order Complete and Products Ordered event names.

## Testing

Test results can be found [here](https://docs.google.com/document/d/1-7HsTAGabosgiLbxY-O8qZuWZ4pF4x2UPsscOUrtEXQ/edit)

Diff in order completed event before and after

image

Diff in ordered product event before and after.

We were repeating Order complete properties which is not necessary.

image

### DIFF
--- a/packages/destination-actions/src/destinations/klaviyo/orderCompleted/index.ts
+++ b/packages/destination-actions/src/destinations/klaviyo/orderCompleted/index.ts
@@ -39,11 +39,11 @@ function createOrderCompleteEvent(payload: Payload) {
       SKU: sku,
       Name: name,
       Quantity: quantity,
-      ItemPrice: price,
-      RowTotal: price,
+      'Item Price': price,
+      'Row Total': price,
       Categories: [category],
-      ProductURL: product['product_url'] ?? product['productURL'],
-      ImageURL: product['image_url'] ?? product['imageUrl'],
+      'Product URL': product['product_url'] ?? product['productURL'],
+      'Image URL': product['image_url'] ?? product['imageUrl'],
       ...customProps
     }
   })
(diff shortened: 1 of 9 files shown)


packages/destination-actions/src/destinations/klaviyo/orderCompleted/index.ts
```diff
@@ -155,18 +156,108 @@ const action: ActionDefinition<Settings, Payload> = {
...
+  return {
+    data: {
+      type: 'event',
+      attributes: {
+        properties: {
+          Categories: categories,
+          ItemNames: itemNames,
+          // products array is reformatted and sent as items
+          Items: items,
+          ...convertKeysToTitleCase(payload.properties)
```
varadarajan-tw, line 190: I am not sure if i should be doing this. Classic destination only captilizes the reserved/allowlisted fields and leaves the rest as such. 

The only requirement was to add `ItemNames`, `Categories` and `Items`. However, the formatting of property names in these events is inconsistent across [cloud classic destination](https://github.com/segmentio/integrations/blob/c1042ca507274d7a5ccdd4ca8b06bf68b8158642/integrations/klaviyo/lib/mapper.js#L138), [device classic destination](https://github.com/segmentio/analytics.js-integrations/blob/e09a39466bd1288fe496ead63acebb7ee5510d10/integrations/klaviyo/lib/index.js#L268)

[Klaviyo docs](https://developers.klaviyo.com/en/docs/guide_to_integrating_a_platform_without_a_pre_built_klaviyo_integration#placed-order) had all properties capitalized. So, I implemented this.

varadarajan-tw: The reason customers want ItemNames, Categories and Items added at the rootlevel is because only top level properties are available for segmentation.

https://developers.klaviyo.com/en/v2023-09-15/docs/custom_integration_faqs#how-should-i-structure-my-data-for-segmentation-and-flow-filtering

marinhero: > Klaviyo docs had all properties capitalized. So, I implemented this.

This makes sense but it also means that the behavior will change, right? I think we should document this behavior change in the docs of the destination.

marinhero (dismissed):

rvadera12 (approved):

maryamsharif (approved):

packages/destination-actions/src/destinations/klaviyo/orderCompleted/index.ts
varadarajan-tw: Sure. I'll work on updating the docs. Here, I have update the properties field description saying that segment will convert properties to title case by default.

sayan-das-in (approved):
LGTM !!!

PR #2133, opened 2024-07-02
Add array editor option for defaultObjectUI 

pooyaj:
Adding support for editable array of objects to support the following UI: 
Screenshot 2024-07-02 at 1 14 00 PM

## Testing
Testing not required as this is just a type update


nick-Ag (approved):
Just one quick question: Will this view always be the default for object arrays?

pooyaj:
@nick-Ag this won't be the default. The default view will be this one: 
Screenshot 2024-07-02 at 1 29 16 PM

With the default we have to map the parent field to an array in the payload before mapping sub-properties. With the `arrayeditor` we won't be able to map the parent field, but the UI allows composing an array like below by adding elements, and then mapping each sub-property in the element: 
`[{type: ‘car’, color: ‘...’}, {type: ‘bikes, color:’...’}]`

PR #2168, opened 2024-07-11
Update dynamic field types to support dynamic array of objects

pooyaj:
Updating types to support array of objects: 
Screenshot 2024-07-11 at 4 06 40 PM

Note: Since adjusting types to be stricter, had to update the code in a couple of places to make sure the types for the ActionDefinition is accurate 

## Testing
Testing not required as this is a non-functional change.

### DIFF
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -70,6 +70,9 @@ type GenericActionHookBundle = {
   }
 }
 
+// Utility type to check if T is an array
+type IsArray<T> = T extends (infer U)[] ? U : never
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export interface ActionDefinition<
   Settings,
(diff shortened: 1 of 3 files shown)


packages/core/src/destination-kit/action.ts
```diff
@@ -82,16 +85,27 @@ export interface ActionDefinition<
    * This is likely going to change as we productionalize the data model and definition object
    */
   dynamicFields?: {
-    [K in keyof Payload]?: Payload[K] extends object
+    [K in keyof Payload]?: IsArray<Payload[K]> extends never
```
nick-Ag, line 88: This is some gnarly typescript, but I think I know what it's saying 😅

nick-Ag (approved):

packages/core/src/destination-kit/action.ts
pooyaj: Lol, yeah, getting really complex 😵‍💫

PR #2171, opened 2024-07-12
Loops - add mailing lists

askkaz:
_Adds the Loops mailing list feature to Segment_

## Testing

_Tested manually with the action tester, works as expected._


packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
```diff
@@ -49,6 +49,13 @@ const action: ActionDefinition<Settings, Payload> = {
       required: false,
       default: { '@path': '$.traits.lastName' }
     },
+    mailingLists: {
+      label: 'Mailing Lists',
+      description:
+        'An object containing key-value pairs of mailing list IDs and true/false determining if the contact should be added to or removed from each list.',
+      type: 'object',
+      required: false
+    },
```
joe-ayoub-segment, line 58: hi @askkaz - just wondering if you tested this locally and if it worked as expected?

Does this field expect the following shape of data?

```
{
   "list_id_1": true, 
   "list_id_2": false,
    ... 
   "list_id_x": true
}
```

Should some validation be done in the perform() function to ensure that the object is in the correct shape?

There is a way to ensure type validation using a field like this, but the field would need to be an array, like this: 

```
[
  { "list_id": "list_id_1", value: true },
  { "list_id": "list_id_2", value: false },
    ... 
  { "list_id": "list_id_x", value: true },
]
```

The field configuration for this would look like this: 

   ```
 mailingLists: {
      label: 'Mailing Lists',
      description:
        'An an array of objects containing key-value pairs of mailing list IDs and true/false determining if the contact should be added to or removed from each list.',
      type: 'object',
      multiple: true,
      required: false, 
      properties: {
         list_id: {
            label: "List ID",  
            description "List ID for mailing list to add or remove the user from",
            type: "string", 
            required: true
         },
         value: {
            label: "value",  
            description "true indicates that the user is to be added to the list, false will remove the user from the list",
            type: "boolean", 
            required: true
         }
      }
    },
```

Note that this configuration will validate the payload so that mailingLists that are not an array of the correct shape will result in the payload being rejected with an invalid payload error. 

Please indicate your preference and I can deploy the code.

joe-ayoub-segment:
Hi @askkaz thanks for raising the PR. 
Just one comment - please see above.

joe-ayoub-segment:
hi @askkaz - checking in on this PR. is it still something you want to proceed with?

joe-ayoub-segment:
Closing PR for now. @askkaz please feel free to reopen when you have time to follow up. 
Kind regards,
Joe

askkaz:
ah sorry, missed this, we'll follow up!

danrowden:
@joe-ayoub-segment Hi Joe. Thanks for taking a look at this. We agree that your suggestion is better and we'd love to proceed with it

PR #2172, opened 2024-07-12
Iterable Lists, to be used with Audiences only

seg-leonelsanches:
This is different from [Iterable Action Destination](https://github.com/segmentio/action-destinations/tree/main/packages/destination-actions/src/destinations/iterable) because it operates in the scope of audiences only. 

One of our customers asked us to have Segment Audiences synced as Lists in Iterable. The endpoints are:

- https://api.iterable.com/api/docs#lists_subscribe
- https://api.iterable.com/api/docs#lists_create
- https://api.iterable.com/api/docs#lists_unsubscribe

Many implementation aspects of this implementation are very similar to Iterable Action Destination. In this implementation, the audience key (computation key) is used to create the list by name, and also fetch the List ID by name. 

This destination works with only two identifiers: `user_id` and `email`. Other traits can be sent through a `dataFields` property. 

## Testing

I need help testing this, since it is an Audience Destination.

### DIFF
--- a/packages/destination-actions/src/destinations/iterable-lists/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/iterable-lists/__tests__/index.test.ts
@@ -1,0 +1,18 @@
+import nock from 'nock'
+import { createTestEvent, createTestIntegration } from '@segment/actions-core'
+import Definition from '../index'
+
+const testDestination = createTestIntegration(Definition)
+
+describe('Iterable Lists', () => {
+  describe('testAuthentication', () => {
+    it('should validate authentication inputs', async () => {
+      nock('https://your.destination.endpoint').get('*').reply(200, {})
+
+      // This should match your authentication.fields
+      const authData = {}
+
+      await expect(testDestination.testAuthentication(authData)).resolves.not.toThrowError()
+    })
+  })
+})
--- a/packages/destination-actions/src/destinations/iterable-lists/index.ts
+++ b/packages/destination-actions/src/destinations/iterable-lists/index.ts
@@ -1,0 +1,85 @@
+import { IntegrationError, AudienceDestinationDefinition } from '@segment/actions-core'
+import type { AudienceSettings, Settings } from './generated-types'
+
+const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
+  name: 'Iterable Lists',
+  slug: 'actions-iterable-lists',
+  mode: 'cloud',
+  description: 'Sync Segment Engage Audiences to Iterable Lists',
+
+  authentication: {
+    scheme: 'custom',
+    fields: {
+      apiKey: {
+        type: 'string',
+        label: 'API Key',
+        description:
+          "To obtain the API Key, go to the Iterable app and naviate to Integrations > API Keys. Create a new API Key with the 'Server-Side' type.",
+        required: true
+      }
+    },
+    testAuthentication: (request, { settings }) => {
+      return request('https://api.iterable.com/api/lists', {
+        method: 'GET',
+        headers: { 'Api-Key': settings.apiKey }
+      })
+    }
+  },
+
+  audienceFields: {
+    placeholder: {
+      type: 'boolean',
+      label: 'Placeholder Setting',
+      description: 'Placeholder field to create the AudienceSettings object. Please do not change this.',
+      default: true
+    }
+  },
+  audienceConfig: {
+    mode: {
+      type: 'synced', // Indicates that the audience is synced on some schedule
+      full_audience_sync: false // If true, we send the entire audience. If false, we just send the delta.
+    },
+    async createAudience(request, createAudienceInput) {
+      const audienceSettings = createAudienceInput.audienceSettings
+      const settings = createAudienceInput.settings
+      // @ts-ignore type is not defined, and we will define it later
+      const personasSettings = audienceSettings.personas
+      if (!personasSettings) {
+        throw new IntegrationError('Missing computation parameters: Id and Key', 'MISSING_REQUIRED_FIELD', 400)
+      }
+
+      const audienceKey = personasSettings.computation_key
+
+      await request('https://api.iterable.com/api/lists', {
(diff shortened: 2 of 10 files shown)


joe-ayoub-segment:
Hi @seg-leonelsanches thanks for raising this PR for a new Iterable Audience Destination. cc @smultani 

As this is for a new Destination for a significant Partner I'd like to ensure that it's been planned and designed sufficiently so that it will benefit as many customers as possible. Iterable is a Partner who has put a lot of effort into their existing Actions Destination, so I'd like to ask them if they'd like to have some input also. 

**Some design decisions  to discuss:** 
1 Would it make sense to include this Action into the existing Iterable Destination? (I think probably not, but worth discussing)
2. Should we support sending identifiers other than userId/email? If so, should we pick up the additional identifiers automatically, or force the customer to explicitly map them in the dataFields field?
3. Does this really need to be a native Audience Destination? This increases the latency between syncs, and makes the Destination be far less 'real-time'. 
4. If we do go with 'Native Audience' then we should support track() and not identify() calls. AFAIK identify() calls don't get populated with Audience field data. 
5. Presets don't work with Engage. Customers will need to configure the Action unfortunately.
6. Can calls to an Iterable Batch endpoint contain data for multiple Audiences in the same payload? If not, then we need to factor this in to the performBatch code. 
7. Should we make use of the bulk batching pipeline, or the regular batching pipeline?
8. The computation_key, computation_id and external_audience_id values can be accessed without the need to look at the rawData object. 
9. We should prevent any non Audience payload from invoking the perform() and performBatch() functions. 

Also, it would be great to see some evidence of testing showing: 
1. createAudience() => Audience created in Iterable  
2. track() payload => perform() => Audience populated in Iterable
3. track() payloads => performBatch() => Audience populated in Iterable

As a starting point, shall I see if I can line up a call with an Engineer from Iterable? We could discuss the requirements / goals for the Destination, and review the code as is and go from there? 

Thanks,
Joe

packages/destination-actions/src/destinations/iterable-lists/__tests__/index.test.ts
```diff
@@ -0,0 +1,19 @@
...
+describe('Iterable Lists', () => {
+  describe('testAuthentication', () => {
+    it('should validate authentication inputs', async () => {
+      nock('https://api.iterable.com/api').get('/lists').reply(200, {})
+
+      const settings = {
+        apiKey: '12345'
+      }
+
+      await expect(testDestination.testAuthentication(settings)).resolves.not.toThrowError()
```
marinhero, line 16: is this valuable? You are mocking it with a success and expecting it not to fail.

packages/destination-actions/src/destinations/iterable-lists/index.ts
```diff
@@ -0,0 +1,101 @@
...
+        json: {
+          name: audienceKey
+        }
+      })
+      const createAudienceResponseJson = await createAudienceResponse.json()
+
+      return { externalId: createAudienceResponseJson.listId }
+    },
+    async getAudience(_, getAudienceInput) {
+      return { externalId: getAudienceInput.externalId }
```
marinhero, line 86: I assume this method isn't complete. Right?

marinhero (changes requested):
Hey @seg-leonelsanches,

Can you share the entire `curl` request you are using to test this? I would like to see if we are talking about a call to createAudience or an action.

Also, the definition you have for `external_audience_id` doesn't match your payload as you are providing the id within the audience settings and not at the context.personas level.

packages/destination-actions/src/destinations/iterable-lists/__tests__/index.test.ts
seg-leonelsanches: @marinhero This test template came automatically when I scaffolded the destination. The idea is to simply test whether the endpoint is called.

packages/destination-actions/src/destinations/iterable-lists/index.ts
seg-leonelsanches: @marinhero Well, in theory I expect the `externalId` coming from a database, or something like it. I just don't understand how the system internals work.

seg-leonelsanches:
> Hey @seg-leonelsanches,

Can you share the entire curl request you are using to test this? I would like to see if we are talking about a call to createAudience or an action.

To test `createAudience`, I'm using:

```sh
curl --location 'http://localhost:3000/createAudience' \
--header 'Content-Type: application/json' \
--data '{
    "settings": {
        "apiKey": "my-api-token"
    },
    "audienceSettings": {
        "personas": {
            "computation_key": "the_super_mario_bros_super_audience_2"
        }
    }
}'
```

To test `syncAudience`, I use two very similar requests:

`perform`:

```sh
curl --location 'http://localhost:3000/syncAudience' \
--header 'Content-Type: application/json' \
--data-raw '{
    "settings": {
        "apiKey": "my-api-key"
    },
    "payload": {
        "context": {
            "personas": {
                "audience_settings": {
                    "computation_key": "the_super_mario_bros_super_audience_2",
                    "external_audience_id": "4269566"
                }
            }
        },
        "traits": {
            "email": "leonel.sanches@segment.com"
        },
        "userId": "leonel.sanches@segment.com"
    }
}'
```

`performBatch`:

```sh
{
    "settings": {
        "apiKey": "my-api-key"
    },
    "payload": [
        {
            "context": {
                "personas": {
                    "audience_settings": {
                        "computation_key": "the_super_mario_bros_super_audience_2",
                        "external_audience_id": "4269566"
                    }
                }
            },
            "traits": {
                "email": "leonel.sanches@segment.com"
            },
            "userId": "leonel.sanches@segment.com"
        }
    ]
}
```

seg-leonelsanches:
@marinhero @joe-ayoub-segment I think this is ready to be reviewed.

joe-ayoub-segment (approved):

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @seg-leonelsanches this has been deployed!

PR #2177, opened 2024-07-15
[STRATCONN-3963] Allow for dynamic profile field values based on audience key value

tcgilbert:
JIRA: https://segment.atlassian.net/browse/STRATCONN-3963
[Slack Thread](https://twilio.slack.com/archives/C010W2FMYN5/p1720812980453239)

Vineyard Vines have reached out with an issue around the number of mappings they can create per destination. They are new premium plus customer and are creating a lot of new audiences for their campaigns. They have the ability to create 125 audiences and intend to use them all, if not add more.

But due to how the Listrak destination works, they need to configure two mappings per Audience (one for "Audience Entered" and one for "Audience Exited". This is because Listrak does not make use of the typical trait_key and true/false combo that other destinations use. Instead, the audience key is manually set with an "on" value for "Entered" and an "off" value for "Exited".

Screenshot 2024-07-14 at 1 35 20 PM

This PR provides the ability to accomplish the same results with just one mapping. This is by providing an additional option for the Profile Field Value: 

- "on"
- "off"
- "useAudienceKey" (new option): when used, the action will use the value associated with the audience key. 
  - if audience_key is `true` this will set to "on", if `false` this will set value to "off"

As a result, users will be able to configure just one mapping for just the audience_key, and use the `"useAudienceKey"` field value to default to using the audience key value. 

Screenshot 2024-07-15 at 9 35 15 AM
Screenshot 2024-07-15 at 10 47 08 AM

## Testing

**Tested Locally:** 
Screenshot 2024-07-15 at 10 41 54 AM
Screenshot 2024-07-15 at 10 42 19 AM

**Added tests for new functionality:** 
Screenshot 2024-07-15 at 10 49 41 AM

## Next Steps

Fairly certain this is not a breaking change, unless somehow the deployment process affects that. Will need to flag with Listrak and get their insight into this.


joe-ayoub-segment:
Hi @tcgilbert thanks for raising this PR. 

Couple questions on it. Was there any comms with Listrak folks about this change?

I'm not sure if this Destination was designed for Audience syncs, so it might not be so surprising that it's clunky to get working with Audiences. 

I'm thinking it might be best to add a new Action specifically for syncing Audience data. 

There's a fairly straightforward pattern for doing this - which I'm happy to discuss with you. 

Should we loop in the Listrak team so that they can give the 👍 to this change?

tcgilbert:
Hey @joe-ayoub-segment 

> Couple questions on it. Was there any comms with Listrak folks about this change? Should we loop in the Listrak team so that they can give the 👍 to this change?

Yeah didn't want to proceed with this until Listrak gave the thumbs up. I emailed them yesterday explaining the change, and you are cc'd in it "Subject: Segment Listrak Destination Update"


> I'm not sure if this Destination was designed for Audience syncs, so it might not be so surprising that it's clunky to get working with Audiences. I'm thinking it might be best to add a new Action specifically for syncing Audience data

I'd be open to this as well. [Looking at their docs though](https://segment.com/docs/connections/destinations/catalog/actions-listrak/#sync-an-engage-audience), they have detailed instructions for using this action to sync audiences. So I figured this would be a fairly easy way to provide more flexibility, without breaking anything. 

Going to convert this to draft until we hear back.

tcgilbert:
@joe-ayoub-segment got this from Listrak: 

Screenshot 2024-07-16 at 5 32 17 PM

Happy to discuss before merging if other approaches come to mind

tcgilbert:
> Thanks @tcgilbert . I left you some comments on Slack. Can you check please?

will address before EOW

joe-ayoub-segment:
hi @tcgilbert , 

Just noticed a couple of files in a folder called /thomas-test - probably added by mistake?

tcgilbert:
> hi @tcgilbert ,
> 
> Just noticed a couple of files in a folder called /thomas-test - probably added by mistake?

sorry mistake, it has been removed

joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi Thomas, 

Will the customer still need to explicitly paste in the audience_key into the Profile Field Values as per below?
image

I'm not sure I fully understand what is being attempted here. Is it the ablity to support sending a trait boolean (i.e Profile Field Value) with the name of the Audience?

joe-ayoub-segment (changes requested):
See comment

joe-ayoub-segment:
Discussed with @tcgilbert - approving for deploy next Tuesday.

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @tcgilbert this PR has been deployed

PR #2132, opened 2024-07-02
Hide legacy postConversion action that is no longer supported

tcgilbert:
Google Ads Enhanced Conversions Legacy API was scheduled to EOL June 30th, 2024. Google is now saying it will EOL it “imminently.” Google will no longer let data flow to this API. Hiding this action as a result. 

[More context in Slack](https://twilio.slack.com/archives/C03NK2Y2GTC/p1719946735563759?thread_ts=1719844307.469689&cid=C03NK2Y2GTC)

## Testing

- no testing needed

PR #2183, opened 2024-07-16
Revert "Hide legacy postConversion action that is no longer supported"

brennan:
- Reverts segmentio/action-destinations#2132
- These actions should remain visible until we confirm that Google has actually deprecated them. More context here: https://twilio.slack.com/archives/GSUCUMF6J/p1721152825028129


joe-ayoub-segment:
Hi @brennan can we close this old PR?

PR #2187, opened 2024-07-17
Adding support for FieldDisplayModes & disabling certain input methods

pooyaj:
This ticket update types to support these features: 
- **FieldDisplayMode** --> this flag determines how the field will be displayed between three modes:

Normal, the same as current view of the fields 
Screenshot 2024-07-17 at 3 24 55 PM

Expanded, where the label and description will be expanded on top 
Screenshot 2024-07-17 at 3 25 29 PM

Collapsed, where multiple fields will be collapsed behind a dropdown
Screenshot 2024-07-17 at 3 25 56 PM

- **disabledInputMethods**: This is an array that include the list of entry methods that should be disabled for the field. This is to prevent certain fields from being mappable to payload variables ( only allowing static values ) or disabling certain tabs in the picker. 

- Adding a new `sync` category to allow certain fields being displayed in the sync section instead of the main mapping area

**Note**: After merging types, there will be [subsequent PR](https://github.com/segmentio/action-cli/pull/174) into actions-cli to push these flags to `displayMetadata` for the field. 

## Testing

Testing not required as this is just adding new optional types.

### DIFF
--- a/packages/core/src/destination-kit/types.ts
+++ b/packages/core/src/destination-kit/types.ts
@@ -113,7 +113,13 @@ export interface GlobalSetting {
 export type FieldTypeName = 'string' | 'text' | 'number' | 'integer' | 'datetime' | 'boolean' | 'password' | 'object'
 
 /** The supported field categories */
-type FieldCategory = 'identifier' | 'data' | 'internal' | 'config'
+type FieldCategory = 'identifier' | 'data' | 'internal' | 'config' | 'sync'
+
+/** supported input methods when picking values */
+type FieldInputMethods = 'literal' | 'variable' | 'function' | 'enrichment' | 'freeform'
+
+/** display modes for a field */
+type FieldDisplayMode = 'expanded' | 'normal' | 'collapsed'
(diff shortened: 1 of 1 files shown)


packages/core/src/destination-kit/types.ts
```diff
@@ -213,6 +219,19 @@ export interface InputField extends InputFieldJSONSchema {
...
    */
   category?: FieldCategory
+
+  /**
+   * Determines how the field should be displayed in the UI:
+   * - expanded: Fully visible title and description for detailed information.
+   * - normal: One line field, and title, with description hidden behind a tooltip.
+   * - collapsed: Fields grouped behind a dropdown
+   */
+  displayMode?: FieldDisplayMode
```
nick-Ag, line 229: What's the default? assuming normal? Might be worth it to update the comment to point to the default

pooyaj: done 💥

nick-Ag (approved):

PR #2212, opened 2024-07-24
[STRATCONN-3832] Validate Audience Field inputs while creating Audience""

Innovative-GauravKochar:
Reverts segmentio/action-destinations#2129

This Pull request is to add validation of audience field inputs while creating audience.
This PR is also to add personas as separate field in createAudienceInput , removed out of audienceSettings.

Jira ticket:- https://segment.atlassian.net/jira/software/c/projects/STRATCONN/boards/310?assignee=63617339fc0cc7a600b03c6b&selectedIssue=STRATCONN-3832
Related Integrations PR :-https://github.com/segmentio/integrations/pull/2931
## Testing

Testing Completed Successfully [here](https://docs.google.com/document/d/1D2CX5_yRBU0abiOuqfK234KsAfcr76BVLtk2GXSrVX8/edit#heading=h.x9dgiyfuowsl)


packages/core/src/destination-kit/index.ts
```diff
@@ -94,12 +95,15 @@ export type AudienceResult = {
 }
 
 export type AudienceMode = { type: 'realtime' } | { type: 'synced'; full_audience_sync: boolean }
+export type Personas = { computation_id: 'string'; computation_key: 'string' }
```
marinhero, line 98: How did we validate that this type works for all of the existing destinations? What would happen if the Personas field comes with an extra parameter that isn't listed here?

See how this payload looks for example:

Screenshot 2024-07-24 at 12 26 56 PM

packages/browser-destinations/package.json
```diff
@@ -84,7 +84,7 @@
   "size-limit": [
     {
       "path": "dist/web/*/*.js",
-      "limit": "169 KB"
+      "limit": "190 KB"
```
varadarajan-tw, line 87: why the bump here?

varadarajan-tw: ah! seems like you used lodash in actions-core.

varadarajan-tw: If you use loadash in actions-core, it affects bundle size of web destinations. Could you not do that?

packages/core/src/destination-kit/index.ts
```diff
@@ -94,12 +95,15 @@ export type AudienceResult = {
 }
 
 export type AudienceMode = { type: 'realtime' } | { type: 'synced'; full_audience_sync: boolean }
+export type Personas = { computation_id: 'string'; computation_key: 'string' } & { [key: string]: any }
```
varadarajan-tw, line 98: why not?
```suggestion
export type Personas = { computation_id: 'string'; computation_key: 'string', { [key: string]: any }}
```
whats the need for intersecting types here? `&` can be useful if you are combining two predefined types and in this step, we are creating a plain new type.

packages/core/src/destination-kit/index.ts
```diff
@@ -426,6 +430,10 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
     if (!instanceOfAudienceDestinationSettingsWithCreateGet(audienceDefinition.audienceConfig)) {
       throw new Error('Unexpected call to createAudience')
     }
+    //validate audienceField Input
+    if (!isEmpty(createAudienceInput.audienceSettings)) {
```
varadarajan-tw, line 434: Could we use a simple method instead of using lodash in actions-core? If you have any other suggestions, pls do shar.
```
function isEmpty(obj) {
  for (const key in obj) {
     if (Object.prototype.hasOwnProperty.call(obj, key)) {
        return false
     }
  } 
  return true
}
```

Innovative-GauravKochar: We can do using Native js like 

if (createAudienceInput.audienceSettings && Object.keys(createAudienceInput.audienceSettings).length > 0) {
  validateSchema(createAudienceInput.audienceSettings, fieldsToJsonSchema(audienceDefinition.audienceFields));
}

packages/core/src/destination-kit/index.ts
```diff
@@ -94,12 +94,15 @@ export type AudienceResult = {
 }
 
 export type AudienceMode = { type: 'realtime' } | { type: 'synced'; full_audience_sync: boolean }
+export type Personas = { computation_id: 'string'; computation_key: 'string'; [key: string]: any }
```
varadarajan-tw, line 97: nit: Should you use unknown here? You'll probably see a warning in your editor as well.
```suggestion
export type Personas = { computation_id: 'string'; computation_key: 'string'; [key: string]: unknown }
```

packages/core/src/destination-kit/index.ts
```diff
@@ -426,6 +429,10 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
     if (!instanceOfAudienceDestinationSettingsWithCreateGet(audienceDefinition.audienceConfig)) {
       throw new Error('Unexpected call to createAudience')
     }
+    //validate audienceField Input
+    if (createAudienceInput.audienceSettings && Object.keys(createAudienceInput.audienceSettings).length > 0) {
```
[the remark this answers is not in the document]
Innovative-GauravKochar: That's okay. Yes we can add test but test cases for createAudience and getAudience in destination-kit is added in [this](https://github.com/segmentio/action-destinations/pull/2212/files) PR. So we can add that once that is merged otherwise it will be duplicated in both PRs.

joe-ayoub-segment (changes requested):
please also fix for iterable-lists and aws-s3 destinations

image

joe-ayoub-segment (changes requested):
hi @Innovative-GauravKochar can you remove the unnecessary formatting changes please? It's making the PR difficult to review properly.

Innovative-GauravKochar:
Created a separate PR as it recieved many identation changes while resolving conflicts
https://github.com/segmentio/action-destinations/pull/2248

PR #2216, opened 2024-07-25
Mailing list support for Loops

danrowden:
Re-do of #2171 

@joe-ayoub-segment This includes your suggestion for a better mailing list data format.


joe-ayoub-segment (approved):

danrowden:
@joe-ayoub-segment Done!

askkaz:
> @joe-ayoub-segment any update here?

our bad, updating tests

askkaz:
Ok @joe-ayoub-segment we're running into a problem with this approach. We've defined the mailingLists as an array of objects. However, when using the action tester, its creating an object and not an array (of objects):
Screenshot 2024-08-22 at 11 22 13 AM
This works, but it generates an object, not an array of objects:
Screenshot 2024-08-22 at 11 22 53 AM
We're not sure if this is an artifact of the builder or a limitation in Segment where we cannot actually create an array of objects.

If we "fake" it by passing in an array as part of the test event, this also works and passes through an array of objects:
Screenshot 2024-08-22 at 11 25 57 AM

Is this a limitation of the tester or something with how Segment actually works? Will users be able to create an array of objects in the UI or not?

packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
```diff
@@ -49,6 +49,29 @@ const action: ActionDefinition<Settings, Payload> = {
       required: false,
       default: { '@path': '$.traits.lastName' }
     },
+    mailingLists: {
+      label: 'Mailing Lists',
+      description:
+        'An an array of objects containing key-value pairs of mailing list IDs and true/false determining if the contact should be added to or removed from each list.',
+      type: 'object',
+      multiple: true,
```
joe-ayoub-segment, line 57: hi @danrowden - we can't add a new 'required' field as it will break data collection for customers who already have the Integration set up. 

You should set this field to be optional. And then in the perform() function you'll also need to check if the field is defined before trying to access it. 

e.g. 
if(payload.mailingLists !== 'undefined') {
  .. do stuff
}

packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
```diff
@@ -49,6 +49,29 @@ const action: ActionDefinition<Settings, Payload> = {
...
+          description: 'The ID of the mailing list.',
+          type: 'string',
+          required: true
+        },
+        value: {
+          label: 'value',
+          description:
+            'true indicates that the user is to be added to the list, false will remove the user from the list.',
+          type: 'boolean',
+          required: true
```
joe-ayoub-segment, line 71: It's OK to leave these sub property fields of the object as required.

packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
joe-ayoub-segment: Ah I see you are already doing this in the perform() function. Great. 
You just need to make this field optional.

joe-ayoub-segment:
hi @danrowden this is working correctly. 

image

The customer needs to pass an array in the main payload. 

We can make this more explicity by adding a default to the field, like this: 


```
    mailingLists: {
      label: 'Mailing Lists',
      description:
        'An an array of objects containing key-value pairs of mailing list IDs and true/false determining if the contact should be added to or removed from each list.',
      type: 'object',
      multiple: true,
      required: false,
      properties: {
        list_id: {
          label: 'List ID',
          description: 'The ID of the mailing list.',
          type: 'string',
          required: true
        },
        value: {
          label: 'value',
          description:
            'true indicates that the user is to be added to the list, false will remove the user from the list.',
          type: 'boolean',
          required: true
        }
      },
      default: {
        '@arrayPath': [
          '$.properties.mailing_lists',
          {
            list_id: {
              '@path': '$.id'
            },
            value: {
              '@path': '$.value'
            }
          }
        ]
      }
    },
```

This will make the UI look something like this: 

image

It makes the customer more aware that the item needs to be an array, and provides some defaults for how the items in the array should be named. The customer can change these mappings though, of course.

packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
danrowden: I believe it is optional because of the `required: false` below?

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @danrowden the change looks good - however CI is now failing.

danrowden:
@joe-ayoub-segment I noticed today that we are modelling this on a "track" event but it should be for "identify" events, which don't have `properties`. I'll chatting with @askkaz today about this

danrowden:
This is ready for another look @joe-ayoub-segment

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @danrowden this PR has been deployed. Please confirm that the change works as expected.

PR #2224, opened 2024-07-29
RIP: @segment/actions-cli-internal

varadarajan-tw:
The commands in @segment/actions-internal-cli are no longer used and have been ported to [action-cli](https://github.com/segmentio/action-cli). We can now get rid of the package.

We had a dependency on an internal control-plane-client lib and had to use special GH_PAT token and NPM token in our workflows. We also don't need that any longer.

Also removes @segment/build-exp-team from code owners - this team no longer exists.

## Testing

Tested the cli in local. The generate, init commands are working.

Build and test cases are passing.


sayan-das-in (approved):
Thank you for taking this up. 🚀🚀

joe-ayoub-segment (approved):

itsarijitray (approved):

PR #2241, opened 2024-08-06
[Hubspot] Custom Event Action

joe-ayoub-segment:
Adding a new Action to Hubspot
And updating an existing Action. 

Both are not used by customers yet. No unit tests yet. Testing party next week. 

Actions will be behind a feature flag. 

## Testing

Testing has been done locally, and will be tested in Production with a testing 'party' on 19th August. 

Note: unit tests will follow post deploy but before customers have access.


packages/destination-actions/src/destinations/hubspot/customEvent/hubspot-api.ts
```diff
@@ -0,0 +1,489 @@
...
+      const { fullyQualifiedName, name, properties: hsProperties, archived } = response.data
+
+      if (archived) {
+        return mismatchSchemaDiff
+      }
+
+      const schemaDiff = { missingProperties: {} } as SchemaDiff
+
+      for (const propName in schema.properties) {
+        const matchedProperty = hsProperties.find((hsProp) => hsProp.name === propName)
```
varadarajan-tw, line 250: for next iteration - may be it is worth iterating once and constructing a map with key as name. There is a possibility that event definition might contain a lot of properties.

packages/destination-actions/src/destinations/hubspot/customEvent/hubspot-api.ts
```diff
@@ -0,0 +1,489 @@
...
+            const fullyQualifiedName = schemaDiff?.fullyQualifiedName as string
+            const name = schemaDiff?.name as string
+            await this.saveSchemaToCache(fullyQualifiedName, name, schema)
+            await this.sendEvent(fullyQualifiedName, payload)
+            break
+          }
+
+          case 'properties_missing': {
+            if (this.syncMode === 'add') {
+              throw new IntegrationError(
```
varadarajan-tw, line 358: for iteration 2 of this action - should we drop the missing properties if they don't care about rest of the properties?

packages/destination-actions/src/destinations/hubspot/customEvent/hubspot-api.ts
```diff
@@ -0,0 +1,489 @@
...
+            json
+          })
+        )
+      })
+
+      const responses = await Promise.allSettled(requests)
+
+      for (const response of responses) {
+        if (response.status === 'rejected') {
+          const error = response.reason as RegectedResponse
```
varadarajan-tw, line 470: typo??
```suggestion
          const error = response.reason as RejectedResponse
```

varadarajan-tw (commented):
Left few questions but overall the approach looks good. 

For future iterations - is it possible to add few metrics for figuring out how new properties are created per request?

sayan-das-in (approved):
This action is not being used by any customer in Prod. LGTM!
Follow up action: Add unit test cases.

PR #2248, opened 2024-08-06
[STRATCONN-3832] Validate Audience Field inputs while creating Audience

Innovative-GauravKochar:
Reverts segmentio/action-destinations#2129

This Pull request is to add validation of audience field inputs while creating audience.
This PR is also to add personas as separate field in createAudienceInput , removed out of audienceSettings.

Jira ticket:- https://segment.atlassian.net/jira/software/c/projects/STRATCONN/boards/310?assignee=63617339fc0cc7a600b03c6b&selectedIssue=STRATCONN-3832
Related Integrations PR :-https://github.com/segmentio/integrations/pull/2931
## Testing

Testing Completed Successfully [here](https://docs.google.com/document/d/1D2CX5_yRBU0abiOuqfK234KsAfcr76BVLtk2GXSrVX8/edit#heading=h.x9dgiyfuowsl)


packages/core/src/destination-kit/index.ts
```diff
@@ -94,12 +94,15 @@ export type AudienceResult = {
...
 
 export type AudienceMode = { type: 'realtime' } | { type: 'synced'; full_audience_sync: boolean }
+export type Personas = { computation_id: 'string'; computation_key: 'string'; [key: string]: unknown }
 
 export type CreateAudienceInput<Settings = unknown, AudienceSettings = unknown> = {
   settings: Settings
 
   audienceSettings?: AudienceSettings
 
+  personas?: Personas
```
maryamsharif, line 104: In the PR description it says **add personas as separate field in createAudienceInput , removed out of audienceSettings.**

Would this break current implementations that might use the personas object from audienceSettings?

Innovative-GauravKochar: No, it should not . There are 3 destinations which were accessing personas as audienceSettings.personas and I have made necessary changes for those . It's in same PR.Kindly review !

packages/destination-actions/src/destinations/amazon-amc/__tests__/index.test.ts
```diff
@@ -139,9 +155,9 @@ describe('Amazon-Ads (actions)', () => {
         audienceName: 'Test Audience',
         audienceSettings: {
           ...audienceSettings,
-          ttl: 12345678,
+          ttl: '12345678',
           currency: 'USD',
-          cpmCents: 1234
+          cpmCents: '1234'
```
joe-ayoub-segment, line 160: hi @Innovative-GauravKochar why are you changing this test?

packages/destination-actions/src/destinations/amazon-amc/index.ts
```diff
@@ -178,22 +178,14 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
...
-        }
-        payload.metadata.ttl = timeToLive
+        payload.metadata.ttl = ttl
       }
 
       if (cpm_cents && currency) {
-        const cpmCents = Number(cpm_cents)
-        if (!cpmCents) {
-          throw new IntegrationError('CPM_CENTS:-String can not be converted to Number', 'INVALID_CPMCENTS_VALUE', 400)
-        }
```
joe-ayoub-segment, line 192: hi @Innovative-GauravKochar is this validation being removed because it's no longer needed?

packages/core/src/destination-kit/index.ts
```diff
@@ -94,12 +94,15 @@ export type AudienceResult = {
 }
 
 export type AudienceMode = { type: 'realtime' } | { type: 'synced'; full_audience_sync: boolean }
+export type Personas = { computation_id: 'string'; computation_key: 'string'; [key: string]: unknown }
```
marinhero, line 97: There's also the `namespace` field in [there](https://github.com/segmentio/external-audience-manager-service/blob/97b95a968ffdfedad095928f5c2037c24e92886e/internal/gxClient/gxClient.go#L75C2-L79C4) . We should add a comment that references this code in there too.

marinhero (commented):
While I agree with the proposed change I would like to de-risk it further. We are shipping it at once to three different destinations one of which (Yahoo Audiences) is widely used. Do you see value in leaving the original code in place as well as the new one, updating one destination at the time and ultimately removing the old code? That would also help us to not rollout everything together within the same PR.

Or do you guys think I'm being a bit paranoid here @joe-ayoub-segment @maryamsharif ??

packages/destination-actions/src/destinations/amazon-amc/index.ts
Innovative-GauravKochar: Yes @joe-ayoub-segment   , it's no longer needed now

packages/destination-actions/src/destinations/amazon-amc/__tests__/index.test.ts
Innovative-GauravKochar: Hi @joe-ayoub-segment  ,Audience Fields takes an input as an string and validateSchema validates and converts the input as per the type provided. 
Because this is numeric string , it didn't throw an error.If it could be any random string the it would have throw an error.
packages/destination-actions/src/destinations/amazon-amc/__tests__/index.test.ts

joe-ayoub-segment (approved):

joe-ayoub-segment:
> While I agree with the proposed change I would like to de-risk it further. We are shipping it at once to three different destinations one of which (Yahoo Audiences) is widely used. Do you see value in leaving the original code in place as well as the new one, updating one destination at the time and ultimately removing the old code? That would also help us to not rollout everything together within the same PR.
> 
> Or do you guys think I'm being a bit paranoid here @joe-ayoub-segment @maryamsharif ??

I think this makes sense @marinhero . We need to be very sure that this won't cause an incident.

Innovative-GauravKochar:
> While I agree with the proposed change I would like to de-risk it further. We are shipping it at once to three different destinations one of which (Yahoo Audiences) is widely used. Do you see value in leaving the original code in place as well as the new one, updating one destination at the time and ultimately removing the old code? That would also help us to not rollout everything together within the same PR.
> 
> Or do you guys think I'm being a bit paranoid here @joe-ayoub-segment @maryamsharif ??

We can not split [this](https://github.com/segmentio/action-destinations/pull/2248/files) PR that manner .
There are three destinations that were accessing `audienceSettings.personas` (Yahoo-audiences, Iterable-lists, AWS-S3). If we change one of them to access personas via `createAudienceInput.personas` and leave the other two as they are with the old code, it will break during schema validation.
`ValidateSchema` ensures that the resolved JSON for `audienceFields` adheres to the `audienceFields` schema. However, this validation strips out any JSON that is not defined in the schema. We cannot define personas in `audienceFields`, which is why we moved personas out of `audienceSettings` and made it a separate field.I wonder if it could be possible to condition validateSchema for those destinations.

joe-ayoub-segment:
> > While I agree with the proposed change I would like to de-risk it further. We are shipping it at once to three different destinations one of which (Yahoo Audiences) is widely used. Do you see value in leaving the original code in place as well as the new one, updating one destination at the time and ultimately removing the old code? That would also help us to not rollout everything together within the same PR.
> > Or do you guys think I'm being a bit paranoid here @joe-ayoub-segment @maryamsharif ??
> 
> @marinhero @joe-ayoub-segment We can not split this PR that manner . There are three destinations that were accessing `audienceSettings.personas` (Yahoo-audiences, Iterable-lists, AWS-S3). If we change one of them to access personas via `createAudienceInput.personas` and leave the other two as they are with the old code, it will break during schema validation. `ValidateSchema` ensures that the resolved JSON for `audienceFields` adheres to the `audienceFields` schema. However, this validation strips out any JSON that is not defined in the schema. We cannot define personas in `audienceFields`, which is why we moved personas out of `audienceSettings` and made it a separate field.

I had a chat with Gaurav about this PR. He's done testing in Staging, so I think it should be OK to deploy this.

marinhero (approved):

PR #2258, opened 2024-08-08
[SALESFORCE] Add Sync Mode Support for "Case" action

marinhero:
**Note to the reviewer**: it's easier to review this if you go commit by commit instead of going through all the files at once as I separated the ported legacy action from the syncMode implementation.

The destination already implements its own version of Sync Mode via mapping settings.
We will translate the “legacy” sync mode into our new and native Sync Mode. After all, backwards compatibility is not of concern given that we are creating a new action.

The sync modes supported by perform are different from the ones supported by performBatch and the UI is not aware of that. We will offer all of the sync modes for both perform and performBatch. The UI team will make use of conditional field so the options not supported by performBatch are hidden. For now, we will return a non-retryable error code for those scenarios and update the docs to illustrate this difference.

## Testing

Work In Progress

### DIFF
--- a/packages/destination-actions/src/destinations/salesforce/cases2/index.ts
+++ b/packages/destination-actions/src/destinations/salesforce/cases2/index.ts
@@ -1,0 +1,76 @@
+import type { ActionDefinition } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import {
+  bulkUpsertExternalId2,
+  bulkUpdateRecordId2,
+  customFields2,
+  traits2,
+  validateLookup2,
+  enable_batching2,
+  recordMatcherOperator2,
+  batch_size2,
+  hideIfDeleteSyncMode
+} from '../sf-properties'
+import Salesforce, { generateSalesforceRequest } from '../sf-operations'
+
+const OBJECT_NAME = 'Case'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Case',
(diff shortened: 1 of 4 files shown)


packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -240,6 +240,32 @@ export default class Salesforce {
...
+        'Unsupported operation',
+        400
+      )
+    }
+
+    if (syncMode === 'upsert') {
+      return await this.bulkUpsert(payloads, sobject)
+    } else if (syncMode === 'update') {
+      return await this.bulkUpdate(payloads, sobject)
+    } else if (syncMode === 'create') {
```
marinhero, line 264: this should be add

packages/destination-actions/src/destinations/salesforce/cases2/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  recordMatcherOperator2,
+  batch_size2,
+  hideIfDeleteSyncMode
+} from '../sf-properties'
+import Salesforce, { generateSalesforceRequest } from '../sf-operations'
+
+const OBJECT_NAME = 'Case'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Case',
```
maryamsharif, line 20:
```suggestion
  title: 'Case V2',
```

maryamsharif (approved):

PR #2265, opened 2024-08-08
[STRATCONN-4071] Add batching support to Google Enhanced Conversions

sayan-das-in:
This PR adds Batching Support to the following actions of Google Enhanced Conversions
- uploadCallConversion
- uploadClickConversion
- uploadConversionAdjustment
- uploadCallConversion2
- uploadClickConversion2
- uploadConversionAdjustment2

To reduce risk this PR sets `enable_batching` field is hidden and set to false by default.
This would be useful to enable `performBatch` for selected customers.

## Testing
Tested locally with unit tests and on staging.
Staging and Prod test is not possible as it requires a working Google Ads account.

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/constants.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/constants.ts
@@ -1,3 +1,3 @@
 // GEC's max batch size is 2000
 // https://developers.google.com/google-ads/api/docs/best-practices/quotas#conversion_upload_service
-export const GOOGLE_ENHANCED_CONVERSIONS_BATCH_SIZE = 2000
+export const GOOGLE_ENHANCED_CONVERSIONS_BATCH_SIZE = 1500
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/constants.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/constants.ts
@@ -1,0 +1,3 @@
+// GEC's max batch size is 2000
+// https://developers.google.com/google-ads/api/docs/best-practices/quotas#conversion_upload_service
+export const GOOGLE_ENHANCED_CONVERSIONS_BATCH_SIZE = 2000
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
@@ -97,6 +97,26 @@ export async function getCustomVariables(
   )
 }
 
+export function memoizedGetCustomVariables() {
+  const cache: Map<string, Promise<ModifiedResponse<QueryResponse[]>>> = new Map()
+
+  return async (
+    customerId: string,
+    auth: any,
+    request: RequestClient,
+    features: Features | undefined,
+    statsContext: StatsContext | undefined
+  ) => {
+    if (cache.has(customerId)) {
+      return cache.get(customerId)
+    } else {
+      const result = getCustomVariables(customerId, auth, request, features, statsContext)
+      cache.set(customerId, result)
+      return result
+    }
+  }
+}
+
 export async function getConversionActionId(
   customerId: string | undefined,
   auth: any,
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/uploadClickConversion/index.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/uploadClickConversion/index.ts
@@ -226,13 +226,12 @@ const action: ActionDefinition<Settings, Payload> = {
       label: 'Batch Data to Google Enhanced Conversions',
       description:
         'If true, Segment will batch events before sending to Google’s APIs. Google accepts batches of up to 2000 events.',
-      default: true
+      default: false
     },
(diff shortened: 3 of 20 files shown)


packages/destination-actions/src/destinations/google-enhanced-conversions/uploadClickConversion/index.ts
```diff
@@ -230,107 +247,126 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      },
-      userIdentifiers: []
-    }
-    // Add Consent Signals 'adUserData' if it is defined
-    if (payload.ad_user_data_consent_state) {
-      request_object['consent'] = {
-        adUserData: payload.ad_user_data_consent_state
+  const getCustomVariables = memoizedGetCustomVariables()
+
+  const request_objects: ClickConversionRequestObjectInterface[] = await Promise.all(
```
marinhero, line 271: Why do we need a promise here? You are not making any requests and are simply filling and array.

packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
```diff
@@ -97,6 +97,26 @@ export async function getCustomVariables(
...
+  const cache: Map<string, Promise<ModifiedResponse<QueryResponse[]>>> = new Map()
+
+  return async (
+    customerId: string,
+    auth: any,
+    request: RequestClient,
+    features: Features | undefined,
+    statsContext: StatsContext | undefined
+  ) => {
+    if (cache.has(customerId)) {
```
marinhero, line 110: When is this cache getting invalidated? How does it operate between instances of the Integrations Monoservice? Why do we need it?

marinhero (changes requested):
I left a few questions and would like to get the answers to them, as well as the test document for this PR. Additionally, I noticed you skipped the version 2 actions with the same name. If that is currently out of scope, please create a ticket to follow up on those.

One last thing to consider for the existing implementation:

Right now, the perform blocks handle most of our customers needs effectively. Your changes are modifying that code path. To minimize the impact of your changes, consider leaving the perform block intact until we can validate that this new approach solves our problems. This way, we maintain the existing implementation while offering a new option for those who need batching. Although it might result in some duplicated code, I believe it's a worthwhile tradeoff. What do you think?

packages/destination-actions/src/destinations/google-enhanced-conversions/functions.ts
sayan-das-in: ### Context
We have a function called `getCustomVariables` that calls Google APIs to get custom variables defined in that account. In the [earlier implementation](https://github.com/segmentio/action-destinations/blob/06d0d12b4b40ac708b348fd9ad39b403689b1d78/packages/destination-actions/src/destinations/google-enhanced-conversions/uploadCallConversion/index.ts#L141-L147) we were checking if we even need to call this API by checking if payload has custom variables defined.

While porting the logic to batch, there are two options

1. Make an API call regardless Custom Variables are required or not
2. Defer this checking in the loop, but that would make multiple API calls

If we are opting for 2nd variant, we don't have to make a call unless we really need to, but the trade-off is when we need it for multiple payload items, we would end up making multiple calls, that's where the memoization comes in.

### When is this cache getting invalidated?
In order to take advantage of memoization we have to initialize a memoized instance of the function (example here), the cache would be valid within the scope of each `performBatch` block instance and then it would get garbage collected.

### How does it operate between instances of the Integrations Monoservice?
Every instance has it's individual short-lived cache. In-fact the cache is local to every invocation of `performBatch` block after which it would get garbage collected.

### Why do we need it?
Memoization acts as a wrapper around deterministic, but computationally expensive functions or API calls that would generate the same result if same parameters are passed to it. Here in `f(x) => y`, the value of `y` will always be same as long as x remains same.

packages/destination-actions/src/destinations/google-enhanced-conversions/uploadClickConversion/index.ts
sayan-das-in: We are calling the `getCustomVariables` function for each payload, for that we would need to to await for the result. More details about the implementation in the comments below.

packages/destination-actions/src/destinations/google-enhanced-conversions/constants.ts
```diff
@@ -0,0 +1,3 @@
+// GEC's max batch size is 2000
+// https://developers.google.com/google-ads/api/docs/best-practices/quotas#conversion_upload_service
+export const GOOGLE_ENHANCED_CONVERSIONS_BATCH_SIZE = 2000
```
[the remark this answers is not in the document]
sayan-das-in: Sure, I will make this change.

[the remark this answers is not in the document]
sayan-das-in: Changed the batch size to 1500.

varadarajan-tw (approved):

marinhero (approved):
Thank you for updating the tests and de-risking the existing actions by splitting the code. Well done.

joe-ayoub-segment:
hi @sayan-das-in PR deployed

PR #2278, opened 2024-08-12
registering dawn and topsort

joe-ayoub-segment:
Registering 2 new Destinations 

dawn and topsort

## Testing

Testing will be done in Private beta.


nick-Ag (approved):
🚢

PR #2305, opened 2024-08-19
fix for iterable date format bug

joe-ayoub-segment:
Issue raised by @tcgilbert on behalf of customer. 

image

## Testing

Testing done locally: See this video https://drive.google.com/file/d/1vJpW0_156W7y2w2fOJtvcELg3Ok51leV/view?usp=sharing 

Request that @tcgilbert and @KakeJopulsky also test locally.


packages/destination-actions/src/destinations/iterable/utils.ts
```diff
@@ -19,6 +19,23 @@ function dateToIterableDateStringFormat(isoDateStr: string) {
   }
   const date = new Date(isoDateStr)
   if (date instanceof Date && !isNaN(date.valueOf())) {
+    const year = date.getUTCFullYear()
```
craigotis, line 22: 👋 I'm the Hinge Health engineer behind the original report. Thanks for jumping on this so quickly!

I had also pulled down this code, and considered an approach that matched the `isoDateStr` against a separate `yyyy-mm-dd` regex, to avoid unwantingly stripping a timestamp from a datetime that is legitimately at UTC midnight. (Then, if the string matches the regex, you just give it back.) But it wasn't immediately clear to me whether the `isoDateStr` was gaining the timestamp prior to landing in this function - what if someone really sends `2023-05-17T00:00:00.000Z`, do they want it altered?

Will defer to your judgement!

joe-ayoub-segment:
hi @KakeJopulsky I merged your PR into this PR. 
However the unit test named "does not alter a midnight UTC datetime" is failing. 

Any chance you could take a look please?

KakeJopulsky:
👍 LGTM! Tested locally.

craigotis (approved):

sayan-das-in (approved):

joe-ayoub-segment:
hi @KakeJopulsky this PR has been deployed

joe-ayoub-segment:
hi @craigotis this PR has been deployed

craigotis:
@joe-ayoub-segment Wonderful, we tested this morning and confirmed the fix is working as we'd expect. Thank you so much for the quick turnaround and collaboration here! 🙌 💯

PR #2324, opened 2024-08-21
channels-1094: Refactoring diffs

pmunin:

## Testing


packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -154,67 +149,88 @@ export abstract class MessageSendPerformer<
...
-      if (rawCachedResponse) {
-        cachedResponse = CachedValueFactory.fromString(rawCachedResponse)
-      }
-    } catch (error) {
-      if (!(error instanceof CachedValueSerializationError)) {
-        this.logError(`Unexpected error reading cache for messageId: ${messageId}, recipientId: ${recipientId}`, {
-          error
-        })
-        throw error
+    const recipientId = recepient.id!
```
cogwizzle, line 152: We probably shouldn't guarantee this if there is a chance it could be undefined.

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -154,67 +149,88 @@ export abstract class MessageSendPerformer<
...
+    return await this.getOrAddCache(
+      `${messageId}${recipientId.toLowerCase()}`,
+      () => this.sendToRecepient(recepient) as Promise<ModifiedResponse<unknown>>,
```
cogwizzle, line 155: Why do we have to cast this?

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
pmunin: addressed

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
pmunin: not really, addressed

cogwizzle (changes requested):

cogwizzle:
Fixed the last batch of TypeScript issues. I think that if we're going to use patterns like this one we should be very strict with types and ensure that we're writing completely type complete code even though it increases the verbosity of our code. The language patterns you've introduced mirror very strongly typed languages and the more TS hacks we use the more likely it is to break.

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,66 +106,113 @@ export abstract class MessageSendPerformer<
...
+      //if undefined returned, then the value will not be cached
+      stringify: (cacheable: ValueOrError<T>) => string | void
+      //if undefined returned, then the cache is either corrupted or expired and will be re-executed
+      parse: (cachedValue: string) => ValueOrError<T> | void
     }
+  ): Promise<T> {
+    const engageDestinationCache = this.engageDestinationCache
+    if (!engageDestinationCache) return createValue()
+
+    const cacheRead = await getOrCatch(() => engageDestinationCache.getByKey(key))
```
pmunin, line 152: too much code to make typescript happy for questionable value coming out of it
```suggestion
    if (!this.engageDestinationCache) return createValue()

    const cacheRead = await getOrCatch(() => this.engageDestinationCache!.getByKey(key))
```

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,66 +106,113 @@ export abstract class MessageSendPerformer<
...
+    const cacheReadValue = 'value' in cacheRead ? cacheRead.value : undefined
```
pmunin, line 153: same here - too much code and complexity for questionable value. increases cognitive load is the price we pay for perfectionism :)
```suggestion
    const cacheReadValue = cacheRead.value
```

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,66 +106,113 @@ export abstract class MessageSendPerformer<
...
+    } else if (cacheReadValue) {
+      let parsedCache: ValueOrError<T> | undefined = undefined
+      let parsingError: any | undefined = undefined
+      const result = getOrCatch(() => serializer.parse(cacheReadValue))
+      if ('error' in result) {
+        parsingError = result.error
+      }
+      if ('value' in result) {
+        parsedCache = result.value || undefined
+      }
```
pmunin, line 166: same here. this one line 
`const { value: parsedCache, error: parsingError } = getOrCatch(() => serializer.parse(cacheRead.value!))` became this large paragraph with bunch of explicit typings just to make typescript happy

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,66 +106,113 @@ export abstract class MessageSendPerformer<
...
+        this.statsIncr('cache_parsing_error')
+      } else if (parsedCache) {
+        //parsed cache successfully && cache is not expired
+        // parsedValue - either value or error was parsed
+        if ('error' in parsedCache) {
+          this.statsIncr('cache_hit', 1, [`cached_error:${!!parsedCache?.error}`])
+          if (parsedCache?.error) throw parsedCache.error
+        }
+        if ('value' in parsedCache) {
+          return parsedCache.value
```
pmunin, line 181: same here. Also cache_hit metric tag does not make sense in this implementation. I think cache_hit should include erroneous value which we could not parse

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,66 +106,113 @@ export abstract class MessageSendPerformer<
...
+    if ('error' in stringified && stringified.error) {
+      this.logError('cache_stringify_error', { key, error: stringified.error })
+      this.statsIncr('cache_stringify_error')
+    } else if ('value' in stringified && stringified.value) {
+      //result stringified and contains cacheable value - cache it
+      const result = getOrCatch(() => this.engageDestinationCache?.setByKey(key, stringified.value))
+      if ('error' in result && result.error) {
+        this.logError('cache_saving_error', { key, error: result.error })
+        this.statsIncr('cache_saving_error')
+      }
```
pmunin: bug in the code and notably happy typescript didn't help :-)
```suggestion
      const result = await getOrCatch(() => this.engageDestinationCache?.setByKey(key, stringified.value))
      if ('error' in result && result.error) {
        ...
      }
```
also var name result is confusable with line 193. And again, this is less readable code, comparing to what it was:
```
      const { error: cacheSavingError } = await getOrCatch(() => this.engageDestinationCache!.setByKey(key, stringified.value!)
      if (cacheSavingError) {
          ...
      }
      
```

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -523,3 +521,21 @@ export abstract class MessageSendPerformer<
     return body.traits
   }
 }
+
+type Ok<T> = { value: T }
+type Err<E> = { error: E }
+type ValueOrError<T> = Ok<T> | Err<T>
```
pmunin, line 528: looks unnecessary to add so much types, i'd just keep it as it was
```suggestion
type ValueOrError<T> = { value?: T; error?: any } 
```

pmunin (commented):
@cogwizzle plz see comments. I didn't find bugs you were talking about, let's zoom after meetings

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,66 +106,96 @@ export abstract class MessageSendPerformer<
...
+        this.engageDestinationCache!.setByKey(key, stringified.value!)
+      )
+      if (cacheSavingError) {
+        this.logInfo('cache_saving_error', { key, error: cacheSavingError })
+        this.statsIncr('cache_saving_error')
       }
     }
+
+    if (resultError) throw resultError
+    else return result as T
```
iamsuneeth, line 198: is this casting because we haven't specified T when calling `getCatch` method ? . Could we avoid this ?

joe-ayoub-segment:
hi @pmunin @cogwizzle please let me know when this is ready for release.

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
pmunin: no it's because typescript thinks result can be T | undefined based on types. T avoid this we'll have to make code ugly, verbose and runtime error prone

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -155,65 +106,104 @@ export abstract class MessageSendPerformer<
...
+        }
       }
-      this.logError(`Failed to read cache for messageId: ${messageId}, recipientId: ${recipientId}`, {
-        error,
-        message: error.message
-      })
+    )
+  }
+
+  async getOrAddCache<T>(
```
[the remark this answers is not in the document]
pmunin: feel free to suggest your implementation that will keep up the same clarity of where the error came from

PR #2325, opened 2024-08-21
Channels-1094 refactored

pmunin:

## Testing


packages/actions-shared/src/engage/utils/getOrRetry.ts
```diff
@@ -0,0 +1,90 @@
...
+  const retryIntervalMs = args?.retryIntervalMs === undefined ? backoffRetryPolicy() : args?.retryIntervalMs ?? 1000
+  return getOrCatch(
+    async () =>
+      await retry((attempt) => action(attempt), {
+        retries: args?.attempts,
+        async onFailedAttempt(error, attemptCount) {
+          if (args?.onFailedAttempt) await args.onFailedAttempt(error, attemptCount)
+          const intervalMs: number =
+            typeof retryIntervalMs == 'function' ? retryIntervalMs(attemptCount, error) : retryIntervalMs
+          if (intervalMs > 0) await delay(intervalMs)
```
cogwizzle, line 19: Hmm. Interesting. Why not call this function `sleep`? That is what I've seen it called in other languages.

packages/core/src/destination-kit/action.ts
```diff
@@ -82,8 +89,11 @@ type PerformBatchResponse = MaybePromise<MultiStatusResponse> | MaybePromise<unk
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export interface ActionDefinition<
   Settings,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   Payload = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   AudienceSettings = any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   GeneratedActionHookBundle extends GenericActionHookBundle = any
 > extends BaseActionDefinition {
```
cogwizzle, line 98: We should add an explanation as to why we need any here.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/previewApiLookup/apiLookups.ts
```diff
@@ -89,7 +90,7 @@ export const performApiLookup = async (
   tags: string[],
   settings: Settings,
   logger?: Logger | undefined,
-  dataFeedCache?: DataFeedCache | undefined
+  engageDestinationCache?: EngageDestinationCache | undefined
```
cogwizzle, line 93: We don't need this bonus `undefined` do we?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/SendEmailPerformer.ts
```diff
@@ -258,7 +259,7 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload>
       },
       json: mailContent
     }
-    this.statsClient?.set('message_body_size', JSON.stringify(req).length)
+    //this.statsClient?.set('message_body_size', JSON.stringify(req).length) // Commented due to performance issues
```
cogwizzle, line 262: Oof.

packages/destination-actions/src/destinations/engage-messaging-twilio/__tests__/__helpers__/test-utils.ts
```diff
@@ -16,56 +16,72 @@ afterEach(() => {
...
-  // eslint-disable-next-line prefer-rest-params
-  const args = arguments[0] as CreateTestActionArgs
-  timestamp = timestamp ?? new Date().toISOString()
+const defautlSettings = Object.freeze({
+  sourceId: 'e',
+  twilioAccountSID: 'a',
+  twilioApiKeySID: 'f',
+  twilioApiKeySecret: 'b',
+  profileApiAccessToken: 'c'
+})
```
cogwizzle, line 53: Do we have access to faker in this repo? If so we could use faker to generate these fake values.

packages/destination-actions/src/destinations/reddit-audiences/const.ts
```diff
@@ -1,2 +1,2 @@
 export const EMAIL_SCHEMA_NAME = 'EMAIL_SHA256'
-export const MAID_SCHEMA_NAME = 'MAID_SHA256' 
\ No newline at end of file
+export const MAID_SCHEMA_NAME = 'MAID_SHA256'
```
cogwizzle, line 2: I don't care, but this is just increasing the size of the PR and someone might complain about it.

packages/destination-actions/src/destinations/postscript/types.ts
```diff
@@ -9,4 +9,4 @@ export type Subscriber = {
         [AUDIENCE_PROPERTY]: string[]
     },
     id: string
-}
\ No newline at end of file
+}
```
cogwizzle, line 12: I don't care, but this is just increasing the size of the PR and someone might complain about it.

packages/destination-actions/src/destinations/reddit-audiences/syncAudience/types.ts
```diff
@@ -38,4 +38,4 @@ export interface CreateAudienceRespError {
 export interface RedditAccessTokenRequest {
     grant_type: string,
     refresh_token: string
-}
\ No newline at end of file
+}
```
cogwizzle, line 41: I don't care, but this is just increasing the size of the PR and someone might complain about it.

packages/destination-actions/src/destinations/reddit-audiences/types.ts
```diff
@@ -20,4 +20,4 @@ export interface UpdateAudienceReq {
     }
 }
 
-export type Columns = (typeof MAID_SCHEMA_NAME | typeof EMAIL_SCHEMA_NAME)[]
\ No newline at end of file
+export type Columns = (typeof MAID_SCHEMA_NAME | typeof EMAIL_SCHEMA_NAME)[]
```
cogwizzle, line 23: I don't care, but this is just increasing the size of the PR and someone might complain about it.

cogwizzle (commented):
Made some comments about the existing state. I'd like to defer to others if possible to get a final yes or no on this one due to our previous conversations about types.

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -100,7 +110,10 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
           }
       }
     })
-    return await this.requestClient<Data>(url, options)
+    return await this.requestClient<Data>(url, {
+      timeout: this.isRequestTimeoutExtended() ? 60_000 * 5 : 30_000,
```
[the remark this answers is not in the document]
pmunin: they are not really reused, so I didn't see value of extracting it to consts

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -128,6 +141,9 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
   logError(msg: string, metadata?: object) {
     this.logger.logError(msg, metadata)
   }
+  logWarn(msg: string, metadata?: object) {
```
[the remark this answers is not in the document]
pmunin: you probably mean protected, not private (it needs to be used in inherited classes). But again, this would be extra code for no real value. No one outside is supposed to instantiate or access this class (which is abstract btw) and someone decides to - they will have much bigger issues than accessing logger methods, I don't want our code to be polluted with baby guards, you will not see much of this in the rest of the code in our repos

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -142,15 +158,15 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
     this.statsClient.stats({ method: 'incr', metric, value, tags })
   }
 
-  statsHistogram(metric: string, value: number, tags?: string[]) {
+  statsHistogram(metric: string, value: number, tags?: StatsTags) {
     this.statsClient.stats({ method: 'histogram', metric, value, tags })
   }
 
-  statsSet(metric: string, value: number, tags?: string[]) {
+  statsSet(metric: string, value: number, tags?: StatsTags) {
```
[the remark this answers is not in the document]
pmunin: same  as [here](https://github.com/segmentio/action-destinations/pull/2325#discussion_r1774032904)

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
cogwizzle: Just as a label. It makes it a bit easier to read. Not married to this happening, but just wanted to make the recommendation.

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,320 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+   * @throws {IntegrationError} with retry flag set to true
+   */
+  throwRetryableError(message: string, code = 'ETIMEDOUT'): never {
+    const error = new IntegrationError(message, code, 500)
+    error.retry = true
+    throw error
+  }
+
+  createStepLogger(args: LogStepDetails): StepLogger {
+    //TODO: this is for troubleshooting, should be incorporated in track decorator
```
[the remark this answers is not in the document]
pmunin: I'd rather keep it as todo in code, not a big fan of noising up jira with something that is not urgent/critical: it will keep eyesore every backlog refinement, and then the jira ticket may get lost, while this comment will stay here with the code, until someone bump into it and find time to refactor it

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,320 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+       */
+      cacheGroup?: string
+      serializer?: CacheSerializer<T>
+      expiryInSeconds?: number
+      lockOptions?: LockOptions
+      saveRetry?: GetOrRetryArgs
+      onSaveFailed?: (error: Error) => void
+    }
+  ): Promise<T> {
+    const cache_group = options.cacheGroup || this.currentOperation?.parent?.func.name || ''
```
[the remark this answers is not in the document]
pmunin: cacheGroup undefined is unknown. not sure I understand what you suggest

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,320 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+    const log = this.createStepLogger({
+      tags: {
+        cache_group: cache_group,
+        withLock: !!options.lockOptions,
+        last_step: 'init'
+      },
+      logs: { key }
+    })
+
+    if (!this.engageDestinationCache) return log.track('no_cache', () => createValue())
```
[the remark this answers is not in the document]
pmunin: stepLogger.track does very similar to what decorator `@track` does, but for the inline functions:
1) log/stats before executing inline function
2) executes function
3) log/stats after executing function (respecting whether function returned error or value).
I'll add jsdoc to the interface

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,320 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+   * @param code
+   * @throws {IntegrationError} with retry flag set to true
+   */
+  throwRetryableError(message: string, code = 'ETIMEDOUT'): never {
+    const error = new IntegrationError(message, code, 500)
+    error.retry = true
+    throw error
+  }
+
+  createStepLogger(args: LogStepDetails): StepLogger {
```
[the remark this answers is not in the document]
pmunin: if you think about it - literally every single method in this class should be private or protected, but value of doing that is doubtful

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,320 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+      },
+      logs: { key }
+    })
+
+    if (!this.engageDestinationCache) return log.track('no_cache', () => createValue())
+    const cache = this.engageDestinationCache
+
+    const serializer = options.serializer || DefaultSerializer
+
+    const cacheRead = await log.track('cache_reading', () => getOrCatch(() => cache.getByKey(key)))
```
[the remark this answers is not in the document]
pmunin: that suggested code does not look more readable than what we currently have:
```
    const cacheRead = await step.track('cache_reading', () => getOrCatch(() => cache.getByKey(key)))

    if (cacheRead.error) {
      //redis error
      this.throwRetryableError('Error reading cache: ' + cacheRead.error.message)
    }
```

suggested code also implies that panic error will be caught by getOrCatch so i am not sure how that would work.
The purpose of current code is tracked getorcatch, then if error - throw it as retryable

packages/destination-actions/src/destinations/engage-messaging-twilio/__tests__/__helpers__/test-utils.ts
pmunin: I don't think so. 
this file changes were taken from your branch: https://github.com/segmentio/action-destinations/pull/2024/files#diff-8069a172846060b9b18e42ee39ab64955731610f9db26f0c32fa040d78c87e20

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,320 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+      } else {
+        //parsed cache successfully && cache is not expired
+        // parsedValue - either value or error was parsed
+        log.write('cache_hit', { tags: { cache_hit: true, cached_error: !!parsedCache.error } })
+        if (parsedCache.error) {
+          throw parsedCache.error
+        }
+        return parsedCache.value
+      }
+    }
```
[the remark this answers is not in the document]
pmunin: I don't see a way to break it down elegantly, without make a hit to readability. I'll add a comment to code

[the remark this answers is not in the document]
pmunin: to be clear this is not really different strategy. Locking in this case is an extra wrapping logic and using recursion is an elegant way to implement it without duplicating code

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -238,24 +274,68 @@ export abstract class MessageSendPerformer<
...
+    const res = [
+      `space_id:${settings.spaceId}`,
+      `projectid:${settings.sourceId}`,
+      `settings_region:${settings.region}`,
+      `channel:${this.getChannelType()}`
+    ]
+    const correlation_id = this.getCorrelationId()
+    if (correlation_id) res.push(`correlation_id:${correlation_id}`)
+
+    const computation_id = (payload as any).segmentComputationId
```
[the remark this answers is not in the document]
pmunin: good catch. added it to the base payload type

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -369,3 +444,34 @@ export abstract class MessageSendPerformer<
...
+    const parsed = JSON.parse(cachedValue) as ValueOrError<any>
+    if (parsed.error) {
+      const error = new IntegrationError(parsed.error.message, parsed.error.code, parsed.error.status)
+      error.retry = false
+      return { error }
+    } else {
+      return { value: { status: parsed.value } }
+    }
+  }
+}
```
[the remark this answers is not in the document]
pmunin: no, sendToRecepient returns any type

packages/actions-shared/src/engage/utils/delay.ts
```diff
@@ -0,0 +1,7 @@
+export async function delay(ms: number): Promise<void> {
+  return new Promise((resolve) => {
+    setTimeout(() => {
+      resolve()
+    }, ms)
+  })
+}
```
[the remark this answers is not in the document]
pmunin: i didn't find anny

packages/actions-shared/src/engage/utils/getOrRetry.ts
pmunin: not sure what do you mean by
> call this function sleep
JS is a single threaded, non-blocking language, there is no sleep here

packages/core/src/destination-kit/action.ts
pmunin: this is not our team's code, we didn't add any, we just added a comment to make linter happy.
Even if it was our code - I prefer not to add comments where something is intuitive and obvious, which I believe is the case here

packages/destination-actions/src/destinations/postscript/types.ts
pmunin: for records: took a while to figure out how to revert that change. It appeared to be the `.editorconfig` what was used by git hooks. So the solution was to restore from main and then commit it with `--no-verify`

packages/actions-shared/src/engage/utils/getOrRetry.ts
cogwizzle: It just works similar to sleep methods in like python.https://www.geeksforgeeks.org/sleep-in-python/

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
cogwizzle: Ah yeah. Well in this case the getOrCatch is the thing that responds so it would be the response value of `step.track` too right? The overall suggestion was to add some syntactic sugar to the way you could choose to consume the ValueOrError archetype you are going for. If you don't like it you don't have to implement it.

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
cogwizzle: Your call sir. I just was making a recommendation, but one could argue that private and protected don't really protect you. They're forced compliance with a sledge hammer. I have no strong feelings if you don't decide to do it.

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
cogwizzle: 100%. I looked to see if TS had a decorator type for this and it did not. I wish it did though.

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
cogwizzle: Firstly, the variable `cache_group` shouldn't ever really be an empty string as it provides no value in the logging right? Wouldn't it be better to be undefined? I see this is passed into a logger below, but it provides no value to log an empty tag right?
```ts
const cache_group: string | undefined = options.cacheGroup || this.currentOperation?.parent?.func.name
```

Secondly the name of the variable is `cache_group`. Generally in JavaScript or TypeScript projects it would be `cacheGroup`.

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
cogwizzle: I disagree and think it'll be ignored, but that is fine.

pmunin:
@joe-ayoub-segment I would like this to be deployed, but still waiting for approvals

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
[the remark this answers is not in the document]
nick-Ag: Going to second the recommendation to either track this TODO or remove it entirely

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,344 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+    if (!isNone(cacheRead.value)) {
+      //if cache FOUND (getByKey returned not null and not undefined)
+      // trying to parse the cached value
+      const { value: parsedCache, error: parsingError } = step.track('cache_parse', () =>
+        getOrCatch(() => serializer.parse(cacheRead.value!))
+      )
+
+      if (parsingError) {
+        //exception happened while parsing the cache.
+        // Log it and execute as if we don't have cache
```
nick-Ag, line 388: I think this if block is missing the 'Log it and execute as if we don't have cache' code. Empty if block

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,344 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+      getOrCatch(() => createValue())
+    )
+
+    //before returning result - we need to try to serialize it and store it in cache
+    const stringified = step.track('cache_stringified', () =>
+      getOrCatch(() => serializer.stringify(resultError ? { error: resultError } : { value: result }))
+    )
+
+    if (stringified.error) {
+      //
```
nick-Ag, line 414: Why is this if block empty?

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
```diff
@@ -168,6 +184,344 @@ export abstract class EngageActionPerformer<TSettings = any, TPayload = any, TRe
...
+    )
+
+    if (redisError) {
+      this.throwRetryableError('Error acquiring lock: ' + redisError.message)
+    } else if (!lock?.release) {
+      // no redis error and no lock acquired - means it was acquiring timeout
+      step.write('cache_lock_timeout_exit')
+      this.throwRetryableError('Timeout while acquiring lock')
+    }
+    //if lock obtained or there was redis error - execute createValue and finally release lock
```
nick-Ag, line 506: Seems like you handle the redis error case in the first if block?
```suggestion
    //if lock obtained  - execute createValue and finally release lock
```

packages/actions-shared/src/engage/utils/EngageLogger.ts
```diff
@@ -25,33 +25,53 @@ export class EngageLogger extends OperationLogger {
...
-    )
+    try {
+      const msgs = [msg, ...(metadata ? [JSON.stringify(metadata)] : [])]
+      const [firstMsg, ...rest] = msgs
+      this.loggerClient?.info(
+        `TE Messaging: ${this.channelType} ${firstMsg}`,
+        ...rest,
+        JSON.stringify({ ...this.logDetails, ...metadata })
+      )
+    } catch {
```
nick-Ag, line 37: Could we remove these two try/catch blocks then? Wouldn't it be easier to debug if we were throwing when an error occurred rather than completely ignoring the error?

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
pmunin: good catch, fixing it

packages/actions-shared/src/engage/utils/EngageLogger.ts
pmunin: no, we have unreliable behavior of logging, specifically dropped logs during high load. It's still not clear why logs are dropped, infra-observability is aware of it and could not help (see rfc doc for details). This empty catch is to make execution logger-fail-proof

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
pmunin: good catch, there were multuple iterations of changes of this logic, and I forgot to update the comment

packages/actions-shared/src/engage/utils/EngageActionPerformer.ts
pmunin: good catch. We don't need explicit error logging here, because it will be done by step.track. Removing empty block.

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -311,39 +398,33 @@ export abstract class MessageSendPerformer<
...
+      const sendResult = sendResults[0]
+      if (sendResult.status === 'rejected') {
+        throw sendResult.error
       } else {
-        return sendResults[0].result
+        return sendResult.result
       }
     }
 
-    const fulfilled = sendResults.filter((sr) => sr.status === 'fulfilled')
```
nick-Ag, line 321: Was this code block that checks if `fulfilled` has at least one success purposely removed? Wondering why we no longer would want to return in this case?

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
```diff
@@ -311,39 +398,33 @@ export abstract class MessageSendPerformer<
...
       throw AggregateError.create({
         errors: rejected.map((r) => r.error),
-        message: (msg) => 'Failed to send to all subscribed recepients (retryable):' + msg
+        takeCodeAndStatusFromError: firstRetryableError,
+        message: (msg) =>
+          `Failed to send to all subscribed recepients (${firstRetryableError ? 'retryable' : 'not-retryable'}):` + msg
       })
     }
 
-    //NON RETRYABLE CASE:
```
nick-Ag, line 339: Wondering why this non-retryable case was removed and not replaced? Do we no longer handle non-retryable errors?

packages/actions-shared/src/engage/utils/getOrRetry.ts
```diff
@@ -0,0 +1,91 @@
+// import { retry } from '@segment/actions-core/retry' // TODO: fix this import. Currently it gives a runtime error: retry_1.retry is not a function (see below)
```
nick-Ag, line 1: calling out this TODO. Please resolve or remove entirely

packages/actions-shared/src/engage/utils/getOrRetry.ts
```diff
@@ -0,0 +1,91 @@
...
+
+interface RetryOptions {
+  retries?: number
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onFailedAttempt?: (error: any, attemptCount: number) => PromiseLike<void> | void
+}
+
+const DEFAULT_RETRY_ATTEMPTS = 2
+
+export async function retry<T>(
```
nick-Ag, line 68: This looks like you copied it from actions-core. Please resolve the import at the top and avoid duplicating code entirely
https://github.com/segmentio/action-destinations/blob/d7a9a2a1fa8a0d26d5f0228d34fb80c6da0a55ed/packages/core/src/retry.ts#L9

packages/actions-shared/src/engage/utils/getOrRetry.ts
```diff
@@ -0,0 +1,91 @@
+// import { retry } from '@segment/actions-core/retry' // TODO: fix this import. Currently it gives a runtime error: retry_1.retry is not a function (see below)
+import { delay } from './delay'
```
nick-Ag, line 2: for both of these imports (`delay` and `getOrCatch`) could we just define them both in this file and avoid importing? I feel like that would be easier for a future developer to read thru. Instead of needing multiple code windows to understand what each util function does we could just define them all here. I'm assuming `delay` and `getOrCatch` will only be imported by this file

packages/actions-shared/src/engage/utils/getOrRetry.ts
```diff
@@ -0,0 +1,91 @@
...
+import { ValueOrError, getOrCatch } from './getOrCatch'
+
+export async function getOrRetry<T>(
+  action: (attempt: number) => PromiseLike<T> | T,
+  args?: GetOrRetryArgs
+): Promise<ValueOrError<T>> {
+  const retryIntervalMs = args?.retryIntervalMs === undefined ? backoffRetryPolicy() : args?.retryIntervalMs ?? 1000
+  return getOrCatch(
```
nick-Ag, line 10: Do we really need a utility `getOrCatch` method here? This is fairly difficult to read imo

packages/core/src/create-test-integration.ts
```diff
@@ -45,13 +45,18 @@ interface InputData<Settings> {
   auth?: AuthTokens
   /**
    * The features available in the request based on the customer's sourceID;
-   * `features`, `stats`, `logger`, `dataFeedCache`, and `transactionContext` and `stateContext` are for internal Twilio/Segment use only.
+   * `features`, `stats`, `logger`, `engageDestinationCache`, `transactionContext` and `stateContext` are for internal Twilio/Segment use only.
    */
   features?: Features
+  /** Engage internal use only. DO NOT USE. */
```
nick-Ag, line 51: I think `statsContext`, and `logger` can be used outside of the Engage context right? Could you remove the comments above each of those?

packages/core/src/destination-kit/index.ts
```diff
@@ -425,8 +430,8 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
       this.settingsSchema = fieldsToJsonSchema(this.authentication.fields)
     }
 
-    for (const action of Object.keys(destination.actions)) {
-      this.partnerAction(action, destination.actions[action])
+    for (const [name, action] of Object.entries(destination.actions)) {
```
nick-Ag, line 433: Why is this change necessary?

packages/core/src/destination-kit/types.ts
```diff
@@ -66,15 +68,17 @@ export interface ExecuteInput<
   matchingKey?: string
   /** The data needed in OAuth requests */
   readonly auth?: AuthTokens
-  /**
-   * The features available in the request based on the customer's sourceID;
-   * `features`,`stats`, `logger` , `transactionContext` and `stateContext` are for internal Twilio/Segment use only.
-   */
+  /** Engage internal use only. DO NOT USE. */
   readonly features?: Features
+  /** Engage internal use only. DO NOT USE. */
```
nick-Ag, line 73: Same comment here, `logger` and `statsConext` aren't engage specific imo. Just segment internal

packages/core/src/destination-kit/types.ts
```diff
@@ -66,15 +68,17 @@ export interface ExecuteInput<
   matchingKey?: string
   /** The data needed in OAuth requests */
   readonly auth?: AuthTokens
-  /**
```
nick-Ag, line 69: Could you replace `The features available in the request based on the customer's sourceID;` comment. I don't think that needs to be removed

packages/destination-actions/src/destinations/engage-messaging-sendgrid/__tests__/send-email.test.ts
```diff
@@ -88,6 +99,7 @@ describe.each([
           source_id: 'sourceId',
           space_id: 'spaceId',
           user_id: userData.userId,
+          //message_id: 'messageId',
```
nick-Ag, line 102: should this be removed entirely?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/__tests__/send-email.test.ts
```diff
@@ -192,6 +206,7 @@ describe.each([
           timestamp,
           event: 'Audience Entered',
           userId: userData.userId,
+          //messageId: 'messageId',
```
nick-Ag, line 209: should this be removed entirely?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/SendEmailPerformer.ts
nick-Ag: should this be removed entirely?

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/SendEmailPerformer.ts
```diff
@@ -272,7 +273,23 @@ export class SendEmailPerformer extends MessageSendPerformer<Settings, Payload>
 
   @track()
   async getBodyTemplateFromS3(bodyUrl: string) {
-    const { content } = await this.request(bodyUrl, { method: 'GET', skipResponseCloning: true })
+    const content = !this.isFeatureActive('engage-cache-email-template')
```
nick-Ag, line 276: I think the ternary here should be replaced by an if/else statement that assigns to `context`

packages/core/src/destination-kit/types.ts
pmunin: yep it was removed by accident, thanks

packages/core/src/destination-kit/index.ts
pmunin: readability improvement, but will reverted for denoising purposes

packages/core/src/create-test-integration.ts
pmunin: ~1.5 yrs ago - this was created for Engage purposes only. I'm not aware if that was changed, but I don't mind cleaning up those comments

packages/actions-shared/src/engage/utils/getOrRetry.ts
pmunin: not sure what you suggest? getOrRetry needs to return `Promise<ValueOrError>` per current design

packages/actions-shared/src/engage/utils/EngageLogger.ts
```diff
@@ -25,33 +25,53 @@ export class EngageLogger extends OperationLogger {
...
+  }
 
   getErrorMessage(error: unknown) {
     const errorDetails = getErrorDetails(error)
-    let res = `${errorDetails.message}`
-    if (errorDetails.code) res += `. Code: ${errorDetails.code}`
-    if (errorDetails.status) res += `. Status: ${errorDetails.status}`
+    let res = `${errorDetails?.message}`
+    if (errorDetails?.code) res += `. Code: ${errorDetails.code}`
+    if (errorDetails?.status) res += `. Status: ${errorDetails.status}`
```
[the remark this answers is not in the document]
pmunin: other story for null checks? this was to make code safer, as there was a chance where error is undefined

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
pmunin: it's not removed, but combined in new logic. if `takeCodeAndStatusFromError` is not specified it will use first error from the array, which will be non-retryable

packages/actions-shared/src/engage/utils/MessageSendPerformer.ts
pmunin: this behavior was incorrect, if we have ANY errors we should throw it.

packages/actions-shared/src/engage/utils/getOrRetry.ts
pmunin: this something for action-destinations owners to resolve. This import fails in runtime

packages/actions-shared/src/engage/utils/getOrRetry.ts
pmunin: It's opinionated comment, everyone has their own. I have the opposite. the bigger file - the harder it is to review changes and to navigate through the code, much easier to navigate through the tree, than scrolling through the text.

packages/actions-shared/src/engage/utils/getOrRetry.ts
pmunin: This is exactly what the comment above says:
```
/**
 * import { retry } from '@segment/actions-core/retry'
 * ^ This import is not working, gives a runtime error: retry_1.retry is not a function during jest tests,
 * so the following code is a copy of the original code from: @segment/actions-core/retry.
 * Once jest config is fixed to use the correct import, the following code should be removed.
 */
```

This is for owners of the repo to resolve in a separate PR. has nothing to do with current story.

packages/destination-actions/src/destinations/engage-messaging-sendgrid/sendEmail/SendEmailPerformer.ts
pmunin: addressed

nick-Ag (changes requested):
Read thru all the code, thank you for addressing comments so far. 
I do think for the `getOrRetry.ts` file, the broken import: `// import { retry } from '@segment/actions-core/retry' // TODO: fix this import. Currently it gives a runtime error: retry_1.retry is not a function (see below)` needs to be fixed before merging.

If you can't find a way to get that working then please write up a ticket on the stratconn board that describes how to reproduce the issue you're seeing so that we can be aware of it. We shouldn't leave TODOs in the codebase without a solid way to followup on them. Please don't leave items for our team to fix without notifying us of the issue

pmunin:
@nick-Ag fixed it (it was easier to fix than I expected), plz review

nick-Ag (approved):

packages/actions-shared/src/engage/utils/EngageLogger.ts
```diff
@@ -25,33 +25,53 @@ export class EngageLogger extends OperationLogger {
...
-    const [firstMsg, ...rest] = msgs
-    this.loggerClient?.info(
-      `TE Messaging: ${this.channelType} ${firstMsg}`,
-      ...rest,
-      JSON.stringify({ ...this.logDetails, ...metadata })
-    )
+    try {
+      const msgs = [msg, ...(metadata ? [JSON.stringify(metadata)] : [])]
+      const [firstMsg, ...rest] = msgs
+      this.loggerClient?.info(
```
brennan, line 32: Do we need to leave these logs around permanently? If they're to facilitate testing, can we commit to removing them once the changes are validated in production?

joe-ayoub-segment:
hi @brennan there's some core changes in here. Do you want me to deploy this on Tuesday, or would you prefer that it's done by someone in the US during US working hours so that it can be monitored more closely?

packages/actions-shared/src/engage/utils/EngageLogger.ts
pmunin: yes, we need it. this is old logic - logInfo is [feature-flagged](https://flagon.segment.com/families/centrifuge-destinations/gates/engage-messaging-log-info) and by default OFF in Production. We occasionally turn it on for certain workspaces for troubleshooting scenarios

brennan:
@joe-ayoub-segment Please do not merge/deploy this PR as part of the regularly scheduled deploy on October 8. The on-call will deploy this separately in the late morning/early afternoon. Thank you.

PR #2343, opened 2024-08-27
Adds a performBatch method to the Iterable updateUser destination.

craigotis:
Refactors some of the payload transformation code from the perform method, to be reused in the performBatch.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/iterable/updateUser/index.ts
+++ b/packages/destination-actions/src/destinations/iterable/updateUser/index.ts
@@ -11,6 +11,35 @@ import {
 } from '../shared-fields'
 import { convertDatesInObject, getRegionalEndpoint } from '../utils'
 
+interface UserUpdateRequestPayload {
+  email?: string
+  userId?: string
+  dataFields?: {
+    [k: string]: unknown
+  }
+  mergeNestedObjects?: boolean
+}
+
+interface BulkUserUpdateRequestPayload {
+  users: UserUpdateRequestPayload[]
+}
+
+const transformIterableUserPayload: (payload: Payload) => UserUpdateRequestPayload = (payload) => {
+  // Store the phoneNumber value before deleting from the top-level object
+  const phoneNumber = payload.phoneNumber
+  delete payload.phoneNumber
+
+  const formattedDataFields = convertDatesInObject(payload.dataFields ?? {})
+  const userUpdateRequest: UserUpdateRequestPayload = {
+    ...payload,
+    dataFields: {
+      ...formattedDataFields,
+      phoneNumber: phoneNumber
+    }
+  }
+  return userUpdateRequest
+}
+
 const action: ActionDefinition<Settings, Payload> = {
   title: 'Upsert User',
   description: 'Creates or updates a user',
@@ -34,38 +63,28 @@ const action: ActionDefinition<Settings, Payload> = {
     }
   },
   perform: (request, { payload, settings }) => {
-    const { email, userId, dataFields } = payload
-
-    if (!email && !userId) {
+    if (!payload.email && !payload.userId) {
       throw new PayloadValidationError('Must include email or userId.')
     }
 
-    interface UserUpdateRequest {
-      email?: string
-      userId?: string
-      dataFields?: {
-        [k: string]: unknown
-      }
-      mergeNestedObjects?: boolean
-    }
-
-    // Store the phoneNumber value before deleting from the top-level object
-    const phoneNumber = payload.phoneNumber
-    delete payload.phoneNumber
-
-    const formattedDataFields = convertDatesInObject(dataFields ?? {})
-    const userUpdateRequest: UserUpdateRequest = {
-      ...payload,
-      dataFields: {
-        ...formattedDataFields,
-        phoneNumber: phoneNumber
-      }
-    }
+    const updateUserRequestPayload: UserUpdateRequestPayload = transformIterableUserPayload(payload)
 
     const endpoint = getRegionalEndpoint('updateUser', settings.dataCenterLocation as DataCenterLocation)
     return request(endpoint, {
       method: 'post',
-      json: userUpdateRequest,
+      json: updateUserRequestPayload,
+      timeout: Math.max(30_000, DEFAULT_REQUEST_TIMEOUT)
+    })
+  },
+  performBatch: (request, { settings, payload }) => {
+    const bulkUpdateUserRequestPayload: BulkUserUpdateRequestPayload = {
+      users: payload.map(transformIterableUserPayload)
+    }
+
+    const endpoint = getRegionalEndpoint('bulkUpdateUser', settings.dataCenterLocation as DataCenterLocation)
+    return request(endpoint, {
+      method: 'post',
+      json: bulkUpdateUserRequestPayload,
       timeout: Math.max(30_000, DEFAULT_REQUEST_TIMEOUT)
     })
   }
--- a/packages/destination-actions/src/destinations/iterable/updateUser/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/iterable/updateUser/__tests__/index.test.ts
@@ -5,115 +5,175 @@ import Destination from '../../index'
 const testDestination = createTestIntegration(Destination)
 
 describe('Iterable.updateUser', () => {
-  it('works with default mappings', async () => {
-    const event = createTestEvent({ type: 'identify' })
+  describe('perform', () => {
+    it('works with default mappings', async () => {
+      const event = createTestEvent({ type: 'identify' })
 
-    nock('https://api.iterable.com/api').post('/users/update').reply(200, {})
+      nock('https://api.iterable.com/api').post('/users/update').reply(200, {})
 
-    const responses = await testDestination.testAction('updateUser', {
-      event,
-      useDefaultMappings: true
-    })
-
-    expect(responses.length).toBe(1)
-    expect(responses[0].status).toBe(200)
-  })
-
-  it('throws an error if `email` or `userId` are not defined', async () => {
-    const event = createTestEvent({
-      type: 'identify',
-      traits: {
-        firstName: 'Johnny',
-        lastName: 'Depp'
-      }
-    })
-
-    await expect(
-      testDestination.testAction('updateUser', {
-        event
+      const responses = await testDestination.testAction('updateUser', {
+        event,
+        useDefaultMappings: true
       })
-    ).rejects.toThrowError(PayloadValidationError)
-  })
 
-  it('maps phoneNumber in dataFields', async () => {
-    const event = createTestEvent({
-      type: 'identify',
-      userId: 'user1234',
-      traits: {
-        phone: '+14158675309'
-      }
+      expect(responses.length).toBe(1)
+      expect(responses[0].status).toBe(200)
     })
 
-    nock('https://api.iterable.com/api').post('/users/update').reply(200, {})
-
-    const responses = await testDestination.testAction('updateUser', {
-      event,
-      useDefaultMappings: true
-    })
-
-    expect(responses.length).toBe(1)
-    expect(responses[0].options.json).toMatchObject({
-      userId: 'user1234',
-      dataFields: {
-        phoneNumber: '+14158675309'
-      }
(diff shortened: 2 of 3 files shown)


packages/destination-actions/src/destinations/iterable/updateUser/index.ts
```diff
@@ -34,38 +63,28 @@ const action: ActionDefinition<Settings, Payload> = {
...
-      }
+    const endpoint = getRegionalEndpoint('updateUser', settings.dataCenterLocation as DataCenterLocation)
+    return request(endpoint, {
+      method: 'post',
+      json: updateUserRequestPayload,
+      timeout: Math.max(30_000, DEFAULT_REQUEST_TIMEOUT)
+    })
+  },
+  performBatch: (request, { settings, payload }) => {
+    const bulkUpdateUserRequestPayload: BulkUserUpdateRequestPayload = {
```
craigotis, line 80: Note I chose **not to** enforce the `payload.email || payload.userId` check here, since the behavior was semi-undefined. Should we fail the whole batch? Just remove that one erroneous entry?

Felt _less surprising_ to just let Iterable treat it as a failure and process the rest of the batch.

TheFacto (approved):

packages/destination-actions/src/destinations/iterable/updateUser/index.ts
```diff
@@ -34,38 +63,28 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      users: payload.map(transformIterableUserPayload)
     }
 
-    const endpoint = getRegionalEndpoint('updateUser', settings.dataCenterLocation as DataCenterLocation)
+    const endpoint = getRegionalEndpoint('bulkUpdateUser', settings.dataCenterLocation as DataCenterLocation)
     return request(endpoint, {
       method: 'post',
-      json: userUpdateRequest,
+      json: bulkUpdateUserRequestPayload,
       timeout: Math.max(30_000, DEFAULT_REQUEST_TIMEOUT)
```
joe-ayoub-segment, line 88: hi @KakeJopulsky - if we're going use the Iterable bulk endpoint is there a specific timeout we should use?

joe-ayoub-segment:
hi @craigotis and @KakeJopulsky 

Craig thanks for raising this PR to add batch support to the updateUser Action for the Iterable Destination. 

hi @KakeJopulsky - Do you know if the bulk API was missing when this Destination was built? Just wondering why we never used it! 

Anyway thanks again Craig for the PR. 

I'll need @KakeJopulsky 's approval before this can be deployed. 
Also, we may need to add in some addition details.

1. Jake, what is the max batch size for this bulk API? I may need to specify this in the configuration. 
2. Jake, do you know what timeout value we should specify, if any, for calls to the bulk API?
3. Jake, should we make the bulk call the default call for the updateUser Action?
4. Craig, you asked about how to handle multiple responses. We're building that capability at the moment but it's not yet available for use. For now your implementation makes sense. i.e less validation of individual events -> send all events. 

@KakeJopulsky - would you be able to review this PR please, and clarify if we should apply batching to any of the other Actions?

Thank you both!!

craigotis:
Thanks @joe-ayoub-segment, I had very similar questions!

As you can see in the screencast, the bulk update endpoint for Iterable returns a payload that looks something like:
```
{
  "successCount": 947,
  "failCount": 4,
  "invalidEmails": [...],
  "invalidUserIds": [...],
  "filteredOutFields": [...],
  "createdFields": [...]
}
```
And while that might be visible in the debugger/any response-displaying tooling, it seems that currently, this data can't be more transparently displayed. **Which is fine,** we just haven't used any batch-enabled Segment destinations (yet) in our travels, so I think I'm more interested in knowing ahead of time _how_ Segment deals with the `performBatch` returns. Today, the `perform`, for 30,453 records, will report status that looks like:

361960272-5cf2cb60-16a3-4258-8f34-40c3f72ac84e

And I'm curious what happens (as an example) if we throw those same 30k records at this batch-enabled destination. If they get broken into chunks of 1,000 and we make ~30 calls to Iterable, do we see `Successful: 30 | Failed: 0`, or is Segment smart enough to render this panel differently?

craigotis:
@joe-ayoub-segment Oh, one other note - once we get this across the line, we'll come back again with batching for the Track Event Iterable destination as well. We just wanted to learn the ropes and not over-complicate this first PR.

joe-ayoub-segment:
> @joe-ayoub-segment Oh, one other note - once we get this across the line, we'll come back again with batching for the Track Event Iterable destination as well. We just wanted to learn the ropes and not over-complicate this first PR.

Sounds good @craigotis . 

Just need @KakeJopulsky 's input...

joe-ayoub-segment:
hi @craigotis I've still not heard back from @KakeJopulsky. I've followed up on Slack now as well. 
Apologies for the delay. If I don't hear from Jake today I'll message some other folks at Iterable.

packages/destination-actions/src/destinations/iterable/updateUser/index.ts
KakeJopulsky: Under normal circumstances 3s should be sufficient!

KakeJopulsky:
Hey @craigotis @joe-ayoub-segment, thank you for putting this together!

**Batch sizing:** There is no fixed limit on the number of records per request, but the maximum request size is 4MB.
**Timeouts:** Under normal circumstances 3s should be sufficient. The limit is 5 req/s per project. 

The bulk endpoint is not ideal for real-time event tracking. Therefore, I recommend keeping the /users/update endpoint as the default.

The track event action could be updated to support batching as well.

KakeJopulsky:
LGTM! I pulled @craigotis's branch and tested locally.

craigotis:
> Hey @craigotis @joe-ayoub-segment, thank you for putting this together!
> 
> **Batch sizing:** There is no fixed limit on the number of records per request, but the maximum request size is 4MB. **Timeouts:** Under normal circumstances 3s should be sufficient. The limit is 5 req/s per project.
> 
> The bulk endpoint is not ideal for real-time event tracking. Therefore, I recommend keeping the /users/update endpoint as the default.
> 
> The track event action could be updated to support batching as well.

Thanks @KakeJopulsky! Based on that...

1. Personally I think the **30 second** timeout currently implemented in this destination action is a safe default - 3 seconds (or was that a typo?) feels a bit tight, assuming we might occasionally want to be prepared for _not normal_ circumstances.
2. Eyeballing a few 4kb JSON payloads, I think Segment's default batch size of 1,000 records (@joe-ayoub-segment can you confirm?) would _probably_ be okay for most use cases, though I'm not sure what the best mitigation strategy is here. (Sizing it up in the implementation and making multiple Iterable calls feels risky, what if one succeeds and one doesn't, is that a "failed" or "successful" batch in Segment's eyes?) Can Segment drop the default batch size for this destination to 500 records? Or, better yet, could Segment allow for **configurable** batch sizes based on the destination mapping configuration, in the future?
3. On the topic of _defaults_, I know we can update the `ActionDefinition` to [optionally define](https://github.com/segmentio/action-destinations/tree/main?tab=readme-ov-file#batching-requests) how `enable_batching` is presented, but if we **don't** do that, what default does Segment assume for this configuration?

KakeJopulsky:
> 1. Personally I think the **30 second** timeout currently implemented in this destination action is a safe default. 3 seconds (or was that a typo?) feels a bit tight, assuming we might occasionally want to be prepared for _not normal_ circumstances.
> 2. ... think Segment's default batch size of 1,000 records (@joe-ayoub-segment can you confirm?) would _probably_ be okay for most use cases ...

@craigotis 
1. I misunderstood the timeout earlier—30 seconds as the default is sufficient!
2. +1 Generally, 1000 records should be manageable. The challenge is that more fields in each record increase the data footprint. I anticipate a mix of users performing partial versus full record updates.

joe-ayoub-segment:
> that might be visible in the debugger/any response-displaying tooling, it seems that currently, this data can't be more transparently displayed. **Which is fine,** we just haven't used any batch-enabled Segment destinations (yet) in our travels, so I think I'm more interested in knowing ahead of time _how_ Segment deals with the `performBatch` returns. Today, the `perform`, for 30,453 records, will report status that looks like

Hi @craigotis - currently Segment's performBatch() function will return a single response code, which will result in an all-or-nothing response in the delivery tab of the Destination. It's not ideal, and we are implementing the ability to handle multi-status responses, however that won't be implemented in time for this PR.

joe-ayoub-segment:
> 2\. o 500 records? Or, better yet, could Segment allo

Hi @craigotis , we 'can' make the batch size configurable by the customer, however I think it would be better to set the Max batch size to 500 and then hide the field so customers can't change it. 

 We wouldn't want customers setting the batch size in excess of 1000, and there's currently no way to impose a limit if we do expose the batch_size field to customers.

joe-ayoub-segment:
hi @craigotis , looks like we have approval from @KakeJopulsky . 

Can you please add the following field definitions to the Action please?
```
    enable_batching: {
      label: 'Enable Batching',
      description: 'When enabled, Segment will send data to Iterable in batches of up to 500',
      type: 'boolean',
      required: false,
      default: false
    },
    batch_size: {
      label: 'Batch Size',
      description: 'Maximum number of events to include in each batch. Actual batch sizes may be lower.',
      type: 'number',
      unsafe_hidden: true,
      required: false,
      default: 500
    }
```

This will enable batching, but will leave it disabled by default. 
It will also set the max batch size to 500, and will hide the batch_size field so customers can't change the setting. 
Finally, both of these fields will be 'optional', so it won't impact customers who have already set up their Destinations. 

After adding the config please be sure to run the './bin/run generate:types' command, and commit the changes. 

After that we can deploy!

craigotis:
> hi @craigotis , looks like we have approval from @KakeJopulsky .
> 
> Can you please add the following field definitions to the Action please?
> 
> ```
>     enable_batching: {
>       label: 'Enable Batching',
>       description: 'When enabled, Segment will send data to Iterable in batches of up to 500',
>       type: 'boolean',
>       required: false,
>       default: false
>     },
>     batch_size: {
>       label: 'Batch Size',
>       description: 'Maximum number of events to include in each batch. Actual batch sizes may be lower.',
>       type: 'number',
>       unsafe_hidden: true,
>       required: false,
>       default: 500
>     }
> ```
> 
> This will enable batching, but will leave it disabled by default. It will also set the max batch size to 500, and will hide the batch_size field so customers can't change the setting. Finally, both of these fields will be 'optional', so it won't impact customers who have already set up their Destinations.
> 
> After adding the config please be sure to run the './bin/run generate:types' command, and commit the changes.
> 
> After that we can deploy!

Done and done, thanks for the snippet! Saved me some doc diving. 
d44e80e

joe-ayoub-segment:
hi @craigotis this PR has been deployed. Can you confirm that it works as you expect please?

craigotis:
> hi @craigotis this PR has been deployed. Can you confirm that it works as you expect please?

Hey @joe-ayoub-segment , we don't have a ton of visibility through the Segment UI, but while we see the `Enable Batching` option in the mapping and enabled it:
image

When we reset and re-ran our sync with approximately 160k test records, it looks like it's going to take about 2 hours, which is as long as (if not longer than) it took with the non-batching approach. (And much longer than I'd expect. For batches of 500 records, and 5 req/s, my ballpark was 1-2 minutes.)

KakeJopulsky:
@craigotis, I reviewed the logs on my end and found that the batch sizes varied, ranging from 1 to 256 per request.

craigotis:
> @craigotis, I reviewed the logs on my end and found that the batch sizes varied, ranging from 1 to 256 per request.

Oof - @joe-ayoub-segment any metrics on your side that might hint as to the median batch size? Or thoughts on how we can get that closer to the max batch size?

Our job finally finished, but it was actually _slower_ than the non-batch destination action. 😢 

Screenshot 2024-09-04 at 6 00 18 PM

marinhero:
Hi @craigotis, let me take a look into it and I'll get back to you in an hour or so. Thanks!

marinhero:
Hey team, looks like the `max_batch` size value is below the threshold we have for our batching algorithm to kick-in. I'm going to increase its size and run a deploy. Expect this to be fixed within the next 48 hours. Thank you.

PR #2358, opened 2024-09-02
[STRATCONN-4162]:Add min/max validations for number type

itsarijitray:
JIRA: https://segment.atlassian.net/browse/STRATCONN-4062

In an effort to expose batch size to the end customers for the batching destinations, Batch size min/max limit needs to be put in place in all the batch supported action destinations to ensure stability of the Integration.

https://docs.google.com/document/d/1jhAJ3JNcFHs7n9l4w8CVOohU52ZRqFuFX2qGrsjRSkk/edit

Screenshot 2024-09-13 at 5 31 24 PM
Screenshot 2024-09-13 at 5 33 38 PM

## Testing


marinhero:
How will this affect the existing destinations that have batching enabled? will they start to see it or only when explicitly indicated?

itsarijitray:
> How will this affect the existing destinations that have batching enabled? will they start to see it or only when explicitly indicated?

If the batch_size field is explicitly declared and with a min/max config, only then this is validation runs. Existing definitions will not be impacted as these are optional properties

packages/core/src/__tests__/schema-validation.test.ts
```diff
@@ -239,4 +239,22 @@ describe('validateSchema', () => {
       }
     `)
   })
+
+  it('should validate min/max for number type fields', () => {
```
marinhero, line 243: NIT: It would be great to see some tests with more extreme conditions like the ones we talked about in the actions PR.

marinhero (approved):

PR #2370, opened 2024-09-04
Enable bulk_batching for Iterable updateUser action

marinhero:
Follow up to https://github.com/segmentio/action-destinations/pull/2343 to enable bulk batching for Iterable updateUser action.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/iterable/updateUser/index.ts
+++ b/packages/destination-actions/src/destinations/iterable/updateUser/index.ts
@@ -74,7 +74,7 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'number',
       unsafe_hidden: true,
       required: false,
-      default: 500
+      default: 1001
     }
   },
   perform: (request, { payload, settings }) => {


joe-ayoub-segment:
Thanks for the PR @marinhero. I'll schedule for review. cc @craigotis

joe-ayoub-segment:
hi @marinhero why is this set to 1001? I understand that Iterable has a max batch size of 1000. Maybe @KakeJopulsky can confirm.

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @marinhero and @craigotis . I merged this so that it will go out as soon as someone runs a deploy. 
Let's confirm that 1001 is a safe max batch size for Iterable though, and make a change if it's not.

craigotis:
Thanks @joe-ayoub-segment and @marinhero - also curious about the 1001 (vs. 1000 or some other number), but I think we're fine from an Iterable perspective, since [the docs](https://api.iterable.com/api/docs#users_bulkUpdateUser) only seem to impose a 4MB limit on the request body itself, not necessarily on the number of records it contains.

marinhero:
@craigotis the deploy is now completed. Please delete your existing mapping, create a new one and run a test. Let us know how it goes!

craigotis:
@marinhero That looks _much_ better. We'll continue testing with a larger dataset and keep you posted.
image

joe-ayoub-segment:
hi @craigotis and @KakeJopulsky - 
I"m glad the change seems to be having the desired effect. 

However we need to verify that the max batch size for Iterable is 1001 or under. Is this the case?

KakeJopulsky:
> hi @craigotis and @KakeJopulsky - I"m glad the change seems to be having the desired effect.
> 
> However we need to verify that the max batch size for Iterable is 1001 or under. Is this the case?

The maximum batch size is a 4MB payload, with no limit on the number of records/users per request. Is there anyway to ensure batches stay under 4MBs?

joe-ayoub-segment:
> @joe-ayoub-segment
> 
> > However we need to verify that the max batch size for Iterable is 1001 or under. Is this the case?
> 
> The maximum batch size is a 4MB payload, with no limit on the number of records/users per request. Is there anyway to ensure batches stay under 4MBs?

Hi Jake the only way could be to measure the size of the outgoing payload before sending it. 
However 4MB is very generous and is very unlikely to be breached. 

Segment's max event size is 32kb per event, so 32 X 1001 =  ~3MB. 

I think we're more than save at the 1001 event max limit. 

Thanks for your help.

PR #2390, opened 2024-09-11
StackAdapt Audiences destination

AlliterativeAlice:
This is to add a StackAdapt audiences destination to enable customers to sync their Segment Engage audiences to StackAdapt so that they can target their programmatic ad campaigns to their Segment Engage audiences.

## Testing


packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/index.ts
```diff
@@ -0,0 +1,176 @@
...
+        }
+      }
+    }
+  }
+}
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Forward Profile',
+  description: 'Forward new or updated user profile to StackAdapt',
+  defaultSubscription: 'type = "identify" or type = "alias" or type = "track"',
```
joe-ayoub-segment, line 32:
```suggestion
  defaultSubscription: 'type = "identify" or type = "track"',
```

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/index.ts
```diff
@@ -0,0 +1,176 @@
...
+      description: 'The StackAdapt advertiser to add the profile to.',
+      type: 'string',
+      required: true,
+      dynamic: true
+    }
+  },
+  dynamicFields: {
+    advertiser_id: async (request, { page }): Promise<DynamicFieldResponse> => {
+      // Even though its typescript type is string, in testing I found page can be a number so I use == here
+      page = !page || page == '1' ? '' : page
```
joe-ayoub-segment, line 125: I'm not sure if pagination is fully implemented for dynamic fields on the front end website yet. I don't believe we have any Integrations using it yet. 

How long will the max size be for the returned list of choices?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+
+interface Mapping {
+  incoming_key: string
+  destination_key: string
+  data_type: string
+  is_pii: boolean
+}
+
+export async function performForwardProfiles(request: RequestClient, events: Payload[]) {
+  if (events.length === 0) return
```
joe-ayoub-segment, line 47: nit: shouldn't be needed, but OK.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+  const fieldsToMap: Set<string> = new Set(['userId'])
+  const fieldTypes: Record<string, string> = { userId: 'string' }
+  const advertiserId = events[0].advertiser_id
```
joe-ayoub-segment, line 50: Looks like you want advertisserId to be static across all events. 
The customer can actually override the returned values from the dynamic function call, and send any value they like through the UI for this field. 

Options to work around this are: 

1. We can edit the advertiser_id field to restrict inputs other than from the returned dynamic list call. OR
2. We can move advertiser_id to the global settings so that it's static across the instance of the Integration. 

If you want to go with option 1, please add this line to the advertiser_id field defintion: 

```
disabledInputMethods: ['literal', 'variable', 'function', 'freeform', 'enrichment']
```

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+  const profileUpdates = events.flatMap((event) => {
+    const profile: Record<string, string | number | undefined> = {
+      userId: event.user_id
+    }
+    if (event.event_type === 'track' && event.segment_computation_class !== 'audience') {
+      // Filter out track events that are not audience enter/exit events until we have support for event tracking on our side
+      return []
```
joe-ayoub-segment, line 57: IIRC you want designed this Action to cover 2 scenarios: 
1. send Audience data to Stackadapt. (i.e Audience name and membership info as traits).
2. send user profile data to Stackadapt. 

If the above is correct, then here are the 3 main scenarios you should cover: 
1. regular track() events which contain user profile details in the $.context.traits location. 
2. regular identify() events which contain user profile details in the $.traits location.  
3. audience track() or audience identify() calls which can contain audience trait details in either the $.properties or $.traits locations respectively. 

You can also support alias() if you like, but those do not have traits in them.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// From https://www.30secondsofcode.org/js/s/string-case-conversion/
+function toCamelCase(str: string) {
+  const s = str
+    .match(/[A-Z]{2,}(?=[A-Z][a-z]+[0-9]*|\b)|[A-Z]?[a-z]+[0-9]*|[A-Z]|[0-9]+/g)
+    ?.map((x) => x.slice(0, 1).toUpperCase() + x.slice(1).toLowerCase())
+    .join('')
+  return s ? s.slice(0, 1).toLowerCase() + s.slice(1) : ''
+}
```
joe-ayoub-segment, line 170: You can use this if you like: 

import camelCase from 'lodash/camelCase'

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+
+function sha256hash(data: string) {
+  const hash = createHash('sha256')
+  hash.update(data)
+  return hash.digest('hex')
+}
+
+function getType(value: unknown) {
+  if (value instanceof Date) return 'date'
```
joe-ayoub-segment, line 159: Will a Date object ever be passed into this function call?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+  const fieldTypes: Record<string, string> = { userId: 'string' }
+  const advertiserId = events[0].advertiser_id
+  const profileUpdates = events.flatMap((event) => {
+    const profile: Record<string, string | number | undefined> = {
+      userId: event.user_id
+    }
+    if (event.event_type === 'track' && event.segment_computation_class !== 'audience') {
+      // Filter out track events that are not audience enter/exit events until we have support for event tracking on our side
+      return []
+    } else if (event.event_type === 'alias') {
```
joe-ayoub-segment, line 58: I'm finding some of the logic a little tricky to follow. Personally I find using destructuring a nice way to do this sort of thing: 

    const {
      advertiser_id, 
      segment_computation_key, 
      segment_computation_class,
      segment_computation_id, 
      segment_computation_id, 
      event_type, 
      previous_id, 
      user_id, 
      traits_or_props
    } = event
    
    const isAudience = segment_computation_class === 'audience'
    
    const { 
        audience_key,  
        [segment_computation_key]: action,       // true or false 
         ...rest                            // all the other user profile details you want to collect 
    } =  traits_or_props

joe-ayoub-segment:
Thanks for raising this PR @AlliterativeAlice . I think it's mostly there, though I think there's a few things which could be done to make it clearer to follow. 

We could also add a 'preset' which will initialize for customers who connect the Action to a non Engage Source. I can add this code myself after you've completed the code. 

Happy to chat about any of the comments or if you have any questions.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,170 @@
...
+        // Field type should be the most specific type of the values we've seen so far, use string if there is a conflict of types
+        if (event?.traits_or_props?.[key] || event?.traits_or_props?.[key] === 0) {
+          const type = getType(event.traits_or_props[key])
+          if (fieldTypes[camelCaseKey] && fieldTypes[camelCaseKey] !== type) {
+            fieldTypes[camelCaseKey] = 'string'
+          } else {
+            fieldTypes[camelCaseKey] = type
+          }
+        }
+      }
```
joe-ayoub-segment, line 89: can we extract this to a separate function?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
AlliterativeAlice: Is it possible to have items in the `traits` or `properties` object that are dates? If so, yes.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
AlliterativeAlice: We always wanted `advertiser_id` to be a global setting but found that we couldn't use a dynamic list in a global setting (this is why I asked if it was possible to have dynamic global settings in our Slack channel), so I think we will need to go with option 1.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/index.ts
AlliterativeAlice: We don't impose a max limit technically but I am checking internally to see what is the highest number of advertisers any client of ours actually has.

AlliterativeAlice: Potentially the number of advertisers could be in the hundreds - do you have a limit on your side for the number of options in a dynamic field?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
joe-ayoub-segment: It's not possible to have Date objects in traits or properties. However there can be date strings in there in ISO 8601 format. i.e "2024-08-28T11:47:26.883Z" or "2024-08-28". 

If you need to detect date strings you'll have to do it manually in the code.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/index.ts
joe-ayoub-segment: There's no limit on our side. Hundreds should be fine. 
If you wanted you could sort the items alphabetically to make it easier for the customer to find the item they are looking for?

joe-ayoub-segment:
Hi @AlliterativeAlice - please let me know when you'd like me to re-review.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/index.ts
AlliterativeAlice: For now we will load just the first page (first 100) which should be sufficient for our private beta client but we'll need to revisit before going public. I'll see if we can increase the max page size for this API on our end to load all the advertisers in a single request. Otherwise we will iterate through all pages upfront.

AlliterativeAlice:
> Hi @AlliterativeAlice - please let me know when you'd like me to re-review.

I've updated the PR based on your comments so please re-review.

packages/destination-actions/src/destinations/stackadapt-audiences/index.ts
```diff
@@ -0,0 +1,65 @@
...
+                name
+              }
+              scopes
+            }
+          }
+        }
+      }`
+
+      const res = await request(domain, {
+        body: JSON.stringify({ query: scopeQuery })
```
joe-ayoub-segment, line 37: Oops I missed this the first time around. 

can you add this line to the request please?

```
        throwHttpErrors: false
```

joe-ayoub-segment: This allows you to handle non 2XX responses yourself.

packages/destination-actions/src/destinations/stackadapt-audiences/types.ts
```diff
@@ -0,0 +1,15 @@
+export interface AdvertiserScopesResponse {
```
joe-ayoub-segment, line 1: I can't see where this is referenced in code. Is it used?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardProfile/functions.ts
```diff
@@ -0,0 +1,175 @@
...
+      segment_computation_class,
+      segment_computation_id,
+      event_type,
+      previous_id,
+      user_id,
+      traits_or_props
+    } = event
+
+    const isAudienceEvent = segment_computation_class === 'audience'
+    const audienceKey = (traits_or_props.audience_key as string) ?? segment_computation_key
```
joe-ayoub-segment, line 64:
```suggestion
    const audienceKey = segment_computation_key ?? undefined 
```

joe-ayoub-segment: The audience_key is in a couple of locations in the payload. The segment_computation_key field is suffient to collect its value.

joe-ayoub-segment:
Thanks for making these changes @AlliterativeAlice . 

I ran the code locally and still have some questions. Would you be available to jump on a Worksing Session so we can get this ready for deploy?

There's some things we could do to make it clearer to the customer how data gets mapped from the Segment event to Stackadapt, but it's a little hard to explain over a Github comment. 

There's also some standard Segment user traits which we could set up default mappings for. 

I suggest scheduling a 1h workshop here: https://calendly.com/joe_ayoub/integration-workshop?back=1 

I'm pretty sure we'll be good to deploy the code after the workshop. 

Kind regards,
Joe

packages/destination-actions/src/destinations/stackadapt-audiences/types.ts
AlliterativeAlice: Yes, it's used in `testAuthentication`

AlliterativeAlice:
Thanks for the call yesterday. I've updated the destination code to use different actions for profile and audience events. Please take another look and let me know if more changes are needed.

joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @AlliterativeAlice PR deployed.

PR #2393, opened 2024-09-11
[Hubspot] - unit tests and caching for new Actions

joe-ayoub-segment:
unit tests and caching for new Actions
also adds code in core to pass subscriptionMetadata 

## Testing

Ran Hadron tests to check that caching didn't break any core functionality.


packages/core/src/create-test-integration.ts
```diff
@@ -46,6 +53,7 @@ interface InputData<Settings> {
   dataFeedCache?: DataFeedCache
   transactionContext?: TransactionContext
   stateContext?: StateContext
+  subscriptionMetadata?: SubscriptionMetadata
```
varadarajan-tw, line 56: Could we also update README.md?
https://github.com/segmentio/action-destinations?tab=readme-ov-file#the-perform-function

and call out this is for internal use only

varadarajan-tw (commented):
LGTM! Could we update README.md before we merge this?

varadarajan-tw (approved):

PR #2414, opened 2024-09-17
[WIP] Update mailingLists type for Loops

tcgilbert:
This PR modifies the mailingLists field from an array of objects to a single object. Since users can't manually provide objects for the array, the previous structure was not very useful. It's unlikely that event payloads would include an array of mailingLists objects by default.

With this update, users can now easily input multiple mailling list IDs when creating a mapping, which will look like the following:
Screenshot 2024-09-17 at 3 06 45 PM

This is more in line with the request parameters outlined in [Loops API docs](https://loops.so/docs/api-reference/update-contact#request).

### Considerations

I did run some queries to see if any users are actively making use of mailingLists with the action currently and there is one customer that is. That being the case we need to either: 

- Create a new v2 action
- Add an additional field (one that uses array of objects, and one that is just an object)
- Notify the customer that is making use of this field of the change they will need to put in place
    - while this option is not desirable, the use may be happy to accomodate the change as they are using a hacky work around to get the mailing lists ids that they want in the mapping
 
Screenshot 2024-09-17 at 4 46 22 PM
Screenshot 2024-09-17 at 4 46 57 PM

## Testing


danrowden:
Thanks for this Thomas. We'll try it out.

How can we also test the UI you show in your screenshot? Action Tester has a different UI

tcgilbert:
> How can we also test the UI you show in your screenshot? Action Tester has a different UI

Afraid the actions tester won't mock the production UI accurately. That said the "Custom Attributes" field is the same type, so it will look the same in the UI: 

Screenshot 2024-09-18 at 7 18 18 AM

Only difference is that the `defaultUI` option for mailingLists will default to showing the "Edit Object" pane of the mapping

danrowden:
This is what I see in Actions Tester. There is no way for me to add custom values to the right side of `=` here. Am I missing something in the set up to get it to work?

Screenshot 2024-09-18 at 14 52 39

tcgilbert:
@danrowden here is a video of me inputting the key values: https://www.loom.com/share/928e1ca07771436989d466de9f5362ee?sid=b5e39c08-061e-47c6-a425-73df8a893093

if it helps happy to jump on a quick call to talk it through, lmk.

danrowden:
Thanks for this! That helps

tcgilbert:
> @tcgilbert And just to confirm, users can send in mailing list data in an `identify()` call eg https://loops.so/docs/integrations/segment#sending-data-to-segment

@danrowden yes, they could send with an identify call. If so it would likely be included in the traits object: 

```json
"traits": {
    "mailingLists": {
        "list_id_1": true
    }
}
```

The user would have the flexibility to map this object to Mailing lists field as they see fit with the drafted option

> For our existing customer on the current version, would this require adding a new field like `mailingListData` (named different), ask the customer to switch over and then remove the existing `mailingLists` in a future update?

Yes, we'd need to leave the current field as is for now, and not remove it until we could confirm it is not in use. The other option would be to create a cloned action with the name "v2", that has the updated mailingList field as an object.

danrowden:
@tcgilbert We have checked our logs and haven't found any recent activity for calls from Segment containing `mailingLists` data (anything with that key have an empty array). We are happy to make this change ASAP so it can roll out. We'd really like to avoid creating a new V2 action

tcgilbert:
> @tcgilbert We have checked our logs and haven't found any recent activity for calls from Segment containing `mailingLists` data (anything with that key have an empty array). We are happy to make this change ASAP so it can roll out. We'd really like to avoid creating a new V2 action

Hey @danrowden - @joe-ayoub-segment and I discussed, and we think it will fine to simply overwrite the existing field. If you have tested my recommendation and it works for you I can re-open this PR. 

Otherwise, feel free to take this as a reference and submit your own PR that updates the `maillingLists` field

danrowden:
@tcgilbert I tested sending data in traits, both mapping them in the UI and without mapping and found some things to fix, which I've done in this PR: https://github.com/segmentio/action-destinations/pull/2464

tcgilbert:
thanks @danrowden. I see what you mean about the `mailingLists` in traits now. I've added my review [here](https://github.com/segmentio/action-destinations/pull/2464) and will close this one out.

PR #2464, opened 2024-09-27
Allow mailingLists in traits

danrowden:
This PR modifies the mailingLists field from an array of objects to a single object. Since users can't manually provide objects for the array, the previous structure was not very useful. It's unlikely that event payloads would include an array of mailingLists objects by default.

With this update, users can now easily input multiple mailling list IDs when creating a mapping

`mailingLists` added to traits can also be sent to the APIm with our without manual mapping in the UI.


packages/destination-actions/src/destinations/loops/createOrUpdateContact/index.ts
```diff
@@ -146,10 +122,27 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    if (typeof obj[key] !== 'boolean') {
+      if (obj[key] === 'true') {
+        result[key] = true
+      } else if (obj[key] === 'false') {
+        result[key] = false
+      }
+    } else {
+      result[key] = obj[key]
+    }
+  }
```
tcgilbert, line 146:
```suggestion
for (const key in obj) {
    const value = obj[key];
    
    if (typeof value === 'string') {
      if (value === 'true') {
        result[key] = true;
      } else if (value === 'false') {
        result[key] = false;
      }
    } else if (typeof value === 'boolean') {
      result[key] = value;
    }
  }
```

Getting a type issue here with the build, perhaps try the above^: 
Screenshot 2024-09-27 at 1 20 33 PM

tcgilbert (changes requested):

joe-ayoub-segment:
hi @tcgilbert will you be reviewing this?

tcgilbert (approved):

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @danrowden and @tcgilbert - PR deployed. Please confirm you are happy with the update.

danrowden:
@joe-ayoub-segment Yes, happy thanks! I've tested it and it works

PR #2474, opened 2024-10-01
[PE-102] Add Snap Audience destination

tcgilbert:
This pull request adds a new Snap Audience destination. 

- [planning doc](https://docs.google.com/document/d/1Uq6AroCG4CaFNOw9sq0On3PYluLD9h0dLPhNK8YOTF8/edit)

## Testing

- ✅  Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- ✅ Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)


joe-ayoub-segment:
Thanks for the PR @tcgilbert ! 

I did a quick review but would like to catch up in person to walk through the details. Most of my comments are 'nits': 

1. customAudienceName or audienceName fields - how will this work in the UI? WIll the customer only see the customAudienceName field? If so, what value populates the audienceName field in the createAudience function?

2. Nit: Maybe we could define types for the request and responses for the createAudience and getAudience functions? It's not something I've done in the past, but have started doing it recently and it helps with code clarify. 

3. Should we go with oauth-managed or oauth2? oauth-managed is where the oauth config is directly added into the Developer Portal. oauth2 request that we manually add these values into chamber. 

4. Nit: We could simplify the code by throwing errors directly from the validateAndExtractIdentifier function. 

5. Nit: normalizeAndHashEmail and normalizeAndHashMobileId do the same thing. Could be merged together. 

6. Nit: A nice pattern to use is to have a single function which is able to handle both input from perform() and performBatch. Essentially the code will handle an array of payloads regardless of whether or not it was called from perform() or performBatch(). This means there's only one code path to maintain and test. You could still check for multiple payloads and throw a batchErrorMessage if you like.  

7. Nit: We could defined a type for request and responses for adding/removing users from an Audience. 

8. I'm not sure if I understand how collecting the advertising_id will work? The default mapping is to context.device.advertisingId but then we're also asking the customer to configure ID enrichment for ios.idfa and android.idfa ?

tcgilbert:
> * customAudienceName or audienceName fields - how will this work in the UI? WIll the customer only see the customAudienceName field? If so, what value populates the audienceName field in the createAudience function?

This just gives the option for the user to use a name that is different from the Segment Audience name if they want. Tbh not sure if that is a valid need, so okay with removing. 

> * Nit: Maybe we could define types for the request and responses for the createAudience and getAudience functions? It's not something I've done in the past, but have started doing it recently and it helps with code clarify.

Done.

> * Should we go with oauth-managed or oauth2? oauth-managed is where the oauth config is directly added into the Developer Portal. oauth2 request that we manually add these values into chamber.

Was thinking oauth2 since we created it

> * Nit: We could simplify the code by throwing errors directly from the validateAndExtractIdentifier function.

Have a couple questions on this as this function is used in batching too. Can talk it through

> * Nit: normalizeAndHashEmail and normalizeAndHashMobileId do the same thing. Could be merged together.

Done

> * Nit: A nice pattern to use is to have a single function which is able to handle both input from perform() and performBatch. Essentially the code will handle an array of payloads regardless of whether or not it was called from perform() or performBatch(). This means there's only one code path to maintain and test. You could still check for multiple payloads and throw a batchErrorMessage if you like.

Yeah this is good feedback. I've updated to use this format

> * Nit: We could defined a type for request and responses for adding/removing users from an Audience.

done

> * I'm not sure if I understand how collecting the advertising_id will work? The default mapping is to context.device.advertisingId but then we're also asking the customer to configure ID enrichment for ios.idfa and android.idfa ?


This has to do with the `.` notation used for properties, can talk through what would be ideal here

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @tcgilbert , I'll deploy this PR today. 
However it's not going to work until ADS is updated as well. The Destination needs to be registered in ADS as an Audience Destination. FYI @marinhero

joe-ayoub-segment:
hi @tcgilbert PR deployed. Will you follow up with next steps please?

tcgilbert:
> hi @tcgilbert PR deployed. Will you follow up with next steps please?

thanks will continue testing e2e

PR #2477, opened 2024-10-02
Add API versioning to Customer Match User List

maryamsharif:
Add the function to change between CANARY version when we need to upgrade.

## Testing


packages/core/src/destination-kit/index.ts
```diff
@@ -114,6 +114,8 @@ export type CreateAudienceInput<Settings = unknown, AudienceSettings = unknown>
   audienceName: string
```
marinhero, line 114: You probably don't need this here anymore since you have https://github.com/segmentio/action-destinations/pull/2499/files

I think it'll be automatically removed when the other PR is merged but just double check.

marinhero (approved):

packages/core/src/destination-kit/index.ts
maryamsharif: Yeah I added it to get the build to pass but I am going to remove it now.

joe-ayoub-segment:
hi @maryamsharif PR deployed

PR #2480, opened 2024-10-04
AAD-427 Fix group call: pass through userId and anonymousId

accoilmj:
We noticed problems with our group call becase userId/anonymousId were not being passed through. This PR adds those to the group payload. I've also added some tests to make sure it shows up.

In our live testing we also notice the old action "Post to Accoil" is still in the mapping configuration even though it was previously removed as this action isn't public yet. Is there a way to clean that up or should it be reinstated?

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/accoil-analytics/__tests__/__snapshots__/snapshot.test.ts.snap
+++ b/packages/destination-actions/src/destinations/accoil-analytics/__tests__/__snapshots__/snapshot.test.ts.snap
@@ -2,6 +2,7 @@
 
 exports[`Testing snapshot for actions-accoil-analytics destination: group action - all fields 1`] = `
 Object {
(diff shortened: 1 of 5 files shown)


joe-ayoub-segment (approved):

accoilmj:
Thanks for the review @joe-ayoub-segment. Do I need to take any further steps to have this merged?

joe-ayoub-segment:
hi @accoilmj - you don't need to do anything further. The change will be deployed later today. 
Best regards,
Joe

accoilmj:
Much appreciated @joe-ayoub-segment

joe-ayoub-segment:
hi @accoilmj PR deployed. Please confirm that the change works as expected

accoilmj:
Thanks @joe-ayoub-segment we're seeing events come through as expected now.

PR #2485, opened 2024-10-07
Salesforce custom object external

seg-leonelsanches:
This modification:

- Updates Salesforce base API from version v53 to v62;
- Adds a new mapping to upsert a custom object by its Id.

## Testing


packages/destination-actions/src/destinations/salesforce/customObjectExternalId/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+      return salesforceInstance.customObjectName()
+    }
+  },
+  perform: async (request, { settings, payload }) => {
+    const sf: Salesforce = new Salesforce(settings.instanceUrl, await generateSalesforceRequest(settings, request))
+
+    return sf.upsertCustomObject(payload, payload.customObjectName)
+  }
+}
```
rvadera12, line 50: Is there a reason why we are not implementing a performBatch method here?

packages/destination-actions/src/destinations/salesforce/sf-operations.ts
```diff
@@ -113,7 +114,7 @@ interface CreateJobResponseData {
   id: string
 }
 
-interface SObjectsResponseData {
+export interface SObjectsResponseData {
```
rvadera12, line 117: Why are we changing this to exports, and can we check backwards compatibility? If this interface is already in use, changing it to an export might affect existing consumers of the code. Ensure there are no breaking changes.

rvadera12 (changes requested):
Left a couple of comments. Can we also clean up the PR to only include the changes you made and not other syntax changes from other destinations.

packages/destination-actions/src/destinations/salesforce/customObjectExternalId/index.ts
seg-leonelsanches: @rvadera12 The batch doesn't make sense for this endpoint. Each record requires one separate API call. The API endpoint format is `/services/data/v61.0/sobjects/${entityType}/${entityKey}/${entityKeyValue}`. There's no way we can aggregate N records in 1 API call.

joe-ayoub-segment:
hi @seg-leonelsanches just doing some tidy up - can we close this PR?

PR #2008, opened 2024-04-24
Add Magellan AI (Action) Destination

leboshi:
Initial PR adding an action destination for [Magellan AI](https://www.magellan.ai)'s measurement and attribution platform ([more information](https://help.magellan.ai/)).

## Testing

- [X] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [X] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

Tested locally, and we *think* we properly constructed all the complex e-commerce events. 🤞

## Notes

PR for documentation incoming.  Documentation points out that semantically, the `Signed Up` event maps most closely to our `Lead` action, but our concept of leads would fit better with your e-commerce events.

Thank you, and let us know what next steps you need from us!

PR #2495, opened 2024-10-09
[Magellan AI] Add client metadata mappings

leboshi:
Adds mappings for IP address and user agent to all actions. The fields are required, but default mappings are provided.

(It makes more sense now why [you asked](https://github.com/segmentio/action-destinations/pull/2008#pullrequestreview-2030726344) whether we need to send user identifiers in our action calls... We assumed the end user's IP address and user agent would be passed through instead of the requests coming from your servers, and somehow we never noticed that in our live testing until now. Sorry, and thanks!)

## Testing

Again, the two new fields are required, but default mappings are provided. The only case where that might be an issue is the user agent field on the iOS Segment library, relevant for the `install` and `thirdPartyEvent` actions, which is already documented and should already have been accounted for by any users already using that action.

### DIFF
--- a/packages/destination-actions/src/destinations/magellan-ai/__tests__/__snapshots__/snapshot.test.ts.snap
+++ b/packages/destination-actions/src/destinations/magellan-ai/__tests__/__snapshots__/snapshot.test.ts.snap
@@ -3,12 +3,14 @@
 exports[`Testing snapshot for actions-magellan-ai destination: addToCart action - all fields 1`] = `
 Object {
   "currency": "NIO",
+  "ip": "36.184.84.10",
   "productId": "1ycI^2r",
   "productName": "1ycI^2r",
   "productType": "1ycI^2r",
   "productVendor": "1ycI^2r",
   "quantity": -64992310240215.04,
   "token": "1ycI^2r",
+  "ua": "1ycI^2r",
   "value": -64992310240215.04,
   "variantId": "1ycI^2r",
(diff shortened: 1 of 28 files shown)


tcgilbert:
hey @leboshi - the update looks good to me. Though the validation check is failing: 

Screenshot 2024-10-11 at 9 54 40 AM

Could you update the types and push the changes? 

```
# cmd to run
./bin/run generate:types --path ./packages/destination-actions/src/destinations/magellan-ai/index.ts                                                                                                                             
```

leboshi:
Oops. Thanks. Got the snapshots, forgot the types. Just pushed!

tcgilbert (approved):
@leboshi Thanks for adding the types

For future reference, I believe that adding the new required fields would still be a breaking change even with the defaults. Deploying the updated destination code wouldn't update a user's saved mappings on our backend to include the defaults. 

That said, considering your integration is in private beta this should be fine. The next deployment is scheduled for Wednesday the 16th, we'll reach out to you then

joe-ayoub-segment (approved):
Approving based on @tcgilbert 's review.

joe-ayoub-segment:
hi @leboshi PR deployed. Please confirm that you are happy with the changes.

leboshi:
Thanks for getting that out! We tested on our end, and everything's behaving as expected now. We've notified our beta users and are waiting for confirmation from them, but I think we're all set!

PR #801, opened 2022-10-07
Transform special traits to Mixpanel reserved properties

zihejia:
This PR adds transformation of special traits including 'created', 'email', 'firstName', 'lastName', 'name', 'username', and 'phone' to Mixpanel reserved properties for compatibility with Mixpanel classic integration(https://segment.com/docs/connections/destinations/catalog/mixpanel/#group-using-device-mode). 

This PR also removed the misplaced `segment_source_name` property for the `$set` call. 

## Testing

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

PR #2500, opened 2024-10-11
[PE-103] Update group action to use created_at as a reserved key

tcgilbert:
Currently, our Mixpanel group action only treats the `name` trait field as a reserved property for Mixpanel.

Screenshot 2024-10-11 at 12 48 23 PM

- [Reserved Properties in Mixpanel](https://docs.mixpanel.com/docs/data-structure/property-reference/reserved-properties)

[A customer has raised a concern](https://segment.zendesk.com/agent/tickets/521549) that they would like the created field to be handled in the same way as name. As currently, when `createdAt`, `created`, or `created_at` are not sent as `$created`, these fields are added to group profiles as regular profile properties instead of being treated as reserved properties.

The ask from the customer is to update the group action to operate in the same way that the identify action does regarding reserved properties. Similar to this update: https://github.com/segmentio/action-destinations/pull/801

image

## Testing

- ✅  Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- ✅ Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)

### before update
Screenshot 2024-10-11 at 12 22 48 PM

### after update
Screenshot 2024-10-11 at 12 22 28 PM

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/groupIdentifyUser/__tests__/index.test.ts
@@ -13,7 +13,7 @@ describe('Mixpanel.groupIdentifyUser', () => {
     const event = createTestEvent({
       timestamp,
       groupId: 'test-group-id',
-      traits: { hello: 'world', company: 'Mixpanel', name: 'test' }
+      traits: { hello: 'world', company: 'Mixpanel', name: 'test', created_at: timestamp }
     })
 
     nock('https://api.mixpanel.com').post('/groups').reply(200, {})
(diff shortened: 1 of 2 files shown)


joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @tcgilbert thanks for doing this. 
I see no issues with it as it replicates a pre-existing pattern from the identify() Action, and you showed that it tested successfully.

joe-ayoub-segment:
hi @tcgilbert this PR has been deployed. Please confirm you are happy with the change!

PR #2511, opened 2024-10-17
[Iterable Lists] Adding logic to not create an audience if an existing audience already exists with the same name.

seg-leonelsanches:
After testing Iterable Lists destination with the customer, we found out that the audience lifecycle tries to create an audience more than once, which it generates errors on Iterable, preventing the audience to sync correctly. 

Here I add logic to first verify whether the corresponding list already exists. If it exists, the logic returns its id.

## Testing


joe-ayoub-segment (approved):

joe-ayoub-segment:
PR deployed

PR #2513, opened 2024-10-18
fix preset mappings

tcgilbert:
Currently, you can't create the destination because the preset mappings are missing values for required fields, this fixes that. 

```
"external_audience_id is missing a value, but is required."
Field: settings.external_audience_id

"audienceKey is missing a value, but is required."
Field: settings.audienceKey

"props is missing a value, but is required."
Field: settings.props

"enable_batching is missing a value, but is required."
Field: settings.enable_batching
```

## Testing

N/A


joe-ayoub-segment (approved):

joe-ayoub-segment:
Hi @tcgilbert this PR has been deployed. Please confirm you are happy with the change!

PR #2516, opened 2024-10-18
[Facebook Conversions] Adding Test Event Code to all mappings.

seg-leonelsanches:
Customers don't want to set the Test Event Code in Settings for all mapping. Suppose a customer with 10+ mappings: is it a good approach setting a Test Event Code when the intention is to test one mapping?

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/addToCart.test.ts
+++ b/packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/addToCart.test.ts
@@ -484,6 +484,65 @@ describe('FacebookConversionsApi', () => {
       )
     })
 
+    it('should send test_event_code if present in the mapping', async () => {
+      nock(`https://graph.facebook.com/v${API_VERSION}/${settings.pixelId}`).post(`/events`).reply(201, {})
+
+      const event = createTestEvent({
+        event: 'Product Added',
+        userId: 'abc123',
+        timestamp: '1631210000',
+        properties: {
+          action_source: 'email',
+          currency: 'USD',
+          value: 12.12,
+          email: 'nicholas.aguilar@segment.com',
+          traits: {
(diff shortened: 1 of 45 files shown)


joe-ayoub-segment:
hi @seg-leonelsanches thanks for the PR. 
The code looks good to me, but someone from Stratconn will need to review. 
Could you also add a unit test for this please? and also show proof of testing. 

hi @brennan - another SA enhancement raised by @seg-leonelsanches . Could someone from the Stratconn team please own reviewing this?

packages/destination-actions/src/destinations/facebook-conversions-api/fb-capi-properties.ts
```diff
@@ -212,3 +212,11 @@ export const event_id: InputField = {
     '@path': '$.messageId'
   }
 }
+
+export const test_event_code: InputField = {
```
marinhero, line 216: All right, let's break this down:

- not a required field so existing mappings won't break
- no default value so it's not expected

looking good so far!

packages/destination-actions/src/destinations/facebook-conversions-api/__tests__/addToCart.test.ts
```diff
@@ -484,6 +484,65 @@ describe('FacebookConversionsApi', () => {
       )
     })
 
+    it('should send test_event_code if present in the mapping', async () => {
```
marinhero, line 487: Nice! and we are not adding the original use case because it's already being tested.

marinhero (approved):
Well done! :shipit:

joe-ayoub-segment:
Will deploy next Tuesday. No deploys this week due to election.

joe-ayoub-segment:
Hi @seg-leonelsanches PR deployed

PR #2521, opened 2024-10-21
[STRATCONN-4282] Braze validation and multi-status fix

sayan-das-in:
Context
Earlier Braze required on the the 3 fields to be present to be considered an event to be valid, they were `braze_id`, `user_alias` and `external_id`. Later `email` was added, however this validation was missing from the MultiStatus flow.

PR Contents
- Adds email as one of the required fields
- Fixes couple of error messages to represent that email is one the fields used in validation
- Fixes a MultiStatus bug where `sent` was to the original batch payload instead of what was actually sent.

## Testing

Testes locally.

### DIFF
--- a/packages/destination-actions/src/destinations/braze/__tests__/multistatus.test.ts
+++ b/packages/destination-actions/src/destinations/braze/__tests__/multistatus.test.ts
@@ -46,6 +46,15 @@ describe('MultiStatus', () => {
             externalId: 'test-external-id'
           }
         }),
+        // Valid Event
+        createTestEvent({
+          type: 'identify',
+          receivedAt,
+          traits: {
+            name: 'Example User Two',
+            email: 'user@example.com'
+          }
+        }),
         // Event without any user identifier
         createTestEvent({
           type: 'identify',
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/braze/__tests__/multistatus.test.ts
```diff
@@ -276,21 +316,26 @@ describe('MultiStatus', () => {
...
 
-      // The third event fails as pre-request validation fails for not having a valid user identifier
-      expect(response[2]).toMatchObject({
-        status: 400,
-        errortype: 'PAYLOAD_VALIDATION_FAILED',
-        errormessage: 'One of "external_id" or "user_alias" or "braze_id" is required.',
-        errorreporter: 'DESTINATION'
-      })
+      // // The third event fails as it doesn't have any products
+      // expect(response[2]).toMatchObject({
```
brennan, line 325: Should this be commented out?

packages/destination-actions/src/destinations/braze/utils.ts
```diff
@@ -179,7 +184,7 @@ export async function sendBatchedTrackEvent(
   })
 
   await handleBrazeAPIResponse(
-    transformPayloadsType(payloads),
+    payloads as object as JSONLikeObject[],
```
brennan, line 187: This change seems stylistic. How risky is it? Should we leave it out to reduce the number of variables we're changing in this PR?

sayan-das-in: This is simply a type definition. The transformPayloadsType function was previously used only to change the type. However, since type information is removed during the compilation from TypeScript to JavaScript, this would result in an extra, non-functional function being left behind.

packages/destination-actions/src/destinations/braze/__tests__/multistatus.test.ts
sayan-das-in: Thank you for pointing this out. I had accidentally commented it out during testing but have now uncommented it.

brennan (approved):

PR #2522, opened 2024-10-21
[DO NOT PUSH TO PROD] - S3 batch id settings

mayur-pitale:

## Testing


joe-ayoub-segment:
hi @mayur-pitale and @nick-Ag - I'm not sure if this is going to do what you are hoping it will do. 

I believe that there can be multiple performBatch() calls per sync, which will result in different uuids being generated. 

Are you sure that this is worth testing in Stage?

PR #2536, opened 2024-10-27
[MAIN] [STRATCONN] 4261 Mixpanel Multistatus support

AnkitSegment:
Multistatus support is added for Mixpanel destinations
JIRA TICKET: https://segment.atlassian.net/browse/STRATCONN-4261

[Testing Docs](https://docs.google.com/document/d/1WrPuqWdyM1gyN1wTLt2SYNO_cpfpnn69uNc8IgvTJJ4/edit?usp=sharing)

## Testing

Screenshot 2024-10-28 at 2 07 38 PM

Include any additional information about the testing you have completed to


packages/destination-actions/src/destinations/mixpanel/common/utils.ts
```diff
@@ -99,3 +101,54 @@ export function cheapGuid(maxlen?: number) {
...
+        sent: data,
+        body: events[data.index]
+      })
+    })
+  }
+  if (apiResponse.data.code !== 200 && apiResponse.data.code !== 400) {
+    for (let i = 0; i < payloadCount; i++) {
+      multiStatusResponse.setErrorResponseAtIndex(i, {
+        status: apiResponse.data.code,
+        errormessage: apiResponse.data.error ?? 'Payload validation error',
```
varadarajan-tw, line 146: Should we tag this error as unknown error from mixpanel?
```suggestion
        errormessage: apiResponse.data.error ?? 'Unknown error from mixpanel',
```

packages/destination-actions/src/destinations/mixpanel/common/utils.ts
```diff
@@ -99,3 +101,54 @@ export function cheapGuid(maxlen?: number) {
...
+        sent: events[i]
+      })
+    }
+
+    apiResponse.data.failed_records?.map((data) => {
+      multiStatusResponse.setErrorResponseAtIndex(data.index, {
+        status: 400,
+        errormessage: data.message,
+        sent: data,
+        body: events[data.index]
```
varadarajan-tw, line 138: Shouldn't this be the other way round? sent should be the events and body should be data?

varadarajan-tw (approved):

joe-ayoub-segment:
hi @AnkitSegment and @varadarajan-tw . 

Ankit and I just reviewed PR with @spjtls9 from Mixpanel. 
2 main suggestions were: 
1. Can we keep the perform() and performBatch() functions using the exact same code path? 
    - i.e they call the same functions, with an array of payloads being passed in. 
2. Can we put the new code path and old code paths behind a feature flag, and deploy this code dark, then roll out in a controlled manner? 
    - This will allow for testing in Production, and for a quick and safe rollback should issues arise.

AnkitSegment:
> hi @AnkitSegment and @varadarajan-tw .
> 
> Ankit and I just reviewed PR with @spjtls9 from Mixpanel. 2 main suggestions were:
> 
> 1. Can we keep the perform() and performBatch() functions using the exact same code path?
>    
>    * i.e they call the same functions, with an array of payloads being passed in.
> 2. Can we put the new code path and old code paths behind a feature flag, and deploy this code dark, then roll out in a controlled manner?
>    
>    * This will allow for testing in Production, and for a quick and safe rollback should issues arise.

Hey @joe-ayoub-segment ,
1. I've updated the code to ensure that perform() and performBatch() now follow the same code path. Both functions pass an array of payloads to the same underlying logic, as requested.

2. We already have the `cloudevents-spec-v2-all` feature flag, which controls multistatus responses based on user tiers. This allows for a controlled rollout and testing in production.
If you think we need a separate or more granular flag separately for this destination, I’m happy to discuss and make the necessary adjustments.

joe-ayoub-segment (commented):
hi @AnkitSegment , 
Just had a look through the code. 
I'm not familiar with Hubspot's API or this Destination so I'm not able to properly review. But I do have a couple of questions I'm sure you can easily answer ;). 

**Question 1:** The the handleMixPanelApiResponse function, why is this condition present?

if (apiResponse.data.code === 400 || apiResponse.data.code === 200) {
 // do stuff
}
if (apiResponse.data.code !== 200 && apiResponse.data.code !== 400) {
 // do other stuff
}

Is it because If a 400 response is returned it's still possible that some events were successfully delivered? And conversely, if a 200 response is returned then there can also be some failures?

**Question 2**: for the trackPurchase function, why are there 2 arrays, events and sentEvents?

Why is this happening?

    sentEvents.push(purchaseEvents as object as JSONLikeObject)
    events.push(...purchaseEvents)

I assume sentEvents is just a duplicate array which keeps track of the events which were actually sent (so you have a reference for the index of each event when you process the response)?

joe-ayoub-segment:
Also did you make a decision regarding how to roll this out? Will it be behind a feature flag?

AnkitSegment:
Hello @joe-ayoub-segment ,
Answer 1. We have different conditions for statuses 400 and 200 because the response can either indicate full success or partial success with failed records. Any other response would mean the entire batch has failed.

Answer 2. Yes, sendEvents is to track the events that were actually sent.

varadarajan-tw (approved):
LGTM

packages/destination-actions/src/destinations/mixpanel/trackPurchase/index.ts
```diff
@@ -49,15 +50,38 @@ const getPurchaseEventsFromPayload = (payload: Payload, settings: Settings): Mix
...
+  const sentEvents: JSONLikeObject[] = []
+  payload.forEach((value) => {
+    const purchaseEvents = getPurchaseEventsFromPayload(value, settings).flat()
+    sentEvents.push(purchaseEvents as object as JSONLikeObject)
+    events.push(...purchaseEvents)
+    return purchaseEvents
   })
+  const response = await callMixpanelApi(request, settings, events, false)
+  const multiStatusResponse = handleMixPanelApiResponse(payload.length, response, sentEvents)
+  console.log('features', features && features['mixpanel-multistatus'])
```
joe-ayoub-segment, line 64: Should this console.log be removed?

packages/destination-actions/src/destinations/mixpanel/trackPurchase/index.ts
```diff
@@ -49,15 +50,38 @@ const getPurchaseEventsFromPayload = (payload: Payload, settings: Settings): Mix
...
+) => {
+  return await request<MixpanelTrackApiResponseType>(
+    `${getApiServerUrl(settings.apiRegion)}/import?strict=${settings.strictMode ?? `1`}`,
+    {
+      method: 'post',
+      json: events,
+      headers: {
+        authorization: `Basic ${Buffer.from(`${settings.apiSecret}:`).toString('base64')}`
+      },
+      throwHttpErrors: throwHttpErrors
```
joe-ayoub-segment, line 82: Is the addition of this line going to change anything when the Feature Flag is off for the user?

packages/destination-actions/src/destinations/mixpanel/trackPurchase/index.ts
```diff
@@ -49,15 +50,38 @@ const getPurchaseEventsFromPayload = (payload: Payload, settings: Settings): Mix
...
+  const events: MixpanelEvent[] = []
+  const sentEvents: JSONLikeObject[] = []
+  payload.forEach((value) => {
+    const purchaseEvents = getPurchaseEventsFromPayload(value, settings).flat()
+    sentEvents.push(purchaseEvents as object as JSONLikeObject)
+    events.push(...purchaseEvents)
+    return purchaseEvents
   })
+  const response = await callMixpanelApi(request, settings, events, false)
+  const multiStatusResponse = handleMixPanelApiResponse(payload.length, response, sentEvents)
```
joe-ayoub-segment, line 63: Shouldn't this line be behind a feature flag? Looks like it's going to be executed even when the flag is off for the user. Part of using a feature flag is to enable us to roll back to a code path which is no different to what was live prior to the deploy.

joe-ayoub-segment:
hi @AnkitSegment thanks for making the update to add the FF. 
It still looks to me like some new code paths are being introduced even for users who have the flag set to off. Is there a deliberate reason for this?

packages/destination-actions/src/destinations/mixpanel/trackPurchase/index.ts
varadarajan-tw: My initial understanding was that we wanted to hide only the response we return back. But I agree that putting the entire parsing logic within feature flag is better if we are unsure about the feature 👍

joe-ayoub-segment:
hi @AnkitSegment please let me know when you'd like me to rereview.

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @AnkitSegment PR deployed

PR #2581, opened 2024-11-11
[Hubspot] - STRATCONN-4331 - trim white space from string properties

joe-ayoub-segment:
This PR adds code to trim properties for for Hubspot Web Mode Actions which interact with Objects.

For background on the need for this PR please see https://segment.atlassian.net/browse/STRATCONN-4331

## Testing

Unit test added.


pooyaj (approved):

joe-ayoub-segment:
deployed

PR #2594, opened 2024-11-15
[Drip] - new Destination - Drip

joe-ayoub-segment:
New Drip Destination

## Testing

Unit tests included.


jacobpmeyer:
> hi @jacobpmayer - this is the new PR for the Drip Integration which I'll deploy on Tuesday.
> 
> Can you clarify that no other user identifiers such as userId or anonymousId should be sent to Drip? Does Drip only accept email addresses as the main user identifier?

Our users have an `account_id`, but it's true that we only identify their customers/subscribers by their email address.

itsarijitray (approved):

joe-ayoub-segment:
deployed

PR #2606, opened 2024-11-20
Migrate rokt-audiences to new authentication method

GoodmanBen:
## Testing
- [x] Updated [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing). Successfully checked authentication and test event submission:
image
image
- [x] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change.  --> This integration is in private beta and we have established a migration plan with the one active private beta user.

### DIFF
--- a/packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/index.ts
+++ b/packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/index.ts
@@ -62,12 +62,12 @@ const action: ActionDefinition<Settings, Payload> = {
     }
   },
 
-  perform: (request, { payload }) => {
-    return processPayload(request, [payload])
+  perform: (request, { settings, payload }) => {
+    return processPayload(request, settings, [payload])
   },
 
-  performBatch: (request, { payload }) => {
-    return processPayload(request, payload)
+  performBatch: (request, { settings, payload }) => {
+    return processPayload(request, settings, payload)
   }
 }
 
--- a/packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/index.ts
+++ b/packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/index.ts
@@ -37,7 +37,7 @@ const action: ActionDefinition<Settings, Payload> = {
         '@if': {
           exists: { '@path': '$.context.traits.email' },
           then: { '@path': '$.context.traits.email' },
-          else: { '@path': '$.email' }
+          else: { '@path': '$.traits.email' }
         }
       }
     },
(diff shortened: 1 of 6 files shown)


joe-ayoub-segment:
hi @GoodmanBen thanks for raising this draft PR. 
I won't review it until you ask me to - however I had a quick scan and noticed that all of the auth settings are changing. 
This will be a breaking change, and will prevent any data collection for customers who are already using the Integration. 
Customers will need to reconfigure auth settings before data collection to Rokt can resume. 
Just want to make sure you are aware of this.

GoodmanBen:
> hi @GoodmanBen thanks for raising this draft PR. I won't review it until you ask me to - however I had a quick scan and noticed that all of the auth settings are changing. This will be a breaking change, and will prevent any data collection for customers who are already using the Integration. Customers will need to reconfigure auth settings before data collection to Rokt can resume. Just want to make sure you are aware of this.

Hi @joe-ayoub-segment thank you, this has been tested and is ready for review. Yes, thank you for flagging. We understand that this is a breaking change. This extension is currently in private beta and we have coordinated with the one user on how to transition to the new authentication approach.

packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/index.ts
```diff
@@ -37,7 +37,7 @@ const action: ActionDefinition<Settings, Payload> = {
         '@if': {
           exists: { '@path': '$.context.traits.email' },
           then: { '@path': '$.context.traits.email' },
-          else: { '@path': '$.traits.email' }
+          else: { '@path': '$.email' }
```
joe-ayoub-segment, line 40:
```suggestion
          else: { '@path': '$.traits.email' }
```

joe-ayoub-segment: $.traits.email is the the location where the email address will be when Engage outputs an identify() call from the Audience. 
$.context.traits.email is the the location where the email address will be when Engage outputs an track() call from the Audience.

joe-ayoub-segment (dismissed):

packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/index.ts
GoodmanBen: Ah got it, accepting suggestion then merging!

joe-ayoub-segment (approved):

joe-ayoub-segment:
We're in a deploy freeze so next deploy scheduled for 3-Dec

PR #2608, opened 2024-11-20
[Hubspot] [STRATCONN-4366] - correcting custom object reference to use canonical identifier

joe-ayoub-segment:
Update to force Hubspot customEvent Action to reference Custom Objects via the canonical identifier rather than the friendlier name. The friendlier name is being deprecated by Hubspot in 2025. 

## Testing

Tested locally
image


sayan-das-in (approved):

joe-ayoub-segment:
deployed

PR #2611, opened 2024-11-21
[STRATCONN-4213] - Update hotfix release scripts

varadarajan-tw:
This PR applies some fixes to hotfix release scripts

**1. Tag commit with package versions on the hotfix branch itself.**

If we tag the hotfix merge commit on main branch with package version (i.e package tags), the next publish won't work as lerna relies on git tags to figure out the last published commit and decide if new package publish is needed.  
To solve this,

- the special package version tagging step after hotfix merge has been removed from publish.yml workflow
- the if condition to skip adding package tags while running `scripts/generate-release-tags.sh` has been removed so that package version tagging is done on the hotfix branch itself.

**2. Fixes Hotfix Github Release Notes**

Previously, notes of the hotfix release created in Github contained list of PRs merged to main. Instead, it should not list any PRs from main branch because hotfix doesn't include those PRs already on main. It only contains the fix applied on top of the previous release. So, we simply skip generating PR list for Hotfix releases. 

## Testing

Testing completed successfully 

For 1,  I am simply removing a job which was introduced just for hotfix but it is no longer required. So, no testing is required.
For 2,  Here is an [example release created](https://github.com/segmentio/action-destinations/releases/tag/hotfix-2024-11-21.1) while testing.

### DIFF
--- a/scripts/github-action/create-github-release.js
+++ b/scripts/github-action/create-github-release.js
@@ -1,14 +1,22 @@
 // This is a github action script and can be run only from github actions. To run this script locally, you need to mock the github object and context object.
 module.exports = async ({ github, context, core, exec }) => {
-  let { GITHUB_SHA, DRY_RUN } = process.env
+  let { GITHUB_SHA, DRY_RUN, GITHUB_REF } = process.env
+
+  // Get the branch name from the GITHUB_REF
+  const branchName = GITHUB_REF.replace('refs/heads/', '')
+  const isHotfixBranch = branchName.startsWith('hotfix/')
 
   if (Boolean(DRY_RUN)) {
     core.info('Running in dry-run mode')
   }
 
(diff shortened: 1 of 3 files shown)


scripts/github-action/create-github-release.js
```diff
@@ -1,14 +1,22 @@
 // This is a github action script and can be run only from github actions. To run this script locally, you need to mock the github object and context object.
 module.exports = async ({ github, context, core, exec }) => {
-  let { GITHUB_SHA, DRY_RUN } = process.env
+  let { GITHUB_SHA, DRY_RUN, GITHUB_REF } = process.env
```
joe-ayoub-segment, line 3: is this already configured ?

joe-ayoub-segment (approved):

PR #2617, opened 2024-11-29
[Taboola Audiences] - Supporting computation key to be used when creating Audiences

joe-ayoub-segment:
Adding a new setting field to allow for computation key to be used during Audience creation. 

Required by some customers who are migrating from a Custom Function, and need the Audience names to match so they don't have duplicate Audiences created in Taboola. 

## Testing

Unit tests added.


joe-ayoub-segment:
PR deployed

PR #2629, opened 2024-12-06
[LinkedIn Audiences] Necessary fixes for `updateAudience` mapping.

seg-leonelsanches:
- User Email is not always `context.traits.email`;
- Bypass `email` if it is already hashed.

## Motivation

- Some customers have emails as normal traits, not identifiers;
- Some customers have different spellings for their emails;
- Some customers use trait enrichment for emails, and they come as properties;
- Some customers don't send plain emails to Segment: they already hash emails using SHA256.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
+++ b/packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
@@ -88,7 +88,6 @@ async function processPayload(
   payloads: Payload[],
   statsContext: StatsContext | undefined
 ) {
-  // validate(settings, payloads)
   validate(settings)
 
   const linkedinApiClient: LinkedInAudiences = new LinkedInAudiences(request)
@@ -124,7 +123,6 @@ async function processPayload(
   return res
 }
 
-// function validate(settings: Settings, payloads: Payload[]): void {
 function validate(settings: Settings): void {
   if (!settings.send_google_advertising_id && !settings.send_email) {
     throw new IntegrationError(
--- a/packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
+++ b/packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
@@ -46,6 +46,16 @@ const action: ActionDefinition<Settings, Payload> = {
         '@path': '$.context.device.advertisingId'
       }
     },
+    source_segment_id: {
+      label: 'LinkedIn Source Segment ID',
+      description:
+        "A Segment-specific key associated with the LinkedIn DMP Segment. This is the lookup key Segment uses to fetch the DMP Segment from LinkedIn's API.",
+      type: 'string',
+      unsafe_hidden: true, // This field is hidden from customers because the desired value always appears at '$.properties.audience_key' in Personas events.
+      default: {
+        '@path': '$.properties.audience_key'
+      }
+    },
     personas_audience_key: {
       label: 'Segment Engage Audience Key',
       description:
@@ -88,7 +98,7 @@ async function processPayload(
   payloads: Payload[],
   statsContext: StatsContext | undefined
 ) {
-  validate(settings)
+  validate(settings, payloads)
 
   const linkedinApiClient: LinkedInAudiences = new LinkedInAudiences(request)
 
@@ -123,7 +133,16 @@ async function processPayload(
   return res
 }
 
-function validate(settings: Settings): void {
+function validate(settings: Settings, payloads: Payload[]): void {
+  const isAutoOrUndefined = ['AUTO', undefined].includes(payloads[0]?.dmp_user_action)
+  if (isAutoOrUndefined && payloads[0].source_segment_id !== payloads[0].personas_audience_key) {
+    throw new IntegrationError(
+      'The value of `source_segment_id` and `personas_audience_key` must match.',
+      'INVALID_SETTINGS',
+      400
+    )
+  }
+
   if (!settings.send_google_advertising_id && !settings.send_email) {
     throw new IntegrationError(
       'At least one of `Send Email` or `Send Google Advertising ID` must be set to `true`.',
--- a/packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
+++ b/packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
@@ -29,9 +29,12 @@ const action: ActionDefinition<Settings, Payload> = {
       label: 'User Email',
       description: "The user's email address to send to LinkedIn.",
       type: 'string',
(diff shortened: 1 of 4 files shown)


brennan:
@seg-leonelsanches for now, the only changes I would be comfortable making are:

- Exposing the `email` field to customers (but not changing the default)
- Checking whether email addresses are hashed

Happy to review and approve a PR with only these two changes.

`I don't know any customer that wants to add a mapping per audience in Engage.` - For now, this is required to ensure that bulk batching works correctly for this destination.

seg-leonelsanches:
> @seg-leonelsanches for now, the only changes I would be comfortable making are:
> 
> * Exposing the `email` field to customers (but not changing the default)
> * Checking whether email addresses are hashed
> 
> Happy to review and approve a PR with only these two changes.
> 
> `I don't know any customer that wants to add a mapping per audience in Engage.` - For now, this is required to ensure that bulk batching works correctly for this destination.

Thank you, @brennan. As requested, I reverted the changes regarding `source_segment_id`.

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
```diff
@@ -29,9 +29,12 @@ const action: ActionDefinition<Settings, Payload> = {
       label: 'User Email',
       description: "The user's email address to send to LinkedIn.",
       type: 'string',
-      unsafe_hidden: true, // This field is hidden from customers because the desired value always appears at path '$.context.traits.email' in Personas events.
       default: {
-        '@path': '$.context.traits.email'
+        '@if': {
+          exists: { '@path': '$.traits.email' },
+          then: { '@path': '$.traits.email' },
+          else: { '@path': '$.context.traits.email' }
```
seg-leonelsanches, line 36: @brennan This does the exact same thing as before, but checking an extra field.

seg-leonelsanches: UI Preview: 
image

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
```diff
@@ -212,9 +224,15 @@ function getUserIds(settings: Settings, payload: Payload): Record<string, string
   const users = []
 
   if (payload.email && settings.send_email === true) {
+    let email = payload.email
+    const isHashed = new RegExp(/[0-9abcdef]{64}/gi).test(payload.email)
+    if (!isHashed) {
```
brennan, line 229: Can we add a unit test that exercises this code path, please?

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
```diff
@@ -212,9 +224,15 @@ function getUserIds(settings: Settings, payload: Payload): Record<string, string
   const users = []
 
   if (payload.email && settings.send_email === true) {
+    let email = payload.email
+    const isHashed = new RegExp(/[0-9abcdef]{64}/gi).test(payload.email)
```
brennan, line 228: Can we add a unit test with a pre-hashed email?

seg-leonelsanches: This is done.

brennan:
@joe-ayoub-segment When we deploy, this will require a push. The only metadata value to be updated is the default mapping for the `email` field: https://github.com/segmentio/action-destinations/pull/2629/files#diff-79d0d51e489f761c254e9fbbe921303e9a89b54400da18a98f42cdc6e8487c3fR32-R36

cc @rvadera12

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
marinhero: Hi @seg-leonelsanches , quick question, why aren't we favoring the existing path value to be the one that gets check first instead of making it the fallback? That would feel like a more incremental change to me. What would you say?

```
{
    '@if': {
        exists: { '@path': '$.context.traits.email' },
        then: { '@path': '$.context.traits.email' },
        else: { '@path': '$.traits.email' }
    }
}
```

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
marinhero, line 228: instead of implementing this, please use the [Hashing Utils](https://github.com/segmentio/action-destinations/blob/6955920f1a4c6194e86ce6486a510f82a64b7d4e/packages/core/src/hashing-utils.ts#L24) that are used in other destinations. Thanks!

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
seg-leonelsanches: @marinhero My first version of this PR had the hardcoded audience key removed, therefore the precedence to work with traits. 

I think the precedence doesn't matter that much here. I prefer to not have this modified right now, since it was very difficult to get it approved on time before the code freeze.

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
```diff
@@ -168,15 +174,15 @@ async function createDmpSegment(
...
+): Promise<string> {
   statsContext?.statsClient?.incr('oauth_app_api_call', 1, [...statsContext?.tags, `endpoint:create-dmpSegment`])
   const res = await linkedinApiClient.createDmpSegment(settings, payload)
   const headers = res.headers.toJSON()
   return headers['x-linkedin-id']
 }
 
 function extractUsers(settings: Settings, payloads: Payload[]) {
-  const elements: Record<string, any>[] = []
+  const elements: Record<string, unknown>[] = []
```
marinhero, line 185: Good call 🙇🏽

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/index.ts
```diff
@@ -197,12 +203,18 @@ function getAction(payload: Payload) {
 
   if (dmp_user_action === 'ADD') {
     return 'ADD'
-  } else if (dmp_user_action === 'REMOVE') {
+  }
+
+  if (dmp_user_action === 'REMOVE') {
```
marinhero, line 208: NIT: I would actually discourage unneeded changes on this destination due to its high usage. Up to you wether to keep them or not.

marinhero (changes requested):
There are some welcome changes here. I'll request an update due to the redundant hashing function. Let's use the Hashing Utils instead.

marinhero (approved):

brennan (approved):

joe-ayoub-segment:
hi @seg-leonelsanches PR deployed. cc @brennan

PR #2640, opened 2024-12-12
TikTok - STRATCONN-5346 - changing default object UI view for content field

joe-ayoub-segment:
Changing the default field UI view so that it displays the key:value pairs for the TikTok Content field. 

This doesn't fix any bug - it might help customers understand how to use the field better. 

## Testing

None needed.


itsarijitray (approved):

joe-ayoub-segment:
deployed

PR #2645, opened 2024-12-12
Update: Format Date type objects to permissible Iterable date string

TheFacto-HH:
Hello 👋 

We noticed when attempting to map the `receivedAt` property from Segment's UI to an Iterable destination field, Iterable was dropping the property (we used Iterable's Data-Schema-Management to define the property as a `Date` ahead of time). We discovered that this is very likely due to `receivedAt` being a `Date` which was not handled by the [`convertDatesInObject`](https://github.com/segmentio/action-destinations/blob/6955920f1a4c6194e86ce6486a510f82a64b7d4e/packages/destination-actions/src/destinations/iterable/utils.ts#L36) function.

We're looking for feedback on this update which would check against `Date` types and format them to be compatible with Iterable's permitted date string format.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/iterable/updateUser/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/iterable/updateUser/__tests__/index.test.ts
@@ -117,6 +117,46 @@ describe('Iterable.updateUser', () => {
         }
       })
     })
+
+    it('should format date objects to permissible format', async () => {
+      const d = new Date()
+      const event = createTestEvent({
+        type: 'identify',
+        userId: 'user1234',
+        receivedAt: d,
+        traits: {
+          phone: '+14158675309'
+        }
+      })
+
+      nock('https://api.iterable.com/api').post('/users/update').reply(200, {})
+
+      const responses = await testDestination.testAction('updateUser', {
+        event,
+        mapping: {
+          phoneNumber: {
+            '@path': '$.traits.phone'
+          },
+          userId: {
+            '@path': '$.userId'
+          },
+          dataFields: {
+            receivedAtDateString: {
+              '@path': '$.receivedAt'
+            }
+          }
+        }
+      })
+
+      expect(responses.length).toBe(1)
+      expect(responses[0].options.json).toMatchObject({
+        userId: 'user1234',
+        dataFields: {
+          phoneNumber: '+14158675309',
+          receivedAtDateString: `${d.toISOString().replace('T', ' ').split('.')[0]} +00:00`
+        }
+      })
+    })
   })
   describe('performBatch', () => {
     it('works with default mappings', async () => {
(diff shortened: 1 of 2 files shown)


joe-ayoub-segment:
hi @TheFacto-HH thanks for raising this draft PR. 
I had a quick look at the code (even though it's not ready for review) and can see that this line: 

```
} else if (value instanceof Date) {
      obj[prop] = dateToIterableDateStringFormat(value.toISOString())
    } 
```

isn't going to do anything, as value will never be a Date object. 

In order to help you with the issue you've been having, could you please tell me the following: 
1. Which field are you passing the value into?
2. What value are you passing into the field?
3. What expected outcome do you wish to happen?

Kind regards,
Joe

TheFacto-HH:
Hey @joe-ayoub-segment , thanks for the early review!

> isn't going to do anything, as value will never be a Date object.

We noticed that if we mapped `receivedAt` to a destination property in Iterable, Iterable was dropping it. This seems to be because Segment was propagating an actual `Date` object to the destination instead of formatting it to an Iterable-supported date string. I noticed that receivedAt - per Segment's types - [can, in fact, be a Date](https://github.com/segmentio/action-destinations/blob/d3e328117ce3f3efe4f8793fa7b2a2a8898d5f32/packages/core/src/segment-event.ts#L142).

If you remove the line that you mentioned and re-run the unit test I added:

Screenshot 2024-12-13 at 10 43 51 AM

joe-ayoub-segment:
Thanks for explaining @TheFacto-HH. 
I still can't see how a date object can be passed to a field in the Destination in Production though. 
The JSON payload gets sent through a pipeline built from multiple microservices before it gets passed into the perform() function. At this stage the Date object will have been converted into an ISO 8601 string. 

Could you show me reproduction steps for the issue you are having please, if possible with screenshots of the mapping fields and an example payload? I don't doubt your claim - I'm just trying to verify it. 

The change looks safe, but I'd like to ensure that it actually addresses a fault before I merged it. 

Thanks,
Joe

TheFacto-HH:
Sure thing!

**Segment Mapping**

Screenshot 2024-12-13 at 11 38 17 AM

(`receivedAt` is mapped to `member360DeletedAt`)

**Iterable Data Schema**

Screenshot 2024-12-13 at 11 33 40 AM

(`member360DeletedAt` is defined as a `date` and **User field lock is on**, meaning unrecognized fields will be dropped).

**Iterable User Fields**

Screenshot 2024-12-13 at 11 35 56 AM

(Notably, `member360DeletedAt` is not present after sending the test event [and the status was `200`]).

joe-ayoub-segment:
hi @TheFacto-HH - looks like we won't be deploying this PR until the new year.

craigotis (approved):

joe-ayoub-segment:
hi @TheFacto-HH thanks for following up. I've been out of office ill most of this week (am still out). This PR should go out Monday or Tuesday next week. Apologies for the delay.

joe-ayoub-segment:
hi @TheFacto-HH PR deployed.

PR #2647, opened 2024-12-13
[SFMC] Remove stats for SFMC

harsh-joshi99:
Remove the stats code for SFMC since we have completed the investigation regarding uniqueness of key and id of events in payload.

We are only seeing 1 key/id per payload hence all events in a payload have similar key/id according to the stats.
image (2)

## Testing
Testing is not required since we are removing code only added for checking stats.

- [ ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality - NR
- [ ] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) - NR
- [ ] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change. - NR
- [ ] [Segmenters] Tested in the staging environment - NR
- [ ] [Segmenters] [If applicable for this change] Tested for regression with Hadron. - NR


joe-ayoub-segment:
hi @harsh-joshi99 PR deployed

PR #2653, opened 2024-12-17
[Marketo] - Add MultiStatus Support

sayan-das-in:
This PR adds MultiStatus support to Marketo destination.

## Testing

Testing in progress

### DIFF
--- a/packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
+++ b/packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
@@ -28,6 +28,7 @@ import {
   CREATE_LIST_ENDPOINT,
   GET_LIST_ENDPOINT
 } from './constants'
+import { JSONLikeObject } from '@segment/actions-core/*'
 
 // Keep only the scheme and host from the endpoint
 // Marketo UI shows endpoint with trailing "/rest", which we don't want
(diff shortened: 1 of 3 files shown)


packages/destination-actions/src/destinations/marketo-static-lists/__tests__/multistatus.test.ts
```diff
@@ -0,0 +1,354 @@
+import { SegmentEvent, createTestEvent, createTestIntegration } from '@segment/actions-core'
+import nock from 'nock'
+import MarketoStaticList from '../index'
+// import { BrazeTrackUserAPIResponse } from '../utils'
```
varadarajan-tw, line 4:
```suggestion
// import { BrazeTrackUserAPIResponse } from '../utils'
```

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
```diff
@@ -66,26 +76,93 @@ export async function addToList(
...
+  statsContext?.statsClient?.incr('addToAudience.success', payloads.length, statsContext?.tags)
+
+  // Build a MultiStatusResponse and return it
+  const multiStatusResponse = new MultiStatusResponse()
+
+  for (let i = 0; i < payloads.length; i++) {
+    multiStatusResponse.setSuccessResponseAtIndex(i, {
+      status: 200,
+      // CSV data could be as large as 10 MB, so we truncate it to upto 50 characters
+      sent: `${csvData.substring(0, 50)}...`,
```
varadarajan-tw, line 141: Could we set CSV representation of each row instead of trimming and setting the first 50 characters? Sent should be set the same way for both success and failure.

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
```diff
@@ -66,26 +76,93 @@ export async function addToList(
...
+  const url =
+    api_endpoint +
+    BULK_IMPORT_ENDPOINT.replace('externalId', list_id).replace('fieldToLookup', payloads[0].lookup_field)
+
+  const response = await request<MarketoBulkImportResponse>(url, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'multipart/form-data; boundary=--SEGMENT-DATA--'
+    },
+    body: createFormData(csvData)
```
varadarajan-tw, line 124: Should we set throwHttpErrors false here? The request client might throw exception for non-2xx status. The next block seems to handle errors returned with 2xx status.

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
```diff
@@ -124,6 +201,77 @@ export async function removeFromList(
...
+  }
+  statsContext?.statsClient?.incr('removeFromAudience.success', payloads.length, statsContext?.tags)
+
+  // Build a MultiStatusResponse and return it
+  const multiStatusResponse = new MultiStatusResponse()
+
+  for (let i = 0; i < payloads.length; i++) {
+    multiStatusResponse.setSuccessResponseAtIndex(i, {
+      status: 200,
+      sent: '',
```
varadarajan-tw, line 266: Any reason why`sent` isn't set?

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
```diff
@@ -124,6 +201,77 @@ export async function removeFromList(
...
+
+  const getLeadsUrl =
+    api_endpoint +
+    GET_LEADS_ENDPOINT.replace('field', payloads[0].lookup_field).replace(
+      'emailsToFilter',
+      encodeURIComponent(usersToRemove)
+    )
+
+  // Get lead ids from Marketo
+  const getLeadsResponse = await request<MarketoGetLeadsResponse>(getLeadsUrl, {
```
varadarajan-tw, line 229: Same as above. Should we set throwHttpErrors false here? The request client might throw exception for non-2xx status. The next block seems to handle only for 2xx status.

varadarajan-tw (commented):
Greate work! Left a couple of questions around error handling.

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
sayan-das-in: Marketo always returns 200 status but agreed that we can set the `throwHttpErrors` to false here.

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
sayan-das-in: Updated the code.

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
sayan-das-in: This is a `DELETE` call and the body is technically empty. The item to be deleted is already in the URL.

sayan-das-in: We pass on an id as a query parameter in the URL, maybe we can record something like `id=1` in the sent parameters.

packages/destination-actions/src/destinations/marketo-static-lists/functions.ts
sayan-das-in: Made a change to record each row of the CSV in the sent field.

varadarajan-tw (approved):
Confirmed with @sayan-das-in that Marketo Static Lists always returns 200. ([ref](https://experienceleague.adobe.com/en/docs/marketo-developer/marketo/rest/bulk-import/bulk-lead-import
))
Thank you for addressing the comments! Nice work ✨

PR #2655, opened 2024-12-17
[TOPSORT] Action Definitions for Banner Events

agustinespildora:
> [!NOTE]  
> Ready to review.

# Topsort Banner Events

This is a destination so that eCommerce sites that already use Segment (to track actions) can integrate Topsort events for sponsored listings seamlessly. With this PR we now add the events integration for banner ads too. 

## Intro to Topsort Banners
Topsort allows to monetize through banner auction ads served in your site. To learn more about Topsort you can visit our [docs](https://docs2.topsort.com).
This destination already helps you track the 3 main events Topsort needs to serve the most relevant sponsored listing ads: `impressions`, `clicks` and `purchases`. Here is the new events we map for the banner events:

1. **Banner Impression**: Every time a user sees a Topsort promoted banner.
2. **Banner Click**: If the user then clicks on the banner, then a topsort `click` should be triggered with the banner `resolvedBidId`.
3. **Product Impression**: If the banner redirects to a PLP: every time a user sees a Topsort promoted product in the landing page of the banner. If the banner redirects to a PDP: every time the user sees the Topsort promoted banner.
4. **Product Click**: If the banner redirects to a PLP: every time a user clicks or adds to the cart a Topsort promoted product in the landing page of the banner. If the banner redirects to a PDP: every time the user clicks the Topsort promoted banner.
5. **Purchase**: [Already mapped] Finally, when an order is completed and confirmed to be successful, a `purchase` event should be reported to Topsort. You can report a purchase whether it has promoted products or not. We will take care of filtering relevant promoted products inside the completed order given the information about promoted clicks. For more information on banner attribution with Topsort you can go to https://api.docs.topsort.com/api-reference/examples/sponsored-banners/attribution

## Changes

- [x] Added additional attribution required fields
- [x] Added banner impression event presets

#### Click event with additional attribution in local Action Tester
image

#### Impression event with additional attribution in local Action Tester
image

#### Banner Impression event with local Action Tester (updated with last changes Jan 7th)
image

#### Banner Click event with local Action Tester (updated with last changes Jan 7th)
image


agustinespildora:
Regarding the `Required Field Check`
The existing actions were added only optional fields
The new actions are not used right now by the only client using this destination, so that failed check can be ignored
@tcgilbert

tcgilbert (dismissed):

tcgilbert:
@agustinespildora appreciate for the thorough PR! Looks good to me. 

> Regarding the Required Field Check

Yeah this fails when coming from a ofrked branch, so no worries there. That said, we're going into a deploy freeze for the Xmas and New Years holidays, so this PR won't go out until the New Year.

agustinespildora:
That's a pitty, no chance to be able to slip this into some hot-fix or something?
The sooner the better, we would really appreciate it 🙏🏼 
@tcgilbert

packages/destination-actions/src/destinations/topsort/bannerImpression/index.ts
```diff
@@ -0,0 +1,59 @@
...
+      default: {
+        '@path': '$.properties.resolvedBidId'
+      }
+    }
+  },
+  perform: (request, { payload, settings }) => {
+    const client = new TopsortAPIClient(request, settings)
+    return client.sendEvent({
+      impressions: [payload]
+    })
```
joe-ayoub-segment, line 55: hi @agustinespildora - how is this Action any different to the Impression Action? It seems to send the exact same payload to Topsort.

packages/destination-actions/src/destinations/topsort/bannerClick/index.ts
```diff
@@ -0,0 +1,59 @@
...
+      default: {
+        '@path': '$.properties.resolvedBidId'
+      }
+    }
+  },
+  perform: (request, { payload, settings }) => {
+    const client = new TopsortAPIClient(request, settings)
+    return client.sendEvent({
+      clicks: [payload]
+    })
```
joe-ayoub-segment, line 55: hi again @agustinespildora (and @tcgilbert ) - why is the payload for this new Action identical to the payload send in the existing click Action?

If they are the same, then a new Action might not be the best way to do this. You could just add a new Preset which uses the 'click' Action but sets the subscription to be sets the subscription to 'type = "track" and event = "Banner Clicked"'.

tcgilbert:
@agustinespildora - Yes, @joe-ayoub-segment raises a great point. If there’s no functional difference between some of the actions (e.g., "Banner Click" vs. "Banner Impression"), it’s probably best to consolidate them into a single action. You could then create multiple presets to differentiate how the single action can be used

packages/destination-actions/src/destinations/topsort/bannerImpression/index.ts
agustinespildora: Hey @joe-ayoub-segment, this event sends the same fields as the Impression action, and thus to the API there is no difference. But the value of the resolvedBidId is decoded differently on our side, and there is a difference on the ecommerce side too, because it needs to detect a Banner impression and send a Banner resolvedBidId instead of a product resolvedBidId. That's why we decided to name it as a different event.

packages/destination-actions/src/destinations/topsort/bannerClick/index.ts
agustinespildora: I see your point now. I can do that change as I believe makes sense. Thanks for the tip.

agustinespildora: This is done

packages/destination-actions/src/destinations/topsort/bannerImpression/index.ts
agustinespildora: This is addressed and done

tcgilbert (approved):

joe-ayoub-segment (approved):

joe-ayoub-segment:
hi @agustinespildora PR deployed. cc @tcgilbert

PR #2671, opened 2025-01-09
[SendGrid]- sendEmail Action - change to default mappings for cc and bcc fields

joe-ayoub-segment:
SendGrid - sendEmail Action - change to default mappings for cc and bcc fields 

## Testing
Tested locally.


joe-ayoub-segment:
Deployed

PR #2674, opened 2025-01-10
[STRATCONN-5399]-Handles refresh token retry for multistatus responses

varadarajan-tw:
This PR handles 401 refresh token retries for multistatus responses. Previously, we threw 401 exception for refresh token retries. However, for handling multistatus, developers now return a `MultistatusResponse` object from `performBatch` block. This PR ensures that if any of the events in multistatus response has 401 status, the batch will be retried with new refresh token for oauth enabled destinations.

Related but not dependent - https://github.com/segmentio/integrations/pull/3079

## Testing

Testing completed successfully in staging - [Test Doc](https://docs.google.com/document/d/1Q-sIszXK_1EFDarVemfCknOxsmRcOm9AWI0LmU0KFGQ/edit?tab=t.0)

### DIFF
--- a/packages/core/src/destination-kit/index.ts
+++ b/packages/core/src/destination-kit/index.ts
@@ -483,7 +483,7 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
       return await audienceConfig?.createAudience(requestClient, createAudienceInput)
     }
 
-    const onFailedAttempt = async (error: ResponseError & HTTPError) => {
+    const onFailedAttempt = async (error: ResponseError | HTTPError) => {
       settings = await this.handleAuthError(error, settings)
     }
     return await retry(run, { retries: 2, onFailedAttempt })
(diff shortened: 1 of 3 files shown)


packages/core/src/destination-kit/index.ts
```diff
@@ -901,7 +901,7 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
       return result
     }
 
-    const onFailedAttempt = async (error: ResponseError & HTTPError) => {
+    const onFailedAttempt = async (error: ResponseError | HTTPError) => {
```
varadarajan-tw, line 904: error can either be ResponeError or HTTPError. It can never be a mix of both.

packages/core/src/destination-kit/index.ts
```diff
@@ -928,12 +928,39 @@ export class Destination<Settings = JSONObject, AudienceSettings = JSONObject> {
...
+        const isOAuthDestination =
+          this.authentication?.scheme === 'oauth2' || this.authentication?.scheme === 'oauth-managed'
+        if (attemptCount <= MAX_ATTEMPTS && has401Errors && isOAuthDestination) {
+          await this.handleAuthError(settings, options)
+          return false
+        }
+      }
+      return true
+    }
+    return await retry<Result[]>(run, { retries: MAX_ATTEMPTS, onFailedAttempt, shouldRetry })
```
varadarajan-tw, line 963: Another approach I could have taken is to throw an exception if any of the events failed with 401 error and re-use the existing flow. However, that means we would end up reporting other 400 errors as 401 as well. Therefore, decided to go ahead with this approach of intentionally handling multistatus response. Open to any other thoughts.

itsarijitray (approved):

sayan-das-in (approved):

PR #2684, opened 2025-01-16
Fix Mixpanel distinct_id for anonymous events

nunofgs:
This PR adds support for anonymous Mixpanel events by allowing the `distinct_id` property to be empty, [as documented](https://developer.mixpanel.com/reference/import-events#propertiesdistinct_id) by Mixpanel:

> If the event is not associated with any user, set distinct_id to the empty string. Events with an empty distinct_id will be excluded from all behavioral analysis.

Before this PR (when distinct_id is empty):

Screenshot 2025-01-16 at 11 12 42

After this PR:

Screenshot 2025-01-16 at 11 12 58

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/mixpanel/trackEvent/functions.ts
+++ b/packages/destination-actions/src/destinations/mixpanel/trackEvent/functions.ts
@@ -17,12 +17,24 @@ export function getEventProperties(payload: Payload, settings: Settings): Mixpan
     browserVersion = getBrowserVersion(payload.userAgent)
   }
   const integration = payload.context?.integration as Record<string, string>
+
+  // When the `distinct_id` property is undefined we add the property in as a blank string,
+  // so that we can support Mixpanel events that aren't tied to a specific user.
+  //
+  // The `distinct_id` property is, according to documentation, supposed to
+  // ALWAYS be present, either with an identifying value or with an empty
+  // string value to denote that the event should be excluded from behavioral
+  // analysis.
+  //
+  // @see https://developer.mixpanel.com/reference/import-events#propertiesdistinct_id
+  const distinct_id = payload.distinct_id ?? ''
+
   return {
     time: time,
     ip: payload.ip,
     id: payload.distinct_id,
     $anon_id: payload.anonymous_id,
-    distinct_id: payload.distinct_id,
+    distinct_id,
     $app_build_number: payload.app_build,
     $app_version_string: payload.app_version,
     $app_namespace: payload.app_namespace,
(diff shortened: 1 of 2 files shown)


joe-ayoub-segment:
Hi @brennan - this is an external PR for an internally built Destination. Can you triage please?

joe-ayoub-segment:
Hi @nunofgs , 
Apologies for the crazy delay with this PR. 
I was on extended leave and am back now. Looks like this PR fell through the cracks. 

Is it still something you'd like progressed?

If so can you help me understand the implications of deploying this change?
- If we pass an empty string for the distinct_id, will is suddenly result in anonymous user profiles showing up in customer's Mixpanel instances? 
- If so: 
   - Which identifier will these anonymous profiles have? 
   - Will deploying this change affect the customer's analytics in any way? 
   - Will there be any other impact of this change to customers who have been using Segment and Mixpanel in this way for the last few years?

This PR might fix a bug - however it's important to understand if customers have become used to, or rely on their data being collected in this way now. If we fix this bug, might it cause issues for some customers?

I'm keen to get your thoughts on this. 

Kind regards,
Joe

joe-ayoub-segment:
image

joe-ayoub-segment:
hi @nunofgs just following up on this - do you still want to progress this PR?

joe-ayoub-segment:
hi @spjtls9 - I wonder if you could take a look at this PR from @nunofgs please? It's a very small change, but it would be great to have your sign off on it.

packages/destination-actions/src/destinations/mixpanel/trackEvent/functions.ts
```diff
@@ -17,12 +17,24 @@ export function getEventProperties(payload: Payload, settings: Settings): Mixpan
...
+  // @see https://developer.mixpanel.com/reference/import-events#propertiesdistinct_id
+  const distinct_id = payload.distinct_id ?? ''
+
   return {
     time: time,
     ip: payload.ip,
     id: payload.distinct_id,
     $anon_id: payload.anonymous_id,
-    distinct_id: payload.distinct_id,
+    distinct_id,
```
spjtls9, line 37:
```suggestion
    distinct_id: payload.distinct_id ?? ''
```

spjtls9 (changes requested):
Yup this makes sense. Thank you for the change. I have one change request for syntax

nunofgs:
Hi Joe! Sorry I missed the notification on this thread.

> If so can you help me understand the implications of deploying this change?

I have no affiliation with Mixpanel so I'm probably not qualified to answer. Very valid questions though! I simply encountered this bug when attempting to send an anonymous event to Mixpanel.

> This PR might fix a bug - however it's important to understand if customers have become used to, or rely on their data being collected in this way now. If we fix this bug, might it cause issues for some customers?

Hyrum's Law indeed. Worth checking with the Mixpanel team for sure.

I'll make the requested change by @spjtls9 👍

joe-ayoub-segment (approved):

PR #2654, opened 2024-12-17
update: update VWO SmartCode to version 2.1

zeeshan-vwo:
This PR updates the VWO SmartCode in the project to version 2.1, replacing the outdated version to ensure compatibility with VWO’s latest features and improvements. The update is backward-compatible, with testing successfully completed in a local environment. No new test cases are required.

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [x] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change.
- [x] [Segmenters] Tested in the staging environment
- [x] [Segmenters] [If applicable for this change] Tested for regression with Hadron.

PR #2686, opened 2025-01-16
updating smart code snippet to not cause opacity 0 in hide element style

segment-anthony:
There was a bug in VWO that was fix in legacy repository to which caused a 'flashing' issue. This was caused by the hide element style where opacity 0 was causing a quick 'flash' in the website. 

## Testing

### DIFF
--- a/packages/browser-destinations/destinations/vwo/src/init-script.ts
+++ b/packages/browser-destinations/destinations/vwo/src/init-script.ts
@@ -8,74 +8,173 @@ export function initScript({
   useExistingJquery = false,
   isSpa = 1
 }) {
-  window._vwo_code =
-    window._vwo_code ||
+  window._vwo_code ||
     (function () {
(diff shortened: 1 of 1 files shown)


joe-ayoub-segment:
hi @segment-anthony thanks for raising this. 
Is there any way you can test that this works before I deploy it on Monday?
With a PR like this I can't review the code in a meaningful way (I won't understand what the change it), so to reduce risk it's best to have some testing done. Would this be possible?

joe-ayoub-segment:
Regarding testing. 
@seg-leonelsanches indicated that he tested this PR: https://github.com/segmentio/action-destinations/pull/2654 
which is identical this current new PR, apart from a css single line.

image


https://github.com/user-attachments/assets/1c3a2a05-1362-400e-997a-3ab49eafcf98

joe-ayoub-segment (approved):

PR #1450, opened 2023-07-27
[ACT-817] Add new preset type

valerieernst:
This PR adds the concept of categorizing presets as either automatic (i.e. automatically add when a user creates a destination) or associated with a specific pre-defined event.

See the README updates for more info.

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [ ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [ ] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

PR #2714, opened 2025-01-30
[LAB-1682] Create presets for each standard audience action destination

drew-thompson:
This PR adds presets (of the `specificEvent` kind) for each of the standard (not full sync) _audience_ action destinations in our catalog. These presets will be used in filtering out actions which are extraneous to event emitter creation, leaving only the ones used for syncing profiles.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
@@ -178,10 +178,9 @@ const destination: AudienceDestinationDefinition<Settings> = {
   presets: [
     {
       name: 'Entities Audience Membership Changed',
-      partnerAction: 'uploadConversionAdjustment',
+      partnerAction: 'userList',
       mapping: {
-        // TODO: See if we use the original or the v2 action
-        ...defaultValues(uploadConversionAdjustment.fields),
+        ...defaultValues(userList.fields),
         properties: {
           '@path': '$.properties'
         }
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
@@ -1,4 +1,4 @@
-import { AudienceDestinationDefinition } from '@segment/actions-core'
+import { AudienceDestinationDefinition, defaultValues } from '@segment/actions-core'
 import type { Settings } from './generated-types'
 import postConversion from './postConversion'
 import uploadCallConversion from './uploadCallConversion'
@@ -174,7 +174,22 @@ const destination: AudienceDestinationDefinition<Settings> = {
     uploadClickConversion2,
     uploadCallConversion2,
     userList
-  }
+  },
+  presets: [
+    {
+      name: 'Entities Audience Membership Changed',
+      partnerAction: 'uploadConversionAdjustment',
+      mapping: {
+        // TODO: See if we use the original or the v2 action
+        ...defaultValues(uploadConversionAdjustment.fields),
+        properties: {
+          '@path': '$.properties'
+        }
+      },
+      type: 'specificEvent',
+      eventSlug: 'warehouse_audience_membership_changed_identify'
+    }
+  ]
 }
 
 export default destination
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
@@ -179,9 +179,7 @@ const destination: AudienceDestinationDefinition<Settings> = {
     {
       name: 'Entities Audience Membership Changed',
       partnerAction: 'userList',
-      mapping: {
-        ...defaultValues(userList.fields)
-      },
+      mapping: defaultValues(userList.fields),
       type: 'specificEvent',
       eventSlug: 'warehouse_audience_membership_changed_identify'
     }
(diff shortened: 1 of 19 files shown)


packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
```diff
@@ -174,7 +174,22 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
     uploadCallConversion2,
     userList
-  }
+  },
+  presets: [
+    {
+      name: 'Entities Audience Membership Changed',
+      partnerAction: 'uploadConversionAdjustment',
+      mapping: {
+        // TODO: See if we use the original or the v2 action
```
SethNuteTwilio, line 183: Do we need to follow-up on this?

packages/destination-actions/src/destinations/taboola-actions/index.ts
```diff
@@ -77,7 +77,10 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
       full_audience_sync: false
     },
     async createAudience(request, createAudienceInput) {
-      const audienceName = createAudienceInput.settings?.audience_identifier === 'computation_key' ? createAudienceInput.personas?.computation_key : createAudienceInput.audienceName
+      const audienceName =
```
SethNuteTwilio, line 80: Nested ternary 😭

SethNuteTwilio (dismissed):
LGTM 👍

packages/destination-actions/src/destinations/google-enhanced-conversions/index.ts
drew-thompson: Yes we do 😄

rvadera12 (commented):
From my understanding, the new actions will be visible to all users, regardless of whether they are using the entities flow or not. Given that adding these presets essentially creates new actions with predefined property values, it might lead to confusion among users who don’t need these actions or aren’t familiar with the concept.

Would it be possible to hide these new actions from the UI for all customers to avoid adding unnecessary complexity? Alternatively, perhaps we could include a more detailed explanation of these actions in the docs or setup flow, so users understand their purpose and context better.

This would help streamline the user experience and prevent any potential confusion when these actions appear in the UI.

Just considering the user flow, we've received numerous questions and feedback in the past regarding multiple actions that essentially perform the same function, which has led to confusion.

CC: @smultani

drew-thompson:
> From my understanding, the new actions will be visible to all users, regardless of whether they are using the entities flow or not. Given that adding these presets essentially creates new actions with predefined property values, it might lead to confusion among users who don’t need these actions or aren’t familiar with the concept.
> 
> Would it be possible to hide these new actions from the UI for all customers to avoid adding unnecessary complexity? Alternatively, perhaps we could include a more detailed explanation of these actions in the docs or setup flow, so users understand their purpose and context better.
> 
> This would help streamline the user experience and prevent any potential confusion when these actions appear in the UI.
> 
> Just considering the user flow, we've received numerous questions and feedback in the past regarding multiple actions that essentially perform the same function, which has led to confusion.
> 
> CC: @smultani

Hi @rvadera12! We're not adding any new `actions`, just `presets` that have the field `type: 'specificEvent'`, which is something recently built for this specific purpose. As a quick example, [customer.io](https://github.com/segmentio/action-destinations/blob/c1b21ce224657b4de0995ed72eea45d1f9336626/packages/destination-actions/src/destinations/customerio/index.ts) has many of these `presets` defined, all pointing to the `trackEvent` action, and each of those presets doesn't cause a new action to be presented to customers.

You can see the PR which added `specificEvent` (which makes it an internal mechanism) in https://github.com/segmentio/action-destinations/pull/1450.

rvadera12:
> > From my understanding, the new actions will be visible to all users, regardless of whether they are using the entities flow or not. Given that adding these presets essentially creates new actions with predefined property values, it might lead to confusion among users who don’t need these actions or aren’t familiar with the concept.
> > Would it be possible to hide these new actions from the UI for all customers to avoid adding unnecessary complexity? Alternatively, perhaps we could include a more detailed explanation of these actions in the docs or setup flow, so users understand their purpose and context better.
> > This would help streamline the user experience and prevent any potential confusion when these actions appear in the UI.
> > Just considering the user flow, we've received numerous questions and feedback in the past regarding multiple actions that essentially perform the same function, which has led to confusion.
> > CC: @smultani
> 
> Hi @rvadera12! We're not adding any new `actions`, just `presets` that have the field `type: 'specificEvent'`, which is something recently built for this specific purpose. As a quick example, [customer.io](https://github.com/segmentio/action-destinations/blob/c1b21ce224657b4de0995ed72eea45d1f9336626/packages/destination-actions/src/destinations/customerio/index.ts) has many of these `presets` defined, all pointing to the `trackEvent` action, and each of those presets doesn't cause a new action to be presented to customers.
> 
> You can see the PR which added `specificEvent` (which makes it an internal mechanism) in #1450.

Got it! Thanks for clarifying, I didn't know we had the capability to hide presets if it is an engage flow.

marinhero (approved):

itsarijitray:
@drew-thompson @SethNuteTwilio 
Some validation steps are [failing in the CI](https://github.com/segmentio/action-destinations/actions/runs/13146783121/job/36686558595?pr=2714). Can you look into this?
Also, why are you adding properties node to presets?

marinhero (approved):
We've reached internal consensus and you'll be deployed in 2/18. Thank you for your patience.

PR #2741, opened 2025-02-14
Fix inconsistencies between node v18 and v22

sayan-das-in:
This PR enables actions destination to be built with both Node 18 and Node 22 by addressing inconsistencies in dependencies, typescript fixes, linting issues and unit tests.

### List of configuration changes
Changed `"useUnknownInCatchVariables": false` in `packages/destination-actions/tsconfig.build.json` as with the typescript upgrade, the error in `catch(error)` is being treated as `unknown` by default, breaking builds and tests. Setting this parameter to false treats the error of type any.

### List of Dependency upgrades

1. @types/node - `^18.11.15` -> `^22.13.1`
2. typescript - `4.3.5` -> `4.9.5`
3. ssh2-sftp-client - `9.1.0` -> `10.0.3`
4. ts-loader - `^9.2.6` -> `^9.5.2`

## Testing

Tested locally and on github actions, no other E2E tests were done as this change doesn't alter any destination logic.
A detailed test document is available [here](https://docs.google.com/document/d/1-j8N3njiKv0SvHxtwWt0KwW2M6-XqBN7Ljv0uo5wsvM/edit?tab=t.yxd7mg12qec9).

These changes have been deployed on staging and being monitored with mme e2e tests.


packages/core/package.json
```diff
@@ -51,7 +51,7 @@
     "package.json"
   ],
   "engines": {
-    "node": ">=14"
+    "node": ">=18.12"
```
silesky, line 54: Why not "^18.12 || ^22.13" Like the others?

sayan-das-in: This one is intentional as the `app` repo depends on this package which runs on 20.x, and without this `yarn install` would complain.

pooyaj (dismissed):

marinhero (approved):

pooyaj (approved):

PR #2772, opened 2025-02-27
[required fields v2] Fixes an issue with schema validation in a specific case

nick-Ag:
This PR fixes an issue I found with conditionally required fields where this condition did not generate a correct schema:

```
    isSendable: {
      label: 'Is Sendable',
      type: 'boolean',
      ...
    },
    sendableCustomObjectField: {
      label: 'Sendable Custom Object Field',
      type: 'string',
      required: {
        match: 'all',
        conditions: [{ fieldKey: 'isSendable', operator: 'is', value: true }]
      }
    },
    sendableSubscriberField: {
      label: 'Sendable Subscriber Field',
      type: 'string',
      required: {
        match: 'all',
        conditions: [{ fieldKey: 'isSendable', operator: 'is', value: true }]
      }
    },
```
When 'isSendable' was not included in the mapping at all, the validation would fail and incorrectly require both 'sendableCustomObjectField' and 'sendableSubscriberField'
## Testing

### DIFF
--- a/packages/core/src/destination-kit/fields-to-jsonschema.ts
+++ b/packages/core/src/destination-kit/fields-to-jsonschema.ts
@@ -109,6 +109,7 @@ const simpleConditionToJSONSchema = (
 
   return {
     if: {
+      required: [dependantFieldKey],
       properties: { [dependantFieldKey]: dependantValueToJSONSchema }
     },
     then: generateThenStatement(fieldKey)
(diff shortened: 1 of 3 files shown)


packages/core/src/destination-kit/fields-to-jsonschema.ts
```diff
@@ -109,6 +109,7 @@ const simpleConditionToJSONSchema = (
 
   return {
     if: {
+      required: [dependantFieldKey],
```
nick-Ag, line 112: This basically requires that the field actually exist for the if statement to evaluate to true. Without it the if statement will be true when the field is missing

rvadera12 (approved):

nick-Ag:
Merging this but not deploying right away

PR #2785, opened 2025-03-10
[STRATCONN-5488] - Adds batch config field to action definition

varadarajan-tw:
This PR adds new batch config field to action definition. This will allow devs to consicely define batch options. Most of the destinations define these batch options as hidden fields. So, this new definition makes them hidden by default. To override, builders simply have to define a field with appropriate keys and then customize the option further.

- batch bytes - max number of bytes allowed in batch
- batch keys - the keys based on which batching has to be done.
- batch size - max number of events to include in a batch
- hideEnableBatching - hides the default generated enable_batching option and sets default value to true.

**Does it impact to existing configurations?**

No impact to existing batch configurations. This PR adds a new convenient way to configure batch related configurations and batch keys moving forward.

If these configurations are defined in both batch settings and as fields, yarn validate run from CI would through an error. This is to ensure we don't end up defining values in two places.

This PR  is no-op without action-cli PR - https://github.com/segmentio/action-cli/pull/211

Documentation will be updated after we launch batch_keys feature.

## Testing

Tested changes validate.ts 

image

image

image

### DIFF
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -13,8 +13,7 @@ export const RESERVED_FIELD_NAMES = [
   'password',
   'secretkey',
   'secret',
-  'securitytoken',
-  'batch_keys'
+  'securitytoken'
 ]
 
 export const OAUTH_SCHEME = 'oauth2'
--- a/packages/cli/src/constants.ts
+++ b/packages/cli/src/constants.ts
@@ -13,7 +13,8 @@ export const RESERVED_FIELD_NAMES = [
   'password',
   'secretkey',
   'secret',
-  'securitytoken'
+  'securitytoken',
+  'batch_keys'
 ]
 
 export const OAUTH_SCHEME = 'oauth2'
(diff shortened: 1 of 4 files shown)


packages/cli/src/constants.ts
```diff
@@ -13,7 +13,8 @@ export const RESERVED_FIELD_NAMES = [
   'password',
   'secretkey',
   'secret',
-  'securitytoken'
+  'securitytoken',
+  'batch_keys'
```
varadarajan-tw, line 17: We don't want anyone to define batch_keys as part of action fields.

packages/core/src/destination-kit/types.ts
```diff
@@ -396,3 +396,27 @@ export type ResultMultiStatusNode =
...
+   * If you wish to expose this to the customer, consider defining this as a field and customize it.
+   */
+  batchSize?: number
+
+  /**
+   * Whether or not to enable batching for this action by default.
+   * This will hide the enable_batching field in the UI and set default state to true.
+   * If you wish to expose this to the customer, consider defining this as a field and customize it.
+   */
+  hideEnableBatching?: boolean
```
varadarajan-tw, line 421: is this naming good? Any other good names?

PR #2791, opened 2025-03-12
[STRATCONN-5488] Introduces batch_keys field (pilot)

varadarajan-tw:
This PR 
- updates `fields` type in ActionDefinition to specify specific shape for batch_keys
- adds some validations for batch_keys field so that we can catch configuration errors earlier.

## Testing

image
image

**Scenario: 2**
image
image

### DIFF
--- a/packages/core/src/destination-kit/action.ts
+++ b/packages/core/src/destination-kit/action.ts
@@ -46,8 +46,8 @@ export type RequestFn<Settings, Payload, Return = any, AudienceSettings = any, A
 
 interface ReservedInputFields {
   batch_keys?: {
-    label: 'Batch Keys'
-    description: 'The mapping keys to use to batch events together. Events with the same values for these keys will be batched together.'
+    label: string
+    description: string
     type: 'string'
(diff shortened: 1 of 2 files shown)


packages/core/src/destination-kit/action.ts
```diff
@@ -44,6 +44,20 @@ export type RequestFn<Settings, Payload, Return = any, AudienceSettings = any, A
   data: ExecuteInput<Settings, Payload, AudienceSettings, ActionHookInputs>
 ) => MaybePromise<Return>
 
+interface ReservedInputFields {
+  batch_keys?: {
+    label: 'Batch Keys'
```
nick-Ag, line 49: Do we want builders to be able to customize label & description? I'm wondering if the value of these could be different from destination to destination?

varadarajan-tw: Sure. Let me change that. We anyways don't want these to be visible (atleast in the near future). So, it doesn't matter what they label this field.

nick-Ag (approved):

PR #2798, opened 2025-03-17
Ortto Action Destination

xitonix:

## Testing


joe-ayoub-segment:
hi @xitonix thanks for raising this PR for a new Integration. Apologies for the delay reviewing this - we've had some staffing availability issues lately. 

I'm going to review this in the next 24h.

packages/destination-actions/src/destinations/ortto/common-fields.ts
```diff
@@ -0,0 +1,144 @@
...
+    description: `Ortto audience ID`,
+    type: 'string',
+    default: {
+      '@if': {
+        exists: {
+          '@path': '$.context.personas.audience_settings.audience_id'
+        },
+        then: { '@path': '$.context.personas.audience_settings.audience_id' },
+        else: { '@path': '$.context.personas.external_audience_id' }
+      }
```
joe-ayoub-segment, line 139: Note to self: I need to check that these default mappings make sense. 
@xitonix - why do you want to check the location '$.context.personas.audience_settings.audience_id' ?
The Audience ID should always be located at '$.context.personas.external_audience_id'

packages/destination-actions/src/destinations/ortto/ortto-client.ts
```diff
@@ -0,0 +1,209 @@
...
+    })
+  }
+
+  // Audiences
+
+  createAudience = async (settings: Settings, audienceName: string): Promise<Audience> => {
+    if (!audienceName) {
+      throw new PayloadValidationError(Errors.MissingAudienceName)
+    }
+    const url = this.getEndpoint(settings.api_key).concat('/audiences/create')
```
joe-ayoub-segment, line 95: Is the call to this API idempotent?
i.e what happens if the Audience has already been created? Will it simply return the pre-existing Audience details? Only reason I ask is if the customer creates an Audience, then deletes it, then recreates it with the same name.

packages/destination-actions/src/destinations/ortto/utils.ts
```diff
@@ -0,0 +1,8 @@
+export const TEST_API_KEY = 'pau-key'
+export function cleanObject<T extends {}>(obj: T): Partial<T> {
+  return Object.fromEntries(
+    Object.entries(obj)
+      .filter(([key, _]) => key !== '')
+      .map(([key, value]) => [key, value instanceof Object && !(value instanceof Array) ? cleanObject(value) : value])
+  ) as Partial<T>
+}
```
joe-ayoub-segment, line 8: Would something like this work better? It doesn't use recursion: 

```
function cleanObject<T>(obj: T): Partial<T> {
  return JSON.parse(
    JSON.stringify(obj, (key, value) =>
      key === '' ? undefined : value
    )
  );
}
```

packages/destination-actions/src/destinations/ortto/ortto-client.ts
```diff
@@ -0,0 +1,209 @@
...
+      if (event.namespace === 'ortto.com') {
+        continue
+      }
+      if (!event.audience_id) {
+        event.audience_id = hookAudienceID
+      }
+      filtered.push(cleanObject(event))
+    }
+    if (filtered.length == 0) {
+      return
```
joe-ayoub-segment, line 73: Maybe throw an error if there are no valid payloads?

joe-ayoub-segment (changes requested):
hi @xitonix thanks for this substantial PR for a new Integration! 

I think the next step should be for you to look through the comments I left, and then for us to catch up 1:1. 

I have some questions about the design of the Integration. 

In particular how Contacts are added and removed from Audiences. 

If I recall correctly you are based in Australia? I'm in the UK timezone but can meet at a time that works for you between 10am UK time and 9pm UK time. Please suggest a time and I'll forward you an invite. I can meet tomorrow (Friday), Monday or Tuesday. 

Best regards,
Joe

packages/destination-actions/src/destinations/ortto/common-fields.ts
```diff
@@ -0,0 +1,144 @@
...
+      email: {
+        label: 'Email',
+        description: "The contact's email address",
+        placeholder: 'john.smith@example.com',
+        type: 'string',
+        format: 'email'
+      },
+      phone: {
+        label: 'Phone Number',
+        description: "The contact's phone number",
```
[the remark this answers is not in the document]
xitonix: Country code isn’t mandatory in our system, as we fall back to an account-level default when it's not provided—and our customers are aware of this. That said, it's still highly recommended, and I’ll update the description to reflect that.

packages/destination-actions/src/destinations/ortto/ortto-client.ts
xitonix: That's correct — if an audience with the same name already exists in our system, it will be returned.

packages/destination-actions/src/destinations/ortto/utils.ts
xitonix: Doesn't seem like it. Looks like `parse` doesn't like `undefined`
2025-05-02 at 09 27 08

packages/destination-actions/src/destinations/ortto/ortto-client.ts
xitonix: I don’t think we should. If an individual payload is truly invalid, we already return a `PayloadValidationError`. In this case, it’s reasonable to silently drop track events sent by Ortto— as users can send activities from Ortto to Segment, and silent dropping will prevent infinite flow of data. It's a valid scenario imo, not an error state.

packages/destination-actions/src/destinations/ortto/ortto-client.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+    const url = this.getEndpoint(settings.api_key).concat('/identify')
+    return this.request(url, {
+      method: 'POST',
+      json: cleaned
+    })
+  }
+
+  sendActivities = async (settings: Settings, payloads: EventPayload[], hookAudienceID: string) => {
```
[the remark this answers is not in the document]
xitonix: Certainly, mate. At the moment, our endpoints mimic that behavior by returning something like this:
```
...
received: 100,
errors: {
  "error message": [even_id_1,...., event_id_N]
}
```
The HTTP status is determined based on the number of errors:

- 400 if all events failed,
- 202 if everything succeeded,
- 206 if it’s a partial success (some valid, some invalid).

I’m happy to revisit the logic if there’s a better approach you’d recommend.

packages/destination-actions/src/destinations/ortto/trackActivity/index.ts
```diff
@@ -0,0 +1,204 @@
...
+          label: 'City',
+          type: 'string',
+          allowNull: true
+        },
+        post_code: {
+          label: 'Postcode',
+          type: 'string',
+          allowNull: true
+        }
+      },
```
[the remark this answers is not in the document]
xitonix: Good catch! I likely forgot to use the common fields for this action. I’ve now also extracted the hook input/output fields into the same file to keep things consistent. Hopefully they work as expected—though I’m not sure how to test them locally since the Action Tester doesn’t display those fields.

packages/destination-actions/src/destinations/ortto/trackActivity/index.ts
```diff
@@ -0,0 +1,204 @@
...
+                message: (err as APIError).message ?? 'Unknown Error',
+                code: (err as APIError).status?.toString() ?? 'Unknown Error'
+              }
+            }
+          }
+        }
+        return {}
+      }
+    }
+  },
```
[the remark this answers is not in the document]
xitonix: That's correct, @joe-ayoub-segment. The goal is to allow users to optionally sync contacts with an Ortto Audience, even if they don’t use the Audience feature in Segment.

> I know it’s been a while since we last met, so my memory might be a little rusty (apologies!).

No worries, mate. When we last spoke, we didn’t have plans to support Audiences, but it's been a feature requested by our mutual customers.

> And you're looking to ensure that the Audience is created on your platform before any user-tracked activity data is sent to it. Is that correct?

Yes, that's right, **only if the user chooses to**. Otherwise, they can simply use an existing audience by providing the `audience_id` field. Both options are optional.

> If so, are you expecting customers to fill in both the audience_id field and the inputFields.name field with the same value?

No, that’s not the case. If the user doesn’t have an existing audience and wants to create one when saving the mapping, they can do so by providing a `name` via `retlOnMappingSave`. If the `audience_id` field is set and a new audience is created on save, the existing audience selected by `audience_id` will take precedence.

At least that’s how I understand the behaviour of `retlOnMappingSave`. Please let me know if I’m on the wrong track!

xitonix:
Hi @joe-ayoub-segment,

Thank you so much for the detailed review — I really appreciate the time you took to go through everything. I’ve gone ahead and addressed/responded to your comments.

Regarding the 1:1, mornings on your end would work best for me as your 9 AM is 6 PM here in Sydney. Unfortunately, I won’t be able to make it this evening, but I’m available either Monday or Tuesday morning UK time. Please feel free to schedule a meeting for whichever of those days suits you best.

Thanks again for the review — looking forward to catching up!

Best regards,
Alex

packages/destination-actions/src/destinations/ortto/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  },
+  extendRequest({ settings }) {
+    if (process?.env?.ORTTO_API_KEY) {
+      settings.api_key = process?.env?.ORTTO_API_KEY
+    }
+    return {
+      headers: {
+        Authorization: `Bearer ${settings.api_key}`
+      },
+      timeout: Math.max(30_000, DEFAULT_REQUEST_TIMEOUT)
```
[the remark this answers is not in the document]
joe-ayoub-segment: Thanks @xitonix I think it would be a good idea to not use the timeout for the moment. If needed we can re-apply it.

packages/destination-actions/src/destinations/ortto/common-fields.ts
joe-ayoub-segment: OK cool - we can focus on this when we meet in person next week.

packages/destination-actions/src/destinations/ortto/utils.ts
joe-ayoub-segment: Let's leave as is then.

packages/destination-actions/src/destinations/ortto/ortto-client.ts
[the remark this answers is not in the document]
joe-ayoub-segment: Thanks for explaining. 
It would be ideal if an array of response codes were provided to indicate which requests were successful or not. But we could still partially implement multistatus responses for cases where some payloads fail validation (which prevents a request from being sent to your platform). 
Let's chat about this when we meet next week.

packages/destination-actions/src/destinations/ortto/trackActivity/index.ts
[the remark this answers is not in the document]
joe-ayoub-segment: Let's also chat about this when we meet!

joe-ayoub-segment:
Thanks @xitonix - I've scheduled a call at 9am my time next Monday. Feel free to add any other attendees.

joe-ayoub-segment:
hi @xitonix - would you be able to schedule a call with me using this Calendly link please?

https://calendly.com/joe_ayoub/integration-workshop

I look forward to catching up, 
Joe

xitonix:
Hey @joe-ayoub-segment,
As discussed, I’ve removed the Engage-related actions and added support for multi-status responses.
I’ll add Audience support in a separate PR/integration.
Could you please take another look when you have a moment? Thanks! 🙏🏻

PR #2815, opened 2025-03-20
fix pinterest-conversions destination item_price field type

ikraamg:
The Type of the item_price is a string according to the documentation and we get an error as the type is automatically converted to as number. The value field type is also corrected as a string.

https://developers.pinterest.com/docs/api/v5/events-create/

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/pinterest-conversions/pinterest-capi-custom-data.ts
+++ b/packages/destination-actions/src/destinations/pinterest-conversions/pinterest-capi-custom-data.ts
@@ -14,7 +14,7 @@ export const custom_data_field: InputField = {
       label: 'Value',
       description:
         'Total value of the event. E.g. if there are multiple items in a checkout event, value should be the total price of all items',
-      type: 'number'
+      type: 'string'
     },
     content_ids: {
       label: 'Content IDs',
@@ -35,7 +35,7 @@ export const custom_data_field: InputField = {
         },
         item_price: {
           label: 'Price',
-          type: 'number',
+          type: 'string',
           description: 'The price of the Item'
         },
         quantity: {


joe-ayoub-segment:
hi @ikraamg, 
Thanks for the PR. 
It looks like you want to change the type for 2 fields. I don't recommend doing this, however maybe we can change how these fields are processed before being forwarded to Pinterest - if there is an issue. 

**Looking at the first field:** custom_data_field.value 

If seems that this field already gets stringified before being sent to Pinterest. This happens [here](http://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/pinterest-conversions/reportConversionEvent/index.ts#L203). 

image

So I don't believe that there's an issue with the first field. 

**Looking at the second field:** custom_data_field.contents.$.item_price : 
It looks like this isn't converted into a string before being sent. The entire JSON object is [sent as is](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/pinterest-conversions/reportConversionEvent/index.ts#L205). 

I can see from [Pinterest's documentation](https://developers.pinterest.com/docs/api/v5/events-create/) that this value should be passed as a string. 

image

I think however that simply changing the type of this second field won't work. Instead we'll need to cast it into a string. 

So something like this will need to be done: 

Before: 

`contents: payload.custom_data?.contents,`

After: 

```
contents: payload.custom_data?.contents?.map(item => ({
  ...item,
  item_price: item.item_price != null ? String(item.item_price) : item.item_price
})),
```

I'm going to pass this PR on to the Engineer who wrote the Integration. 

Thank you for bringing this to our attention @ikraamg !

Kind regards,
Joe

ikraamg:
@joe-ayoub-segment Thanks for your feedback here!

>Looking at the first field: custom_data_field.value
> If seems that this field already gets stringified before being sent to Pinterest. This happens [here](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/pinterest-conversions/reportConversionEvent/index.ts#L203).

This is an issue because if the field is undefined then the the string with value 'undefined' is sent to pinterest and it shows a warning that this is invalid.

joe-ayoub-segment:
> @joe-ayoub-segment Thanks for your feedback here!
> 
> > Looking at the first field: custom_data_field.value
> > If seems that this field already gets stringified before being sent to Pinterest. This happens [here](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/pinterest-conversions/reportConversionEvent/index.ts#L203).
> 
> This is an issue because if the field is undefined then the the string with value 'undefined' is sent to pinterest and it shows a warning that this is invalid.

That's great context thank you @ikraamg. I assume that's happening because of [this line](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/pinterest-conversions/reportConversionEvent/index.ts#L203): 

`value: String(payload?.custom_data?.value),` 

We could fix it as follows: 

```
value: typeof payload?.custom_data?.value === 'number'
  ? String(payload.custom_data.value)
  : payload?.custom_data?.value,
```

Will you be updating the PR?

joe-ayoub-segment:
hi @brennan - I'd like to get your opinion on how to triage this ticket. 

The ticket was raised on the Pinterest Destination by a 3rd party Developer who does not work for Pinterest. 
The bug fix in the PR won't fix the issue which was raised - but I've outlined a possible fix to the issue in comments above. 

Should I or someone else from the team take responsibility to code, test and deploy a fix for the issue raised in this PR?

joe-ayoub-segment:
Raised an internal PR to fix this https://github.com/segmentio/action-destinations/pull/2972
cc @ikraamg

PR #2828, opened 2025-03-25
Eventbridge enhancements

mayur-pitale:
1. Throw an error when DetailType is longer then 128 char or not a string
Otherwise PutEvents silently fails. Event delivery shows 100% success, but EB discards the events.
Question: will Segment discard the entire batch if 1 or 2 events from the batch fail PayloadValidation? I.e. if 1 or 2 events have DetailType > 128 chars.
2. Add PutEvents response validation.
PutEvents returns 200 if HTTP request is successful, regardless of whether payload succeeded or failed validation. If payload fails validation, AWS adds an error message to the 200 response body. We need to fetch response body, check for errors and return an error or errors if an event or events failed AWS validation
See AWS docs [here](https://docs.aws.amazon.com/eventbridge/latest/APIReference/API_PutEvents.html#API_PutEvents_ResponseElements)

## Testing
- Added/Updated [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- Tested batching locally


varadarajan-tw (changes requested):
Left couple of minor comments that Id like to be addressed before shipping this change. Also, it'll be better if we split actions-core change into another PR and add some tests.
cc: @nick-Ag

varadarajan-tw (approved):

PR #2851, opened 2025-04-04
Update Rokt-Audiences Destination to support emailsha256 values

GoodmanBen:
Per customer request, we are adding the ability to specifically sync sha256-hashed email addresses as a Rokt audience list.

## Testing

  - Added test checking that a request with only sha256 email addresses pass validation
Ran successful request with a sha256 email value:
segment-test-email-sha256-working

Ran an e2e test with just an email address value as before and successfully sent the data to Rokt:
segment-test-email-working


packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/custom-audience-operations.ts
```diff
@@ -49,27 +52,51 @@ const getCustomAudienceOperations = (payload: Payload[], settings: Settings): Cu
...
-        list: list,
-        action: action,
-        emails: emails
+      let custom_audience_op: CustomAudienceOperation
+      if (is_hashed) {
+        custom_audience_op = {
+          accountId: settings.accountid,
+          list: list,
+          action: action,
+          emails: emptyList,
```
nick-Ag, line 87: nit: couldn't this just be `emails: []` (and on L96: `sha256s: []`)?

nick-Ag:
Hi @GoodmanBen - a couple things relating to the failing Github checks:
* Please run `yarn types` in your command line and commit the result - it should update a `generated-types.ts` file and will fix the "Validate" Check.
* Since you added a `email_sha256` as a conditionally required field it is triggering the 'Required Field Check'. This should be safe to ignore in this case since email_sha256 is only required when `email` is undefined, and since `email` was previously required then any valid/enabled customer mappings for this destination should still function as intended

packages/destination-actions/src/destinations/rokt-audiences/upsertCustomAudiences/custom-audience-operations.ts
GoodmanBen: Yep, change made.

GoodmanBen:
@nick-Ag thank you for the review. I have run yarn types, and I agree with your comment around `email_sha256`, it should be safe to ignore that warning for our use case.

nick-Ag:
> @nick-Ag thank you for the review. I have run `yarn types`, and I agree with your comment around `email_sha256`, it should be safe to ignore that warning for our use case.

Hey, looks like generating the types uncovered a build issue that was previously hidden. You'll be able to replicate the step locally by running `yarn build` Looks like fixing that would be the last thing before an approve

GoodmanBen:
Ah I see, thanks for flagging, added some more defensiveness and was able to run `yarn build` successfully.

GoodmanBen:
> Hi @GoodmanBen - we recently disabled those failing tests, if you merge main in they should start working on this branch. You are also free to ignore the synk failure as that's expected for partner PRs

Thanks @nick-Ag, sounds good just merged master back in

nick-Ag (approved):

PR #2856, opened 2025-04-07
[Recombee] Add internalAdditionalData field to actions

mstieranka:
This PR adds a new field, `internalAdditionalData`, to all applicable actions within the Recombee destination.

The goal is for us to be able to track the types of events that are sent using our destination, and being able to use this data in our analytics, as well as assist our customers with any issues they may be having.

We set this new field up with `unsafe_hidden` set to `true`, as end users should never be able to modify the values within this field. If this parameter doesn't prevent the user from changing the values within this object, or if there's a better way to obtain `$.type` and `$.event` within actions, please let us know.

Additionally, the original `additionalData` field (that should be present to all users) has had its `defaultObjectUI` set to `keyvalue:only`, as we want to discourage users from setting the value of this field to be an entire object.

## Testing

These changes were tested end-to-end to ensure the event data is sent in all applicable actions, and unit test snapshots were modified to account for the above-mentioned changes.

The new field is not required, meaning this is not a breaking change.

### DIFF
--- a/packages/destination-actions/src/destinations/recombee/addBookmark/index.ts
+++ b/packages/destination-actions/src/destinations/recombee/addBookmark/index.ts
@@ -46,12 +46,25 @@ const action: ActionDefinition<Settings, Payload> = {
   },
   perform: async (request, data) => {
     const client = new RecombeeApiClient(data.settings, request)
-    await client.send(new AddBookmark(data.payload))
+    await client.send(payloadToBookmark(data.payload))
   },
   performBatch: async (request, data) => {
     const client = new RecombeeApiClient(data.settings, request)
-    await client.send(new Batch(data.payload.map((payload) => new AddBookmark(payload))))
+    await client.send(new Batch(data.payload.map((payload) => payloadToBookmark(payload))))
   }
+}
+
+function payloadToBookmark(payload: Payload): AddBookmark {
+  return new AddBookmark({
+    userId: payload.userId,
+    itemId: payload.itemId,
+    timestamp: payload.timestamp,
+    recommId: payload.recommId,
+    additionalData: {
+      ...(payload.internalAdditionalData ?? {}),
+      ...(payload.additionalData ?? {})
+    }
+  })
 }
 
 export default action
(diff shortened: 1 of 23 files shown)


packages/destination-actions/src/destinations/recombee/addBookmark/index.ts
```diff
@@ -46,12 +46,25 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    userId: payload.userId,
+    itemId: payload.itemId,
+    timestamp: payload.timestamp,
+    recommId: payload.recommId,
+    additionalData: {
+      ...(payload.internalAdditionalData ?? {}),
+      ...(payload.additionalData ?? {})
+    }
+  })
+}
```
joe-ayoub-segment, line 68: Hi @mstieranka thanks for raising this PR. 
My main question regards the addition of these new wrapper functions. Why are they needed? 

For example why can the additional work of merging the new internalAdditionalData into the additionalData object not be done in the AddBookmark class?

joe-ayoub-segment (changes requested):

packages/destination-actions/src/destinations/recombee/addBookmark/index.ts
joe-ayoub-segment: Thanks @mstieranka - it might be neater to have the logic all in one place here if possible. Are you OK updating it please?

mstieranka:
Hi @joe-ayoub-segment , I refactored the implementation to be all within the class constructors, as you suggested.

I also added a few more PayloadValidationErrors, so that the error message is more descriptive than the one from the API. I generated a new instance of `Chance` in these tests to achieve this, since I saw it used in another destination. Hope that's okay.

joe-ayoub-segment:
Hi @mstieranka I'm still not following the PR properly. Maybe we could catch up on a call to walk through it? Here is my Calendly link: https://calendly.com/joe_ayoub

Can you schedule a call please?

joe-ayoub-segment:
Talked to @mstieranka who explained the changes. All makes sense. Tests included. 
@mstieranka to make a couple of minor updates, and then notify me when he is done. After that good to deploy.

mstieranka:
Hi @joe-ayoub-segment , I implemented the minor updates we talked about and we've also done an internal round of code review to make sure this is what we want to go out.

This PR is therefore ready for you to merge and release.

joe-ayoub-segment (approved):

mstieranka:
Hi @joe-ayoub-segment , just tested the change and everything seems good. Thanks.

PR #2858, opened 2025-04-07
add intercom_user_jwt

eredi93:
Hello 👋 ,

At Intercom we are in the process of rolling out a new way of securing the Intercom messenger by using JWT instead of IdV (we are currently in open beta).
This PR is adding the new field `intercom_user_jwt`.

## Testing

I added a test for the new filed

image


joe-ayoub-segment (changes requested):
hi @eredi93 thanks for the PR. 
It looks straightforward. 

Just to explain how I understand this change: You'd like to allow customers to include a User JSON Web Token in payloads being sent to Segment. The token will then be included in the requests made to Intercom's API.

As for the approach taken, the customer will need to update their tracking code to ensure that the new intercom_user_jwt field is populated. For some customers this might mean needing to get a developer involved. 

A way around this might be to write some code in the perform() function which automatically finds the JWT and includes it in the payload to Intercom (is the JWT in a cookie or local storage?). You could then simply add a new Field allowing the customer to specify if they want the JWT to be used. 

Note sure if the above is feasible, but thought I'd mention it as it might be a nicer customer experience. 

Please let me know how you'd like to proceed. Happy with either option. 

Kind regards,
Joe

joe-ayoub-segment:
hi again @eredi93 - looks like some CI checks are failing. 
Do you mind running this command and committing the updates to the PR please?

`yarn types`

eredi93:
@joe-ayoub-segment Thanks for looking at the PR.

your understanding of the change is correct. Unfortunately this is a field the customer will need to add themselves. The JWT ins't stored in any cookie and is used to boot the Intercom messenger (we use the JWT to log in the correct user). This JWT token needs to be generated by the customer with a secret they can in their Intercom workspace https://app.intercom.com/a/apps/_/settings/workspace/security?tab=messenger-security

This is a help centre article that explain how to use JWTs https://www.intercom.com/help/en/articles/10589769-authenticating-users-in-the-messenger-with-json-web-tokens-jwts 

We are in the process of deprecating Identity Verification. This field will behave the same way as we are currently asking customers to implement identity verification: https://www.intercom.com/help/en/articles/2047-install-intercom-with-segment#h_9600887271
The main difference here is that instead of sending to us a `user_hash` https://github.com/segmentio/action-destinations/blob/main/packages/browser-destinations/destinations/intercom/src/identifyUser/__tests__/index.test.ts#L34-L39 they would send a JWT via `intercom_user_jwt` (allowing the customer to protect more than the user_id, allowing expiry of the token etc.)

I also just ran `yarn types` and I pushed a commit with the updated types.

joe-ayoub-segment:
Thanks for explaining @eredi93. 
So this will essentially replace the user_hash field which is currently in use. 
I'm happy to deploy this PR as is.  
Will there be any follow up changes needed for this?

joe-ayoub-segment (approved):

eredi93:
Hi @joe-ayoub-segment ,

I mainly just copied what we have for IdentityVerification (user_hash), as this is an evolution of that I wanted to keep the same pattern making it relatively easy for customers to upgrade.

I'm happy to take a different approach if you think I should though.

eredi93:
@joe-ayoub-segment are there any blocker/open questions? or you think we can get this merged in?

joe-ayoub-segment:
hi @eredi93 - this PR has been deployed. Please confirm that you are happy with the change.

PR #2872, opened 2025-04-15
[TikTok] Add error when no audience found

maryamsharif:
Jira - https://segment.atlassian.net/browse/STRATCONN-5682

When an audience is not found it causes the error `Cannot read properties of undefined (reading 'audience_details'` making it hard to debug. This PR adds a more clear error message.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/tiktok-audiences/index.ts
+++ b/packages/destination-actions/src/destinations/tiktok-audiences/index.ts
@@ -139,7 +139,11 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
         throw new IntegrationError('Invalid response from get audience request', 'INVALID_RESPONSE', 400)
       }
 
-      const externalId = r.data['list'][0]['audience_details']['audience_id']
+      const externalId = r.data?.list?.[0]?.audience_details?.audience_id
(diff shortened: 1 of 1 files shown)


packages/destination-actions/src/destinations/tiktok-audiences/index.ts
```diff
@@ -139,7 +139,11 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
         throw new IntegrationError('Invalid response from get audience request', 'INVALID_RESPONSE', 400)
       }
 
-      const externalId = r.data['list'][0]['audience_details']['audience_id']
+      const externalId = r.data?.list?.[0]?.audience_details?.audience_id
```
marinhero, line 142: Rows with empty external ID, correct? The new UI flow should reduce these occurrences

marinhero (approved):

packages/destination-actions/src/destinations/tiktok-audiences/index.ts
maryamsharif: This is for if we created an external audience but it was then deleted or something changed where the saved external ID is no longer available in TikTok.

PR #2873, opened 2025-04-17
Added Functionaility for Custom Attributes and Subscribers API from Attentive

duynguyen100:
Added functionaility for action destinations to support mappings of Custom Attribute and Subscribers API, Unit tested all passed! 

## Testing

@brennan

### DIFF
--- a/packages/destination-actions/src/destinations/attentive/customattributes/index.ts
+++ b/packages/destination-actions/src/destinations/attentive/customattributes/index.ts
@@ -1,0 +1,128 @@
+import { ActionDefinition, PayloadValidationError } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { CustomAttribute, User } from './types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Custom Attributes',
+  description: 'Send custom attributes to Attentive.',
+  defaultSubscription: 'type = "identify"',
+  fields: {
+    userIdentifiers: {
+      label: 'User Identifiers',
+      description:
+        'At least one identifier is required. Custom identifiers can be added as additional key:value pairs.',
+      type: 'object',
+      required: true,
+      additionalProperties: true,
+      defaultObjectUI: 'keyvalue:only',
+      properties: {
+        phone: {
+          label: 'Phone',
+          description: "The user's phone number in E.164 format.",
+          type: 'string',
+          required: false
+        },
+        email: {
+          label: 'Email',
+          description: "The user's email address.",
+          type: 'string',
+          format: 'email',
+          required: false
+        },
+        clientUserId: {
+          label: 'Client User ID',
+          description: 'A primary ID for a user. Should be a UUID.',
+          type: 'string',
+          format: 'uuid',
+          required: false
+        }
+      },
+      default: {
+        phone: {
+          '@if': {
+            exists: { '@path': '$.context.traits.phone' },
+            then: { '@path': '$.context.traits.phone' },
+            else: { '@path': '$.properties.phone' }
+          }
+        },
+        email: {
+          '@if': {
+            exists: { '@path': '$.context.traits.email' },
+            then: { '@path': '$.context.traits.email' },
+            else: { '@path': '$.properties.email' }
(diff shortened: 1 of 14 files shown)


packages/destination-actions/src/destinations/attentive/subscribers/index.ts
```diff
@@ -0,0 +1,88 @@
...
+          type: 'string',
+          required: true
+        },
+        country: {
+          label: 'Country',
+          type: 'string',
+          required: true
+        }
+      }
+    }
```
joe-ayoub-segment, line 57: Segment actually provides the locale of a client side device in most cases. It doesn't do this for server side SKDs though. 

The locale data is usually located here in a Segment payload: 

`$.context.locale`

The value usually looks like this: 'en-US', where the first 2 characters are the language and the last 2 characters are the country. 

See https://segment.com/docs/connections/spec/common/#context 

You could go with a single locale field that collects this value. You could then parse it and use it as you like in the perform() function. 

For example: 

```
locale: {
      label: 'Locale',
      description: 'User locale (language and country)',
      type: 'string',
      required:false,
      default: {'@path': '$.context.locale'}
}
```

packages/destination-actions/src/destinations/attentive/customattributes/index.ts
```diff
@@ -0,0 +1,128 @@
+import { ActionDefinition, PayloadValidationError } from '@segment/actions-core'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+import { CustomAttribute, User } from './types'
+
+const action: ActionDefinition<Settings, Payload> = {
+  title: 'Custom Attributes',
+  description: 'Send custom attributes to Attentive.',
```
joe-ayoub-segment, line 8:
```suggestion
  description: 'Send Custom Attributes to Attentive.',
```

packages/destination-actions/src/destinations/attentive/index.ts
```diff
@@ -35,16 +37,34 @@ const destination: DestinationDefinition<Settings> = {
...
+    {
+      name: 'Track Custom Attributes',
+      subscribe: 'type = "identify"', // Trigger custom attributes on identify events
+      partnerAction: 'customAttributes',
+      mapping: defaultValues(customAttributes.fields),
+      type: 'automatic'
+    },
+    {
+      name: 'Subscribe User',
+      subscribe: 'type = "identify" and traits.phone != null', // Trigger on identify with `subscribed: true`
```
joe-ayoub-segment, line 64: hi @duynguyen100 is the logic here correct?

joe-ayoub-segment (changes requested):
Hi @duynguyen100 thanks for the PR. 

I think I understand what you are intending to do here. 
One Action is to append attributes to a user. Another Action is to subscribe a user. 
Both of these Actions are invoked with an identify() call. 

Are 2 different Actions really needed for this? I'm wondering if a single Action could collect all the user attributes, and would also have a field where the customer can indicate if the user should be subscribed. You could then make 2 calls to your API to update the attributes and to subscribe the user?

Also do these API calls upsert a user?
For example if a user doesn't already exist, will the customAttributes Action create the user before adding the custom attributes to the user? Same question for the subscribers API call.

I'd be happy to discuss these design questions on a call if you can't make the decision yourself. 

Please also have a look at the other comments I left inline. 

Best regards,
Joe

joe-ayoub-segment:
hi again @duynguyen100 tried to yarn build the branch and got this error: 

      src/destinations/attentive/customAttributes/index.ts(4,10): error TS2724: '"./types"' has no exported member named 'CustomAttribute'. Did you mean 'CustomAttributes'?
      src/destinations/attentive/subscribers/index.ts(4,35): error TS2307: Cannot find module './types' or its corresponding type declarations.
error Command failed with exit code 1.

Can you address this also please?

joe-ayoub-segment:
hi @duynguyen100 can you schedule a call with me here please? https://calendly.com/joe_ayoub

joe-ayoub-segment:
Closing as I opened an [internal PR](https://github.com/segmentio/action-destinations/pull/2922/) for this

PR #2884, opened 2025-04-25
Add initial status to drip destination

stevedrip:
The default behavior for `status` is a footgun for users. Prefer the more forgiving `initial_status` instead. When combined with `status_updated_at` this will updated a person's status as if the change happened at that timestamp. When `status_updated_at` is missing, any existing status will take priority.

Previously, we assumed that the user would always want to update the status when identifying a person. We defaulted to unsubscribing, and at the current timestamp. This has proven to be a poor assumption in practice. With this new behavior, if the `status_updated_at` was intended to be passed but missing, we won't modify the person's status in drip.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/drip/identify/index.ts
+++ b/packages/destination-actions/src/destinations/drip/identify/index.ts
@@ -15,6 +15,7 @@ const person = (payload: Payload) => {
     email: payload.email,
     ip_address: payload.ip,
     phone: payload.phone,
+    initial_status: payload.initial_status,
     status: payload.status,
     status_updated_at: payload.status_updated_at,
     tags: payload.tags?.split(',').map((tag) => tag.trim()),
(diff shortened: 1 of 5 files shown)


packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -57,19 +58,26 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'string',
       default: { '@path': '$.traits.phone' }
     },
-    status: {
-      description: "The person's subscription status.",
-      label: 'Status',
+    initial_status: {
+      description: "The person's subscription status if newly identified.",
```
stevedrip, line 62: I feel like a better description is possibly here. I don't particularly like the one in our Drip developer guide though.

Perhaps there is wording that is better than what we have in either place.

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -57,19 +58,26 @@ const action: ActionDefinition<Settings, Payload> = {
...
         }
       }
     },
+    status: {
+      description: "The person's subscription status. Overrides initial_status.",
+      label: 'Status',
+      required: false,
+      type: 'string',
+      default: { '@path': '$.traits.status' }
+    },
```
jacobpmeyer, line 80: Did you decide to keep status around? If so should we add test coverage for it?

stevedrip: I did! Except for one line showing the new defaults, all the prior test coverage remains.

joe-ayoub-segment:
to @jacobpmeyer and Developer Rotation person - I'll take this PR so no need to follow up on it. 

hi @stevedrip - thanks for the PR. 
The code looks good. There is a failing from another Integration though, so can you please pull in the latest code from Segment's main branch please?

Thank you!
Joe

joe-ayoub-segment (approved):
Reviewed - looks good.

joe-ayoub-segment:
This PR is ready to be deployed.

varadarajan-tw:
@stevedrip - This PR has been deployed

PR #2891, opened 2025-04-28
Require `filename_prefix` for s3 destination

JasonSooter:
## Before this PR
- Mapping field `filename_prefix` NOT marked as required

## After this PR
- Mapping field `filename_prefix` is now a required field

## Reason that prompted this change
- If no filename prefix is supplied, files will land in s3 with only a date

## Testing

- [x] Updated [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) to accommodate the required `filename_prefix field

### DIFF
--- a/packages/destination-actions/package.json
+++ b/packages/destination-actions/package.json
@@ -43,6 +43,7 @@
     "@amplitude/ua-parser-js": "^0.7.25",
     "@aws-sdk/client-eventbridge": "^3.741.0",
     "@aws-sdk/client-s3": "^3.600.0",
+    "@aws-sdk/client-sts": "^3.797.0",
     "@bufbuild/protobuf": "^2.2.3",
     "@segment/a1-notation": "^2.1.4",
     "@segment/actions-core": "^3.151.0",
--- a/packages/destination-actions/src/destinations/s3/fields.ts
+++ b/packages/destination-actions/src/destinations/s3/fields.ts
@@ -201,7 +201,7 @@ export const commonFields: ActionDefinition<Settings>['fields'] = {
     label: 'Filename prefix',
     description: `Prefix to append to the name of the uploaded file.`,
     type: 'string',
-    required: false
+    required: true
   },
   delimiter: {
     label: 'Delimeter',
--- a/packages/destination-actions/src/destinations/s3/functions.ts
+++ b/packages/destination-actions/src/destinations/s3/functions.ts
@@ -29,7 +29,7 @@ export async function send(payloads: Payload[], settings: Settings, rawMapping: 
   await s3Client.uploadS3(
     settings,
     fileContent,
-    payloads[0]?.filename_prefix ?? '',
+    payloads[0]?.filename_prefix,
     payloads[0]?.s3_aws_folder_name ?? '',
     payloads[0]?.file_extension
   )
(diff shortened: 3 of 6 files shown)


joe-ayoub-segment:
hi @JasonSooter thanks for the PR. 

We need to be super cautious when converting a non-required field to be a required field. It will break the Integration for any customer who is already using the Integration. Were you aware of this?

If this is still something you would like to do you will need to liaise with all customers who already use the Integration to let them know that data collection for their Destination will stop as soon as this change is deployed. They would then need to go in and update the Mapping field settings and hit the save button. 

Please confirm if this is still something you'd like to go ahead with or if you would like to explore a different way to address the issue.

JasonSooter:
Hi @joe-ayoub-segment,
Yes, on the same page regarding the care required.

Unsure if you were aware of the conversation in [this Slack conversation](https://twilio.slack.com/archives/C07EZRLG1RA/p1745540559343449), but I was under the assumption that only Fidelity was using it currently (as it was built for them) and that `file_prefix` was in use for all of their mappings.

However, both of those details turned out to be false after digging in more deeply.

- [17 workspaces use the destination at this point](https://docs.google.com/spreadsheets/d/1sLqWF29YEryn3zniC6WAqLZz6tNRXSjtIONsnSfqwgc/edit?gid=0#gid=0&range=A2:A18) using the Snowflake query below
- Fidelity does indeed have some mappings where `file_prefix` is not filled in. In that particular case, they're using `AWS Subfolder Name` to partition files exported to S3 from an Audience

The right move is potentially to require at least one of the fields to exist, but unsure how to do that, and it expands beyond the scope given the focus to release as Public Beta.

The original prompting for my focus was JPMC onboarding with the destination. With neither field required, you sync a bunch of files with only the timestamp as a filename. Not ideal for the 40 Million member audience we synced (400 files). But we'll just have to caution them to watch out for this gotcha

Given this situation, I'm closing the PR.

cc @brennan as this is off the table for Public Beta release

image

```sql
"SELECT
    workspace_slug,
    workspace_id
FROM
    ""MAIN_PROD"".""CITRUS"".""DESTINATIONS_MASTER""
WHERE
    destination_slug = 'actions-s3'
    AND DESTINATION_ENABLED = TRUE
    AND DESTINATION_VOLUME IS NOT NULL
group by
    workspace_slug,workspace_id;"
```

joe-ayoub-segment:
hi @JasonSooter - we can actually add a conditional required attribute to the fields in question if you like. Happy to chat in person if needed.

JasonSooter:
Hi @joe-ayoub-segment,

Happy to discuss. This does indeed sound like a good route forward

I sent you a Slack DM (but ack that you may not be feeling well). Just let me know when is a good time for you where our schedules overlap.

Also, meanwhile, if you're aware of a good example where this `conditional required attribute` is implemented that would illustrate what needs to be done, that might be enough for me to go on.

joe-ayoub-segment:
Shared a video via Slack with @JasonSooter explaining how to use conditional required values in Fields. This might be a way to fix this situation in a simpler and more reliable way.

joe-ayoub-segment:
Hey @JasonSooter ,
I seem to have lost the link to the Slack convo where we're discussing this ticket.  
Did you take a look at the video I sent? Did it make sense?

JasonSooter:
Hey @joe-ayoub-segment
I did watch the video ([slack thread](https://twilio.slack.com/archives/C07EZRLG1RA/p1746465153673389?thread_ts=1745540559.343449&cid=C07EZRLG1RA)). And indeed, it does seem like a solid solution.

I was curious though, instead of it potentially causing events to fail, do you see a way that it would just prevent saving the mapping if one or the other isn't filled in?

JasonSooter:
Thanks for confirmation @joe-ayoub-segment.

Given that unfortunate case, I don't see it worth trying to solve if we only catch it after the audience sync has occurred because you still end up with all the downsides.

PR #2902, opened 2025-05-07
Explicitely handle CreateAudienceErrors for Facebook Custom Audiences

marinhero:
In the logs we are seeing `500 - Internal Server Errors` coming from the CreateExternalAudience method of this destination. Upon further analysis, we found that the V3 handler in the Integrations Service wraps the error when it comes without an explicitly set status code:

```
 try {
      const intResponse = await invokeIntegration(instance, method, payload)
      result.success = {
        status: intResponse.status,
        body: intResponse.body
      }
    } catch (e) {
      result.err = e

      // Absence of a status or http response code means this error was internal
      const isInternal = !e.status && !e.code

      if (isInternal) {
        // Log unexpected errors from the integration in sentry
        const summary = summarize(this.integrationSlug, result.err, result.requests)
        summary.version = VERSION_SHA
        sentry.captureIntegrationException(
          result.err,
          integrationCreationName,
          result.body,
          result.context.projectId,
          summary
        )
        result.err.code = 'INTERNAL'
        result.err.status = 500
      } else if (!result.err.status) {
        // If no status is provided, go ahead and give a default status
        result.err.status = 400
      }
```

We want to keep the original status code and not label these errors as `INTERNAL` so EAMS can handle them appropriately. And thus, wrapping the `request` call in a try/catch block will guarantee that we can handle the error accordingly. Moreso, if the error looks like this:

```
 {"error":{"message":"Permissions error","type":"OAuthException","code":200,"error_subcode":1870050,"is_transient":false,"error_user_title":"Business Account Needed to Create\/Edit This Audience","error_user_msg":"To create or edit a Custom Audience made from a customer list, your admin needs to add this ad account to a business.","fbtrace_id":"x"}}
```

EAMS will react to it and render a user friendly error message.

## Testing

Tested completed successfully via EAMS/MONOSERVICE/ACTIONS combo. I ran  EAMS to form the request that centrifuge creates, then monoservice to recieve it and subsequently edited the actions package in the monoservice with this code to test it.

Screenshot 2025-05-07 at 4 42 57 PM

### DIFF
--- a/packages/destination-actions/src/destinations/facebook-custom-audiences/index.ts
+++ b/packages/destination-actions/src/destinations/facebook-custom-audiences/index.ts
@@ -74,14 +74,27 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
           body: new URLSearchParams(payload)
         })
       } catch (err) {
-        let status = err.status || err.code
+        let message = err.response?.content || err.message
+        let userTitle: string | undefined, userMsg: string | undefined
 
-        if (!status && err.response && err.response.status) {
-          status = err.response.status
+        if (typeof message === 'string') {
+          try {
+            const parsed = JSON.parse(message)
+
+            if (parsed?.error) {
+              userTitle = parsed.error.error_user_title
+              userMsg = parsed.error.error_user_msg || parsed.error.error_user_message
+            }
+          } catch (e) {
+            // No-Op. Add the error message to the message variable.
+          }
         }
 
-        const message = err.response?.content || err.message
-        throw new IntegrationError(message, 'CREATE_AUDIENCE_FAILED', status || 400)
+        if (userTitle || userMsg) {
+          message = `${userTitle ? userTitle + ': ' : ''}${userMsg || ''}`.trim()
+        }
+
+        throw new IntegrationError(String(message), 'CREATE_AUDIENCE_FAILED', 400)
       }
 
       const r = await response.json()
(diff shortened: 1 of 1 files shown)


packages/destination-actions/src/destinations/facebook-custom-audiences/index.ts
```diff
@@ -49,29 +49,37 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
...
+          body: new URLSearchParams(payload)
+        })
+      } catch (err) {
+        let status = err.status || err.code
+
+        if (!status && err.response && err.response.status) {
+          status = err.response.status
+        }
+
+        let message = err.response?.content || err.message
```
maryamsharif, line 80:
```suggestion
        const message = err.response?.content || err.message
```

packages/destination-actions/src/destinations/facebook-custom-audiences/index.ts
```diff
@@ -49,29 +49,37 @@ const destination: AudienceDestinationDefinition<Settings, AudienceSettings> = {
...
+      try {
+        response = await request(createAudienceUrl, {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/x-www-form-urlencoded'
+          },
+          body: new URLSearchParams(payload)
+        })
+      } catch (err) {
+        let status = err.status || err.code
```
maryamsharif, line 74: If we already know the error can be found in `error_user_msg`, if we explicitly return that here would it show up in the UI?

maryamsharif (approved):

nick-Ag (approved):

rvadera12 (approved):

PR #2905, opened 2025-05-08
Introduces Liquid template directives

nick-Ag:
This PR introduces a new LiquidJS template directive to mapping-kit. 
The design doc for this new feature can be found here: https://docs.google.com/document/d/1hX4J__JWfBWjq3IO3jiDlX3ilEQ_a9dYS_vDtSdl3FM/edit?tab=t.0#heading=h.6686ryyf0sqt 

## Testing

### Local Testing:
Locally tested against the Salesforce destination:

Screenshot 2025-05-14 at 3 33 22 PM

Screenshot 2025-05-14 at 3 35 58 PM

### Stage Testing:
This change was deployed to staging as a sanity check - no further testing was done as this is not currently accessible from the UI


packages/core/src/mapping-kit/liquid-directive.ts
```diff
@@ -0,0 +1,78 @@
+import { Liquid } from 'liquidjs'
+
+const liquidEngine = new Liquid({
+  renderLimit: 500, // 500 ms
+  parseLimit: 1e8, // 100M characters,
```
pooyaj, line 5: this should be 1000, right?

packages/core/src/mapping-kit/value-keys.ts
```diff
@@ -309,7 +325,8 @@ export function getFieldValueKeys(value: FieldValue): string[] {
           ...getRawKeys(input['@transform'].apply),
           ...getRawKeys(input['@transform'].mapping)
         ],
-        '@excludeWhenNull': (_: ExcludeWhenNullDirective) => ['']
+        '@excludeWhenNull': (_: ExcludeWhenNullDirective) => [''],
+        '@liquid': (input: LiquidDirective) => [input['@liquid']]
```
pooyaj, line 329: we may have to extract all the variables that are references within the liquid template here. Should be doable with some regex-ing.

pooyaj (dismissed):
Couple of comments, otherwise looks great 🔥🔥

packages/core/src/mapping-kit/liquid-directive.ts
nick-Ag: yes!

pooyaj (approved):
🚢🚢🚢

PR #2932, opened 2025-05-21
Add batch_size to identify fields in Drip (Actions)

jacobpmeyer:
Adding `batch_size` to the `identify` fields

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/drip/identify/index.ts
+++ b/packages/destination-actions/src/destinations/drip/identify/index.ts
@@ -118,7 +118,7 @@ const action: ActionDefinition<Settings, Payload> = {
       json: { subscribers: [person(payload)] }
     })
   },
-  performBatch: async (request, { settings, payload }) => {
+  performBatch: (request, { settings, payload }) => {
     const subscribers = payload.map(person)
 
     return request(`https://api.getdrip.com/v2/${settings.accountId}/subscribers/batches`, {
--- a/packages/destination-actions/src/destinations/drip/identify/index.ts
+++ b/packages/destination-actions/src/destinations/drip/identify/index.ts
@@ -107,7 +107,13 @@ const action: ActionDefinition<Settings, Payload> = {
       type: 'integer',
       required: false,
       unsafe_hidden: true,
-      default: 1000,
+      default: {
+        '@if': {
+          exists: { '@path': '$.traits.batch_size' },
+          then: { '@path': '$.traits.batch_size' },
+          else: 1000
+        }
+      },
       minimum: 1,
       maximum: 1000
     }
--- a/packages/destination-actions/src/destinations/drip/identify/index.ts
+++ b/packages/destination-actions/src/destinations/drip/identify/index.ts
@@ -98,6 +98,16 @@ const action: ActionDefinition<Settings, Payload> = {
       required: false,
       type: 'string',
       default: { '@path': '$.context.timezone' }
+    },
+    batch_size: {
+      label: 'Batch Size',
+      description: 'Maximum number of events to include in each batch. Actual batch sizes may be lower.',
+      type: 'integer',
+      required: false,
+      unsafe_hidden: true,
+      default: 100,
+      minimum: 1,
+      maximum: 1000
     }
   },
   perform: (request, { settings, payload }) => {
--- a/packages/destination-actions/src/destinations/drip/identify/index.ts
+++ b/packages/destination-actions/src/destinations/drip/identify/index.ts
@@ -1,8 +1,6 @@
 import type { ActionDefinition } from '@segment/actions-core'
 import type { Settings } from '../generated-types'
 import type { Payload } from './generated-types'
-import { MultiStatusResponse } from '@segment/actions-core'
-import type { JSONLikeObject } from '@segment/actions-core'
 
 const person = (payload: Payload) => {
   return {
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -98,6 +100,16 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    },
+    batch_size: {
+      label: 'Batch Size',
+      description: 'Maximum number of events to include in each batch. Actual batch sizes may be lower.',
+      type: 'integer',
+      required: false,
+      unsafe_hidden: true,
+      default: 1000,
+      minimum: 1,
+      maximum: 1000
```
jacobpmeyer, line 112: I looked around for the correct `type` here and I saw several implementations used `'number'`. I was having an issue updating snapshots with that type though. It's still set to `'number'` in the generated types, but I managed to get the tests passing by changing the type here to `'integer'`

Let me know if that is incorrect.

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -98,6 +100,22 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      type: 'integer',
+      required: false,
+      unsafe_hidden: true,
+      default: {
+        '@if': {
+          exists: { '@path': '$.traits.batch_size' },
+          then: { '@path': '$.traits.batch_size' },
+          else: 1000
+        }
+      },
```
joe-ayoub-segment, line 116:
```suggestion
      default: 1000,
```

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -106,12 +124,25 @@ const action: ActionDefinition<Settings, Payload> = {
...
     })
   },
-  performBatch: (request, { settings, payload }) => {
-    const subs = payload.map(person)
-    return request(`https://api.getdrip.com/v2/${settings.accountId}/subscribers/batches`, {
-      method: 'POST',
-      json: { batches: [{ subscribers: subs }] }
-    })
+  performBatch: async (request, { settings, payload }) => {
+    const batchSize = payload[0]?.batch_size ?? 1000
```
joe-ayoub-segment, line 128:
```suggestion
    const batchSize = payload.length
```

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -98,6 +100,22 @@ const action: ActionDefinition<Settings, Payload> = {
       required: false,
       type: 'string',
       default: { '@path': '$.context.timezone' }
+    },
+    batch_size: {
+      label: 'Batch Size',
```
joe-ayoub-segment, line 105: Please add this line - this should prevent the user from applying a mapping from a payload value: 

`disabledInputMethods: ['variable', 'function', 'enrichment']`

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -106,12 +124,25 @@ const action: ActionDefinition<Settings, Payload> = {
...
+  performBatch: async (request, { settings, payload }) => {
+    const batchSize = payload[0]?.batch_size ?? 1000
+    const multiStatusResponse = new MultiStatusResponse()
+    for (let i = 0; i < payload.length; i += batchSize) {
+      const chunk = payload.slice(i, i + batchSize)
+      const subs = chunk.map(person)
+      const response = await request(`https://api.getdrip.com/v2/${settings.accountId}/subscribers/batches`, {
+        method: 'POST',
+        json: { batches: [{ subscribers: subs }] }
+      })
```
joe-ayoub-segment, line 136: Can you define a response type you expect back from the request?

`const response = await request<BatchSubscribeResp>(`https://api.getdrip.com/v2/${settings.accountId}/subscribers/batches``

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -106,12 +124,25 @@ const action: ActionDefinition<Settings, Payload> = {
...
+      chunk.forEach((_, index) => {
+        multiStatusResponse.pushSuccessResponse({
+          status: response.status,
+          sent: subs[index],
+          body: response.data as JSONLikeObject
+        })
+      })
```
joe-ayoub-segment, line 143: are there any scenarios where a failure could be returned for some of the items in the batch?

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -106,12 +124,25 @@ const action: ActionDefinition<Settings, Payload> = {
...
+    }
```
joe-ayoub-segment, line 144: hi @jacobpmeyer 
Segment's pipeline will ensure that the size of the `payload` array is already chunked to `batch_size`. There's no need for you to do this in code. 

Also, it seems that you are not inspecting the response values from your requests. Shouldn't your responses tell you if each item in the batch request was good or not?

Finally, if any items are not successful, you should update the multiStatusResponse to indicate which items in the batch failed.

Take a look at the multistatus docs here: https://github.com/segmentio/action-destinations/blob/main/docs/multistatus.md 

Also there are a few implementations in the repo.

stevedrip: @jacobpmeyer - perhaps we move multistatus out of this PR to a follow-up, and evaluate its merit? I don't think our usage of multistatus will be super natural because we do less validation on batches, and I'd prefer that it didn't block this PR in any case.

jacobpmeyer: hi @joe-ayoub-segment Thanks for the info regarding segment's handling of the chunking. That definitely simplifies implementation. 

I'd originally assumed that batch_size would be handled under the hood, but I wasn't seeing multiple requests in the return for my `curl` during local testing so I though maybe I needed to handle the logic for the extra requests. 

In any case, I went ahead and removed the `MultiStatusResponse` code based on @stevedrip's comment and after looking over the doc again I'm not sure we would've gotten much value out of it anyway with the current Drip API. So it's essentially just a light cleanup of the code that was already there.

Let me know if what's there now looks okay.

packages/destination-actions/src/destinations/drip/identify/index.ts
```diff
@@ -107,10 +119,11 @@ const action: ActionDefinition<Settings, Payload> = {
     })
   },
   performBatch: (request, { settings, payload }) => {
-    const subs = payload.map(person)
+    const subscribers = payload.map(person)
+
     return request(`https://api.getdrip.com/v2/${settings.accountId}/subscribers/batches`, {
       method: 'POST',
-      json: { batches: [{ subscribers: subs }] }
+      json: { batches: [{ subscribers }] }
```
stevedrip, line 126:
```suggestion
      json: { batches: [{ subscribers: subscribers }] }
```

joe-ayoub-segment:
Hi @jacobpmeyer   PR deployed. Please confirm that you are happy with the change.

PR #2938, opened 2025-05-23
[Ortto] - allow additional in traits field

joe-ayoub-segment:
Need to allow additional fields in the traits field.

## Testing
Not needed - in Private Beta


xitonix (approved):
Thanks for the fix @joe-ayoub-segment. Much appreciated. I also have some small cleanups which I will put in a new PR on Monday. Thanks again mate.

sayan-das-in (approved):

PR #2365, opened 2024-09-03
Per Google, GCLID, GBRAID and WBRAID can't be used at the same time.

seg-leonelsanches:
Per Google:

> "Considering that for iOS, `gclid` is sent if your user has given consent on iOS, and `gbraid` if the user has not given consent, it makes no sense to send Advanced Conversions email data requesting consent if `gbraid` is sent at the same time which implies no consent."

Our integration allows a customer to send all the three parameters together, resulting in cryptic errors in the destination.

## Testing

- [ ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [X] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [Segmenters] Tested in the staging environment

PR #2940, opened 2025-05-23
[Google Enhanced Conversions]- adding additional validation for gclid wbraid and gbraid

joe-ayoub-segment:
Rewrite of this [PR](https://github.com/segmentio/action-destinations/pull/2365) but with a different approach which should work for performBatch as well as individual events. 

Change in functionality as follows: 

- These 2 Actions are changed: `uploadClickConversion, uploadClickConversion2`
- Changes for both are the same. 
- Change 1: only one of `gclid, wbraid or gbraid` fields should be accepted. For perform() we throw an error if this condition not met. For performBatch() we remove additional fields in order of priority but we don't throw an error. 
- Change 2: `customVariables` object should not be included in payload if either `gbraid` or `wbraid` present. 

For Change 1, it would have been preferable to use field conditions to validate, however the field condition capability isn't flexible enough to do this, and are not compatible with unit tests. I've messaged Nick AG about this. If the field conditions capability is improved then this PR should be revisited. 

## Testing

Multiple unit tests updated and new tests added.

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/__tests__/uploadClickConversion.test.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/__tests__/uploadClickConversion.test.ts
@@ -697,9 +697,7 @@ describe('GoogleEnhancedConversions', () => {
         }
       })
 
-      expect(responses[0].options.body).toMatchInlineSnapshot(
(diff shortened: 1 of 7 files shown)


joe-ayoub-segment:
PR deployed

PR #2996, opened 2025-06-13
[Google Ads] Fixes Journeys use cases with the userList action

nick-Ag:
This PR updates the userList 'external_id_type' hook input field to no longer be required. Since that field already has a default this is an acceptable remediation for an issue this destination was having when connected to Journeys. Previously when connected to Journeys the RETL Mapping Save hook defined for this action would not be shown to users (as intended) but the required 'external_id_type' field would still be enforced, making it impossible to publish a Journey with this action. 
Now that field is not required, but we will still default to 'CONTACT_INFO' if a user doesn't fill it out. 

## Testing
Tested successfully in stage by connecting a Journey to Google Ads UserList action
Screenshot 2025-06-13 at 16 09 06
Screenshot 2025-06-13 at 16 34 59 (1)

### DIFF
--- a/packages/destination-actions/src/destinations/google-enhanced-conversions/userList/generated-types.ts
+++ b/packages/destination-actions/src/destinations/google-enhanced-conversions/userList/generated-types.ts
@@ -76,7 +76,7 @@ export interface RetlOnMappingSaveInputs {
   /**
    * Customer match upload key types.
    */
-  external_id_type: string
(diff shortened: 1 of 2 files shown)


joe-ayoub-segment (approved):
Already approved by @marinhero . LGTM.

joe-ayoub-segment:
Hi @nick-Ag PR deployed

PR #3052, opened 2025-07-07
[Amazon CAPI] Removing dataType from custom attributes and removing amazonClickId and ImpressionId fields

bansrav:
_A summary of your pull request, including the what change you're making and why._

1. Checked the Amazon CAPI code and discussed about custom attributes usage with our product manager. We don't use datatype from custom attributes. We only ingest name and value attributes so it's safe to remove datatype. Also, only brand, category, productId and attr1 - attr10 custom attributes are used for reporting, so added this in description.
2. Also removing amazonClickId and ImpressionId fields which are not needed for now.

## Testing

_Include any additional information about the testing you have completed to
ensure your changes behave as expected. For a speedy review, please check
any of the tasks you completed below during your testing._

- [ ] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality
- [ ] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing)
- [ ] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change.
- [ ] [Segmenters] Tested in the staging environment
- [ ] [Segmenters] [If applicable for this change] Tested for regression with Hadron.

PR #3053, opened 2025-07-08
[Amazon CAPI] - Changing how custom attribute field works

joe-ayoub-segment:
This is a breaking change for customers who have the Amazon CAPI Destination installed ( for private beta testing )

Custom Attributes field is being converted into a regular object field, defaulting to key:value UI display. 

Types for each attribute will now be inferred by Segment before being sent to Amazon. 
null and undefined - these will be dropped
number, boolean, object - these will be stringified and sent
string - these will be sent

Customers already using the Integration will need to reconfigure the Custom Attributes field.  

## Testing

Tested locally

### DIFF
--- a/packages/destination-actions/src/destinations/amazon-conversions-api/trackConversion/__tests__/utils.test.ts
+++ b/packages/destination-actions/src/destinations/amazon-conversions-api/trackConversion/__tests__/utils.test.ts
@@ -590,14 +590,71 @@ describe('trackConversion utils', () => {
           amznAdStorage: 'GRANTED',
           amznUserData: 'GRANTED'
         },
-        customAttributes: [
-          { name: 'color', dataType: 'STRING', value: 'blue' },
-          { name: 'size', dataType: 'STRING', value: 'medium' }
-        ],
+        customAttributes: {
+          color: 'blue',
+          size: 'medium',
+          age: 30,
+          stringNum: "9099", // should stay as string
+          stringBool: 'true', // should stay as string
+          stringNegBool: 'false', // should stay as string
+          isMember: true,
+          preferences: { newsletter: true, notifications: false }, // should stringify object
+          tags: ['premium', 'loyalty', 9099], // should stringify array
+          iAmNull: null, // should be ignored
+          iAmUndefined: undefined // should be ignored
+        },
         enable_batching: true
       }
 
       const result = prepareEventData(payload, settings)
+
+      const customAttributes = [
+        {
+          dataType: "STRING",
+          name: "color",
+          value: "blue"
+        },
+        {
+          dataType: "STRING",
+          name: "size",
+          value: "medium"
+        },
+        {
+          dataType: "NUMBER",
+          name: "age",
+          value: "30"
+        },
+        {
+          dataType: "STRING",
+          name: "stringNum",
+          value: "9099"
+        },
+        {
+          dataType: "STRING",
+          name: "stringBool",
+          value: "true"
+        },
+        {
+          dataType: "STRING",
+          name: "stringNegBool",
+          value: "false"
+        },
+        {
+          dataType: "BOOLEAN",
+          name: "isMember",
+          value: "true"
+        },
+        {
+          dataType: "STRING",
+          name: "preferences",
+          value: "{\"newsletter\":true,\"notifications\":false}"
+        },
+        {
+          dataType: "STRING",
+          name: "tags",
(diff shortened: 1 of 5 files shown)


packages/destination-actions/src/destinations/amazon-conversions-api/trackConversion/utils.ts
```diff
@@ -363,6 +366,28 @@ export function prepareEventData(payload: Payload, settings: Settings): EventDat
...
   }
 
+  // Process custom attributes
+  const customAttributeArray: CustomAttributeV1[] = []
+
+  Object.entries(customAttributes ?? {}).forEach(([key, value]) => {
+    if (value === undefined || value === null) return
+    let dataType: 'STRING' | 'NUMBER' | 'BOOLEAN' = 'STRING'
+
+    if (typeof value === 'number') {
```
bansrav, line 376: can we simplify this logic based on the below context. Just pass everything as string value. Ideally, the dataType should be hidden in our CAPI spec.

```
I checked the CAPI v2 code on our side and discussed about the usage of datatype with our product manager. In code, we never use datatype field. We only ingest name and value attributes from custom attributes. Our product manager also confirmed that we can safely remove the datatype. Also, only brand, category, productId and attr1 - attr10 custom attributes are used for reporting
```

packages/destination-actions/src/destinations/amazon-conversions-api/trackConversion/__tests__/utils.test.ts
```diff
@@ -590,23 +590,80 @@ describe('trackConversion utils', () => {
           amznAdStorage: 'GRANTED',
           amznUserData: 'GRANTED'
         },
-        customAttributes: [
-          { name: 'color', dataType: 'STRING', value: 'blue' },
-          { name: 'size', dataType: 'STRING', value: 'medium' }
-        ],
+        customAttributes: {
```
bansrav, line 593: this is not a real world use0case for CAPI. Only brand, category, productId and attr1 to attr10 custom attributes are supported and used for reporting therefore all numbers, booleans or attributes not falling into the this criteria are simply ignored.

packages/destination-actions/src/destinations/amazon-conversions-api/trackConversion/generated-types.ts
```diff
@@ -119,19 +119,8 @@ export interface Payload {
    * Custom attributes associated with the event to provide additional context.
    */
   customAttributes?: {
```
bansrav, line 121: can we add description here similar to what I have in here - https://github.com/segmentio/action-destinations/pull/3052/files

It's important to emphasize on the attributes being supported

`description: 'Custom attributes associated with the event to provide additional context. Only brand, category, productId and attr1 - attr10 custom attributes are used for reporting.'`

joe-ayoub-segment:
hi @bansrav I incorporated your comments. Please take a look. I also removed the 2 fields you removed in your [PR](https://github.com/segmentio/action-destinations/pull/3052).

packages/destination-actions/src/destinations/amazon-conversions-api/trackConversion/utils.ts
```diff
@@ -363,6 +366,24 @@ export function prepareEventData(payload: Payload, settings: Settings): EventDat
...
   }
 
+  // Process custom attributes
+  const customAttributeArray: CustomAttributeV1[] = []
+
+  Object.entries(customAttributes ?? {}).forEach(([key, value]) => {
+    if (value === undefined || value === null) return
+
+    if (!CustomAttributeV1Names.includes(key as CustomAttributeV1Name)) {
+      throw new PayloadValidationError(
```
bansrav, line 376: This won't work. We want to ingest all key/value pairs from clients as there will be use-cases coming later this year related to audience creation from these custom attributes for CAPI, but only brand, category, productId, attr1-attr-10 are used for reporting. So, here is what we should do

1. Display brand, category, productId and attr1-attr-10 explicitly in Segment UI under custom Attributes but also provide option to customer to add more key/value pairs (whatever key name they want) if they like. 
2. Send all those custom attributes to CAPI v2.

joe-ayoub-segment: Thanks for the feedback @bansrav - I've updated the PR as per my understanding of your requirements.

bansrav (approved):

itsarijitray (approved):

PR #3056, opened 2025-07-09
[MAIN] [STRATCONN] Added Session Plugin for Amplitude

AnkitSegment:

In Amplitude, I’ve updated the session plugin by setting hidden to false, allowing users to configure the session time and enable or disable allowSessionTracking. With this change, users can now also track session start and session end events.

## Testing

Screenshot 2025-07-08 at 10 48 52 PM

Screenshot 2025-07-09 at 10 28 42 AM

Screenshot 2025-07-09 at 10 28 52 AM


packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -32,14 +52,53 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
   title: 'Session Plugin',
   description: 'Generates a Session ID and attaches it to every Amplitude browser based event.',
   platform: 'web',
-  hidden: true,
```
varadarajan-tw, line 35: Should we add any documentation and ask our customers not to disable this?

joe-ayoub-segment:
I've asked some people with AJS knowledge to take a look at this PR.

packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -3,17 +3,38 @@ import { UniversalStorage } from '@segment/analytics-next'
...
 import type { Payload } from './generated-types'
+import { Analytics } from '@segment/analytics-next'
 
 function newSessionId(): number {
   return now()
 }
 
-function now(): number {
-  return new Date().getTime()
+function startSession(analytics: Analytics, eventName = 'session_started') {
```
joe-ayoub-segment, line 12: Hi @AnkitSegment , 
Shouldn't we use a default event name written in Capital Case, like all Segment events?
e.g. "Session Started"

packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -3,17 +3,38 @@ import { UniversalStorage } from '@segment/analytics-next'
...
+  analytics
+    .track(eventName, {})
+    .then(() => true)
+    .catch(() => true)
+}
+
+function endSession(analytics: Analytics, eventName = 'session_ended') {
```
joe-ayoub-segment, line 19: Same here. 
Shouldn't we use a default event name written in Capital Case, like all Segment events?
e.g. "Session Ended"

packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -32,14 +53,53 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
...
+      type: 'boolean',
+      default: false,
+      required: false,
+      description:
+        'Generate session start and session end track() events. These events will be sent to the Javascript Source and will be forwarded on to any connected Destinations.'
+    },
+    sessionStartEvent: {
+      label: 'Session Start Event',
+      type: 'string',
+      default: 'session_started',
```
joe-ayoub-segment, line 75: Should we use "Session Started" as the default?

packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -32,14 +53,53 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
...
+            operator: 'is',
+            value: true
+          }
+        ]
+      }
+    },
+    sessionEndEvent: {
+      label: 'Session End Event',
+      type: 'string',
+      default: 'session_ended',
```
joe-ayoub-segment, line 91: Should we use "Session Ended" as the default event name?

packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -32,14 +53,53 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
...
-  hidden: true,
   defaultSubscription: 'type = "track" or type = "identify" or type = "group" or type = "page" or type = "alias"',
   fields: {
     sessionLength: {
       label: 'Session Length',
       type: 'number',
       required: false,
       description: 'Time in milliseconds to be used before considering a session stale.'
+    },
+    allowSessionTracking: {
```
joe-ayoub-segment, line 64: Just semantics - but I think the name of this field and the description is a little misleading. 

We're already tracking sessions by adding the sessionId value into the payload. This happens automatically and this functionality should not change (AFAIK!). 

What we're doing is adding the ability to do the following: 

1. Specify if 'Session Started' and 'Session Ended' track() events should be triggered from the Segment Javascript library. 
2. We're allowing the customer to specify a custom session lifetime - defaulting to 30 minutes. 

Maybe the field could be called something like: 

triggerSessionEvents

And the description could be something like 

"if set to true, 'Session Started' and 'Session Ended' events will be triggered from the user's browser. These events will be forwarded to all connected Destinations.'

packages/browser-destinations/destinations/amplitude-plugins/src/sessionId/index.ts
```diff
@@ -32,14 +53,53 @@ const action: BrowserActionDefinition<Settings, {}, Payload> = {
   title: 'Session Plugin',
   description: 'Generates a Session ID and attaches it to every Amplitude browser based event.',
   platform: 'web',
-  hidden: true,
   defaultSubscription: 'type = "track" or type = "identify" or type = "group" or type = "page" or type = "alias"',
   fields: {
     sessionLength: {
```
joe-ayoub-segment, line 58: Perhaps we should add a 30 second default value here too. i.e on the sessionLength field itself. This will make it easier for the customer to understand the default behaviour. 

We could also update the description to be: 

We could also update the description of the field to indicate what the default value is.

joe-ayoub-segment (changes requested):

joe-ayoub-segment (approved):

PR #3082, opened 2025-07-15
[Dotdigital] - new actions

joe-ayoub-segment:
New Actions for Dotdigital Destination 

## Testing

Tested locally and not in use by customers.


joe-ayoub-segment:
PR deployed

PR #3085, opened 2025-07-16
[STRATCONN-6006] - CM360 handle 200 responses with errors

joe-ayoub-segment:
See https://twilio-engineering.atlassian.net/browse/STRATCONN-6006

CM360 platform can return 20X responses which can contain errors. Currently these types of responses are not handled, so the customer is not able to observe the error conditions and cannot therefor resolve them. 

This PR handles 20X responses which have errors and rethrows a 400.  

## Testing

Tested locally and unit tests added.


joe-ayoub-segment:
PR deployed

PR #3088, opened 2025-07-16
s3_aws_region marked required

nk1107:
This change marks the s3_aws_region field as required property
image (1)

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
+++ b/packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
@@ -32,7 +32,8 @@ const action: ActionDefinition<Settings, Payload> = {
     s3_aws_region: {
       label: 'AWS Region (S3 only)',
       description: 'Region where the S3 bucket is hosted.',
-      type: 'string'
+      type: 'string',
+      required: true
     },
     audience_key: {
       label: 'LiveRamp Audience Key',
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/liveramp-audiences/audienceEnteredS3/index.ts
```diff
@@ -32,7 +32,8 @@ const action: ActionDefinition<Settings, Payload> = {
     s3_aws_region: {
       label: 'AWS Region (S3 only)',
       description: 'Region where the S3 bucket is hosted.',
-      type: 'string'
+      type: 'string',
+      required: true
```
joe-ayoub-segment, line 36: hi @nk1107 I know this is only a draft PR, but just need to let you know that this will break the Integration for any customer who is already using it who isn't sending a value for this field. We need to be super careful about changing a field from being optional to required. Happy to discuss if needed.

sayan-das-in: Hi @joe-ayoub-segment , for LiveRamp this is a required field and if the data is missing it would anyways fail at the end of the pipeline where we lack visibility. The idea of this is to push those errors further up in the pipeline so that users know what is failing and why right at integrations.

sayan-das-in: The screenshot attached to this PR confirms there are no users who has this field set to empty.

sayan-das-in (approved):

joe-ayoub-segment:
Hi @nk1107 PR deployed

PR #3101, opened 2025-07-22
STRATCONN-6008 - [Twilio Messaging] - WIP tags support

joe-ayoub-segment:
WIP - add support to pass tags with a message. 

NOTE: still need to confirm with Twilio team to check that this is done correctly. 

## Testing
Unit test added, but full end to end testing still required.


joe-ayoub-segment:
PR deployed

PR #3104, opened 2025-07-24
[algolia-insights] add support for auth user token and product list viewed

jkaho:
Add support for the Algolia Insights API optional [`authenticatedUserToken`](https://www.algolia.com/doc/api-reference/api-methods/viewed-object-ids/#parameters) field.

Add support for the Segment [`Product List Viewed`](https://segment.com/docs/connections/spec/ecommerce/v2/#product-list-viewed) event.

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/algolia-insights/productListViewedEvents/index.ts
+++ b/packages/destination-actions/src/destinations/algolia-insights/productListViewedEvents/index.ts
@@ -46,15 +46,22 @@ export const productListViewedEvents: ActionDefinition<Settings, Payload> = {
     userToken: {
       type: 'string',
       required: true,
-      description: 'The ID associated with the user.',
+      description:
+        'The ID associated with the user. If a user is authenticated, this should be set to the same value as the Authenticated User Token',
       label: 'User Token',
       default: {
         '@if': {
-          exists: { '@path': '$.userId' },
-          then: { '@path': '$.userId' },
-          else: { '@path': '$.anonymousId' }
+          exists: { '@path': '$.anonymousId' },
+          then: { '@path': '$.anonymousId' },
+          else: { '@path': '$.userId' }
         }
       }
+    },
+    authenticatedUserToken: {
+      type: 'string',
+      description: 'The authenticated ID associated with the user.',
+      label: 'Authenticated User Token',
+      default: { '@path': '$.userId' }
     },
     timestamp: {
       type: 'string',
@@ -103,7 +110,8 @@ export const productListViewedEvents: ActionDefinition<Settings, Payload> = {
       queryID: data.payload.queryID,
       objectIDs: data.payload.objectIDs,
       timestamp: data.payload.timestamp ? new Date(data.payload.timestamp).valueOf() : undefined,
-      userToken: data.payload.userToken
+      userToken: data.payload.userToken,
+      authenticatedUserToken: data.payload.authenticatedUserToken
     }
     const insightPayload = { events: [insightEvent] }
 
--- a/packages/destination-actions/src/destinations/algolia-insights/productListViewedEvents/index.ts
+++ b/packages/destination-actions/src/destinations/algolia-insights/productListViewedEvents/index.ts
@@ -1,0 +1,124 @@
+import type { ActionDefinition, Preset } from '@segment/actions-core'
+import { defaultValues } from '@segment/actions-core'
+import { AlgoliaBehaviourURL, AlgoliaProductViewedEvent, AlgoliaEventType } from '../algolia-insight-api'
+import type { Settings } from '../generated-types'
+import type { Payload } from './generated-types'
+
+const segmentEventName = 'Product List Viewed'
+
+export const productListViewedEvents: ActionDefinition<Settings, Payload> = {
+  title: `${segmentEventName} Events`,
+  description:
+    'Product list viewed events act as a positive signal for associated record objects — the associated Product IDs. Query ID is optional and indicates that the view events are the result of a search query.',
+  fields: {
+    objectIDs: {
+      label: 'Product IDs',
+      description: 'Product IDs of the viewed items.',
+      type: 'string',
+      multiple: true,
+      required: true,
+      default: {
+        '@arrayPath': ['$.properties.products', { '@path': '$.product_id' }]
+      }
+    },
+    index: {
+      label: 'Index',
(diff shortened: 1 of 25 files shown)


joe-ayoub-segment:
Thanks for the PR @jkaho . I'll schedule time to review.

joe-ayoub-segment:
Hi @jkaho I had a quick scan. Did you mean to include the file at this location? packages/browser-destinations/destinations/algolia-plugins/src/algoliaInsights/generated-types.ts

joe-ayoub-segment:
Hi again @jkaho , I reviewed the PR. 

Regarding adding the new field `authenticatedUserToken`: 

- Looks like this field is to capture an identifier for a user when the user is known to the customer. Does this mean that the `userToken` field was always intended to capture an anonymous user's anonymous identifier? because it used to try and capture the main identifier and only fell back to the anonymous one. 
- How will this change affect customers who are already sending the Segment userId (which is the id for users who are known to the customer) to the `userToken` field? They won't be populating the `authenticatedUserToken` field (so it will be undefined).  
- Are you able to provide evidence of testing this please?

Regarding the new Action: 

- I don't think the default mapping used for `ObjectIds` field will work. Have you tested it locally?
- Are you able to provide evidence of testing this please?

Best regards,
Joe

joe-ayoub-segment:
hi again @jkaho can you please also run `yarn types` and commit the updated file please?

jkaho:
Hey @joe-ayoub-segment, appreciate the review!

* We added the [`authenticatedUserToken`](https://www.algolia.com/doc/guides/sending-events/concepts/usertoken/#persistent-user-token) field a couple of years ago now, before which the `userToken` field was intended to capture either the auth or anon token. The addition of the new auth field meant we changed the way we recommend users send us user identifiers, though this wasn't a breaking change. If users don't set the auth token field, we will take the value from the user token field (which is required). Again, this change was made a little while ago now and neglecting to add support for it to our Segment connector was an oversight on our part. Is there anything you would recommend in terms of documentation to mitigate potential confusion?
* I have tested the object IDs mapping via the Actions Tester, and it does seem to map the correct values as per the highlighted portions in the screenshot below. Have I misinterpreted something here? Though, I am noticing that the input badge doesn't show in the user configuration UI..

Screenshot 2025-08-12 at 10 32 37 am

joe-ayoub-segment:
> Hey @joe-ayoub-segment, appreciate the review!
> 
> * We added the [`authenticatedUserToken`](https://www.algolia.com/doc/guides/sending-events/concepts/usertoken/#persistent-user-token) field a couple of years ago now, before which the `userToken` field was intended to capture either the auth or anon token. The addition of the new auth field meant we changed the way we recommend users send us user identifiers, though this wasn't a breaking change. If users don't set the auth token field, we will take the value from the user token field (which is required). Again, this change was made a little while ago now and neglecting to add support for it to our Segment connector was an oversight on our part. Is there anything you would recommend in terms of documentation to mitigate potential confusion?
> * I have tested the object IDs mapping via the Actions Tester, and it does seem to map the correct values as per the highlighted portions in the screenshot below. Have I misinterpreted something here? Though, I am noticing that the input badge doesn't show in the user configuration UI..
> 
> Screenshot 2025-08-12 at 10 32 37 am

Hi @jkaho , that's correct, the 'badge' contains the mapping details the user gets to see in the UI. Our UI doesn't support this type of mapping, so while the mapping itself might work, it breaks the UI. 

I suggest changing the mapping to something like this

```
{
        '@path': '$.properties.products.product_ids'
}
```

packages/destination-actions/src/destinations/algolia-insights/productListViewedEvents/index.ts
```diff
@@ -0,0 +1,144 @@
...
+    products: {
+      label: 'Product Details',
+      description:
+        'The viewed products. Populates the ObjectIDs field in the Algolia Insights API. Each object must contain a product_id field.',
+      type: 'object',
+      defaultObjectUI: 'keyvalue',
+      multiple: true,
+      required: true,
+      properties: {
+        product_id: { label: 'product_id', type: 'string', required: true }
```
joe-ayoub-segment, line 23: Please add a description for product_id.

joe-ayoub-segment: If you want to prevent the customer from adding additional values into items in the product array, you can add this: 

  `additionalProperties: false,`

joe-ayoub-segment:
Thanks @jkaho this looks good. Only a nit issue left. Please see the comments. 
Best regards,
Joe

packages/destination-actions/src/destinations/algolia-insights/productListViewedEvents/index.ts
jkaho: Ah awesome, thanks for the feedback 🙂

joe-ayoub-segment (approved):

joe-ayoub-segment:
Thanks @jkaho I've labelled it for deployment. I'll notify you once it's out. It might be up to 7 days, as we usually only deploy this repo weekly and we missed this week's deployment.

jkaho:
@joe-ayoub-segment is there anything for me to action with the failing CI?

joe-ayoub-segment:
Hi @jkaho yes looks like snapshot tests are failing. Can you run this command and commit the changes please?

`yarn types `


FAIL  src/destinations/algolia-insights/productListViewedEvents/__tests__/snapshot.test.ts
    ● Testing snapshot for AlgoliaInsights's productListViewedEvents destination action: › required fields
  
      expect(received).toMatchSnapshot()
  
      Snapshot name: `Testing snapshot for AlgoliaInsights's productListViewedEvents destination action: required fields 1`
  
      - Snapshot  - 0
      + Received  + 5
  
      @@ -6,10 +6,15 @@
              "eventType": "view",
              "index": "lx7Lx",
              "objectIDs": Array [
                "lx7Lx",
              ],
      +       "products": Array [
      +         Object {
      +           "product_id": "lx7Lx",
      +         },
      +       ],
              "timestamp": 1674843786677,
              "userToken": "lx7Lx",
            },
          ],
        }

jkaho:
@joe-ayoub-segment if you're around, it would be awesome if we were able to get this out in the next deployment (which I believe is tomorrow?) 🙏

joe-ayoub-segment (approved):

PR #3106, opened 2025-07-24
Stratconn 6058/batch size

nk1107:
JIRA - https://twilio-engineering.atlassian.net/jira/software/c/projects/STRATCONN/boards/2317?assignee=712020%3Ae753d890-5047-4206-aaa4-9693b6ada1b0&assignee=712020%3A087e2fcd-dfdf-4343-a761-04f54a014b74&selectedIssue=STRATCONN-6058

Summary of the Ticket - Customers (Siemens) are running into issues with likely due to batch size being too large for Customer.io .

Change Requested - Add editable batch size input in the actions for Customer.io
This change adds `enable_batch` and `batch_size` properties in all the actions of actions-customerio
## Testing

With batch size greater than 100
image

With batch size less than 2
image

### DIFF
--- a/packages/destination-actions/src/destinations/customerio/utils.ts
+++ b/packages/destination-actions/src/destinations/customerio/utils.ts
@@ -116,8 +116,6 @@ type BasePayload = {
   primary?: Identifiers
   secondary?: Identifiers
   timestamp?: string | number
-  enable_batching?: boolean
-  batch_size?: number
 }
 
 export const buildPayload = <Payload extends BasePayload>({ action, type, payload }: RequestPayload<Payload>) => {


packages/destination-actions/src/destinations/customerio/customerio-properties.ts
```diff
@@ -0,0 +1,19 @@
...
+    default: true
+  },
+  batch_size: {
+    label: 'Batch Size',
+    description: 'Maximum number of events to include in each batch. Actual batch sizes may be lower.',
+    type: 'number',
+    required: false,
+    minimum: 2,
+    maximum: 15,
+    default: 10
```
joe-ayoub-segment, line 17: Are you sure of these limits @nk1107 ? Is this documented somewhere?

joe-ayoub-segment (dismissed):

packages/destination-actions/src/destinations/customerio/customerio-properties.ts
nk1107: I am referring to this documentation: https://docs.customer.io/integrations/api/track/#operation/batch

It says : Your batch request must be smaller than 500kb. Each of the requests within the batch must also be 32kb or smaller.

So for `max_batch` size i am considering the edge case where all the requests in the batch are 32 kb which gives us 15 for the `max_batch` size. I do am unsure if this works do let me know what more can i do to confirm this number

joe-ayoub-segment:
Ah OK. that's the batch size for inbound API calls to Segment. You don't
need to worry about this.
The min and max batch sizes are you are setting relate to how many payloads
will be passed into the performBatch() function. This should be based off
of how many payloads the Customer.io API can handle.
So if Customer.io is able to handle up to 1000 events per batch, then we
would set the max to 1000, and maybe the default also. Can you find this
out please, and amend the PR?
Thanks,
Joe

joe-ayoub-segment:
Hi @nk1107 please ping me on Slack or email when you'd like a re-review.

packages/destination-actions/src/destinations/customerio/utils.ts
```diff
@@ -116,10 +116,23 @@ type BasePayload = {
...
+  const {
+    convert_timestamp,
+    person_id,
+    anonymous_id,
+    email,
+    object_id,
+    object_type_id,
+    timestamp,
+    enable_batching,
+    batch_size,
```
joe-ayoub-segment, line 133: Same question here. We usually don't actually use the data from these fields in the perform() and performBatch() in any way. So I don't think you need to deconstruct them from the payload object. Just ignore them.

packages/destination-actions/src/destinations/customerio/utils.ts
```diff
@@ -116,10 +116,23 @@ type BasePayload = {
   primary?: Identifiers
   secondary?: Identifiers
   timestamp?: string | number
+  enable_batching?: boolean
+  batch_size?: number
```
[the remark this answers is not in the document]
nk1107: Hey @joe-ayoub-segment while testing using the action-tester I noticed that these fields are coming in with payload and then these are going through the rest variable and into the body of the request. That's why i added them in this type and then deconstructed them. Let me know if i am missing anything.

packages/destination-actions/src/destinations/customerio/utils.ts
joe-ayoub-segment: I'm pretty sure you can remove this, as you are not actually making use of the values from these fields in the perform() and performBatch() functions.

packages/destination-actions/src/destinations/customerio/utils.ts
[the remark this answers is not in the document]
joe-ayoub-segment: I'm pretty sure you can remove this, as you are not actually making use of the values from these fields in the perform() and performBatch() functions.

packages/destination-actions/src/destinations/customerio/utils.ts
nk1107: Added the changes

joe-ayoub-segment (approved):

PR #3112, opened 2025-07-25
[Ortto Audiences] - Bug fix

joe-ayoub-segment:
createAudience function is not working.

## Testing

Not in use by customers yet.


xitonix (approved):
As you mentioned in your email, I doubt it would make any difference but doesn't harm to give it a shot. LGTM

sayan-das-in (approved):

PR #3126, opened 2025-07-30
[Epsilon] - Required field missing default mapping - bug

joe-ayoub-segment:
Bug fix for Epsilon Destination. 

## Testing

Not needed. Not in use by customers yet.


joe-ayoub-segment:
PR deployed.

PR #3181, opened 2025-08-14
[Epsilon] - making company magic a separate setting field

joe-ayoub-segment:
Adding new field. New Integration not in use by customers. 

## Testing

Not in use by customers. Unit test updated.


joe-ayoub-segment:
PR deployed

PR #3204, opened 2025-08-27
STRATCONN-6008 - [Twilio Messaging] - FB Messenger support

joe-ayoub-segment:
PR to add FB Messenger support to Twilio Messaging Destination

## Testing
Unit tests to be added
Testing in production with feature flag


varadarajan-tw:
Just left couple of comments. Also, it seems like codecov/patch is failing because test coverage isn't sufficient enough. pls see if we can increase test coverage

PR #3215, opened 2025-08-28
StackAdapt Audience Destination: Configure advertiser under basic settings instead of mappings

Vanessa-SSY:

## Testing

### DIFF
--- a/packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
+++ b/packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
@@ -1,4 +1,4 @@
-import { RequestClient, InvalidAuthenticationError } from '@segment/actions-core'
+import { RequestClient } from '@segment/actions-core'
 import { Payload } from './generated-types'
 import { Settings } from '../generated-types'
 import { GQL_ENDPOINT, EXTERNAL_PROVIDER, sha256hash, stringifyJsonWithEscapedQuotes, stringifyMappingSchemaForGraphQL } from '../functions'
--- a/packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
+++ b/packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
@@ -1,5 +1,6 @@
-import { RequestClient } from '@segment/actions-core'
+import { RequestClient, InvalidAuthenticationError } from '@segment/actions-core'
 import { Payload } from './generated-types'
+import { Settings } from '../generated-types'
 import { GQL_ENDPOINT, EXTERNAL_PROVIDER, sha256hash, stringifyJsonWithEscapedQuotes, stringifyMappingSchemaForGraphQL } from '../functions'
 
 const audienceMapping = stringifyMappingSchemaForGraphQL([
@@ -30,12 +31,13 @@ const profileMapping = stringifyMappingSchemaForGraphQL([
   }
 ])
 
-export async function performForwardAudienceEvents(request: RequestClient, events: Payload[]) {
-  const advertiserId = events[0].advertiser_id
+export async function performForwardAudienceEvents(request: RequestClient, events: Payload[], settings: Settings) {
+  const advertiserId = settings.advertiser_id
+
   const profileUpdates = events.map((event) => {
     const { segment_computation_key: audienceKey, segment_computation_id: audienceId, user_id, traits_or_props } = event
 
-    const { [audienceKey]: action } = traits_or_props
+    const action = traits_or_props[audienceKey]
     return {
       userId: user_id,
       audienceId,
--- a/packages/destination-actions/src/destinations/stackadapt-audiences/index.ts
+++ b/packages/destination-actions/src/destinations/stackadapt-audiences/index.ts
@@ -65,7 +65,6 @@ const destination: DestinationDefinition<Settings> = {
               }))
               .sort((a: any, b: any) => a.label.localeCompare(b.label))
             
-            console.log('=============choices', choices)
             return { choices }
           } catch (error: any) {
             return {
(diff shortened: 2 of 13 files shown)


packages/destination-actions/src/destinations/stackadapt-audiences/index.ts
```diff
@@ -19,6 +19,14 @@ const destination: DestinationDefinition<Settings> = {
         description: 'Your StackAdapt GQL API Token',
         type: 'string',
         required: true
+      },
+      advertiser_id: {
+        label: "Advertiser ID",
+        description: "The StackAdapt advertiser ID to add the profile to. The value in this field field can also be overridden at the Action level via the Action field of the same name.",
+        type: 'string', 
+        required: true,
```
joe-ayoub-segment, line 27: Should be required:false.

packages/destination-actions/src/destinations/stackadapt-audiences/index.ts
```diff
@@ -59,16 +67,23 @@ const destination: DestinationDefinition<Settings> = {
...
     }
   },
-  onDelete: async (request, { payload }) => {
+  onDelete: async (request, { payload, settings }) => {
     const userId = payload.userId
     const formattedExternalIds = `["${userId}"]`
     const syncId = sha256hash(String(userId))
+    const advertiserId = settings.advertiser_id
+    if(!advertiserId) {
+      throw new InvalidAuthenticationError("Advertiser value must be provided in either the main Settings Advertiser field or at the Action level Advertiser field.")
```
joe-ayoub-segment, line 76:
```suggestion
      throw new InvalidAuthenticationError("To delete a user, the Advertiser field in Settings should be populated.")
```

Vanessa-SSY:
We will not enable onDelete if we can't have the dropdown for advertiserIds in a single place that can be referenced for both `perform` and `onDelete`. It's going to cause unnecessary confusion for our clients.

Vanessa-SSY:
> > It's not in use by clients at the moment. The traffic is from our internal testing. Could you help confirm if we can have the same dropdown for advertiserIds under `Settings` as the one we have now under `Mappings UI`? @joe-ayoub-segment
> 
> Ah OK if no customers are using it, we can just keep the advertiserIds field in Settings / authentication, and then delete the advertiserIds in the Mappings UI.

Yes, we can surely do that but more importantly we want to keep the UI intuitive for clients. I tried to have the dropdown for advertiser IDs in Settings / authentication but it doesn't seem to work. If we can get the dropdown working, we would still prefer having it in the Mappings UI instead of an input box in Settings

joe-ayoub-segment:
Hi @Vanessa-SSY , 
If you can get the dynamic dropdown working in the global setting then you should be good to go. 
The pattern for implementing the dynamic dropdown for a global setting is a little different (not sure why, I apologise!). 

Can you try something like this please?

```
  authentication: {
    scheme: 'custom',
    fields: {
      myDynamicField: {
        label: 'My Dynamic Field',
        description: 'This field loads options dynamically',
        type: 'string',
        required: true,
        dynamic: async (request, { settings }) => {
          // Make API call using current settings to fetch choices
          const response = await request('https://api.example.com/options')
          return {
            choices: response.data.map(item => ({
              label: item.name,
              value: item.id
            }))
          }
        }
      }
    }
  }

```

Vanessa-SSY:
> > > Hi @Vanessa-SSY , If you can get the dynamic dropdown working in the global setting then you should be good to go. The pattern for implementing the dynamic dropdown for a global setting is a little different (not sure why, I apologise!).
> > > Can you try something like this please?
> > > ```
> > >   authentication: {
> > >     scheme: 'custom',
> > >     fields: {
> > >       myDynamicField: {
> > >         label: 'My Dynamic Field',
> > >         description: 'This field loads options dynamically',
> > >         type: 'string',
> > >         required: true,
> > >         dynamic: async (request, { settings }) => {
> > >           // Make API call using current settings to fetch choices
> > >           const response = await request('https://api.example.com/options')
> > >           return {
> > >             choices: response.data.map(item => ({
> > >               label: item.name,
> > >               value: item.id
> > >             }))
> > >           }
> > >         }
> > >       }
> > >     }
> > >   }
> > > ```
> > 
> > 
> > I tried with the code you suggested above but still seeing advertiser ID under `Basic Settings` being a stale input box instead of a dropdown populated based on API key. Would we be able to populate the advertiser list if API key is being provided under the same tab? I added some logs in the dynamic but don't see it being triggered. I pushed my code for your reference. @joe-ayoub-segment
> 
> Ah OK - I wouldn't worry about testing it in the testing tool. That tool is so out of date it hardly works. We're going to be building a new test tool.
> 
> One way you should be able to test it is to use Postman. When you start up the testing tool you will see a list of endpoints you can hit at your localhost:3000
> 
> For example:
> 
> image
> If you call those endpoints you should be able to invoke the dynamic function that populates the dropdown list.
> 
> If you have any trouble with this please schedule a call with me - I'd be happy to figure it out in a working session. https://calendly.com/joe_ayoub
> 
> Thanks Joe

I have tried to use postman but still don't see the dynamic function being triggered. I would like to schedule a call with you some time tomorrow if that works for you. I'm free any time before 4PM EST. Please let me know you availability. @joe-ayoub-segment

joe-ayoub-segment:
Hi Vanessa,

I just sent you an invite. Let me know if you'd prefer another time.
Best regards
Joe

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
```diff
@@ -1,5 +1,6 @@
-import { RequestClient } from '@segment/actions-core'
+import { RequestClient, InvalidAuthenticationError } from '@segment/actions-core'
```
joe-ayoub-segment, line 1: I think the reference to InvalidAuthenticationError can be removed as it's not used.

PR #3219, opened 2025-09-01
[LinkedIn Audiences] Hide enable batching

varadarajan-tw:
Hide the `enable batching` flag. LinkedIn Audiences has very low rate limits. We shouldn't allow the customer to toggle the `enable_batching` flag.

Only one customer had batching disabled and we enabled it for them via ctlplane API.

## Testing

Testing not required as this is an UX only config change.


joe-ayoub-segment:
Hi @varadarajan-tw PR deployed

PR #3244, opened 2025-09-08
[Vibe] Add Vibe Audience destination

Max0u:
Hello everyone,

After my discussions with `Ga(niprasad) Niprasad` and `Timothy Koeth`, we agreed to move forward with a native action destination.

Our goal is to provide your users with a destination to Vibe Audience, enabling them to retarget their users.

I’ve mainly followed the same logic as `facebook-custom-audience`.

The new endpoints are not yet deployed, but once they are, I’ll update the base URL. Everything has been tested locally so far, and I’d also like to test the segment creation part locally if possible. Is it possible?

From my understanding, your users will need to enter their Vibe ID and the auth token (which we’ll provide in our UI). Segment will then create all Engage segments on our side and push batches of profiles based on the mapping the user sets. Could you confirm if that’s correct?

Thanks in advance for reviewing this!

## Testing


joe-ayoub-segment:
PR deployed

PR #3266, opened 2025-09-11
STRATCONN-6177 - [GA4 Web] - domain setting

joe-ayoub-segment:
GA4 Web Destination - adding new global setting to allow for the domain to be changed. 

## Testing

Unit tests added. Tested locally. 

**Default domain:** 
image

**Custom domain:** 
image


joe-ayoub-segment:
PR deployed

PR #3318, opened 2025-09-26
Ortto: Make message ID optional

xitonix:
Make Message ID and Timestamp optional to support rETL, which cannot provide a Message ID. Without a Message ID, activities from rETL cannot be deduplicated using the usual key, so this change allows the integration to accept events while acknowledging that duplicates may occur.

## Testing


joe-ayoub-segment (approved):

xitonix:
Looking good @joe-ayoub-segment. Thank you.

PR #3331, opened 2025-10-01
[Stackadapt Audiences] - making marketing_status for non Audience payloads

joe-ayoub-segment:
@Vanessa-SSY indicated that `marketing_status  should be in common-fields.ts`

## Testing

To be tested in Production by @Vanessa-SSY 
Unit tests updated.


Vanessa-SSY:
I have tested both forwardProfile and forwardAudienceEvent. Both are working!
Thanks @joe-ayoub-segment !

Vanessa-SSY:
Hi @joe-ayoub-segment , I just checked with our internal team and realize that they are using a different API field for the marketing status so the following change is needed:
Screenshot 2025-10-02 at 8 14 22 PM

I have tested the above change with your PR and it's working.

PR #3671, opened 2026-03-17
STRATCONN-6562 - [Facebook Custom Audiences] - rewrite

joe-ayoub-segment:
Updating the Facebook Custom Audiences integration in preparation for GA. 

This PR rewrites the sync action and surrounding destination code for Facebook Custom Audiences to improve code organization, correctness, and test coverage.

  Key Changes
  - Audience Membership now passed into the perform() and performBatch() functions (from actions-core). 

  Refactoring & Code Organization
  - Extracted all field definitions into fields.ts (both top-level and sync/), hook input/output definitions into sync/fields.ts, hook logic into sync/hook-functions.ts, API operations (createAudience,
  getAudience, parseFacebookError, getApiVersion) into functions.ts, and sync business logic into sync/functions.ts
  - Removed the monolithic fbca-operations.ts (FacebookClient class) and fbca-properties.ts, replacing them with focused, composable modules
  - Moved US_STATE_CODES and SCHEMA_PROPERTIES into sync/constants.ts where they are used; cleaned up top-level constants.ts
  - Moved presets array into its own presets.ts file
  - Added a types.ts file (FacebookDataRow, AudienceJSON, PayloadMap) and top-level types.ts for Facebook error/response shapes

  Sync Mode
  - Changed default sync mode from upsert to mirror
  - Added mirror as a sync mode option (alongside upsert and delete)
  - mirror mode routes individual payloads to add or delete buckets based on audienceMembership flags, issuing parallel POST/DELETE requests to the Facebook API

  Error Handling
  - Replaced ad-hoc error string construction with a structured parseFacebookError helper that extracts error_user_title, error_user_msg, FB message, and error code from nested Facebook API error responses
  - createAudience and getAudience now return { data, error } result objects instead of throwing, letting callers use IntegrationError with proper ErrorCodes
  - getAudience now requests fields=id,name and validates audience ID match

  Audience Settings
  - engageAdAccountId is now optional (falls back to retlAdAccountId from destination settings)
  - Updated description copy for both ad account ID fields to clarify override behavior
  - Switched createAudience from URLSearchParams form-encoding to JSON body

  Tests
  - Replaced single __tests__/index.test.ts for the sync action with four focused suites: upsert.test.ts, delete.test.ts, mirror.test.ts, and legacy.test.ts
  - Added sync/__tests__/functions.test.ts for unit-testing payload transformation and hashing logic
  - Added top-level __tests__/audience-creation.test.ts and __tests__/canary.test.ts
  - Added defaultSubscription: 'type = "track" or type = "identify"' to the sync action

## Testing

Tested in Staging with Facebook Custom Audiences. 
image

## Security Review

## New Destination Checklist


AnkitSegment (approved):

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -0,0 +1,294 @@
...
+    await request(`${BASE_URL}/${getApiVersion(features, statsContext)}/${audienceId}/users`, {
+      method,
+      json
+    })
+    for (let i = 0; i < indices.length; i++) {
+      const originalIndex = indices[i]
+      msResponse.setSuccessResponseAtIndex(originalIndex, {
+        status: 200,
+        body: payloads[i] as unknown as JSONLikeObject,
+        sent: {
```
abhandage, line 77: **[Critical] PII echoed in plaintext in MultiStatusResponse success body**

`payloads[i]` contains unhashed email, phone, name, DOB, gender, city, state, zip, and country. Hashing only happens inside `getData()` when building the Facebook-bound JSON. The `body` field is new in this PR — the old code never put the raw payload into the response. These values will appear in delivery logs and internal observability tooling accessible to operators.

**Fix:** Replace with the hashed row or omit `body` entirely:

```ts
// Option A: use the already-hashed row
body: json.payload.data[i] as unknown as JSONLikeObject,

// Option B: omit body entirely — sent already captures what was transmitted
```

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -0,0 +1,294 @@
...
+}
+
+export function validate(payloads: Payload[], audienceId: unknown, audienceMemberships?: AudienceMembership[]): string | undefined {
+  
+  if(!Array.isArray(audienceMemberships)){
+    return 'Audience membership details for batch missing.'
+  }
+
+  if(Array.isArray(audienceMemberships) && audienceMemberships.length !== payloads.length){
+    return 'Audience membership details count does not match batch payload count.'
```
abhandage, line 159: **[High] Length equality check silently drops all valid payloads when AJV filters any event**

`audienceMemberships` is computed from the full `bundle.data` array (N events) before AJV schema validation. AJV then filters out invalid payloads, producing a shorter `payloads` array (N-K events). By the time `performBatch` is called, `audienceMemberships.length === N` but `payloads.length === N-K`.

This check then calls `returnErrorResponse` which marks every remaining valid payload as failed with `'Audience membership details count does not match batch payload count.'`

Since `externalId` is `required: true`, a single event missing that field in a batch causes all other valid events in the batch to be silently rejected.

**Fix:** Remove this upfront length-equality check. The per-item `undefined` check at the `audienceMemberships?.[index]` call site (line ~33) already handles missing entries safely.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -0,0 +1,294 @@
...
+  if (['female', 'woman', 'girl', 'f'].includes(lowerCaseValue)) return 'f'
+
+  return value
+}
+
+export function normalizeMonth(value: string): string {
+  const normalizedValue = value.replace(/\s/g, '').trim()
+
+  if (normalizedValue.length === 2 && typeof Number(normalizedValue) === 'number') {
+    return normalizedValue
```
abhandage, line 228: **[High] `normalizeMonth` does not zero-pad single-digit months, corrupting SHA-256 hashes**

`normalizeMonth('5')` returns `'5'` instead of `'05'`. The guard checks `normalizedValue.length === 2`, which is false for single-digit inputs. They fall through to the month-name lookup, fail to match, and the original string is returned unmodified.

Facebook requires `DOBM` in `MM` format — `sha256('5')` will never match `sha256('05')`, silently degrading match rates for any customer whose `birth.month` field contains integer-valued months.

Additionally, `typeof Number(normalizedValue) === 'number'` is a JavaScript tautology (`typeof NaN === 'number'`), so non-numeric two-character strings like `'ab'` also bypass the month-name lookup.

**Fix:**
```ts
if (/^\d{1,2}$/.test(normalizedValue)) {
  return normalizedValue.padStart(2, '0')
}
```

Add test cases: `normalizeMonth('5')` → `'05'` and `normalizeMonth('ab')` → fallthrough to name lookup.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/index.ts
```diff
@@ -1,312 +1,52 @@
...
-
-        return {
-          error: {
-            message: 'Invalid operation',
-            code: 'INVALID_OPERATION'
-          }
-        }
+      inputFields: retlHookInputFields,
+      outputTypes: retlHookOutputTypes,
+      performHook: async (request, { settings, hookInputs }) => {
```
abhandage, line 19: **[High] `performHook` ignores `engageAdAccountId`, always uses `retlAdAccountId`**

`settings.retlAdAccountId` is passed unconditionally to `performHook`. But `audienceConfig.createAudience` lower in this file correctly resolves `engageAdAccountId ?? retlAdAccountId`.

If a customer sets `engageAdAccountId` to a different ad account, the hook will list and create audiences under the wrong account while actual event sync targets the correct one — they'll see audiences from the wrong account in the dropdown, and new audiences will be created in the wrong place.

**Fix:**
```ts
performHook: async (request, { settings, hookInputs, audienceSettings }) => {
  const adAccountId = (audienceSettings?.engageAdAccountId ?? settings.retlAdAccountId) as string
  // ...pass adAccountId instead of settings.retlAdAccountId
```

The same fix is needed in `getExistingAudienceIdChoices` in `hook-functions.ts` (it also reads only `settings.retlAdAccountId`).

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/hook-functions.ts
```diff
@@ -0,0 +1,109 @@
...
+  return {
+    error: {
+      message: 'Invalid operation value. Must be existing or create.',
+      code: ErrorCodes.RETL_ON_MAPPING_SAVE_FAILED
+    }
+  }
+}
+
+export async function getExistingAudienceIdChoices(request: RequestClient, { settings }: { settings: Settings }) {
+  const { retlAdAccountId: adAccountId } = settings
```
abhandage, line 72: **[High] `getExistingAudienceIdChoices` ignores `engageAdAccountId`**

This function uses only `settings.retlAdAccountId` to fetch the audience list. If a customer has configured `engageAdAccountId` to a different ad account, the dropdown will show audiences from the wrong account. This is paired with the same issue in `sync/index.ts`'s `performHook`.

**Fix:** Accept and prefer `audienceSettings?.engageAdAccountId` if present:
```ts
export async function getExistingAudienceIdChoices(
  request: RequestClient,
  { settings, audienceSettings }: { settings: Settings; audienceSettings?: AudienceSettings }
) {
  const adAccountId = audienceSettings?.engageAdAccountId ?? settings.retlAdAccountId
  // ...
}
```

packages/destination-actions/src/destinations/facebook-custom-audiences/constants.ts
```diff
@@ -1,56 +1,7 @@
...
-  ['vermont', 'vt'],
-  ['virginia', 'va'],
-  ['washington', 'wa'],
-  ['westvirginia', 'wv'],
-  ['wisconsin', 'wi'],
-  ['wyoming', 'wy']
-])
 export const API_VERSION = 'v24.0'
+
 export const CANARY_API_VERSION = 'v24.0'
```
abhandage, line 3: **[High] `CANARY_API_VERSION` equals `API_VERSION` — canary routing is dead code with illusory test coverage**

Both constants are `'v24.0'`. The `getApiVersion()` function in `functions.ts` has no effect today — it always returns `'v24.0'` regardless of the feature flag.

Worse, `canary.test.ts` works around this by mocking `CANARY_API_VERSION` to `'v25.0'`. When someone eventually bumps this constant to actually roll out the canary, the test suite won't catch regressions — tests never run against the real production constant value.

**Fix options:**
1. If v25.0 is ready: set `CANARY_API_VERSION = 'v25.0'` and remove the `jest.mock` override in `canary.test.ts`
2. If not ready yet: add a comment here making the holding-pattern intentional and documenting when/how to bump it

Don't leave this silent — the next person to touch this file won't know why the test mock exists.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/fields.ts
```diff
@@ -0,0 +1,278 @@
...
+      }
+    }
+  },
+  mobileAdId: {
+    type: 'string',
+    label: 'Mobile Advertising ID',
+    description:
+      'User’s Apple IDFA, Android Ad ID, or Facebook app scoped ID. Keep hyphens (ex: AB1234CD-E123-12FG-J123)',
+    category: 'hashedPII'
+  },
```
abhandage, line 177: **[Medium] `mobileAdId`, `appId`, `pageId` tagged `hashedPII` but sent to Facebook in plaintext**

These fields are correctly transmitted to Facebook unhashed (per the Facebook Custom Audiences API spec), but the `category: 'hashedPII'` label is factually wrong. This will mislead future maintainers into believing hashing is applied before transmission, and the test at `functions.test.ts` explicitly asserts "passes mobileAdId through without hashing" — directly contradicting the field label.

**Fix:** Change `category` to `'identifier'` (or remove it) and add a brief comment:
```ts
// Facebook Custom Audiences API requires MADID, app_ids, and page_ids
// to be sent in plaintext — do NOT hash these values.
mobileAdId: {
  category: 'identifier',
  ...
}
```
Same fix applies to `appId` and `pageId`.

packages/destination-actions/src/destinations/facebook-custom-audiences/functions.ts
```diff
@@ -0,0 +1,116 @@
...
+    const { message } = parseFacebookError(error as FacebookResponseError)
+    return { error: { message, code: ErrorCodes.GET_AUDIENCE_FAILED } }
+  }
+}
+
+export function getApiVersion(features?: Features, statsContext?: StatsContext): string {
+  const { statsClient, tags } = statsContext || {}
+  const version = features && features[FACEBOOK_CUSTOM_AUDIENCE_FLAGON] ? CANARY_API_VERSION : API_VERSION
+  tags?.push(`version:${version}`)
+  statsClient?.incr(`actions_facebook_custom_audience`, 1, tags)
```
abhandage, line 110: **[Low] `getApiVersion` mutates a shared `tags` array, double-counting metrics in mirror mode**

In mirror mode, `send()` calls `sendRequest` twice (once for POST, once for DELETE) with the same `statsContext` reference. Both calls invoke `getApiVersion(features, statsContext)` synchronously, each pushing `'version:vX.X'` onto the same `tags` array and incrementing the counter. Result: duplicate tag entries and 2× metric count for every mirror-mode batch.

**Fix:** Call `getApiVersion` once in `send()` before the `Promise.all`, pass the version string as a param to `sendRequest`, and remove the side-effectful metric emission from inside `getApiVersion`.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
joe-ayoub-segment: normalizeMonth was introduced in commit 0a8d5b9f3 (Aug 2024) with the exact same buggy logic. 
PR #3671 (current PR) just moved it from fbca-properties.ts (or constants.ts) to sync/functions.ts. 
The single-digit padding bug was already there before it's pre-existing, not something #3671 introduced.

JIRA ticket raised to follow up: https://twilio-engineering.atlassian.net/browse/STRATCONN-6851

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
joe-ayoub-segment: Addressed in https://github.com/segmentio/action-destinations/pull/3825. 
The length check in the Facebook Custom Audience destination is a valid defensive guard. It correctly rejects the case where something unexpected happens upstream.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
joe-ayoub-segment: This is deliberate. Other destinations do this. Redaction occurs later after this.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -0,0 +1,294 @@
...
+  payloads.forEach((payload, i) => {
+    setErrorResponse(msResponse, payload, 400, i, isBatch, errormessage, errortype)
+  })
+  return msResponse
+}
+
+export function getAudienceId(payload: Payload, hookOutputs?: { retlOnMappingSave?: { outputs?: { audienceId?: string } } }): string {
+  const { retlOnMappingSave: { outputs: { audienceId: hookAudienceId = undefined } = {} } = {} } = hookOutputs ?? {}
+  const { external_audience_id: payloadAudienceId } = payload
+  return (hookAudienceId ?? payloadAudienceId) as string
```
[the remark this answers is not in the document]
joe-ayoub-segment: Cosmetic type safety. If it's undefined at runtime you'd get an error downstream either way.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/index.ts
```diff
@@ -1,312 +1,52 @@
...
+        const { operation, audienceName, existingAudienceId } = hookInputs
+        const { retlAdAccountId: adAccountId } = settings
+
+        return await performHook(
+          request,
+          adAccountId,
+          operation as string | undefined,
+          audienceName as string | undefined,
+          existingAudienceId as string | undefined
+        )
```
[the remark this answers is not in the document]
joe-ayoub-segment: No. The canary flag is dead code right now (both versions are v24.0). When they actually bump the canary version, this can be addressed then.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/hook-functions.ts
```diff
@@ -0,0 +1,109 @@
...
+  const { choices, error } = await getAllAudiences(request, adAccountId)
```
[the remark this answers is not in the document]
joe-ayoub-segment: No. The canary flag is dead code right now (both versions are v24.0). When they actually bump the canary version, this can be addressed then.

packages/destination-actions/src/destinations/facebook-custom-audiences/functions.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const parts = [
+    error_user_title && `error_user_title: "${error_user_title}"`,
+    error_user_msg && `error_user_msg: "${error_user_msg}"`,
+    fbMessage && `fbmessage: "${fbMessage}"`,
+    outerMessage && `message: "${outerMessage}"`,
+    code && `code: "${code}"`
+  ]
+
+  const message = parts.filter(Boolean).join('. ').trim() || 'An unknown error occurred while communicating with Facebook API.'
```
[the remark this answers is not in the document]
joe-ayoub-segment: No. It works, just verbose. Cosmetic difference in error messages.

packages/destination-actions/src/destinations/facebook-custom-audiences/index.ts
```diff
@@ -1,22 +1,24 @@
...
   description: 'The Facebook Custom Audiences destination.',
-
   authentication: {
     scheme: 'oauth2',
     fields: {
-      retlAdAccountId: adAccountId
+      retlAdAccountId: 
+      {
+        ...adAccountId,
+        description: 'Your advertiser account id. Read [more](https://www.facebook.com/business/help/1492627900875762). This is required to set up the connection, but can be overriden using the Engage Audience setting named "Advertiser Account ID".'
```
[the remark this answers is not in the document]
joe-ayoub-segment: No. One-character typo in a settings description. Trivial. Can be fixed in another PR.

packages/destination-actions/src/destinations/facebook-custom-audiences/functions.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+    if (externalId !== id) {
+      return { error: { message: `Audience not found. Audience ID mismatch. Expected: ${externalId}, Received: ${id}`, code: ErrorCodes.GET_AUDIENCE_FAILED } }
+    }
+    return {
+      data: {
+        externalId: id,
+        name: name || ''
+      }
+    }
```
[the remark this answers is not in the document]
joe-ayoub-segment: No. Same class as #1 — type narrowing nit.

packages/destination-actions/src/destinations/facebook-custom-audiences/functions.ts
joe-ayoub-segment: No. Metrics are slightly inflated but not actionable. Low priority.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/fields.ts
joe-ayoub-segment: No. Functionally correct, just a misleading label

packages/destination-actions/src/destinations/facebook-custom-audiences/constants.ts
joe-ayoub-segment: No. It's intentionally a holding pattern. Not causing harm.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/index.ts
joe-ayoub-segment: Audience creation happens differently for RETL Vs Engage. 
With RETL - performHook us called. 
With Engage - createAudience is called. 

So this comment is incorrect.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/hook-functions.ts
joe-ayoub-segment: Audience creation happens differently for RETL Vs Engage. 
With RETL - performHook us called. 
With Engage - createAudience is called. 

So this comment is incorrect.

PR #3739, opened 2026-04-20
[Amplitude Cohorts] - setting field format update

joe-ayoub-segment:
Unable to test new amplitude cohorts destination due to setting field not displaying. This is because the field has format:email, which is not supported in the UI. 

## Testing
Not needed. This is a completely new integration

## Security Review

## New Destination Checklist


itsarijitray (approved):

joe-ayoub-segment:
deployed

PR #3754, opened 2026-04-27
feat(linkedin-audiences): upgrade API to 202604 with feature flag

harsh-joshi99:
## Summary

Upgrades LinkedIn Audiences API from **202505** to **202604**, deployed behind feature flag `linkedin-audiences-canary-version`.

- LinkedIn supports each version for a minimum of one year. Version 202505 will be sunset in May 2026 — this upgrade keeps the destination within the support window.
- No breaking changes affect the endpoints used by this destination.

FLAGON: https://flagon.segment.com/families/centrifuge-destinations/gates/linkedin-audiences-canary-version
### Version Management
- ✅ Updated `versioning-info.ts` with canary version `202604`
- ✅ Added `LINKEDIN_CANARY_API_VERSION`, `FLAGON_NAME`, and `getApiVersion(features)` to `constants.ts`
- ✅ Updated `extendRequest` in `index.ts` to use `getApiVersion(features)`
- ✅ Updated `LinkedInAudiences` class (`api.ts`) to accept `features` and use `getApiVersion` for `batchUpdate` version header
- ✅ Updated `updateAudience` `perform`/`performBatch` to pass `features` through `processPayload`
- ✅ Feature flag constant: `linkedin-audiences-canary-version`

### Drive-by Fix
- ✅ Corrected `LinkedIn-Version` header casing in `api.ts` and tests — previously `Linkedin-Version` (lowercase `i`), which was a pre-existing inconsistency from #3472. LinkedIn's own API docs and all other occurrences in this repo use `LinkedIn-Version` (capital `I`).

### Testing
- ✅ All existing tests pass (34 tests)
- ✅ Added 2 feature flag tests (stable + canary) for `updateAudience`
- ✅ Test results: **4 test suites passed, 36 tests passed**

## Breaking Changes

No breaking changes affect the `dmpSegments` or `adAccountUsers` endpoints used by this destination between 202505 and 202604.

### Informational Changes (not impacting this destination)
- **BATCH_CREATE partial validation (202502+)**: Enhanced error reporting with per-element `batchIndex` identifiers. This was introduced before our current stable version (202505) and is already in effect.
- **Predictive Audiences API (202511)**: New private API, not relevant to this destination.
- **Buyer Groups targeting facet (202603)**: New ad targeting facet, not relevant to this destination.
- **POST /media/upload sunset (Sep 16, 2025)**: This destination uses BATCH_CREATE on `/dmpSegments/{id}/users`, not file uploads.

Full analysis: `breaking-changes-analysis.md` in the destination directory.

### Manual Testing Required
- [x] Test with feature flag `linkedin-audiences-canary-version` disabled (stable 202505 version used)
- [x] Test with feature flag `linkedin-audiences-canary-version` enabled (canary 202604 version used)
- [x] Verify `LinkedIn-Version` header value in outbound requests matches expected version

[Stage testing.](https://docs.google.com/document/d/1ec6WOmOZjwORFRPTyb8uXVXDbNIh6l9gdi648qzglhA/edit?usp=sharing)

### Automated Testing
- [x] Unit tests passing (36/36)
- [x] Feature flag behavior verified via header matching in tests

## Rollout Plan

1. **Phase 1**: Merge PR, feature flag off by default (stable 202505 unchanged)
2. **Phase 2**: Enable `linkedin-audiences-canary-version` for internal testing
3. **Phase 3**: Gradual rollout to subset of customers
4. **Phase 4**: Full rollout → promote 202604 to stable, remove feature flag
5. **Phase 5**: Clean up old version constant and feature flag

## Risk Assessment

**Risk Level**: LOW

- No breaking changes to the endpoints this destination uses
- Feature flag allows instant rollback to 202505
- Comprehensive test coverage for both versions

---


arnav777dev:
# Comprehensive PR Review

## Bug Analysis (3 Independent Passes)

### Critical / High

**[HIGH] `personas_audience_key` dynamic default breaks the batching contract** *(all 3 passes flagged this)*
`updateAudience/fields.ts` — The field description explicitly states the value \"must be a hard-coded string variable in order for batching to work properly.\" The new `@if` default resolves dynamically at runtime from `$.properties.audience_key` → `$.context.personas.computation_key`. If events in the same batch resolve to different values at those paths (e.g., mixed event types, absent fields, casing differences), batch grouping silently partitions incorrectly or merges events that should be separate. This is a data integrity risk that won't surface until events are misrouted in LinkedIn's API.

**[HIGH] Silent null audience key for existing customers**
`updateAudience/fields.ts` — Customers who previously had no value configured would have received an explicit mapping error. With the new default, if `$.properties.audience_key` is absent **and** `$.context.personas.computation_key` is also absent, the field resolves to `undefined` at runtime. The `required: true` check is satisfied at configuration time, so the null key reaches the API call silently — events dropped or misrouted with no clear error signal.

**[HIGH] Double header set — fragility in `extendRequest` + `batchUpdate`**
`index.ts` and `api.ts` — `extendRequest` sets `LinkedIn-Version` globally for all requests; `api.ts::batchUpdate` also sets it explicitly per-request. Both call `getApiVersion(features)` independently. This works today because both receive the same features value, but creates fragility: any future caller that constructs `LinkedInAudiences` with different/stale `features` will send the wrong version to exactly the endpoint that matters most. The explicit override in `batchUpdate` should either be removed (trusting `extendRequest`) or have a comment explaining why it cannot.

### Medium

**[MEDIUM] Loose feature flag truthiness check — string `\"false\"` enables canary**
`constants.ts` — `features[FLAGON_NAME]` is truthy for any non-empty value, including the string `\"false\"` (a common return value from LaunchDarkly and similar SDKs when a flag is disabled). Use strict equality: `features[FLAGON_NAME] === true`.

**[MEDIUM] `features` availability in `extendRequest` needs framework verification**
`index.ts` — The `extendRequest` callback's destructured shape is typically `{ auth, settings }`. If the framework does not inject `features` into this callback, `features` will always be `undefined` at this path and the flag will silently never activate from `extendRequest`. Verify the framework's `ExtendRequest` type includes `features`.

**[MEDIUM] Flag evaluated independently in two places — potential mid-batch inconsistency**
`index.ts` and `api.ts` — If flag state toggles during a gradual rollout while a batch is in-flight, `extendRequest` and `batchUpdate` could resolve to different versions within a single logical operation. LinkedIn treats these as distinct schema contracts; mixed-version headers in one workflow could produce partial failures with no clear error signal.

### Low

**[LOW] Header casing fix is not flag-gated**
The `Linkedin-Version` → `LinkedIn-Version` rename in `api.ts` is unconditional — it takes effect for all traffic regardless of the feature flag. This makes it harder to isolate as a cause if a regression occurs. (HTTP headers are case-insensitive per spec so practical risk is low, but the fix is bundled outside the flag boundary.)

---

## Test Coverage Analysis

### Well Covered
- **Happy path for both version branches**: nock `matchHeader` assertions confirm the correct `LinkedIn-Version` header is sent to both `dmpSegments` (GET) and `dmpSegments/.../users` (POST) for both stable and canary paths.
- **Feature propagation through the full stack**: Both tests exercise `perform` → `processPayload` → `LinkedInAudiences` constructor, confirming features reach the API client.

### Missing

**[MISSING: HIGH] `getApiVersion()` — features-without-flag branch untested**
3 distinct branches exist: `undefined`, `{}` (flag key absent), and `{ [FLAGON_NAME]: true }`. Only branches 1 and 3 are exercised. Branch 2 is the most common real-world misconfiguration.

**[MISSING: HIGH] `features: { [FLAGON_NAME]: false }` explicit false case**
Explicitly disabling the flag differs from passing no features at all in implementation. Neither test covers `false`. If the truthiness check ever changes, this regression surface is invisible.

**[MISSING: HIGH] `personas_audience_key` default mapping is completely untested**
Every test passes an explicit `mapping: { personas_audience_key: 'personas_test_audience' }`. The new `@if` fallback chain (`properties.audience_key` → `context.personas.computation_key`) has zero test coverage. This is a user-facing data correctness concern.

**[MISSING: MEDIUM] Assertion depth**
Both feature flag tests assert only `resolves.not.toThrowError()`. nock's `matchHeader` will cause a request mismatch if the header is wrong, which surfaces as an error — so the version check is implicitly tested. But the response body, segment lookup logic, and payload structure are not inspected; a regression in the POST body would pass silently.

**[MISSING: MEDIUM] `extendRequest` not tested in isolation**
`LinkedIn-Version` from `extendRequest` is only tested through full action execution. A direct unit test on `extendRequest` with/without features would pinpoint failures faster and document the intended interface contract.

**[MISSING: LOW] `useDefaultMappings: true` with no explicit `personas_audience_key`**
No test omits the `mapping` override entirely and relies on the new default mapping to populate `personas_audience_key`. This leaves the default-mapping regression surface uncovered.

---

🤖 Generated with [Claude Code](https://claude.ai/code) — 3 parallel bug passes + 1 test coverage pass

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/fields.ts
```diff
@@ -1,129 +1,135 @@
-
 import { InputField } from '@segment/actions-core'
 
-export const fields: Record<string, InputField> ={
```
mdkhan-tw, line 4: I expect no change here. How will it impact existing mapping config?

mdkhan-tw (commented):
LGTM otherwise, please review the Copilot comments

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/fields.ts
harsh-joshi99: The change is that `personas_audience_key` now has a default mapping. For new connections, this auto-populates the field so customers don't have to manually copy and paste the audience key.

For existing customers, `push-defaults` only backfills fields where the value is undefined, so it won't affect anyone who already has the field set. Additionally, `push-defaults` supports targeting specific workspace IDs, we can push the default only to customers who want it rather than performing a blanket rollout.

mdkhan-tw: We need to remove whitespace changes in this file to avoid confusion

mdkhan-tw (approved):
Please resolve Copilot comments before merging the PR

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/fields.ts
```diff
@@ -1,129 +1,135 @@
...
+    description:
+      'The `audience_key` of the Engage audience you want to sync to LinkedIn. This value must be a hard-coded string variable, e.g. `personas_test_audience`, in order for batching to work properly.',
+    type: 'string',
+    required: true,
+    default: {
+      '@if': {
+        exists: { '@path': '$.properties.audience_key' },
+        then: { '@path': '$.properties.audience_key' },
+        else: { '@path': '$.context.personas.computation_key' }
       }
```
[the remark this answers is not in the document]
harsh-joshi99: All events for a given Engage audience carry the same value for `computation_key`.
The default change does not make a difference in pre-existing functionality.

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/fields.ts
```diff
@@ -1,129 +1,135 @@
...
-        '@path': '$.event'
+    }
+  },
+  personas_audience_key: {
+    label: 'Segment Engage Audience Key',
+    description:
+      'The `audience_key` of the Engage audience you want to sync to LinkedIn. This value must be a hard-coded string variable, e.g. `personas_test_audience`, in order for batching to work properly.',
+    type: 'string',
+    required: true,
+    default: {
```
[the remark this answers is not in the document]
harsh-joshi99: All events for a given Engage audience carry the same value for computation_key.
The default change does not make a difference in pre-existing functionality.

packages/destination-actions/src/destinations/linkedin-audiences/updateAudience/fields.ts
harsh-joshi99: the whitespace changes are linter-enforced.

joe-ayoub-segment:
deployed

PR #3767, opened 2026-04-29
fix(recombee): Fix API inconsistencies

mstieranka:
Hi, we have identified several issues with out Recombee destination, and this pull request fixes all of them.

1. In the actions `Delete Cart Addition` and `Delete Bookmark`, the request to our API was incorrectly setting the timestamp to milliseconds since epoch where seconds were expected. The following changes were made:

- The tests now correctly expect a seconds-based timestamp, and the integration code has been updated to reflect this change.
- In all actions, the optional field `timestamp` has been modified to use the `datetime` type instead of `string`, which more accurately reflects the intended values.
  - **Note:** When testing using the local server, the mapping accepts an arbitrary string field, however, when sending the payload, the current live version ends up returning `HTTP 400` (the API rejects this parameter), whereas this version returns `HTTP 500` (the validation already fails in Segment). I'm not sure if this is a breaking change, since both requests would have failed and the field itself is optional in all mappings. **This is why the _"Tested for backwards compatibility"_ task below is not yet marked as complete.**
- The `timestamp` handling has generally been made more robust to accept any form of date string parseable by `new Date(str)` or a number representing either seconds or milliseconds since epoch. The field description has been changed to reflect this fact.

2. The default timestamp used by Segment (`$.timestamp`) is corrected for clock skew, however, that means that the Recombee API receives a different timestamp than the user sent. Since the `Delete Bookmark` action expects the exact same timestamp as when sending the request, this meant that such a use case wouldn't work. The following changes were made:

- The `timestamp` mapping in `Add*` events has been changed to use `$.properties.timestamp` by default, falling back to `$.timestamp` if not present.
- The `timestamp` mapping in `Delete*` events now also uses `$.properties.timestamp`.
- The field descriptions were updated to clearly state this fact and advise users.

3. In Segment Ecommerce events, the `price` field is stated per-unit, however, the Recombee API expects the `price` and `profit` fields to be stated per the given quantity. Therefore, the destination now multiplies `price` by `amount` to match the expectation of the Recombee API.

## Testing

## Security Review

## New Destination Checklist


joe-ayoub-segment:
Hi @mstieranka please ping me when this PR is ready for review.

mstieranka:
Hi @joe-ayoub-segment , this PR has had some changes added to it and is now ready for review.

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
```diff
@@ -131,15 +156,17 @@ export class SetViewPortion extends Request<SetViewPortionParams> {
...
-    }
-    return url + `&timestamp=${params.timestamp}`
+function getDeleteUrl(interactionType: string, params: DeleteParams): string {
+  const query = new URLSearchParams({
+    userId: params.userId,
+    itemId: params.itemId
+  })
+
+  if (params.timestamp !== undefined && params.timestamp !== null) {
+    query.append('timestamp', String(datetimeToEpochSeconds(params.timestamp)))
```
joe-ayoub-segment, line 174: consider calling datetimeToEpochSeconds into a variable first, then check if it is a valid timestamp before doing query.append.

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
```diff
@@ -85,6 +86,30 @@ function createAddInteractionData<T extends AddInteractionParams & InternalAddit
   }
 }
 
+function datetimeToEpochSeconds(datetime: string | number): number {
```
joe-ayoub-segment, line 89: The millisecond detection heuristic is fragile. The threshold 10000000000 (Nov 2286 in seconds) means any epoch-seconds value after that date would be incorrectly halved. Conversely, millisecond timestamps before Sept 9, 2001 (10000000000 ms = ~115 days from
  epoch) would be treated as seconds. Both are unlikely in practice but the magic number deserves at least a comment explaining the tradeoff.

  The division produces fractional seconds. numValue / 1000 for an odd millisecond value yields something like 1716048000.123. If the consumer expects an integer epoch (as the name "EpochSeconds" suggests), this should be Math.floor(numValue / 1000).

  String(datetime).trim() !== '' is dead code for the number input type. Number('') is 0 which would pass !isNaN, so the check guards against empty strings — fine. But String(someNumber).trim() !== '' is always true for any number input. Not wrong, just
  unnecessary for that branch.

  new Date(datetime) when datetime is a number that somehow failed the first check — this can't actually happen. If datetime is a number, Number(datetime) will never be NaN. So the new Date(datetime) path only runs for non-numeric strings, which is correct but not
   obvious from reading the flow.

  No timezone consideration for string parsing. new Date("2024-01-15") vs new Date("2024-01-15T00:00:00") behave differently (UTC vs local time) depending on the format. If callers pass date-only strings, the result will vary by environment.

  The biggest practical issue is the missing Math.floor — everything else is edge-case or readability.

packages/destination-actions/src/destinations/recombee/commonFields.ts
```diff
@@ -1,5 +1,71 @@
...
+  description: `The UTC timestamp of when the ${interactionName} occurred, in Unix seconds, Unix milliseconds, or ISO-8601 format. When recording interactions you plan to later delete by exact timestamp — whether via this destination or the Recombee API directly — avoid mapping the root \`timestamp\` here, as it may be corrected for clock skew. Use \`properties.timestamp\` instead.`,
+  type: 'datetime',
+  required: false,
+  default: {
+    '@if': {
+      exists: { '@path': '$.properties.timestamp' },
+      then: { '@path': '$.properties.timestamp' },
+      else: { '@path': '$.timestamp' }
+    }
+  },
```
joe-ayoub-segment, line 53: Unless you have a good reason, I would leave the detault as $.timestamp as that's the location where Segment event timestamps are always found.

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
```diff
@@ -85,6 +86,30 @@ function createAddInteractionData<T extends AddInteractionParams & InternalAddit
...
+  const numValue = Number(datetime)
```
joe-ayoub-segment: If user passes "" this will evaluate to 0. It might enter a code path you didn't anticipate below. Please check.

packages/destination-actions/src/destinations/recombee/commonFields.ts
```diff
@@ -1,5 +1,71 @@
...
+  ...fieldProps
+})
+
+export const deleteTimestampField: (interactionName: string, fieldProps?: Partial<InputField>) => InputField = (
+  interactionName,
+  fieldProps = {}
+) => ({
+  label: 'Timestamp',
+  description: `The UTC timestamp of the ${interactionName} to delete, in Unix seconds, Unix milliseconds, or ISO-8601 format. Must match the timestamp used in the ${interactionName} to be deleted. If omitted, all ${interactionName}s for the given \`userId\` and \`itemId\` are deleted.`,
+  type: 'datetime',
```
joe-ayoub-segment, line 63: Note the original field you are now generating via a function originally was not of type: 'datetime'.
Adding type: 'datetime' might break data collection for customers who are already sending in non datetime string to this field. 

If you confirm that this is OK and deliberate that's OK.

packages/destination-actions/src/destinations/recombee/commonFields.ts
```diff
@@ -1,5 +1,71 @@
...
+  ...fieldProps
+})
+
+export const interactionTimestampField: (interactionName: string, fieldProps?: Partial<InputField>) => InputField = (
+  interactionName,
+  fieldProps = {}
+) => ({
+  label: 'Timestamp',
+  description: `The UTC timestamp of when the ${interactionName} occurred, in Unix seconds, Unix milliseconds, or ISO-8601 format. When recording interactions you plan to later delete by exact timestamp — whether via this destination or the Recombee API directly — avoid mapping the root \`timestamp\` here, as it may be corrected for clock skew. Use \`properties.timestamp\` instead.`,
+  type: 'datetime',
```
joe-ayoub-segment, line 45: Note the original field you are now generating via a function originally was not of type: 'datetime'.
Adding type: 'datetime' might break data collection for customers who are already sending in non datetime string to this field.

If you confirm that this is OK and deliberate that's OK.

joe-ayoub-segment (changes requested):
Hi @mstieranka thanks for the PR and apologies for the delay reviewing it. 

It's generally good - however some important points I left comments on. 

If you can respond to those, and potentially fix, that would be great. 

Thanks,
Joe

DanielBreiner:
Hi @joe-ayoub-segment, I am picking this up as @mstieranka is currently OOO. I addressed your comments (see replies) and incorporated some of the suggestions into the code. Let me know if there is anything else.

Daniel

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
DanielBreiner: Recombee API accepts timestamp in seconds with or without the decimal part. If you insist, we can add the Math.floor, but in my opinion we would unnecessarily lose precision (I'd also argue that the name EpochSeconds doesn't necessarily suggest the value is an integer). The other issues I did fix.

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
```diff
@@ -131,15 +156,17 @@ export class SetViewPortion extends Request<SetViewPortionParams> {
...
+    itemId: params.itemId
+  })
+
+  if (params.timestamp !== undefined && params.timestamp !== null) {
+    query.append('timestamp', String(datetimeToEpochSeconds(params.timestamp)))
   }
-  return url
+
+  return `/${interactionType}/?${query.toString()}`
 }
```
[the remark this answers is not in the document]
DanielBreiner: Fixed

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
DanielBreiner: I'd rather not make the branch quietly get ignored if the timestamp is invalid, as that would erase all interactions of a certain type (see https://docs.recombee.com/api#delete-bookmark-param-timestamp), something that's most likely not wanted.

packages/destination-actions/src/destinations/recombee/commonFields.ts
DanielBreiner: As explained in the PR description, delete did not work before correctly because of the value conversion, so this will not break any existing integration. Meaning yes, this change should be OK.

packages/destination-actions/src/destinations/recombee/commonFields.ts
DanielBreiner: Recombee API only accepts ISO 8601-formatted strings as timestamps, which is the same as type: 'datetime', therefore this change only changes what was an API error before to be a validation error now. This is intentional.

packages/destination-actions/src/destinations/recombee/commonFields.ts
DanielBreiner: Segment's $.timestamp has built-in time correction, causing AddBookmark + RemoveBookmark sent from the customer with the same timestamp to not work, as the time corrected values no longer match. This is one workaround, but we're open to other suggestions.

packages/destination-actions/src/destinations/recombee/recombeeApiClient.ts
DanielBreiner: Fixed

joe-ayoub-segment:
Thanks @mstieranka and @DanielBreiner for addressing the comments. 
Would it be possible to attach some proof of testing?
e.g. some test payloads sent from Postman -> through the destination -> to your platform?

DanielBreiner:
Hey @joe-ayoub-segment

We have an internal repo for testing the Segment/Recombee integration. I have sent you read access, let me know if there is anyone else you want me to invite to see it. Link is here: https://github.com/recombee/segment-integration-tests
In branch `validate-pr-3767`, we created additional tests for this PR and set up an environment to allow us to test a specific version of Action Destinations with Recombee. Down below I copied the results of the tests, which are reproducible like this:

- creating a new Recombee Database and copying over its environment variables
- running action-destinations locally (in this case the branch `recombee:recombee-fix-timestamp-type`)
- executing `pnpm run test:local`

```
 PASS  tests/segment-recombee-integration.test.ts (17.303 s)
  Segment <> Recombee Integration
    ✓ Interaction - Detail View + timestamp + additionalData (2786 ms)
    ✓ Interaction - Detail View + recommId (optional, requires Recombee scenario) (3371 ms)
    ✓ Interaction - Detail View + Duration (optional, requires custom mapping) (2782 ms)
    ✓ Interaction - Cart Addition (5740 ms)
    ✓ Interaction - Bookmark (5984 ms)
    ✓ Interaction - Purchase (3204 ms)
    ✓ Interaction - Rating (optional, preset doesn't exist) (3037 ms)
    ✓ Interaction - View Portion (10930 ms)
    ✓ Interaction - Cart Addition + Timestamp (add, delete) (5614 ms)
    ✓ Interaction - Bookmark + Timestamp (add, delete) (5537 ms)
    ✓ Interaction - Detail View + properties.timestamp precedence (2712 ms)
    ✓ Interaction - Detail View + numeric timestamps (Unix seconds, Unix milliseconds) (2698 ms)
    ✓ Interaction - Detail View + timezone-less timestamp (interpreted as GMT) (2721 ms)
    ✓ Interaction - Purchase + multiple items (2676 ms)
    ✓ Merge Users (5602 ms)

Test Suites: 1 passed, 1 total
Tests:       15 passed, 15 total
Snapshots:   0 total
Time:        17.395 s
```

OndraFiedler:
Hi @joe-ayoub-segment,

Do you have an estimate when the changes will be reviewed and deployed?

OndraFiedler:
@joe-ayoub-segment @needcaffeine Can we please get the fix merged? 🙏

joe-ayoub-segment:
Hi @mstieranka thanks for the nudge. 
If possible, could you add a screen recording of testing please?
I just need some proof that you're able to send events through the new code to your platform. 
Best regards,
Joe

joe-ayoub-segment:
Hi @mstieranka just checking you saw my comment above?

mstieranka:
Hi @joe-ayoub-segment , here is video proof of us running this PR against our tests (including optional functionality that requires additional setup within Segment and/or Recombee).

Those tests (which you should also have access to) cover the entire functionality of this Destination, including behavior changed within this PR.

In this case, they are running against a re-implementation of the same API as the `@segment/analytics-node` client, which sends requests to a local instance of the Destination (`./bin/serve`) instead of the Segment cloud.

https://github.com/user-attachments/assets/64cd6c74-d44e-4d39-9c44-076e0b60108f

If you need anything else, let me know.

joe-ayoub-segment:
> Hi @joe-ayoub-segment , here is video proof of us running this PR against our tests (including optional functionality that requires additional setup within Segment and/or Recombee).
> 
> Those tests (which you should also have access to) cover the entire functionality of this Destination, including behavior changed within this PR.
> 
> In this case, they are running against a re-implementation of the same API as the `@segment/analytics-node` client, which sends requests to a local instance of the Destination (`./bin/serve`) instead of the Segment cloud.
> 
>  Segment-3767-test-proof.mp4 
> If you need anything else, let me know.

Thank you @mstieranka I've approved the PR to be deployed when we next roll out changes. This should be in the next few business days. 

I'll notify you on this ticket once everything is live. 
Thanks,
Joe

joe-ayoub-segment (approved):

PR #3779, opened 2026-05-07
include non str traits

monutwilio:
As of now, we only allow string type traits to be synced using Conversation Memory Sync/destination, we have to allow other types as well.

https://twilio-engineering.atlassian.net/browse/STRATCONN-6798

## Testing

1. Non-string traits visible in Destination.
Screenshot 2026-05-07 at 10 36 36 PM

2. Values synced via destination
Screenshot 2026-05-07 at 10 49 27 PM
Screenshot 2026-05-07 at 10 49 35 PM

3. Visible in conversation memory sync
Screenshot 2026-05-07 at 10 40 55 PM

4. Values synced via conversation memory sync
Screenshot 2026-05-08 at 10 13 40 AM

Screenshot 2026-05-08 at 10 13 50 AM

## Security Review

## New Destination Checklist

### DIFF
--- a/packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
@@ -1856,15 +1856,21 @@ describe('Memora.upsertProfile', () => {
         })) as DynamicFieldResponse
 
         expect(result).toBeDefined()
-        // Should exclude identifiers (traits with idTypePromotion) and non-STRING traits
+        // Should exclude identifiers (traits with idTypePromotion) but include all non-identifier traits regardless of dataType
         // All trait groups use traitGroupName.$.traitName format
         expect(result.choices).toEqual([
           { label: 'Contact.firstName', value: 'Contact.$.firstName', description: 'Contact - firstName (STRING)' },
           { label: 'Contact.lastName', value: 'Contact.$.lastName', description: 'Contact - lastName (STRING)' },
+          { label: 'Contact.age', value: 'Contact.$.age', description: 'User age' },
           {
             label: 'PurchaseHistory.Last Purchase Date',
             value: 'PurchaseHistory.$.lastPurchaseDate',
             description: 'Date of last purchase'
+          },
+          {
+            label: 'PurchaseHistory.Total Spent',
+            value: 'PurchaseHistory.$.totalSpent',
+            description: 'Total amount spent'
           },
           {
             label: 'PurchaseHistory.Favorite Category',
--- a/packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
@@ -518,7 +518,7 @@ describe('Memora.upsertProfile', () => {
       expect(profile.traits.PurchaseHistory.favoriteCategory).toBe('Electronics')
     })
 
-    it('should handle multiple trait groups in the same profile', async () => {
+    it('should handle multiple trait groups in the same profile including non-STRING trait types', async () => {
       const event = createTestEvent({
         type: 'identify',
         userId: 'user-891',
@@ -527,7 +527,9 @@ describe('Memora.upsertProfile', () => {
           first_name: 'Bob',
           last_purchase: '2024-02-20',
           loyalty_tier: 'Gold',
-          last_login: '2024-03-01'
+          last_login: '2024-03-01',
+          total_orders: 42,
+          is_subscribed: true
         }
       })
 
@@ -551,7 +553,9 @@ describe('Memora.upsertProfile', () => {
           profile_traits: {
             'Contact.$.firstName': { '@path': '$.traits.first_name' },
             'PurchaseHistory.$.lastPurchaseDate': { '@path': '$.traits.last_purchase' },
+            'PurchaseHistory.$.totalOrders': { '@path': '$.traits.total_orders' },
             'Loyalty.$.tier': { '@path': '$.traits.loyalty_tier' },
+            'Loyalty.$.isSubscribed': { '@path': '$.traits.is_subscribed' },
             'Engagement.$.lastLogin': { '@path': '$.traits.last_login' }
           }
         },
@@ -563,7 +567,10 @@ describe('Memora.upsertProfile', () => {
       expect(profile.traits.Contact.email).toBe('multi@example.com')
       expect(profile.traits.Contact.firstName).toBe('Bob')
       expect(profile.traits.PurchaseHistory.lastPurchaseDate).toBe('2024-02-20')
+      // Non-STRING trait values must be passed through as their native types (not coerced to strings)
+      expect(profile.traits.PurchaseHistory.totalOrders).toBe(42)
       expect(profile.traits.Loyalty.tier).toBe('Gold')
+      expect(profile.traits.Loyalty.isSubscribed).toBe(true)
       expect(profile.traits.Engagement.lastLogin).toBe('2024-03-01')
     })
 
@@ -1666,7 +1673,7 @@ describe('Memora.upsertProfile', () => {
     })
 
     describe('profile_identifiers (dynamic identifiers from all trait groups)', () => {
-      it('should fetch and return identifier traits from all trait groups', async () => {
+      it('should fetch and return identifier traits from all trait groups including non-STRING types', async () => {
         nock(BASE_URL)
           .get(`/${API_VERSION}/ControlPlane/Stores/test-store-id/TraitGroups?pageSize=100&includeTraits=true`)
           .matchHeader('X-Pre-Auth-Context', 'AC1234567890')
@@ -1720,6 +1727,27 @@ describe('Memora.upsertProfile', () => {
                   }
                 },
                 version: 1
+              },
+              {
+                displayName: 'Device',
+                description: 'Device traits',
+                traits: {
+                  deviceId: {
+                    dataType: 'NUMBER',
+                    description: 'Numeric device identifier',
+                    displayName: 'Device ID',
+                    idTypePromotion: 'device_id',
+                    validationRule: null
+                  },
+                  osVersion: {
+                    dataType: 'STRING',
+                    description: 'OS version',
+                    displayName: 'OS Version',
+                    idTypePromotion: null,
+                    validationRule: null
+                  }
+                },
+                version: 1
               }
             ]
           })
@@ -1730,11 +1758,12 @@ describe('Memora.upsertProfile', () => {
         })) as DynamicFieldResponse
 
         expect(result).toBeDefined()
-        // Should return only traits with idTypePromotion set
+        // Should return all traits with idTypePromotion set, regardless of dataType
         expect(result.choices).toEqual([
           { label: 'Contact.email', value: 'Contact.$.email', description: 'Contact - email (email)' },
           { label: 'Contact.phone', value: 'Contact.$.phone', description: 'Contact - phone (phone)' },
-          { label: 'Loyalty.Member ID', value: 'Loyalty.$.memberId', description: 'Loyalty member ID' }
+          { label: 'Loyalty.Member ID', value: 'Loyalty.$.memberId', description: 'Loyalty member ID' },
+          { label: 'Device.Device ID', value: 'Device.$.deviceId', description: 'Numeric device identifier' }
         ])
       })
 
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/memora/upsertProfile/index.ts
```diff
@@ -412,12 +412,12 @@ async function fetchTraitGroupFields(
         const value = `${traitGroupName}.$.${traitName}`
         const label = `${traitGroupName}.${trait.displayName || traitName}`
 
-        if (trait.idTypePromotion && trait.dataType === 'STRING') {
+        if (trait.idTypePromotion) {
```
monutwilio, line 415: **Identifiers also lose the STRING guard here.** Before this change, only STRING traits with `idTypePromotion` could appear as identifier choices. Now a NUMBER (or BOOLEAN, DATE, etc.) trait that happens to have `idTypePromotion` set will also surface as an identifier option.

If Memora identifiers are always strings (email, phone, loyalty_id), the old guard was correct for the identifier branch and only the trait branch needed relaxing. Consider keeping the STRING check on the identifier branch:

```ts
if (trait.idTypePromotion && trait.dataType === 'STRING') {
  // identifier
} else if (!trait.idTypePromotion) {
  // trait (any dataType)
}
```

If non-STRING identifiers are genuinely supported, that's fine — but it should be covered by a test so the intent is explicit.

packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
```diff
@@ -1856,16 +1856,22 @@ describe('Memora.upsertProfile', () => {
         })) as DynamicFieldResponse
 
         expect(result).toBeDefined()
-        // Should exclude identifiers (traits with idTypePromotion) and non-STRING traits
+        // Should exclude identifiers (traits with idTypePromotion) but include all non-identifier traits regardless of dataType
```
monutwilio, line 1895: The test for non-STRING traits only covers the happy-path dynamic-field listing. Consider adding a test that confirms non-STRING trait values (e.g. a NUMBER like `age: 30`) are correctly round-tripped through `perform`/`performBatch` all the way into the `profiles[].traits` payload sent to the Bulk API. That would catch any accidental string-coercion in `buildTraitGroups`.

packages/destination-actions/src/destinations/memora/upsertProfile/index.ts
monutwilio: added test

packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
monutwilio: done

packages/destination-actions/src/destinations/memora/upsertProfile/index.ts
```diff
@@ -412,12 +412,12 @@ async function fetchTraitGroupFields(
...
           const description = trait.description
             ? trait.description
             : `${traitGroupName} - ${trait.displayName} (${trait.idTypePromotion})`
           identifierChoices.push({ label, value, description })
-        } else if (!trait.idTypePromotion && trait.dataType === 'STRING') {
+        } else {
```
monutwilio, line 420: **Logic correctness - broadened `else` branch may silently include unsupported types**

Before this PR, the `else` branch only fired when `!trait.idTypePromotion && trait.dataType === 'STRING'`. Now it fires for every trait that is not an identifier, including `NUMBER`, `BOOLEAN`, and any future or unknown `dataType` values the Memora API may return.

This is the desired intent, but it would be worth confirming whether the Memora Profiles Bulk API actually accepts non-string values for ALL possible `dataType` values (e.g. `OBJECT`, `ARRAY`, `DATETIME`). If any data type is not accepted by the downstream API, silently including it in the dropdown and the request body could cause runtime errors that are hard to debug.

Recommendation: either add an allowlist of known-safe types (e.g. `['STRING', 'NUMBER', 'BOOLEAN']`) and surface unknown types with a warning, or add a comment explaining that the Profiles Bulk API is schema-validated on the server side so no client-side type filtering is needed.

packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
```diff
@@ -1730,11 +1758,12 @@ describe('Memora.upsertProfile', () => {
...
 
         expect(result).toBeDefined()
-        // Should return only traits with idTypePromotion set
+        // Should return all traits with idTypePromotion set, regardless of dataType
         expect(result.choices).toEqual([
           { label: 'Contact.email', value: 'Contact.$.email', description: 'Contact - email (email)' },
           { label: 'Contact.phone', value: 'Contact.$.phone', description: 'Contact - phone (phone)' },
-          { label: 'Loyalty.Member ID', value: 'Loyalty.$.memberId', description: 'Loyalty member ID' }
+          { label: 'Loyalty.Member ID', value: 'Loyalty.$.memberId', description: 'Loyalty member ID' },
+          { label: 'Device.Device ID', value: 'Device.$.deviceId', description: 'Numeric device identifier' }
```
monutwilio, line 1766: **Missing assertion: `osVersion` (non-identifier trait in Device group) is never verified in `profile_traits`**

The new `Device` trait group fixture (added to the `profile_identifiers` test) includes an `osVersion` trait with `idTypePromotion: null`. Under the new logic, this trait should appear in the `profile_traits` choices — but there is no test that confirms it.

Consider adding a separate test (or extending the existing `profile_traits` test fixture) that includes the `Device` group with `osVersion` and verifies `{ label: 'Device.OS Version', value: 'Device.$.osVersion', description: 'OS version' }` appears in the traits choices. Without this, a regression where non-STRING non-identifier traits in the Device group are accidentally filtered would go undetected.

packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
```diff
@@ -563,7 +567,10 @@ describe('Memora.upsertProfile', () => {
       expect(profile.traits.Contact.email).toBe('multi@example.com')
       expect(profile.traits.Contact.firstName).toBe('Bob')
       expect(profile.traits.PurchaseHistory.lastPurchaseDate).toBe('2024-02-20')
+      // Non-STRING trait values must be passed through as their native types (not coerced to strings)
+      expect(profile.traits.PurchaseHistory.totalOrders).toBe(42)
```
monutwilio, line 571: **Consider also asserting the serialized request body types, not just the reconstructed profile object**

This test re-reads `capturedBody` — which is the raw request body captured by nock — and then accesses the nested profile structure. The `capturedBody` is captured as nock intercepts the JSON-serialized body, so `capturedBody.profiles[0].traits.PurchaseHistory.totalOrders` being `42` (number) is a correct verification.

However, it would add confidence to also directly assert the type:
```ts
expect(typeof profile.traits.PurchaseHistory.totalOrders).toBe('number')
expect(typeof profile.traits.Loyalty.isSubscribed).toBe('boolean')
```
This guards against future coercion bugs where the value passes a loose `toBe(42)` check but the type has been inadvertently changed (e.g. if a serialization step ever converts it to a string `'42'` that is later re-parsed).

packages/destination-actions/src/destinations/memora/upsertProfile/index.ts
monutwilio, line 415: **Edge case: identifier with `idTypePromotion` set but empty `description` and no `displayName`**

The `TraitDefinition` interface marks `displayName` as `string` (required), but the API response is typed as `Record<string, TraitDefinition>` populated from an external HTTP response. In practice the API could return a trait with a missing or empty `displayName`. In that case the fallback description for identifiers becomes `"${traitGroupName} - undefined (${idTypePromotion})"`, which would surface confusingly in the UI dropdown.

The same issue exists for traits (line 423). The pre-existing `trait.displayName || traitName` guard on line 413 already protects the `label`, but the `description` fallback on lines 418 and 423 does not apply the same guard.

Suggested fix:
```ts
: `${traitGroupName} - ${trait.displayName || traitName} (${trait.idTypePromotion})`
```
This is pre-existing but the PR changes the else-branch fallback at line 423 in the same way, so it's a good time to address both.

packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
```diff
@@ -1856,16 +1885,22 @@ describe('Memora.upsertProfile', () => {
         })) as DynamicFieldResponse
 
         expect(result).toBeDefined()
-        // Should exclude identifiers (traits with idTypePromotion) and non-STRING traits
+        // Should exclude identifiers (traits with idTypePromotion) but include all non-identifier traits regardless of dataType
         // All trait groups use traitGroupName.$.traitName format
         expect(result.choices).toEqual([
           { label: 'Contact.firstName', value: 'Contact.$.firstName', description: 'Contact - firstName (STRING)' },
           { label: 'Contact.lastName', value: 'Contact.$.lastName', description: 'Contact - lastName (STRING)' },
+          { label: 'Contact.age', value: 'Contact.$.age', description: 'User age' },
```
monutwilio, line 1900: **No `BOOLEAN` dataType exercised in the dynamic-field dropdown tests**

The `profile_traits` test now adds a `NUMBER` trait (`age`, `totalSpent`) and the `profile_identifiers` test adds a `NUMBER` identifier (`deviceId`). The PR title specifically says "include non str traits", implying `BOOLEAN` is also a supported type — and the per-profile test already asserts `is_subscribed: true` arrives correctly.

However, the dynamic-field tests never verify a `BOOLEAN` trait appears in the choices dropdown. Adding one here (e.g. `isSubscribed: { dataType: 'BOOLEAN', ... }` in the Contact or Loyalty fixture) and asserting the corresponding choice is returned would complete the test coverage for the three primary scalar types (`STRING`, `NUMBER`, `BOOLEAN`).

packages/destination-actions/src/destinations/memora/upsertProfile/__tests__/index.test.ts
```diff
@@ -563,7 +567,10 @@ describe('Memora.upsertProfile', () => {
       expect(profile.traits.Contact.email).toBe('multi@example.com')
       expect(profile.traits.Contact.firstName).toBe('Bob')
       expect(profile.traits.PurchaseHistory.lastPurchaseDate).toBe('2024-02-20')
+      // Non-STRING trait values must be passed through as their native types (not coerced to strings)
+      expect(profile.traits.PurchaseHistory.totalOrders).toBe(42)
       expect(profile.traits.Loyalty.tier).toBe('Gold')
+      expect(profile.traits.Loyalty.isSubscribed).toBe(true)
       expect(profile.traits.Engagement.lastLogin).toBe('2024-03-01')
     })
```
monutwilio, line 575: **The new NUMBER/BOOLEAN round-trip test only covers `perform` (single event), not `performBatch`**

The test added at line 521 lives inside `describe('perform (single profile)')` and calls `testDestination.testAction`. The `performBatch` path goes through the same `buildTraitGroups` helper, so the risk of divergence is low — but since the motivation of this PR is to ensure non-STRING values are passed through correctly, a matching batch test would complete the picture.

Consider adding a test in the `performBatch` section that maps `totalOrders: 42` (NUMBER) and `isSubscribed: true` (BOOLEAN) and asserts both values arrive in the bulk request body with their native types intact. This would also guard against a future regression where someone adds type-coercion logic that only affects the batch path.

packages/destination-actions/src/destinations/memora/upsertProfile/index.ts
monutwilio, line 420: **Unguarded `traitGroup.displayName` (group-level) can produce malformed keys**

A few lines above (line 408), `traitGroupName` is assigned directly from `traitGroup.displayName`. That field is typed as `string` in the interface but comes from an external HTTP response — it could be `undefined` or empty. If it is `undefined`, every `value` in both the identifier and trait choices arrays becomes `'undefined.$.traitName'`, which will render as a broken choice in the UI and fail to parse in `buildTraitGroups`.

Suggested guard (not touched by this PR but in the same function being changed):
```ts
const traitGroupName = traitGroup.displayName
if (!traitGroupName) continue  // skip groups with missing name
```

This is distinct from the existing comment about `trait.displayName` at the individual-trait level — this is the group name that forms the entire key prefix.

arnav777dev (approved):

nk1107 (approved):

joe-ayoub-segment:
deployed

PR #3781, opened 2026-05-07
STRATCONN-6802 - [Facebook Custom Audiences] - journeys bugfix

joe-ayoub-segment:
**Summary:** 

Handles audience membership correctly for events generated via from the journeys_step_entered_track preset. 

Accesses raw event data (context.personas.computation_class) to detect journey step events and overrides audienceMembership to true for all users in the batch (add-only, no removals)
  - Gated behind the facebook-custom-audience-actions-journeys-support feature flag for safe rollout
  - Validates that batches are homogeneous — all journey_step or all non-journey_step — and throws a  PayloadValidationError on mixed batches

**Context for this change:** 

Events from Journeys arrive as track events with event: "Journey Step Entered" and context.personas.computation_class: "journey_step".

Unlike audience events, they do not carry a properties[computation_key] boolean indicating membership. Since the sync action doesn't define fields for computation_class or computation_key, we use rawData to inspect the raw event payload and infer membership (always true for journey steps).

## Testing
  - To be tested in Stage. 
  - Unit tests for getJourneysMemberships — undefined/empty input, all journey, all non-journey, mixed batch error, single event
  - Integration tests verifying feature flag gating — journey events succeed with flag enabled, fail with expected error when flag disabled
  - Verify in staging with feature flag enabled against real journey step events
  - Gradual rollout via feature flag in production

## Security Review

## New Destination Checklist

### DIFF
--- a/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
+++ b/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
@@ -188,8 +188,6 @@ export function getJSON(payloads: Payload[]): AudienceJSON {
 
 export function validate(payloads: Payload[], audienceId: unknown, audienceMemberships?: AudienceMembership[]): string | undefined {
   
-  // TODO handle journey step entered 
-
   if(!Array.isArray(audienceMemberships)){
     return 'Audience membership details for batch missing.'
   }
--- a/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
+++ b/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
@@ -5,14 +5,15 @@ import type { JSONLikeObject, AudienceMembership } from '@segment/actions-core'
 import { StatsContext } from '@segment/actions-core/destination-kit'
 import { processHashing } from '../../../lib/hashing-utils'
 import { PayloadMap, AudienceJSON, FacebookDataRow, RawData } from './types'
-import { BASE_URL } from '../constants'
+import { BASE_URL, FACEBOOK_CUSTOM_AUDIENCE_JOURNEYS_FLAGON } from '../constants'
 import { parseFacebookError, getApiVersion } from '../functions'
 import { FacebookResponseError } from '../types'
 
 /* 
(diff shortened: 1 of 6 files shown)


packages/destination-actions/src/destinations/facebook-custom-audiences/sync/types.ts
```diff
@@ -1,5 +1,24 @@
 import { SCHEMA_PROPERTIES } from './constants'
 import { Payload } from './generated-types'
+import { ExecuteInput } from '@segment/actions-core'
+
+export type RawData = {
+  context?: {
+    personas?: {
+      computation_class?: string
+    }
+  }
```
varadarajan-tw, line 9: This information can be obtained from personasContext that is injected from actons-core

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -4,21 +4,61 @@ import { RequestClient, PayloadValidationError, IntegrationError, MultiStatusRes
...
+export function getJourneysMemberships(
+  rawDatas: RawData[] | undefined
+): boolean[] | undefined {
+  if (!rawDatas || (Array.isArray(rawDatas) && rawDatas.length === 0)) {
+    return undefined
+  }
+
+  const isJourneyStep = rawDatas.map((raw) => raw?.context?.personas?.computation_class === 'journey_step')
+  const allJourney = isJourneyStep.every(Boolean)
+  const noneJourney = isJourneyStep.every((v) => !v)
```
varadarajan-tw, line 27: The pipeline automatically batches by personas.computation_key. If that is the case, are these checks required?

varadarajan-tw: This piece of code could be simplified greatly then. Since journeys only support `Audience Entered`, we can safely assume membership to be `Add` for all joureny events for now.

varadarajan-tw (commented):
I am aligned with the solution approach but have few comments on the implementation.

varadarajan-tw (approved):

joe-ayoub-segment:
deployed

PR #3784, opened 2026-05-12
STRATCONN-6602 - [Google Enhanced Conversions] - Journeysv2 support

joe-ayoub-segment:
Replaced by https://github.com/segmentio/action-destinations/pull/3892

Adds Journeys V2 support to the userList (Customer Match) action and makes the batch multi-status response report the actual operation sent to Google for each payload.

  **Changes**

  1. Journeys V2 / audience-membership support

  Behind the actions-google-ec-audience-membership feature flag, add/remove is now driven by the resolved audienceMembership boolean (from properties[computation_key]) in addition to the existing event-name and sync-mode signals. This enables Journeys V2 step-transition
  events (true → add, false → remove) while leaving Engage and Journeys V1 behavior unchanged whether the flag is on or off.

  **2. Default syncMode → mirror**

  Changed the action's default sync mode from add to mirror so new connections keep the list in sync in both directions rather than append-only. This only affects newly created mappings — existing connections retain their saved sync mode.

  **3. Multi-status sent now carries the per-payload operation**

  Successful batch nodes now report the exact operation sent to Google for that payload ({ create: ... } or { remove: ... }) instead of a generic run-job URL, so add vs. remove is observable per item. This change was required to support the end-to-end tests in #3829,
  where the response type is validated.

  ▎ Note: a pre-existing index mis-attribution in the partial-failure path for mixed add/remove batches is documented inline and tracked separately in STRATCONN-6862 (out of scope here).

  **Testing**

  - New test suites covering Engage, Journeys V1, and Journeys V2 flows across single-event and batch paths, under both flag states (add/remove/mixed, validation errors, and Google partial-failure responses).
  - Updated existing userList tests for the new sent shape.
  - Multi-status response type changes are additionally covered by end-to-end tests in #3829.

  **All tests passing.**

## Security Review

## New Destination Checklist


varadarajan-tw:
@joe-ayoub-segment - could you also please add a monitoring plan on a) how we are going to ensure this doesn't breank any existing flows (are their workspaces that you are planning to manually inspect) b) in case something goes wrong, how are we going to rollback?

joe-ayoub-segment:
> @joe-ayoub-segment - could you also please add a monitoring plan on a) how we are going to ensure this doesn't breank any existing flows (are their workspaces that you are planning to manually inspect) b) in case something goes wrong, how are we going to rollback?

I created a [doc](https://docs.google.com/document/d/1oEjFlcZWXWJlZ998j6PomskycJ5ty4VcfwdnkOjb9mM/edit?tab=t.0) + Datadog notebook. Notebook linked from doc.

joe-ayoub-segment:
replaced by https://github.com/segmentio/action-destinations/pull/3892

PR #3785, opened 2026-05-12
Add orchestrate pipeline skill

Adityaravirala:
## Summary
- Adds /orchestrate slash command that runs the full destination generation pipeline
- Pipeline: refined-actions → endpoint-mapping → spec-generator → generate-destination → test-destination-e2e → deploy-staging
- Coordinates between skills using file paths as handoff — no content pasting between steps
- Confirms with user between each step before proceeding


mdkhan-tw (commented):
Can we remove the aws-kinesis and sqs code to avoid confusion

PR #3792, opened 2026-05-14
feat(klaviyo): add email regex validation to addProfileToList

AnkitSegment:
## Summary
- Adds email regex validation (`/^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/`) to the Klaviyo `addProfileToList` action
- Validates email in both single `perform` (throws `PayloadValidationError`) and batch `performBatch` (returns `PAYLOAD_VALIDATION_FAILED` in multi-status) paths
- Invalid emails are filtered out before sending to Klaviyo's bulk import API, preventing unnecessary API calls

## Test plan
- [x] Unit test: invalid email with single-char TLD rejected in single perform path
- [x] Unit test: batch filters out invalid emails and only sends valid profiles to API
- [x] Existing tests continue to pass (valid emails, phone numbers, external IDs)


packages/destination-actions/src/destinations/klaviyo/addProfileToList/index.ts
```diff
@@ -64,6 +70,9 @@ const action: ActionDefinition<Settings, Payload> = {
     if (!email && !external_id && !phone_number) {
       throw new PayloadValidationError('One of External ID, Phone Number and Email is required.')
```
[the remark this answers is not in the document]
AnkitSegment: @copilot verify

harsh-joshi99 (approved):

joe-ayoub-segment:
Hi @AnkitSegment @harsh-joshi99 will this be tested in stage? If so can we remove the "Ready for release" label until this is done please?

joe-ayoub-segment:
> Hi @AnkitSegment @harsh-joshi99 will this be tested in stage? If so can we remove the "Ready for release" label until this is done please?

Hi @AnkitSegment cc @harsh-joshi99 will you be doing stage testing on this? or do you need to check if the email regex is broad enough? I'll deploy this if you give me the thumbs up.

packages/destination-actions/src/destinations/klaviyo/addProfileToList/index.ts
```diff
@@ -62,7 +68,10 @@ const action: ActionDefinition<Settings, Payload> = {
     } = payload
     const phone_number = processPhoneNumber(initialPhoneNumber, country_code)
     if (!email && !external_id && !phone_number) {
-      throw new PayloadValidationError('One of External ID, Phone Number and Email is required.')
+      throw new PayloadValidationError('One of External ID, Phone Number or Email is required.')
+    }
+    if (!validateEmail(email)) {
+      throw new PayloadValidationError('Email must be a valid email address.')
```
varadarajan-tw, line 74: Why can't we use format:email in field definition?

joe-ayoub-segment:
Not deploying until email formatting clarified by @AnkitSegment .

packages/destination-actions/src/destinations/klaviyo/addProfileToList/index.ts
AnkitSegment: `format: 'email'` (via `ajv-formats`) requires only a 1-character TLD (`[a-z0-9]`), so it accepts `user@domain.c`. Our current `EMAIL_REGEX` enforces `\.[a-zA-Z]{2,}`, which rejects that. Since single-char TLDs aren't real-world valid and Klaviyo's API would likely reject them anyway, I'd prefer to keep the explicit regex to catch this client-side. Also worth noting — `format: 'email'` is already set on the shared `email` field in `properties.ts`, so it applies alongside the regex anyway. Happy to go with just `format: 'email'` if you think the stricter check isn't worth it.

varadarajan-tw (approved):

joe-ayoub-segment:
PR deployed

PR #3799, opened 2026-05-20
STRATCONN-6328/[Actions Hubspot]Fix phone number starting with '+'

monutwilio:
Fixes the bug in HubSpot action-destination where phone number in format +1868658690 was getting converted to 868658690.

https://twilio-engineering.atlassian.net/browse/STRATCONN-6328
## Testing

## Security Review

## New Destination Checklist


packages/destination-actions/src/destinations/hubspot/customEvent/functions/validation-functions.ts
```diff
@@ -73,7 +73,7 @@ function cleanPropObj(
     ) {
       // If the value can be cast to a boolean
       cleanObj[cleanKey] = value.toLowerCase().trim() === 'true'
-    } else if (!isNaN(Number(value))) {
+    } else if (!isNaN(Number(value)) && !(typeof value === 'string' && value.trim().startsWith('+'))) {
```
monutwilio, line 76: **[minor]** `value.trim().startsWith('+')` correctly handles the phone number case, but note that `value` has already been passed through `cleanProp(key)` on the key, not the value — the value itself is still raw. A string like `" +61400000000"` (leading space before `+`) would have `value.trim().startsWith('+')` return `true` and be preserved as-is, which is fine. But consider whether it's worth also guarding hex (`"0x1a"`) and scientific notation (`"1e5"`) strings that `Number()` silently accepts — those could also corrupt non-numeric user data in the same way.

packages/destination-actions/src/destinations/hubspot/customEvent/__tests__/validate.test.ts
```diff
@@ -21,7 +21,8 @@ const payload: Payload = {
       key1: 'value1',
       key2: 'value2'
     },
-    custom_prop_arr: ['value1', 'value2']
+    custom_prop_arr: ['value1', 'value2'],
+    custom_prop_phone: '+61400000000'
   }
```
monutwilio, line 26: **[minor]** Good addition, but the bug fix is only exercised by this single value buried in a large fixture. Consider adding a focused test case (or a few assertions) for the edge cases the guard introduces:
- `'+0'` → stays as string
- `'+1.5'` → stays as string  
- `'123'` → still cast to number (regression guard)
- `'+abc'` → stays as string (was already handled, but worth confirming)

This would make it immediately obvious what behavior the fix guarantees if someone revisits this code.

monutwilio (commented):
## Deep Review Summary

**Overall**: The one-line fix is correct and all 148 HubSpot tests pass. A few things worth addressing before merge:

### PR Description [major]
The body is empty — all template checkboxes are unchecked and there's no summary of what the bug was, what changed, or backward-compat impact. At minimum add:
- What the bug was and how to reproduce
- Confirmation that changing `"+5"` from number→string is intentional and won't break existing customer mappings

### `sendCustomBehavioralEvent` scope [major]
Does `sendCustomBehavioralEvent` have similar property-coercion logic? It uses `flattenObject()` rather than `cleanPropObj()` so it isn't affected, but worth a callout in the PR description so reviewers don't ask.

### Inline comments posted on:
- **validation-functions.ts:76** — hex/scientific-notation strings as a follow-up consideration
- **validate.test.ts** — suggest focused edge-case test assertions for `'+0'`, `'+1.5'`, `'123'` (regression), `'+abc'`

arnav777dev (approved):

PR #3812, opened 2026-05-25
Fix StackAdapt Audience destination overwriting mapping issue

Vanessa-SSY:
Fixes a bug where `upsertProfileMapping` could overwrite a customer's profile field schema when later batches omitted custom traits that earlier batches had included.

Previously, `mappingSchemaV2` was built from custom traits present on the resolved batch payload. During audience backfills, traits are sparse across batches (not every user has every mapped field). A batch without a configured custom trait would send a smaller schema and replace the full mapping on StackAdapt's side. Because `upsertProfiles` and `upsertProfileMapping` are processed asynchronously, this caused downstream fields to disappear after initially being registered.

This change builds `mappingSchemaV2` from the destination subscription mapping (rawMapping in the perform context — same pattern as SFTP), combining:
- Standard fields from `PROFILE_DEFAULT_FIELDS`
- Custom trait keys from `rawMapping.custom_traits `(customer-configured in the Segment UI)

Profile values in `upsertProfiles` are unchanged; only schema registration is stabilized.

## Testing

## Security Review

## New Destination Checklist

### DIFF
--- a/packages/destination-actions/src/destinations/stackadapt-audiences/__tests__/functions.test.ts
+++ b/packages/destination-actions/src/destinations/stackadapt-audiences/__tests__/functions.test.ts
@@ -1,4 +1,69 @@
-import { stringifyJsonWithEscapedQuotes, stringifyMappingSchemaForGraphQL } from '../forwardAudienceEvent/functions'
+import {
+  getConfiguredFieldsToMap,
+  getConfiguredFieldTypes,
+  stringifyJsonWithEscapedQuotes,
+  stringifyMappingSchemaForGraphQL
+} from '../forwardAudienceEvent/functions'
+
+describe('getConfiguredFieldsToMap', () => {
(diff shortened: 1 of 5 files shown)


joe-ayoub-segment:
Hi @Vanessa-SSY , 

I had a look at the PR and I don't think the use of rawData is warranted. I attempted a rewrite of your PR [here](https://github.com/segmentio/action-destinations/pull/3813). 

I kept the changes you made, such as removing the  `getType` and `isDateStr` and `updateFieldsToMapAndFieldTypes` functions. And implemented the logic change to figure out the field schemas based on the entire batch's `custom_traits` values. 

Please take a look and see if it does what you need. If it looks OK can you run it locally and test it please? If it needs changes let me know. 

Also please look through the new / changed unit tests to confirm they test what you need. 

Best regards,
Joe

PR #3813, opened 2026-05-26
[Stackadapt Audiences] - fix schema overwrite during sparse batches 

joe-ayoub-segment:
Based off of Partner raised PR by @Vanessa-SSY https://github.com/segmentio/action-destinations/pull/3812/

Fixes a bug where upsertProfileMapping could overwrite a customer's profile field schema when batches were sparse (not every user has every mapped custom trait).

Previously, mappingSchemaV2 was built from custom trait keys present in the batch payloads. During audience backfills, traits are sparse across batches — a batch without a configured custom trait would send a smaller schema, and since upsertProfileMapping does a full replace, this would cause fields to disappear on StackAdapt's side.

This change:
- Builds mappingSchemaV2 from rawMapping.custom_traits (the destination subscription config) so the full set of configured fields is always sent, regardless of what's in the current batch
- Restricts custom_traits to defaultObjectUI: 'keyvalue:only' to prevent customers from mapping a whole JSON object (which would break the rawMapping key extraction)
- Trims custom trait keys and excludes empty/reserved/duplicate keys
- Reorders mutations so upsertProfileMapping runs before upsertProfiles (StackAdapt processes mutations sequentially, so schema is registered before data is ingested)

## Testing

- [x] Unit tests pass locally
- [x] Vanessa tersted locally with sparse batches (see below)

**Unit tests added in this PR:**
1. should include custom trait in schema when only some payloads have it — sparse batch where only one payload has loyalty_tier, confirms it still appears in schema
2. should exclude reserved computation keys from schema — confirms segment_computation_key/segment_computation_id values are filtered out of mappingSchemaV2
3. should include configured custom traits in schema even when no payloads contain them — payload has no custom trait values at all, confirms schema still includes the configured fields from rawMapping
4. should not duplicate standard fields when they appear in custom_traits — customer maps email/first_name in both standard and custom traits, confirms no duplication in schema

**Tests to be done by Vanessa:**
1. Configure 2-3 custom traits mapped via key-value. Send a batch where all users have all traits. Confirm mappingSchemaV2 includes all custom traits.
2. Send a batch where only some users have a particular custom trait. Confirm the schema still includes it.
3. Send a batch where no users have a particular configured custom trait. Confirm the schema still includes it.
4. Confirm reserved keys (computation_key value, computation_id value) are excluded from the schema.
5. Confirm profile data in upsertProfiles is unaffected — correct values still flow through for users who have them.
6. Confirm custom_traits field in the UI no longer allows mapping to a whole JSON object (only key-value editor available).

### DIFF
--- a/packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/__tests__/index.test.ts
@@ -123,8 +123,8 @@ const mockTrackMapping = {
     birth_date: { '@path': '$.context.traits.birth_date' }
   },
   custom_traits: {
-    custom_trait_1: { '@path': '$.traits.custom_trait_1' },
-    custom_trait_2: { '@path': '$.traits.custom_trait_2' }
+    custom_trait_1: { '@path': '$.context.traits.custom_trait_1' },
+    custom_trait_2: { '@path': '$.context.traits.custom_trait_2' }
   },
   computation_class: { '@path': '$.context.personas.computation_class' },
   computation_key: { '@path': '$.context.personas.computation_key' },
@@ -248,8 +248,8 @@ describe('forwardAudienceEvent', () => {
               input: {
                 advertiserId: 23,
                 externalProvider: \\"segment_io\\",
-                syncId: \\"c89d66b1eed79b7dcd6a2bd3745b81be559bb3077d543cdccb57c023394b0044\\",
-                profiles: \\"[{\\\\\\"userId\\\\\\":\\\\\\"user-id\\\\\\",\\\\\\"email\\\\\\":\\\\\\"test@email.com\\\\\\",\\\\\\"first_name\\\\\\":\\\\\\"Saray\\\\\\",\\\\\\"last_name\\\\\\":\\\\\\"James\\\\\\",\\\\\\"phone\\\\\\":\\\\\\"45678765\\\\\\",\\\\\\"address\\\\\\":\\\\\\"123 Barn St\\\\\\",\\\\\\"city\\\\\\":\\\\\\"NYC\\\\\\",\\\\\\"country\\\\\\":\\\\\\"USA\\\\\\",\\\\\\"state\\\\\\":\\\\\\"NY\\\\\\",\\\\\\"postal_code\\\\\\":\\\\\\"29323\\\\\\",\\\\\\"timezone\\\\\\":\\\\\\"EST\\\\\\",\\\\\\"birth_date\\\\\\":\\\\\\"1990-06-15\\\\\\",\\\\\\"audienceId\\\\\\":\\\\\\"aud_123\\\\\\",\\\\\\"audienceName\\\\\\":\\\\\\"first_time_buyer\\\\\\",\\\\\\"action\\\\\\":\\\\\\"exit\\\\\\"}]\\"
+                syncId: \\"7032157de1ee57b5a7f9512fb65fbe2dff58e53c998cc6095199dac55741e6f8\\",
+                profiles: \\"[{\\\\\\"userId\\\\\\":\\\\\\"user-id\\\\\\",\\\\\\"email\\\\\\":\\\\\\"test@email.com\\\\\\",\\\\\\"first_name\\\\\\":\\\\\\"Saray\\\\\\",\\\\\\"last_name\\\\\\":\\\\\\"James\\\\\\",\\\\\\"phone\\\\\\":\\\\\\"45678765\\\\\\",\\\\\\"address\\\\\\":\\\\\\"123 Barn St\\\\\\",\\\\\\"city\\\\\\":\\\\\\"NYC\\\\\\",\\\\\\"country\\\\\\":\\\\\\"USA\\\\\\",\\\\\\"state\\\\\\":\\\\\\"NY\\\\\\",\\\\\\"postal_code\\\\\\":\\\\\\"29323\\\\\\",\\\\\\"timezone\\\\\\":\\\\\\"EST\\\\\\",\\\\\\"birth_date\\\\\\":\\\\\\"1990-06-15\\\\\\",\\\\\\"custom_trait_1\\\\\\":\\\\\\"custom_value_1\\\\\\",\\\\\\"custom_trait_2\\\\\\":\\\\\\"custom_value_2\\\\\\",\\\\\\"audienceId\\\\\\":\\\\\\"aud_123\\\\\\",\\\\\\"audienceName\\\\\\":\\\\\\"first_time_buyer\\\\\\",\\\\\\"action\\\\\\":\\\\\\"exit\\\\\\"}]\\"
               }
             ) {
               userErrors {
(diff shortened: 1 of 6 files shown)


packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
```diff
@@ -40,11 +41,21 @@ export async function send(request: RequestClient, payloads: Payload[], settings
...
 
+  // Build schema from the union of all custom trait keys across all payloads.
+  // This ensures the schema is stable even when individual payloads are sparse.
+  for (const p of payloads) {
+    if (p.custom_traits) {
+      for (const key of Object.keys(p.custom_traits)) {
+        if (!key || reservedKeys.includes(key) || fieldsToMap.has(key)) continue
+        fieldsToMap.add(key)
+        fieldTypes[key] = 'STRING'
+      }
```
Vanessa-SSY: In this case, we're sending the intersection of client-configured mapping and actual mappings of the fields that the batch contains, right? Client mapping alone doesn’t register a field unless at least one profile in that batch resolves a value into custom_traits, right?
It should be okay to not send the full client-configured mapping. The next time the batch with new fields are sent, mappings will be included, right?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/__tests__/index.test.ts
```diff
@@ -301,46 +301,174 @@ describe('forwardAudienceEvent', () => {
...
+        return body
+      })
+      .reply(200, { data: { success: true } })
+
+    const event: Partial<SegmentEvent> = {
+      userId: mockUserId,
+      type: 'identify',
+      traits: {
+        email: mockEmail,
+        first_time_buyer: true,
```
Vanessa-SSY, line 389: Just to fresh my memory: the value of these reserved computation keys will always be true/false which implies an audience enter/exit event, right?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
joe-ayoub-segment: Hi @Vanessa-SSY ,
Here's the main difference between the 2 PRs: 

My PR (#3813): Builds the schema from the union of custom trait keys actually present across the batch payloads. Always accurate for what's in the current batch, but if a batch is sparse (missing some traits), the schema sent will be smaller than the full configured set.

Your PR (#3812): Builds the schema from rawMapping (i.e the destination config), so it always sends the full set of configured fields regardless of what's in the batch. Solves the sparse batch problem, but could break if the customer maps `custom_traits` to a whole object. 

For example: 

Keys mapped separately (key-value UI):
```
  {
    "custom_trait_1": { "@path": "$.traits.custom_trait_1" },
    "custom_trait_2": { "@path": "$.traits.custom_trait_2" },
    "loyalty_tier": "gold"
  }
  Object.keys() → ["custom_trait_1", "custom_trait_2", "loyalty_tier"] — the actual field names.
```

Whole JSON object mapped (variable/freeform):
```
  { "@path": "$.traits" }
  Object.keys() → ["@path"] — just the directive key, not the trait names.
```

The field has defaultObjectUI: 'keyvalue' which means the UI defaults to the key-by-key approach, but nothing prevents a customer from switching to variable mode and mapping the whole object. If they do, your approach might break.

So question for you: 

Is `upsertProfileMapping` additive (merges new fields into the existing schema) or a full replace (overwrites the entire schema each time it's called). This will determine if my approach of sending only what's in the current batch is safe, or whether we need to send the full schema every time to avoid fields disappearing.

Based on your comment, it sounds like `upsertProfileMapping` might be additive; can you confirm?

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/__tests__/index.test.ts
joe-ayoub-segment: computation_key is the snake_cased name of the audience.
computation_id is the internal Segment identifier for the audience. 

Then,  traits_or_props[computation_key] is a boolean which indicates if the user is being added (true) or removed (false) from the audience.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
Vanessa-SSY: `upsertProfileMapping` does a **full replace** and that's why we'd prefer sending the full schema every time. What should help in this case is we can reverse the order of calling `upsertProfiles` and `upsertProfileMapping` since Stackadapt handles queries in the same request sequentially. If we call `upsertProfileMapping` before `upsertProfiles`, we should be able get the correct mapping before ingestion. 
However there's still risk that `upsertProfileMapping` returns handled error without raising,  `upsertProfiles` could still runs.

Vanessa-SSY:
I've run unit testing and manual testing locally on the changes in this PR. The batch-union behavior is working as intended

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
joe-ayoub-segment: hi @Vanessa-SSY - thanks for confirming it's a full replace, that explains the original bug.

So I think your approach is likely the way to go. 
I did a search for customers who already have the integration configured and didn't find any who were mapping the `custom_traits` to a JSON object, which is good. 

So here's what I'm proposing: 

1. Let's go with your approach and use the rawMappings. 
2. Let's update the `custom_traits` field to only work with defaultObjectUI: 'keyvalue:only'. This will prevent customers from mapping it to a JSON object. It means that customers will need to be deliberate when mapping - i.e they will have to map every individual field they want to pass to Stackadapt. 

If you are happy with this, I'll update this current PR accordingly. I'll just need you to do a test once I've updated the code. Would that be OK?

Vanessa-SSY: Yes, we're aligned on this. Thanks for your time and effort! Please let me know once the changes is done so I can do testing locally.

joe-ayoub-segment: I went ahead and updated the PR to work with your approach @Vanessa-SSY . I updated the description to include some tests for you to do, if you have time?

I also added some unit tests. 

If you can take a look, review and let me know if you are happy with the changes - and confirm they work - I'll label this for deploy.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/__tests__/index.test.ts
```diff
@@ -301,46 +301,221 @@ describe('forwardAudienceEvent', () => {
...
-                }
-              }
-        }",
+    expect(requestBody.query).toContain('upsertProfiles')
+    expect(requestBody.query).toContain('upsertProfileMapping')
+    expect(requestBody.query).toContain('custom_trait_1')
+    expect(requestBody.query).toContain('custom_trait_2')
+  })
+
+  it('should include custom trait in schema when only some payloads have it', async () => {
```
Vanessa-SSY, line 310: This test doesn't apply anymore ,right? We should include custom traits regardless payload.

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
Vanessa-SSY: Could you also reverse the order of the mutations so we're calling `upsertProfileMapping` before `upsertProfiles`?

Vanessa-SSY:
I've tested the code changes locally and everything is working as intended (all mappings are sent regardless profile batch). Please update the query order mentioned [here](https://github.com/segmentio/action-destinations/pull/3813#discussion_r3306309741). This minor change shouldn't affect the manual testing I've done so feel free to deploy once that's updated. Thanks for your help @joe-ayoub-segment

packages/destination-actions/src/destinations/stackadapt-audiences/forwardAudienceEvent/functions.ts
joe-ayoub-segment: That's been done @Vanessa-SSY - can you do the testing please?

Vanessa-SSY:
Just ran test locally and it's working perfectly. Feel free to deploy. Thanks!

Vanessa-SSY:
May I know when's the next deploy? @joe-ayoub-segment

joe-ayoub-segment:
Hi @Vanessa-SSY - there's still deployment pipeline work ongoing so I can't deploy code yet unfortunately. 
I've a lot of PRs I need to get deployed but for the moment there's nothing I can do. 

I'll update you once I know more. 
Best regards,
Joe

Vanessa-SSY:
Hi @joe-ayoub-segment , is there any updates on this issue? Is deployment still blocked?

joe-ayoub-segment:
> Hi @joe-ayoub-segment , is there any updates on this issue? Is deployment still blocked?

Hi @Vanessa-SSY yes the freeze is still in place. Once it's lifted I'll schedule this to go out - however there's a lot of competing changes waiting to be shipped so it will take a little longer than usual. 

Applogies for the delay. 
Joe

Vanessa-SSY:
Hi @joe-ayoub-segment, do you have an estimate on when the deployment will be unblocked? For my understanding of this issue, is this blocking deployment to all features in this repo?

joe-ayoub-segment:
> Hi @joe-ayoub-segment, do you have an estimate on when the deployment will be unblocked? For my understanding of this issue, is this blocking deployment to all features in this repo?

Hi @Vanessa-SSY I don't yet, unfortunately.

Vanessa-SSY:
@joe-ayoub-segment , could you help confirm if this issue is affecting our stackadapt destination only or cross all destinations?

joe-ayoub-segment:
> @joe-ayoub-segment , could you help confirm if this issue is affecting our stackadapt destination only or cross all destinations?

Hi @Vanessa-SSY - based on my understand of the issue it's more a nuance of the way Stackadapt works than anything else. 

The majority of platforms will accept partial user profile data and merge it into the profile. But it seems like Stackadapt overwrites everything with the blob it gets sent. Does that sound correct to you?

Best regards,
Joe

Vanessa-SSY:
> > @joe-ayoub-segment , could you help confirm if this issue is affecting our stackadapt destination only or cross all destinations?
> 
> Hi @Vanessa-SSY - based on my understand of the issue it's more a nuance of the way Stackadapt works than anything else.
> 
> The majority of platforms will accept partial user profile data and merge it into the profile. But it seems like Stackadapt overwrites everything with the blob it gets sent. Does that sound correct to you?
> 
> Best regards, Joe

We do merge profiles data on our end. What's getting overwritten/replaced is the field mapping configuration. 

Here's the scenario. Say the client has configured mappings for `field1`, `field2`, and `field3` in Segment.
The old destination code only sends a field's mapping if a profile in the batch actually has that attribute. So:
- A batch with **profile1** (has `field1`, `field2` only) → we send mappings for `field1`, `field2`.
- A later batch with **profile2** (has all three) → we send mappings for `field1`, `field2`, `field3`.

The problem is on our side: upsertProfileMapping and upsertProfiles are async. So **profile2** can get ingested while the mapping in effect is still the earlier `field1/field2`-only schema from **profile1** — and `field3` on **profile2** gets dropped during ingestion.

The fix in this PR is to always send the full mapping schema from the client's UI configuration, regardless of whether each field has a value in the current batch. That keeps the ingested schema complete and removes the race entirely.
Happy to hop on a call if it'd help. Thanks!
Thanks! @joe-ayoub-segment

Vanessa-SSY:
Do you mind sharing the error message you're getting for deploying this PR so I have more insight into what's happening?
@joe-ayoub-segment

joe-ayoub-segment:
Discussed this PR with @Vanessa-SSY on Slack. Notes below. 

Full slack convo. https://segmentpartners.slack.com/archives/C02KL2J8XJA/p1782235222044309

image
image

Vanessa-SSY:
> Discussed this PR with @Vanessa-SSY on Slack. Notes below.
> 
> Full slack convo. https://segmentpartners.slack.com/archives/C02KL2J8XJA/p1782235222044309
> 
> image image

Replied to your questions on slack: https://stackadapt.slack.com/archives/C02KL2J8XJA/p1783351380397999?thread_ts=1782235222.044309&cid=C02KL2J8XJA and left a question regarding the proposed solution.

Vanessa-SSY:
Hi @joe-ayoub-segment , I replied with more questions in the slack thread. Do you possibly have time to go over this issue together in a call? I'm available anytime Wednesday and Thursday and I'm in EST. Please let me know your availability. Thanks

Vanessa-SSY:
Hi @joe-ayoub-segment, I proposed a different time for Monday, can you check if that works for you?

joe-ayoub-segment:
Replace with https://github.com/segmentio/action-destinations/pull/3879

PR #3823, opened 2026-06-04
STRATCONN-6845 [Facebook Custom Audience]: stop a single invalid phone from failing the whole batch (feature-flagged)

abhandage:
## Summary

Facebook Custom Audiences failed an **entire batch** when a single row had a phone number that normalizes to an empty string (e.g. the placeholder `+0000000000`). `normalizePhone()` strips non-numeric characters and leading zeros, producing `''`, which reached `SmartHashing.hash('')` and threw `Cannot hash an empty string`. Because that throw escaped `getData()` before the request was built, **all** events in the batch were rejected instead of just the one bad row. Reported by Nestle (STRATCONN-6845 / STRATCONN-6848).

This PR routes phone hashing through a new `getHashedPhone(phone, features)` helper in `getData()`:

- **`hashPhone()`** (fixed behavior): pre-hashed sha256/hex values pass through unchanged; otherwise the value is normalized and, if normalization leaves an empty string, it is discarded to `''` instead of throwing — so one bad phone no longer fails the whole batch.
- **Feature-flag gated:** the fix is behind the `fb-custom-audience-batch-discard-fix` flag (`centrifuge-destinations` family, Workspace ID). **Flag OFF → original behavior (unchanged); Flag ON → fix.** This enables a controlled, per-workspace rollout for this high-volume destination.

### Behavior

| Input | Flag OFF (default) | Flag ON |
|---|---|---|
| `+0000000000` in a batch | throws → **whole batch fails** (today's behavior) | that row → `''`, rest of the batch succeeds |
| valid phone | hashed (unchanged) | hashed (unchanged) |
| pre-hashed sha256 phone | passed through | passed through |

### Audience with faulty email
image

### Event Delivery
image

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality — flag ON discards an empty-normalizing phone to `''`; flag OFF throws `Cannot hash an empty string`; pre-hashed passthrough (including a digit-free hash that would otherwise normalize away).
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) — `serve` + curl against the `sync` action with a mocked Graph API, confirming an all-good batch succeeds and a bad-phone batch behaves per flag state.
- [x] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change. — With the flag OFF the original code path runs verbatim; no new fields were added, and valid/pre-hashed phones hash identically in both states.

## Security Review

- [x] **Reviewed all field definitions** for sensitive data (API keys, tokens, passwords, client secrets) and confirmed they use `type: 'password'` — no new fields added by this change. Phone (PII) continues to be SHA-256 hashed before it leaves Segment (Facebook's required format), and already-hashed values are still detected and not re-hashed.

## New Destination Checklist

N/A — this is a bug fix to an existing destination, not a new destination.

### DIFF
--- a/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
+++ b/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
@@ -215,7 +215,7 @@ export function getData(payloads: Payload[]): FacebookDataRow[] {
     const row: FacebookDataRow = [
       externalId ?? '',
       email ? processHashing(email.trim().toLowerCase(), 'sha256', 'hex') : '',
-      phone ? processHashing(phone, 'sha256', 'hex', normalizePhone) ?? '' : '',
+      phone ? (normalizePhone(phone) ? processHashing(phone, 'sha256', 'hex', normalizePhone) ?? '' : '') : '',
       year ? processHashing(year.trim(), 'sha256', 'hex') ?? '' : '',
       month ? processHashing(normalizeMonth(month), 'sha256', 'hex') ?? '' : '',
       day ? processHashing(day.trim(), 'sha256', 'hex') ?? '' : '',
--- a/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
+++ b/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
@@ -11,8 +11,8 @@ import {
 import type { JSONLikeObject, AudienceMembership } from '@segment/actions-core'
 import { StatsContext } from '@segment/actions-core/destination-kit'
 import { processHashing, isAlreadyHashed } from '../../../lib/hashing-utils'
-import { PayloadMap, AudienceJSON, FacebookDataRow, RawData } from './types'
-import { BASE_URL, FACEBOOK_CUSTOM_AUDIENCE_JOURNEYS_FLAGON, FLAGON_NAME_BATCH_DISCARD_FIX } from '../constants'
+import { PayloadMap, AudienceJSON, FacebookDataRow } from './types'
+import { BASE_URL, FLAGON_NAME_BATCH_DISCARD_FIX } from '../constants'
 import { parseFacebookError, getApiVersion } from '../functions'
 import { FacebookResponseError } from '../types'
 
--- a/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
+++ b/packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
@@ -10,9 +10,9 @@ import {
 } from '@segment/actions-core'
 import type { JSONLikeObject, AudienceMembership } from '@segment/actions-core'
 import { StatsContext } from '@segment/actions-core/destination-kit'
-import { processHashing } from '../../../lib/hashing-utils'
-import { PayloadMap, AudienceJSON, FacebookDataRow } from './types'
-import { BASE_URL } from '../constants'
+import { processHashing, isAlreadyHashed } from '../../../lib/hashing-utils'
+import { PayloadMap, AudienceJSON, FacebookDataRow, RawData } from './types'
+import { BASE_URL, FACEBOOK_CUSTOM_AUDIENCE_JOURNEYS_FLAGON } from '../constants'
 import { parseFacebookError, getApiVersion } from '../functions'
 import { FacebookResponseError } from '../types'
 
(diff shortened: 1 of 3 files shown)


nk1107:
`hashPhone()` promises that pre-hashed values are passed through unchanged, but it doesn't trim whitespace before `isAlreadyHashed()`. A value like `'  <64-hex>  '` would fail the pre-hash detection (length ≠ 64 / regex `^[a-f0-9]{64}$` doesn't match), hit `normalizePhone()`, and likely become `''`, silently dropping a valid hash.

Trim once up front and use the trimmed value for both `isAlreadyHashed()` and `normalizePhone()`, and return the trimmed value in the pass-through path.

```ts
export function hashPhone(phone?: string): string {
  const trimmed = phone?.trim()
  if (!trimmed) return ''
  if (isAlreadyHashed(trimmed, 'sha256', 'hex')) return trimmed
  const normalized = normalizePhone(trimmed)
  return normalized ? processHashing(normalized, 'sha256', 'hex') : ''
}
```

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -267,7 +267,7 @@ export function getData(payloads: Payload[]): FacebookDataRow[] {
     const row: FacebookDataRow = [
       externalId ?? '',
       email ? processHashing(email.trim().toLowerCase(), 'sha256', 'hex') : '',
-      phone ? processHashing(phone, 'sha256', 'hex', normalizePhone) ?? '' : '',
+      hashPhone(phone),
```
mdkhan-tw: Do we want to add some feature flag for this change? I understand that this is a simple change but we can never be too sure

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -294,6 +294,20 @@ export function normalizePhone(value: string): string {
...
 }
 
+/**
+ * Hashes a phone number for Facebook Custom Audiences.
+ * - Pre-hashed values (valid sha256/hex) are passed through unchanged.
+ * - Otherwise the value is normalized; if normalization leaves an empty string
+ *   (e.g. '+0000000000'), it is treated as missing and returns '' rather than
+ *   letting SmartHashing.hash() throw 'Cannot hash an empty string'.
+ */
+export function hashPhone(phone?: string): string {
```
mdkhan-tw, line 266: Do we have some contract that if the input is already hashed, we will not hash it again?

nk1107: Yes it is mentioned in the docs: https://www.twilio.com/docs/segment/connections/destinations#hashing
and it is implemented here: https://github.com/segmentio/action-destinations/blob/406b355e7f0c5829b486f7da04bb032420c2dc97/packages/destination-actions/src/lib/hashing-utils.ts#L88

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
nk1107: Sure will add one

mdkhan-tw (dismissed):

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -294,6 +294,34 @@ export function normalizePhone(value: string): string {
...
+ * - Otherwise the value is normalized; if normalization leaves an empty string
+ *   (e.g. '+0000000000'), it is discarded to '' instead of letting SmartHashing.hash()
+ *   throw 'Cannot hash an empty string' and fail the whole batch.
+ */
+export function hashPhone(phone?: string): string {
+  if (!phone) return ''
+  if (isAlreadyHashed(phone, 'sha256', 'hex')) return phone
+  const normalized = normalizePhone(phone)
+  return normalized ? processHashing(normalized, 'sha256', 'hex') : ''
+}
```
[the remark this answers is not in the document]
nk1107: processHashing() is a centralised util used by many destinations. Adding the fix there makes the testing scope very wide. This can be viewed as future bug fixes

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -294,6 +294,34 @@ export function normalizePhone(value: string): string {
   return removedNonNumveric.replace(/^0+/, '')
 }
```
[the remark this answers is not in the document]
nk1107: not part of PR

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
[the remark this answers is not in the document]
nk1107: This can be addressed if this required from the customer

monutwilio (approved):

joe-ayoub-segment:
Hi @abhandage PR deployed

PR #3825, opened 2026-06-05
fix(core): filter audienceMembership to match schema-filtered payloads in executeBatch

abhandage:
## Summary

- When one or more events in a batch fail schema validation, the invalid payloads are removed from the `payloads` array before `performBatch` is called
- The `audienceMembership` array was not trimmed to match, causing `validate()` inside `send()` to return `"Audience membership details count does not match batch payload count."` — failing **every** event in the batch instead of just the invalid one
- Fix: filter `audienceMembership` using the same `invalidPayloadIndices` set used to filter payloads, keeping the two arrays in sync

**Root cause** ([`action.ts#L476`](https://github.com/segmentio/action-destinations/blob/main/packages/core/src/destination-kit/action.ts#L476)):
```ts
// before
const audienceMembership = bundle.data.map((d) => resolveAudienceMembership(d, syncMode))

// after
const audienceMembership = bundle.data
  .map((d) => resolveAudienceMembership(d, syncMode))
  .filter((_, i) => !invalidPayloadIndices.has(i))
```

Also: updates error type for audience-related validation errors (missing audience ID, missing membership details) from `PAYLOAD_VALIDATION_FAILED` to `INVALID_AUDIENCE_MEMBERSHIP` to distinguish them from schema validation failures.

## Customer Impact

Any Facebook Custom Audiences sync batch containing even one schema-invalid event (e.g. a user record missing `externalId`) fails entirely — valid users in the same batch are not added to or removed from the audience. Customers see complete batch drops rather than single-event errors, leading to audience under-population and missed ad targeting.

## Unit tests updated
  - **upsert.test.ts**: Updated "missing audience ID" assertions to expect `INVALID_AUDIENCE_MEMBERSHIP` error type.
  - **mirror.test.ts**: Updated "missing audience membership details" and "missing audience ID" assertions to expect `INVALID_AUDIENCE_MEMBERSHIP` error type.
  - **delete.test.ts**: Updated "missing audience ID" assertions (RETL and Engage) to expect `INVALID_AUDIENCE_MEMBERSHIP` error type.

## Unit tests added
  - **upsert.test.ts**: "should succeed for valid events when one event is filtered by schema validation" — batch of 4 events (2 adds, 1 schema failure, 1 removal) verifying that the audienceMembership array is correctly filtered alongside payloads so
  valid events still succeed.
  - **audience-membership.test.ts** (core): Unit tests for `resolveAudienceMembership` logic.

## Staging test plan
  - [ ] Send a batch where one event has a missing required field (e.g. no `externalId`) — confirm valid events in the same batch succeed and the invalid one returns a 400.
  - [ ] Send a batch with no audience ID configured — confirm all events return `INVALID_AUDIENCE_MEMBERSHIP` errors (not `PAYLOAD_VALIDATION_FAILED`).
  - [ ] Send a mixed add/remove batch (Engage-style with `true`/`false` membership) — confirm adds hit POST and removes hit DELETE on Facebook's API.
  - [ ] Send a batch where all events are valid — confirm normal happy-path behavior is unaffected.

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude-cohorts/syncAudience/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/amplitude-cohorts/syncAudience/__tests__/index.test.ts
@@ -58,8 +58,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
       const responses = await testDestination.testAction('syncAudience', {
         event,
         settings,
-        useDefaultMappings: true,
-        features: { 'actions-core-audience-membership': true }
+        useDefaultMappings: true
       })
 
       expect(responses.length).toBe(1)
@@ -110,8 +109,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
       const responses = await testDestination.testAction('syncAudience', {
         event,
         settings,
-        useDefaultMappings: true,
-        features: { 'actions-core-audience-membership': true }
+        useDefaultMappings: true
       })
 
       expect(responses.length).toBe(1)
@@ -164,8 +162,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
         useDefaultMappings: true,
         mapping: {
           amplitude_id: 'amp789'
-        },
-        features: { 'actions-core-audience-membership': true }
+        }
       })
 
       expect(responses.length).toBe(1)
@@ -193,8 +190,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
         testDestination.testAction('syncAudience', {
           event,
           settings,
-          useDefaultMappings: true,
-          features: { 'actions-core-audience-membership': true }
+          useDefaultMappings: true
         })
       ).rejects.toThrowError('ID Type must be specified in Audience Settings')
     })
@@ -225,8 +221,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
           useDefaultMappings: true,
           mapping: {
             user_id: undefined
-          },
-          features: { 'actions-core-audience-membership': true }
+          }
         })
       ).rejects.toThrowError("No User Identifier of type User ID found in payload. Each payload must have a unique ID for the specified ID Type.")
     })
@@ -337,8 +332,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
       const responses = await testDestination.testBatchAction('syncAudience', {
         events,
         settings,
-        useDefaultMappings: true,
-        features: { 'actions-core-audience-membership': true }
+        useDefaultMappings: true
       })
 
       expect(responses.length).toBe(2)
@@ -409,8 +403,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
       const responses = await testDestination.testBatchAction('syncAudience', {
         events,
         settings,
-        useDefaultMappings: true,
-        features: { 'actions-core-audience-membership': true }
+        useDefaultMappings: true
       })
 
       expect(responses.length).toBe(1)
@@ -464,8 +457,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
           ...settings,
           endpoint: 'europe'
         },
-        useDefaultMappings: true,
-        features: { 'actions-core-audience-membership': true }
+        useDefaultMappings: true
       })
 
       expect(responses.length).toBe(1)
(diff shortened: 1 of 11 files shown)


abhandage:
Since payloads are the already-mapped Payload objects (no `event`, `type`, or `context.personas`), `resolveAudienceMembership` returned undefined for every event — breaking membership resolution across every upsert/delete/journeys test. The correct fix is `bundle.data.map(...).filter((_, i) => !invalidPayloadIndices.has(i))` — resolve from the raw events (preserving event/type/context.personas), then drop the indices that failed schema validation.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -48,37 +55,42 @@ export async function send(
...
-  }
-
-  const audienceId = getAudienceId(payloads[0], hookOutputs)
-  const errorMessage = validate(payloads, audienceId, audienceMemberships)
 
-  if (errorMessage) {
-    return returnErrorResponse(msResponse, payloads, isBatch, errorMessage, ErrorCodes.PAYLOAD_VALIDATION_FAILED)
+    const errorMessage = validate(payloads, audienceId, audienceMemberships)
+    if (errorMessage) {
+      return returnErrorResponse(msResponse, payloads, isBatch, errorMessage, ErrorCodes.PAYLOAD_VALIDATION_FAILED)
```
abhandage: We probably need to return a different error code cc: @varadarajan-tw @joe-ayoub-segment

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/__tests__/upsert.test.ts
```diff
@@ -404,5 +404,94 @@ describe('FacebookCustomAudiences.sync - syncMode: upsert', () => {
         body: { externalId: 'user-2', email: 'user2@example.com', enable_batching: true, batch_size: 10000 }
       })
     })
+
+    it('should not produce audience membership count mismatch when one event fails schema validation', async () => {
+      // Event at index 1 has no properties.externalId, so externalId resolves to undefined and
+      // is stripped by removeEmptyValues. externalId is required — that payload is filtered out
+      // by the schema validator before performBatch is called. The audienceMembership array is
+      // built from the already-filtered payloads array, so its length always matches and
+      // validate() does not return "Audience membership details count does not match batch payload count."
```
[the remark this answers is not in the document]
joe-ayoub-segment: Nit.

packages/destination-actions/src/destinations/amplitude-cohorts/syncAudience/__tests__/index.test.ts
```diff
@@ -58,8 +58,7 @@ describe('Amplitude Cohorts - syncAudience', () => {
       const responses = await testDestination.testAction('syncAudience', {
         event,
         settings,
-        useDefaultMappings: true,
-        features: { 'actions-core-audience-membership': true }
+        useDefaultMappings: true
```
varadarajan-tw, line 61: Is this change necessary as part of this PR?

joe-ayoub-segment: It's removing unused flags from unit tests. I think it's valid to remove them as the tests should pass with the flag. 

It was pointed out by Copilot and will likely just keep triggering comments if not removed.

varadarajan-tw: okay.

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
```diff
@@ -71,14 +78,20 @@ export async function send(
...
+    } else if (audienceMembership === true) {
       addMap.set(index, payload)
-    } 
-    else if (audienceMembership === undefined){
-      setErrorResponse(msResponse, payload, 400, index, isBatch, "Audience membership details missing", ErrorCodes.PAYLOAD_VALIDATION_FAILED)
+    } else if (audienceMembership === undefined) {
+      setErrorResponse(
+        msResponse,
+        payload,
+        400,
```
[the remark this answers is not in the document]
varadarajan-tw: @joe-ayoub-segment - should we fix this?

varadarajan-tw (approved):

packages/destination-actions/src/destinations/facebook-custom-audiences/sync/functions.ts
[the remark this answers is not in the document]
varadarajan-tw: The call site at line 69–72 already passes `ErrorCodes.INVALID_AUDIENCE_MEMBERSHIP` explicitly to `returnErrorResponse`. `validate()` only returns a string message — the error code is determined by the caller, not by `validate` itself. No change needed here.

PR #3827, opened 2026-06-09
fix(hubspot): allow string values to match datetime fields in schema comparison

harsh-joshi99:
## Summary

- HubSpot datetime fields accept plain string values such as Unix epoch timestamps (e.g. `"1780382814"` or `"1780382814000"`), but our schema comparison in `compareProps()` was classifying these as `string:text` and rejecting them against HubSpot's `datetime:date` fields
- Added a coercion rule allowing `string:text` (payload-inferred) to match `datetime:date` (HubSpot schema), consistent with the existing rules for `string:text → enumeration:select` and `string:text → string:textarea`
- HubSpot is the right place to validate the actual value — if someone sends an invalid string, HubSpot will 400 them directly

JIRA: https://twilio-engineering.atlassian.net/browse/STRATCONN-6042

## Root Cause

BigQuery RETL serializes timestamp columns as Unix epoch strings. When these arrive in the `upsertObject` action, `format()` correctly identifies them as plain strings (they don't match the ISO datetime regex), tagging them `string:text`. `compareProps()` then sees `string:text` vs `datetime:date` and throws a type mismatch error before the request ever reaches HubSpot.

## Test plan

Local Testing:
Screenshot 2026-06-09 at 11 13 22 AM

- [x] Added unit test to `upsert-property-mismatch.test.ts` confirming epoch string `"1780382814"` passes schema comparison against a `datetime:date` HubSpot field without throwing
- [x] Added unit tests to new `schema-functions.test.ts` covering `format()` type detection for all value types (date string, datetime string, epoch string, number, boolean, plain string)
- [x] All 41 existing `upsertObject` tests pass
- [x] Verified end-to-end in sandbox: `"1780382814000"` sent to `hs_appointment_start` (`datetime:date`) resulted in a 201 Created with HubSpot.


mdkhan-tw (approved):

joe-ayoub-segment:
@harsh-joshi99 PR deployed

PR #3828, opened 2026-06-09
Add Wingify browser destination integration.

zeeshan-vwo:
## Summary

Adds a new **Wingify Web Mode (Actions)** browser destination (`actions-wingify-web`) for device-mode integration with the Wingify experimentation platform.

This destination:
- Loads the **Wingify Async V3.0 SmartCode** from `https://edge.wingify.net`
- Forwards Segment **track** events to Wingify via `Wingify.event`
- Forwards Segment **identify** traits to Wingify via `Wingify.visitor`
- Prefixes event/attribute names with `segment.` for Data360 compatibility

### Changes

| Area | Details |
|------|---------|
| **Package** | `@segment/analytics-browser-actions-wingify` |
| **Slug** | `actions-wingify-web` |
| **Metadata ID** | `69a2f1c8e4b7d63901ab5678` |
| **Actions** | `trackEvent`, `identifyUser` |
| **Settings** | Wingify Account ID, Settings Tolerance, Add Asynchronous SmartCode |

### Files added

- `packages/browser-destinations/destinations/wingify/` — full destination package (init script, actions, types, utilities, unit tests)
- `packages/destinations-manifest/src/index.ts` — manifest registration
- `packages/destinations-manifest/package.json` — package dependency

### SmartCode

Uses Wingify's async SmartCode (`_wingify_code`, `edge.wingify.net/tag/{accountId}.js`) with preconnect and configurable settings tolerance.

---

## Testing

_Include any additional information about the testing you have completed to ensure your changes behave as expected. For a speedy review, please check any of the tasks you completed below during your testing._

  - Destination initialization (SmartCode load + Wingify object)
  - `trackEvent` (with and without properties)
  - `identifyUser` (attribute formatting)
  - _N/A — this is a new destination, not a change to an existing live integration._

---

## Security Review

  - No sensitive auth fields in this browser destination. `wingifyAccountId` is a numeric account identifier (not a secret). No passwords or API keys are collected.

---

## New Destination Checklist

- [ ] Extracted all action API versions to `versioning-info.ts` file. [example](https://github.com/segmentio/action-destinations/blob/main/packages/destination-actions/src/destinations/facebook-conversions-api/versioning-info.ts)
  - _N/A — browser destination with no versioned cloud API endpoints._


joe-ayoub-segment:
Thanks for the PR @zeeshan-vwo . 
I'll let Copilot do a review first and will do a manual review afterwards.

joe-ayoub-segment:
Hi @zeeshan-vwo , 

I had a quick scan through to see what the integration does. Seems straightforward. 
The actions are fine. 
The main problems people have with new Web destinations is the initialization of their client libraries. 

Have you managed to test this locally? Does your library load as expected? When you call track() or identify(), do the correct calls in your library happen?

best regards,
Joe

zeeshan-vwo:
Hi @joe-ayoub-segment ,
Yes, I have tested it locally and it's working fine.

zeeshan-vwo:
@joe-ayoub-segment can you please review and merge this PR. Please let me know if anything is needed from my end.

joe-ayoub-segment:
Hi @zeeshan-vwo apologies for the delay - we have an unusually large backlog of PRs to get through. I can't provide a precice date for when this will be fully reviewed, yet. 

It is on my list so I will get to it. Can you clarify the urgency please?

Kind regards,
Joe

zeeshan-vwo:
Hi @joe-ayoub-segment ,

Thanks for the update.

The urgency is that we're currently rebranding from VWO to Wingify, and this PR contains the required branding changes. Since the rebrand is already live on our side and we'll soon be onboarding customers, we'd appreciate it if the review could be prioritized as soon as possible.

Kind regards,
Zeeshan

joe-ayoub-segment:
Thanks Zeeshan,

What would you like to do with the existing destination(s)?
Will they be deleted / disabled? If so, when?
Would it make sense to rebrand these instead of creating a new one? Would
the code be significantly different?

Best regards,
Joe

zeeshan-vwo:
Hi @joe-ayoub-segment 
Our existing customers are still using the current integration, so we won't be disabling it anytime soon. We plan to migrate customers to the new integration and eventually deprecate the existing one, but we don't have a defined timeline yet.

We considered rebranding the existing destination, but the new integration differs enough that creating a separate destination is the cleaner approach and avoids impacting current customers.

Best Regards,
Zeeshan

packages/destinations-manifest/src/index.ts
```diff
@@ -47,6 +47,7 @@ register('634ef204885be3def430af66', '@segment/analytics-browser-actions-playerz
 register('635ada35ce269dbe305203ff', '@segment/analytics-browser-actions-logrocket')
 register('6372e1e36d9c2181f3900834', '@segment/analytics-browser-actions-wiseops')
 register('637c192eba61b944e08ee158', '@segment/analytics-browser-actions-vwo')
+register('69a2f1c8e4b7d63901ab5678', '@segment/analytics-browser-actions-wingify')
```
joe-ayoub-segment: hi @zeeshan-vwo please remove the change to this line. This ID is generated before we deploy the new integration.

packages/destinations-manifest/package.json
```diff
@@ -54,6 +54,7 @@
     "@segment/analytics-browser-actions-userpilot": "^1.99.0",
     "@segment/analytics-browser-actions-utils": "^1.98.0",
     "@segment/analytics-browser-actions-vwo": "^1.102.0",
+    "@segment/analytics-browser-actions-wingify": "^1.0.0",
```
joe-ayoub-segment: Hi @zeeshan-vwo please remove this line. I'll update the manifest when I register the destination.

joe-ayoub-segment:
Hi @zeeshan-vwo I reviewed the PR, and triggered Claude and Copilot reviews. 

Can you please respond to the comments?

The Claude review is below: 

----------------------------

Thanks for this — the async-queue loading approach is the right pattern, and porting from the existing actions-vwo-web destination is a sensible baseline. However, the port dropped one of VWO's null-safety guards, and the result is a hard crash in
  initialize under the exact conditions Segment runs it. Requesting changes for the one blocking item below.

  🔴 Blocking: initScript throws a TypeError when invoked by Segment

  In init-script.ts, the SmartCode reads document.currentScript and dereferences .nonce without a guard:

  var i = !1,
    o = n.currentScript,   // null when not executing as an inline <script>
    ...
  var code = {
    nonce: o.nonce,        // 💥 TypeError: Cannot read properties of null (reading 'nonce')

  document.currentScript is only populated while the browser is synchronously executing a script element. Segment calls initialize programmatically (it's a function on the destination definition, not inline <script> execution), so
  document.currentScript is null. Reproduced end-to-end against the repo's pinned jsdom:

  initScript THREW: TypeError - Cannot read properties of null (reading 'nonce')

  Consequences:
  - code.init() is never reached, so the SmartCode tag is never injected — the destination silently does nothing.
  - initialize has no try/catch around initScript, so the throw propagates out of load() and rejects, surfacing as an unhandled device-mode load error.

  Why VWO doesn't have this bug — the VWO SmartCode this was copied from guards every nonce access (v = d.querySelector('#vwoCode') returns null safely, and uses nonce: v && v.nonce). The Wingify port guards the two later nonce uses (o.nonce &&
  t.setAttribute(...) in addScript/init) but not the object-literal line. That single unguarded line is the crash.

  Fix (minimal, matches the file's own existing pattern):

  nonce: o && o.nonce,

  🟠 Tests don't catch this, and assertions are too weak

  The addSmartcode: true test calls load() → initialize → initScript, which per the above should reject — please confirm the suite was actually run green, since production browsers will have currentScript === null.

  Separately, the current assertions can't detect a no-op:

  const script = window.document.querySelector('script[src~="..."]') as HTMLScriptElement
  expect(script).toBeDefined()         // querySelector returns null on miss; expect(null).toBeDefined() PASSES
  expect(wingifyObject).toBeDefined()  // window.Wingify = [] is always defined

  toBeDefined() only fails on undefined. Please switch to expect(script).not.toBeNull() and assert script.src contains the tag URL, so this code path is genuinely covered.

  ✅ What looks good (re: non-blocking / race conditions)

  - No document.write; the tag is injected via document.head.appendChild(script) — non-blocking, won't stall Segment's library or the page. preconnect + fetchPriority = 'high' are fine.
  - The async-queue pattern is correct: window.Wingify = window.Wingify || [] plus the event/visitor push([...]) fallbacks mean calls made before the real SDK loads are queued and drained later — no ordering race.
  - The localStorage / JSON.parse path is safe — the throw is caught and the default config is retained.
  - resolveWhen(() => hasOwnProperty('Wingify'), 100) resolves immediately since window.Wingify was just set.

  Minor (non-blocking)

  - Please confirm the trimmed-down SmartCode matches what Wingify currently ships at edge.wingify.net, since vendored copies drift.
  - window.Wingify = window.Wingify || [] assigns [] to a field typed as the Wingify object — mirrors VWO, but worth a yarn typecheck pass since index.ts is not @ts-nocheck'd.

  ---
  Once nonce: o && o.nonce is in place and the test asserts the script tag is actually injected, this should be good to go.

zeeshan-vwo:
Hi @joe-ayoub-segment — pushed fixes for the blocking currentScript / nonce crash (o && o.nonce on all three access points), strengthened the SmartCode injection test (not.toBeNull() + script.src assertion), and aligned performance.getEntriesByName + timer naming. Wingify tests pass locally.

joe-ayoub-segment:
Thanks @zeeshan-vwo . 

**Initialization**

The resolveWhen check on init doesn't really do anything. Right before it, we set window.Wingify = window.Wingify || [], so by the time resolveWhen checks whether window.Wingify exists, it's always already there. It resolves immediately every time and never actually waits for the Wingify SmartCode to load. This matches what the VWO destination does, so it's not a new problem - but it's worth knowing the check is effectively a no-op. It works anyway because calls get queued onto the stub array until the real library loads.

One small thing that's slightly worse than VWO: in the SmartCode, Object.assign(JSON.parse(localStorage.getItem(...)), e) will throw if that localStorage key is empty (because JSON.parse(null) returns null). It's inside a try/catch so it's harmless, but VWO guards it properly. Minor.

**Unit tests**

Coverage is OK for the happy path but misses a few things:

1. The most important gap: the queue fallback in perform is never tested. This line — window.Wingify.event = window.Wingify.event || function(...) { window.Wingify.push(...) } - is what makes events work before the library finishes loading. The tests mock initialize to inject a Wingify object that already has event/visitor, so that fallback branch never runs. Please add a test that sets window.Wingify = [] and checks the call lands in push(['event', ...]).
2. The addSmartcode: false test only checks that window.Wingify is defined — but that's true either way. It should check that the script tag is NOT added to the page (i.e. the
  querySelector for the Wingify script is null).
3. Add a couple of edge cases: empty traits on identify ({}), and a nested/non-string property value on track.

joe-ayoub-segment:
Hi @zeeshan-vwo - can you address this?

🔴 Blocking: initScript throws a TypeError when invoked by Segment

In init-script.ts, the SmartCode reads document.currentScript and dereferences .nonce without a guard:

var i = !1,
o = n.currentScript, // null when not executing as an inline <script>
...
var code = {
nonce: o.nonce, // 💥 TypeError: Cannot read properties of null (reading 'nonce')

document.currentScript is only populated while the browser is synchronously executing a script element. Segment calls initialize programmatically (it's a function on the destination definition, not inline <script> execution), so
document.currentScript is null. Reproduced end-to-end against the repo's pinned jsdom:

initScript THREW: TypeError - Cannot read properties of null (reading 'nonce')

Consequences:

code.init() is never reached, so the SmartCode tag is never injected — the destination silently does nothing.
initialize has no try/catch around initScript, so the throw propagates out of load() and rejects, surfacing as an unhandled device-mode load error.
Why VWO doesn't have this bug — the VWO SmartCode this was copied from guards every nonce access (v = d.querySelector('#vwoCode') returns null safely, and uses nonce: v && v.nonce). The Wingify port guards the two later nonce uses (o.nonce &&
t.setAttribute(...) in addScript/init) but not the object-literal line. That single unguarded line is the crash.

Fix (minimal, matches the file's own existing pattern):

nonce: o && o.nonce,

joe-ayoub-segment (approved):

PR #3833, opened 2026-06-16
fix(hubspot-web): add id=hs-script-loader to prevent duplicate script injection

harsh-joshi99:
## Summary
Adds `{ id: 'hs-script-loader' }` to the HubSpot tracking-script `loadScript` call in the hubspot-web browser destination. HubSpot's own embed uses the `hs-script-loader` id to detect whether its script is already present; without it, the script can be injected twice.

Fixes STRATCONN-6721.

## Root cause
HubSpot's analytics script (loaded transitively by `js.hs-scripts.com/<portalId>.js`) injects its region-specific tracking script through a guard keyed on the DOM id `hs-script-loader`. Confirmed verbatim in HubSpot's own code at `js.hs-analytics.net/analytics/<ts>/<portalId>.js`:

```js
// invoked as: embedHubSpotScript('https://js-na1.hs-scripts.com/<portalId>.js', 'hs-script-loader')
embedHubSpotScript = function (t, e) {
  if (!( ...|| document.getElementById(e)) && this.trackingEnabled) {
    var n = document.createElement("script");
    n.src = t; n.type = "text/javascript"; n.id = e; // e === "hs-script-loader"
    // ...insertBefore
  }
};
```

Segment was injecting `js.hs-scripts.com` **without** that id, so `document.getElementById('hs-script-loader')` returned null, the guard didn't trip, and HubSpot injected a second (`js-na1.hs-scripts.com`) script. Setting the id on Segment's injected script makes the guard find it and skip the duplicate.

**References:**
- HubSpot bootstrap loader: `js.hs-scripts.com/<portalId>.js` (~1.9 KB; loads the analytics script below)
- HubSpot analytics script containing the `embedHubSpotScript` guard: `js.hs-analytics.net/analytics/<ts>/<portalId>.js`
- Jira: [STRATCONN-6721](https://twilio-engineering.atlassian.net/browse/STRATCONN-6721)
- HubSpot docs — [Install the HubSpot tracking code](https://knowledge.hubspot.com/reports/install-the-hubspot-tracking-code) (does not show the snippet itself; the guard lives in the served scripts above)

## Changes
- `src/index.ts`: pass `{ id: 'hs-script-loader' }` to `deps.loadScript(scriptPath, ...)`
- `src/__tests__/index.test.ts`: add a test asserting the injected HubSpot script carries `id="hs-script-loader"`

## Testing
Verified locally via the Action Tester with a real HubSpot portal ID, comparing `main` vs this branch.

| | Scripts injected |
|---|---|
| **Before** (`main`) | **2** — `js.hs-scripts.com` (`id:''`) + duplicate `js-na1.hs-scripts.com` (`id:'hs-script-loader'`) |
| **After** (this branch) | **1** — `js.hs-scripts.com` (`id:'hs-script-loader'`), no duplicate |

**Before — duplicate script injected (`main`, unfixed):**

Before - two HubSpot scripts

**After — single script with `id=hs-script-loader` (this branch):**

After - one HubSpot script

##### Before
Screenshot 2026-08-13 at 11 47 42 PM

 ##### After
Screenshot 2026-08-14 at 3 05 19 PM

- New unit test added for the script `id`.
- ⚠️ Note: the hubspot-web jest suite currently fails to construct its jsdom environment locally (`TypeError: Cannot read properties of undefined (reading 'html')` in `JSDOMEnvironment`) — a pre-existing tooling mismatch (jest 30 vs ts-jest <28), unrelated to this change. CI will run the canonical suite.

## Deploy note
These changes have also been pushed directly to `staging` for testing.

[STRATCONN-6721]: https://twilio-engineering.atlassian.net/browse/STRATCONN-6721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/browser-destinations/destinations/hubspot-web/src/index.ts
```diff
@@ -92,7 +92,7 @@ export const destination: BrowserDestinationDefinition<Settings, Hubspot> = {
       ? 'https://js-eu1.hsforms.net/forms/v2.js'
       : 'https://js.hsforms.net/forms/v2.js'
 
-    await deps.loadScript(scriptPath)
+    await deps.loadScript(scriptPath, { id: 'hs-script-loader' })
```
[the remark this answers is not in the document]
harsh-joshi99: For the bug this PR targets (STRATCONN-6721), the `id` alone is the correct and sufficient fix: the duplicate was being injected by **HubSpot's own loader**, which checks for an existing `#hs-script-loader` and, not finding one, injected a region-specific copy (`js-na1.hs-scripts.com`). Setting the id makes HubSpot's loader detect our script and skip its duplicate. Verified locally with a real portal ID — 2 scripts on `main` → 1 on this branch (screenshots in the PR description).

The "site already ships HubSpot's native embed" case is a real but separate, pre-existing scenario, and a skip-if-`#hs-script-loader`-exists guard would change behavior. I'd prefer not to expand scope here:
- The actual injection lives in shared core `load-script.ts`, not this destination, so a clean guard isn't local to `initialize()`.
- A behavior change to handle native-embed coexistence deserves its own ticket/PR rather than riding along on this targeted fix.

packages/browser-destinations/destinations/hubspot-web/src/index.ts
```diff
@@ -92,7 +92,7 @@ export const destination: BrowserDestinationDefinition<Settings, Hubspot> = {
...
     if (settings.loadFormsSDK) {
```
harsh-joshi99, line 96: > [!IMPORTANT]
> :yellow_circle: **MEDIUM: Fix is asymmetric: loadScript dedups by src, not id**

[Found by 3/3 passes] HubSpot's embed guard skips injection when `document.getElementById('hs-script-loader')` exists, which this PR now satisfies — good. But the runtime's own loadScript() (browser-destination-runtime/src/load-script.ts:3) dedups only by `s.src === src` and returns early only when the found script has a Segment-set `status` attribute of 'loaded'/'loading'. Two consequences: (1) If the customer already has HubSpot's native embed snippet on the page (same src `https://js.hs-scripts.com/{portalId}.js`, id present but no Segment `status` attr), loadScript falls past the dedup branch and injects a SECOND `<script id='hs-script-loader'>` — the opposite direction of the bug being fixed. (2) Two HubSpot Web instances whose src differs (EU `js-eu1.hs-scripts.com` + non-EU `js.hs-scripts.com`, or two portalIds) each create an element carrying the same hardcoded id, yielding a duplicate DOM id and a getElementById target that resolves to whichever is first in document order. Neither is a crash; the common single-instance/same-region case (the STRATCONN-6721 target) is fully fixed.

**Suggestion:** Optionally mirror HubSpot's own guard before injecting — e.g. `if (document.getElementById('hs-script-loader')) return window._hsq` in initialize() — or have loadScript dedup by matching id when one is supplied. For multi-instance support, derive a per-instance id (include portalId/region) instead of a fixed string. If multi-portal-per-page is out of scope for this ticket, no change needed.

packages/browser-destinations/destinations/hubspot-web/src/__tests__/index.test.ts
```diff
@@ -37,6 +37,19 @@ describe('Hubspot Web (Actions)', () => {
...
+      portalId: '12345',
+      subscriptions
+    })
+
+    await event.load(Context.system(), {} as Analytics)
+
+    const loaderScripts = Array.from(document.querySelectorAll<HTMLScriptElement>('script#hs-script-loader'))
+    expect(loaderScripts).toHaveLength(1)
+    expect(loaderScripts[0].src).toBe('https://js.hs-scripts.com/12345.js')
+  })
```
harsh-joshi99, line 51: > [!NOTE]
> :large_blue_circle: **LOW: Test asserts resolved IDL src; would break if URL became relative**

`loaderScripts[0].src` returns the IDL-resolved absolute URL. Because scriptPath is already absolute, resolution is an identity and the assertion holds correctly today. Noted only because it would silently break if the injected URL ever became relative.

**Suggestion:** Optional: assert via `getAttribute('src')` to test the literal injected value rather than the resolved IDL value.

harsh-joshi99 (commented):
> [!NOTE]
> **Self-Review**
> Posted by PR author (`harsh-joshi99`) — self-reviews cannot apply the `APPROVED` label. Intended verdict: **Manual Review Needed :warning:**. Ask a teammate to run `plugin-review` for the `APPROVED` label.

## Code Review Personae Verdict: Manual Review Needed :warning:

The one-line change (adding id='hs-script-loader' to the injected HubSpot tracking script) introduces no crashes, NPEs, resource leaks, races, or security issues, and the new test is correct and not flaky. All 3 bug-finder passes independently converged on a single design concern: loadScript dedups by src/status, not by id, so the fix is asymmetric and can produce a duplicate DOM id in multi-instance edge cases.

Policy scan clean.

---

<details>
<summary><h3>:memo: Deep Code Review (3 Passes) — 0 Critical, 0 High, 1 Medium, 2 Low :yellow_circle:</h3></summary>

Consensus: all 3 independent passes flagged the same line (src/index.ts:95) and the same underlying mechanism — the runtime loadScript() dedups by `s.src === src` (load-script.ts:3) and short-circuits only when an existing matching script carries a Segment `status` attribute of 'loaded'/'loading'. The new `id` is never used for de-duplication. The single-instance / same-region path that STRATCONN-6721 targets is fully fixed (HubSpot's own embed guard does `document.getElementById('hs-script-loader')` and now finds Segment's script). The residual concern is asymmetry: (a) if a customer already has HubSpot's native embed snippet on the page (same src, no Segment `status` attr), Segment still falls through and injects a second `<script id='hs-script-loader'>`; (b) two HubSpot Web instances with differing src (EU + non-EU, or differing portalId) each inject an element with the same hardcoded id, producing a duplicate DOM id and a potentially wrong getElementById target. Both are narrow edge cases, not production crashes. The new test's `toHaveLength(1)` is sound — setup-after-env.ts resets document.head per test and the harness's empty `<script>` has no id, so no cross-test leakage or false match.

### Findings

- :yellow_circle: **MEDIUM: Fix is asymmetric: loadScript dedups by src, not id** (`packages/browser-destinations/destinations/hubspot-web/src/index.ts:95`)
  [Found by 3/3 passes] HubSpot's embed guard skips injection when `document.getElementById('hs-script-loader')` exists, which this PR now satisfies — good. But the runtime's own loadScript() (browser-destination-runtime/src/load-script.ts:3) dedups only by `s.src === src` and returns early only when the found script has a Segment-set `status` attribute of 'loaded'/'loading'. Two consequences: (1) If the customer already has HubSpot's native embed snippet on the page (same src `https://js.hs-scripts.com/{portalId}.js`, id present but no Segment `status` attr), loadScript falls past the dedup branch and injects a SECOND `<script id='hs-script-loader'>` — the opposite direction of the bug being fixed. (2) Two HubSpot Web instances whose src differs (EU `js-eu1.hs-scripts.com` + non-EU `js.hs-scripts.com`, or two portalIds) each create an element carrying the same hardcoded id, yielding a duplicate DOM id and a getElementById target that resolves to whichever is first in document order. Neither is a crash; the common single-instance/same-region case (the STRATCONN-6721 target) is fully fixed.
  **Suggestion:** Optionally mirror HubSpot's own guard before injecting — e.g. `if (document.getElementById('hs-script-loader')) return window._hsq` in initialize() — or have loadScript dedup by matching id when one is supplied. For multi-instance support, derive a per-instance id (include portalId/region) instead of a fixed string. If multi-portal-per-page is out of scope for this ticket, no change needed.
- :large_blue_circle: **LOW: Test asserts resolved IDL src; would break if URL became relative** (`packages/browser-destinations/destinations/hubspot-web/src/__tests__/index.test.ts:50`)
  `loaderScripts[0].src` returns the IDL-resolved absolute URL. Because scriptPath is already absolute, resolution is an identity and the assertion holds correctly today. Noted only because it would silently break if the injected URL ever became relative.
  **Suggestion:** Optional: assert via `getAttribute('src')` to test the literal injected value rather than the resolved IDL value.
- :large_blue_circle: **LOW: Pre-existing: tag.parentElement throws when document has zero <script> tags** (`packages/browser-destination-runtime/src/load-script.ts:45`)
  `getElementsByTagName('script')[0]` is undefined if the page has no script tags, so `tag.parentElement?.insertBefore` would throw 'Cannot read properties of undefined'. NOT introduced or newly exercised by this PR (the id attribute changes nothing on this branch), and the test harness avoids it by appending an empty `<script>` in beforeEach. Flagged for completeness only; out of scope for this PR.
  **Suggestion:** Out of scope. If hardened later, fall back to document.head when no script tag exists.

</details>

<details>
<summary><h3>:shield: Policy Compliance Review — No violations (3 policies evaluated)</h3></summary>

Evaluated 3 policies across all review surfaces. No violations detected.

</details>

### Review Stages Executed

- :white_check_mark: Deep Code Review (3 Passes)
- :white_check_mark: Policy Compliance Review
<sub>Generated by code-review-personae v0.8.7 | Deep Code Review (3 Passes), Policy Compliance Review</sub>

sayan-das-in (approved):

AnkitSegment (approved):

joe-ayoub-segment:
PR deployed.

PR #3846, opened 2026-06-24
fix(ms-bing-ads-audiences): surface real Bing API errors in createAudience

harsh-joshi99:
## Summary

The `createAudience` flow for **Ms Bing Ads Audiences** masks the actual cause of create failures, producing opaque errors that make customer issues impossible to diagnose from logs.

Two distinct gaps:

1. **Untyped HTTP failures escaped uncaptured.** The `POST /Audiences` call had no error handling. When Bing returned a non-2xx (often an empty / non-JSON body), the request client threw a raw error that bubbled up untyped — the platform reported it as a generic `500 / INTERNAL / "Partner API internal error"` with an empty `responseFromDestination`, so Bing's real status/body was never logged.
2. **Unrecognized PartialErrors were swallowed.** Bing returns `200` with a `PartialErrors` array for most create failures. Only the Terms & Conditions code was handled; every other code (e.g. a duplicate audience name) fell through to the opaque `NO_AUDIENCE_ID` error, discarding Bing's actual `ErrorCode`/`Message`.

## Motivating case

A customer (STRATCONN-6886) hit repeated `500 / "Partner API has encountered an internal server error"` on audience creation. The destination discards Bing's actual response (`responseFromDestination:[]` in logs), so the real cause was never visible. This PR does **not** claim to root-cause that customer's failure — it makes the error legible (`Microsoft Bing Ads returned HTTP <status>: <body>`) so we can finally see what Bing is actually returning and debug their real issue once this is in prod.

## Changes

- Add error handling inline in `createAudience` (`index.ts`), keeping the existing flow in place to minimize the diff and regression risk. Only the small, reusable `readResponseBody` helper lives in `utils.ts`.
- Wrap the `POST /Audiences` call so that on `HTTPError` we capture Bing's **actual status and response body** in the thrown error. The body is read from the request client's already-parsed `content` (falling back to `text()`), truncated to 2KB to avoid log bloat.
- Surface **every non-T&C PartialError** with Bing's own `ErrorCode` and `Message` instead of collapsing into `NO_AUDIENCE_ID`.
- All HTTP failures throw `IntegrationError` carrying Bing's status. Non-`HTTPError` failures (timeouts, socket/connection errors) are also wrapped as a retryable `IntegrationError` (status 500) rather than escaping raw. A missing/unknown upstream status defaults to 500 (retryable) rather than 400.

## Retry semantics (note for reviewers)

This replaces the earlier revision's `RetryableError`. Per the repo contract (`core/src/errors.ts`), retry is driven by the **status code**, not the error class: an `IntegrationError` with a **5xx** status is still automatically retried (except 501), while 4xx is not (except 408/423/429). So a transient Bing 5xx on create **still retries** — which is the intended behavior — and a 4xx (e.g. invalid account, duplicate name) surfaces terminally to the user. (An earlier commit message incorrectly implied dropping `RetryableError` made 5xx non-retrying; that was wrong — the status code governs retry regardless of error class.)

## Contract notes

- `createAudience` still returns `AudienceResult = { externalId: string }` on success — unchanged. Failures are thrown, per the framework contract.
- The thrown error's `status` / `code` / `message` are exactly what external-audience-service's gxClient logs, so this directly improves the Grafana visibility for these failures.

## Testing

- Unit tests assert **error type, status, and code** for the non-2xx (400), 5xx-empty-body, unrecognized-PartialError, and connection-failure (no HTTP response) paths.
- Direct unit tests for `readResponseBody` cover every branch (string / object / Buffer / `text()` fallback / `text()` throwing / truncation).
- All 40 tests in the destination pass; lint clean.

## Stage Testing

Deployed to staging and verified end-to-end.

- Triggered a failure by configuring **invalid account IDs** (a quick way to force a create failure). This is **not** the customer's exact scenario — their account IDs are valid — but it exercises the same error path and confirms we now capture Bing's real response.
- In this case Microsoft returned only a **generic error** (`InternalError` / "An internal error has occurred.") with a `TrackingId` and no actionable detail. So for some failures the body itself won't explain the cause — but we now log Bing's actual payload **and the TrackingId**, which is exactly what's needed to debug the customer's real failure (or escalate to Microsoft) once this is in prod. Previously this was an opaque `500 / "Partner API internal error"` with an empty `responseFromDestination`.

**Local API test (direct call to Bing) — Microsoft's raw response:**

Screenshot 2026-06-29 at 6 49 00 PM

**Stage Grafana — error surfaced by this change (Bing's real body + TrackingId):**

Screenshot 2026-06-29 at 5 56 36 PM

## Deep-review follow-ups (deferred, not blocking)

- Bing's `ErrorCode` used as the Segment `code` is higher-cardinality than the repo's usual stable constants (low).
- PII: surfacing the raw Bing body in the message is bounded by 2KB truncation; not a regression vs. the existing sync-path error handling. Not actioned.
- UI still shows generic text (external-audience-service maps 5xx to a generic message regardless of our `message`); improving the customer-facing UI string is a separate change in that service.


packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -95,21 +114,57 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
+        // The Bing Ads API can return a non-2xx (often an empty or non-JSON body) that the
+        // request client throws as an HTTPError. Without this, the raw error escapes untyped
+        // and the platform surfaces an opaque "500 / Bad Request" with no detail. Capture
+        // Bing's actual status and response body so the real cause is visible in logs.
+        if (error instanceof HTTPError) {
+          const status = error.response?.status
+          const body = await readResponseBody(error.response)
+          const message = `Failed to create audience. Microsoft Bing Ads returned HTTP ${status ?? 'unknown'}: ${
+            body || 'no response body'
+          }`
```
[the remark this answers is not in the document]
harsh-joshi99: Done in 1c0631ea3 — truncating the body to a 2KB cap with a `...(truncated)` suffix before interpolating.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -80,44 +75,7 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
-        }
-      }
-
-      // Extract the created audience ID
-      const audienceId = response?.data?.AudienceIds?.[0]
-      if (!audienceId) {
-        throw new IntegrationError('Failed to create audience: No AudienceId returned', 'NO_AUDIENCE_ID', 400)
-      }
-
+      const audienceId = await createBingAudience(request, createAudienceInput?.audienceName)
```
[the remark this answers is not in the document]
harsh-joshi99: Done in 1c0631ea3 — `createBingAudience` now accepts `audienceName?: string`; the validation/throw inside is retained.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/__tests__/index.test.ts
```diff
@@ -102,6 +102,49 @@ describe('Bing Ads Audiences', () => {
...
+      }
+      nock(BASE_URL).post('/Audiences').reply(200, partialErrorResponse)
+
+      await expect(testDestination.createAudience(createAudienceInput)).rejects.toThrowError(
+        new IntegrationError(
+          'Failed to create audience: CampaignServiceDuplicateAudienceName: An audience with this name already exists.',
+          'CampaignServiceDuplicateAudienceName',
+          400
+        )
+      )
```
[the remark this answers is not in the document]
harsh-joshi99: Done in 1c0631ea3 — switched to capturing the thrown error and asserting `instanceof IntegrationError` plus explicit `message`/`code`/`status`.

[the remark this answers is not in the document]
harsh-joshi99: Done in 1c0631ea3 — now captures the error and asserts `instanceof` + `message`/`code`/`status` explicitly instead of `toThrowError(instance)`.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -1,15 +1,34 @@
...
+const readResponseBody = async (response?: Response): Promise<string> => {
+  if (!response) {
+    return ''
+  }
+  try {
+    return (await response.text()) ?? ''
+  } catch {
+    return ''
+  }
+}
```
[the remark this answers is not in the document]
harsh-joshi99: Not changing — this is based on an incorrect assumption about this codebase. `HTTPError.response` here is typed as the Fetch `Response` (see `packages/core/src/request-client.ts`: `response: Response`), not a Got-style object, so `.text()` is guaranteed to exist. The existing `handleHttpError` in this same module already calls `.json()` on it. `tsc --noEmit` passes for the package with no errors in these files.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/utils.ts
```diff
@@ -185,3 +193,103 @@ export const handleHttpError = async (
...
+const readResponseBody = async (response?: Response): Promise<string> => {
+  if (!response) {
+    return ''
+  }
+  try {
+    return (await response.text()) ?? ''
+  } catch {
+    return ''
+  }
+}
```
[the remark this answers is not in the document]
harsh-joshi99: Not changing — no compile error here. `ModifiedResponse extends Response` (`packages/core/src/types.ts`), and `HTTPError.response` is typed `Response`, so `.text()` is valid and `tsc --noEmit` passes. Typing the param as `Response` is consistent with the actual `HTTPError.response` type.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/utils.ts
```diff
@@ -185,3 +193,103 @@ export const handleHttpError = async (
...
+    // Surface every other PartialError with Bing's own ErrorCode and Message so the real cause
+    // (e.g. a duplicate audience name) is visible instead of NO_AUDIENCE_ID.
+    throw new IntegrationError(
+      `Failed to create audience: ${errorObj?.ErrorCode ?? 'UnknownError'}: ${
+        errorObj?.Message ?? 'No error message provided'
+      }`,
+      errorObj?.ErrorCode ?? 'CREATE_AUDIENCE_FAILED',
+      400
+    )
+  }
```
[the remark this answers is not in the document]
harsh-joshi99: Not changing — for the create path this is a non-issue. We send exactly one audience (`Audiences: [{ ... }]`, hardcoded single element), and Bing's `PartialErrors` are indexed to the request array, so at most one error can be returned. Aggregating would be dead code for a single-item request. The framework also invokes `createAudience` once per audience.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -80,44 +75,7 @@ const destination: AudienceDestinationDefinition<Settings> = {
       // API Docs
       // Add Audience: https://learn.microsoft.com/en-us/advertising/campaign-management-service/addaudiences?view=bingads-13&tabs=prod&pivots=rest
       // Customer List: https://learn.microsoft.com/en-us/advertising/campaign-management-service/customerlist?view=bingads-13&tabs=json
-      const audienceName = createAudienceInput?.audienceName
```
mdkhan-tw, line 83: Can we avoid this huge refactor and perform error handling here? It will help me review the code and reduce the total changes we are making as part of PR reducing regression bug risk.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -95,24 +98,65 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
+        // escapes untyped and the platform surfaces an opaque "500 / Bad Request" with no detail.
+        // Capture Bing's actual status and response body so the real cause is visible in logs.
+        if (error instanceof HTTPError) {
+          const status = error.response?.status
+          const body = await readResponseBody(error.response as ModifiedResponse)
+          // Default a missing upstream status to 500 (retryable) rather than 400 — an unknown
+          // status is an ambiguous failure, not a client error.
+          throw new IntegrationError(
+            `Failed to create audience. Microsoft Bing Ads returned HTTP ${status ?? 'unknown'}: ${
+              body || 'no response body'
```
mdkhan-tw: Let's send status and not body which might have sensitive info

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -95,24 +98,65 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
+            }`,
+            'CREATE_AUDIENCE_FAILED',
+            status ?? 500
```
mdkhan-tw: ?? should it be ||

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -95,24 +98,65 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
         }
+
+        // Surface every other PartialError with Bing's own ErrorCode and Message so the real cause
+        // (e.g. a duplicate audience name) is visible instead of NO_AUDIENCE_ID.
+        throw new IntegrationError(
+          `Failed to create audience: ${errorObj?.ErrorCode ?? 'UnknownError'}: ${
+            errorObj?.Message ?? 'No error message provided'
+          }`,
+          errorObj?.ErrorCode ?? 'CREATE_AUDIENCE_FAILED',
+          400
```
mdkhan-tw: 1. why 400 status code?
2. What do PartialError means. Does it means few thing succeeded? Is it a retryable error?

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
harsh-joshi99: The issue is that status only was not helpful for us in debugging the customer issue. 
The body is what carries Microsoft's TrackingId (and ErrorCode), which is the reference the MS support team needs to actually investigate on their end.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
harsh-joshi99: 1. it's a failure that requires user action (e.g. duplicate name, terms not accepted, invalid input) — retrying the identical request won't help, so it's a non-retryable client error.

2. Bing's AddAudiences returns HTTP 200 even on failure, with a PartialErrors[] array describing which items failed. It's "partial" because the API is batch-shaped (you can send multiple audiences). In our case we send exactly one, so a PartialError means that one create failed - nothing succeeded. It is not a "some worked" situation for us.

mdkhan-tw: Can we avoid throwing 400 unless it's a bad request. Can we throw 500 instead, since we aren't sure about category of the error.

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
mdkhan-tw: Can we understand the response structure and only include what's required for debugging and ensuring that they aren't sensitive in nature

harsh-joshi99:
Testing after refactor: 

Error with wrong CID: 
Screenshot 2026-07-06 at 11 46 53 AM

No error with correct ID:
Screenshot 2026-07-06 at 11 50 15 AM

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -95,27 +98,72 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
+          method: 'POST',
+          json
+        })
+      } catch (error) {
+        // The request client throws an HTTPError on a non-2xx response. Without this, the raw error
+        // escapes untyped and the platform surfaces an opaque "500 / Bad Request" with no detail.
+        // Capture Bing's actual status and response body so the real cause is visible in logs.
+        if (error instanceof HTTPError) {
+          const status = error.response?.status
+          const rawBody = await readResponseBody(error.response as ModifiedResponse)
```
mdkhan-tw, line 113: nit: there is no reason for this function to be async

packages/destination-actions/src/destinations/ms-bing-ads-audiences/utils.ts
```diff
@@ -185,3 +185,75 @@ export const handleHttpError = async (
...
+      Errors?: Array<{ ErrorCode?: string; Message?: string }>
+      TrackingId?: string
+    }
+    const parts = (parsed?.Errors ?? []).map(
+      (e) => `${e?.ErrorCode ?? 'UnknownError'}: ${e?.Message ?? 'No error message provided'}`
+    )
+    if (parsed?.TrackingId) {
+      parts.push(`TrackingId: ${parsed.TrackingId}`)
+    }
+    return parts.length ? parts.join('; ') : rawBody
```
mdkhan-tw: We should never return rawBody even in case of failure becuase it might contain sensitive info.

Can we check if rawBody can have sensitive info?

mdkhan-tw (dismissed):

AnkitSegment (dismissed):

packages/destination-actions/src/destinations/ms-bing-ads-audiences/index.ts
```diff
@@ -95,27 +98,72 @@ const destination: AudienceDestinationDefinition<Settings> = {
...
+        )
       }
 
-      // Extract the created audience ID
+      // A 200 with no AudienceId and no PartialError is an ambiguous failure — classify as 500
+      // (retryable) since we have no signal that it's caller-actionable.
       const audienceId = response?.data?.AudienceIds?.[0]
       if (!audienceId) {
-        throw new IntegrationError('Failed to create audience: No AudienceId returned', 'NO_AUDIENCE_ID', 400)
+        throw new IntegrationError('Failed to create audience: No AudienceId returned', 'NO_AUDIENCE_ID', 500)
```
joe-ayoub-segment, line 165: Hi @harsh-joshi99 -maybe a 500 is the correct error, but I don't think it will be retried (see comment above on lines 162/3.

harsh-joshi99: yeah, its not supposed to be, I will fix the comment.

mdkhan-tw (approved):

joe-ayoub-segment:
Hi @harsh-joshi99 PR deployed.

PR #3854, opened 2026-07-03
[Vibe] - new Conversions destination

mdkhan-tw:
Adds a new cloud-mode destination, **Vibe Tracking Event** (slug `actions-vibe-conversions`), that sends conversion events to Vibe via `POST https://t.vibe.co/s2s-conversion/events/segment`.

**What & why:** Vibe wants to receive Segment conversion events (purchases, page views, signups, etc.) for advertiser attribution. This destination maps Segment track/page events to Vibe's single-event conversion API.

### Details

- **Action:** `trackConversion` — 9 mapped fields (`a`, `eid`, `ts`, `ip`, `em`, `ed`, `gid`, `ua`, `url`) plus the `aid` (Pixel ID) settings value injected into every event.
- **Auth:** `scheme: custom`. The Vibe conversion API requires no API key/token; the advertiser is identified by the Pixel ID (`aid`) setting.
- **Presets:** Order Completed → `purchase`, Page Viewed → `page_view`, Signed Up → `signup`.
- **Validation & transforms:**
  - Either `ip` or `em` is required (throws `PayloadValidationError` if both missing).
  - Timestamp must be within the last 7 days; ISO timestamps are converted to UNIX milliseconds.
  - Event type `a` is constrained to the API enum via `choices`.
  - Event data (`ed`) is serialized to a stringified JSON object.
- **No batching:** the Vibe API has no batch endpoint, so `performBatch` is intentionally omitted (single-event POST only).

### TODO before merge

- ⚠️ **Not registered** in `packages/destination-actions/src/destinations/index.ts` yet — the metadata ID must be the production-assigned MongoDB ObjectId created in the control plane, not a placeholder. Register once the destination is created in production.
- **Open questions to confirm with Vibe:** endpoint path `/events/segment` (PRD) vs `/events` (public docs); `em` field and `call` enum value present in PRD but absent from public docs; final destination name (currently "Vibe Tracking Event", TBC); `gid` label/description.

## Testing

- [ ] [If destination is already live] Tested for backward compatibility of destination.

11 Jest tests pass (unit + snapshot), covering the happy path, ISO→UNIX ms conversion, email fallback, both validation errors, enum rejection, presets, and request-body snapshots. Typecheck and lint are clean.

## Security Review

- [x] **Reviewed all field definitions** for sensitive data (API keys, tokens, passwords, client secrets). This API uses no secrets; the only credential-like value is the non-secret Pixel ID.

## New Destination Checklist

- [ ] Extracted all action API versions to `versioning-info.ts` file.


packages/destination-actions/src/destinations/vibe-conversions/index.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const destination: DestinationDefinition<Settings> = {
+  name: 'Vibe Tracking Event',
+  slug: 'actions-vibe-conversions',
+  mode: 'cloud',
+  description: 'Send conversion events to Vibe.',
+
+  authentication: {
+    // Vibe's conversion API requires no API key or token. The advertiser is
+    // identified by the Pixel ID (aid), which is injected into every event.
```
joe-ayoub-segment: let's remove unnecessary comments

packages/destination-actions/src/destinations/vibe-conversions/index.ts
```diff
@@ -0,0 +1,66 @@
+import type { DestinationDefinition } from '@segment/actions-core'
+import { defaultValues } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+
+import trackConversion from './trackConversion'
+
+// Preset mappings reuse the action's default field values, but pin `a` to a
+// preset-specific enum value (the `a` field itself intentionally has no default).
```
joe-ayoub-segment: remove comments.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
```diff
@@ -0,0 +1,139 @@
...
+          exists: { '@path': '$.context.traits.email' },
+          then: { '@path': '$.context.traits.email' },
+          else: { '@path': '$.properties.email' }
+        }
+      }
+    },
+    ed: {
+      label: 'Event Data',
+      description:
+        'Event data. Sent to Vibe as a stringified JSON object. The reserved attributes Price (USD) and Purchase ID are merged into this object.',
```
joe-ayoub-segment:
```suggestion
        'Additional event data to send with the event.',
```

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
```diff
@@ -0,0 +1,139 @@
...
+      type: 'string',
+      required: false,
+      default: {
+        '@if': {
+          exists: { '@path': '$.properties.purchase_id' },
+          then: { '@path': '$.properties.purchase_id' },
+          else: { '@path': '$.properties.order_id' }
+        }
+      }
+    },
```
joe-ayoub-segment: Personally I think these fields should be included as properties in the ed field. 

for example something like this: 

ed: {
      label: 'Event Data',
      description: 'Additional data to be sent with the event.',
      type: 'object',
      additionalProperties: true,
      properties: {
           purchase_id: {
                label: 'Purchase ID',
                description:'Reserved event-data attribute: a unique identifier for the purchase. Merged into Event Data as `purchase_id`.',
                type: 'string'
           },
           price_usd: {
                label: 'Price (USD)',
                description: 'Reserved event-data attribute: the price of the conversion in USD. Merged into Event Data as `price_usd`.',
                type: 'number'
           }
      },
      default: { 
           purchase_id: {'@path': '$.properties.purchase_id'},
           price_usd: { '@path': '$.properties.price_usd' }
       }
    },

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
```diff
@@ -0,0 +1,139 @@
...
+      label: 'Batch Size',
+      description: 'Maximum number of events to include in each batch. Actual batch sizes may be lower.',
+      type: 'number',
+      required: false,
+      unsafe_hidden: true,
+      default: 1000
+    }
+  },
+  perform: (request, { payload, settings }) => {
+    return sendEvent(request, settings, payload)
```
joe-ayoub-segment, line 101: Usually I prefer to have a single shared code path for single events and batch events. The path needs accepts an array of payloads, so we wrap the single payload in an array. 

We then need code branches inside the shared function to handle multistatus responses when dealing with a real batch. 

The last param I added (false) is an isBatch param which can be used in the sendBatch function. 

This is a preference I have - but others may disagree with it. 

```suggestion
    return sendBatch(request, settings, [payload], false)
```

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+// reserved attributes (price_usd, purchase_id) into the event data object.
+function buildEventData(payload: Payload): string | undefined {
+  const { ed, price_usd, purchase_id } = payload
+
+  const merged: Record<string, unknown> = { ...(ed ?? {}) }
+  if (price_usd !== undefined && price_usd !== null) merged.price_usd = price_usd
+  if (purchase_id !== undefined && purchase_id !== null) merged.purchase_id = purchase_id
+
+  if (Object.keys(merged).length === 0) return undefined
+  return JSON.stringify(merged)
```
joe-ayoub-segment:
```suggestion
  if (Object.keys(merged).length === 0) { 
     return undefined
  }
  return JSON.stringify(merged)
```

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+// is sent as its own request. Validation and delivery are isolated per event so
+// one invalid or failed event does not fail the whole batch.
+export async function sendBatch(
+  request: RequestClient,
+  settings: Settings,
+  payloads: Payload[]
+): Promise<MultiStatusResponse> {
+  const multiStatusResponse = new MultiStatusResponse()
+
+  await Promise.all(
```
joe-ayoub-segment: This is a bad idea. We should only implement performBatch if the destination API actually has a dedicated batch endpoint to hit. In this case all we're doing is sending a load of single events using Promise.all. Imagine if we get a batch of 1000 events, we'll end up sending 1000 separate post requests :(.

mdkhan-tw: Agreed — removed `performBatch`, `sendBatch`, and the `enable_batching`/`batch_size` fields. The destination now only implements single-event `perform`, since Vibe has no batch endpoint. Fixed in c11a1f2 (and prior).

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
mdkhan-tw: Done — `price_usd` and `purchase_id` are now nested as properties inside the `ed` object field (with `additionalProperties: true`), exactly as suggested. Removed the two separate top-level fields.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
mdkhan-tw: Applied — updated the `ed` description to 'Additional data to be sent with the event.'

packages/destination-actions/src/destinations/vibe-conversions/index.ts
mdkhan-tw: Removed.

packages/destination-actions/src/destinations/vibe-conversions/index.ts
mdkhan-tw: Removed.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
mdkhan-tw: Applied the multi-line brace style throughout the helper.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
mdkhan-tw: Since `performBatch` has been removed entirely (no batch endpoint on Vibe), there's now a single `perform` path only, so the shared-path/`isBatch` refactor no longer applies here. Happy to revisit if Vibe adds a batch endpoint later.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+    throw new PayloadValidationError('Either `ip` (IP Address) or `em` (Email) is required.')
+  }
+
+  const tsMs = toUnixMs(ts)
+  if (tsMs !== undefined) {
+    const now = Date.now()
+    if (now - tsMs > MAX_EVENT_AGE_MS) {
+      throw new PayloadValidationError('`ts` (timestamp) must be within the last 7 days.')
+    }
+  }
```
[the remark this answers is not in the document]
mdkhan-tw: Fixed — added an explicit `tsMs > now` guard that throws 'timestamp cannot be in the future' before the 7-day-age check. Added a test covering it.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    throw new PayloadValidationError('Either `ip` (IP Address) or `em` (Email) is required.')
+  }
+
+  const tsMs = toUnixMs(ts)
+  if (tsMs !== undefined) {
+    const now = Date.now()
+    if (now - tsMs > MAX_EVENT_AGE_MS) {
+      throw new PayloadValidationError('`ts` (timestamp) must be within the last 7 days.')
+    }
+  }
```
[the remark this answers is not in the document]
mdkhan-tw: Fixed — now rejects future timestamps with an explicit guard (`tsMs > now`). Test added.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  return ms
+}
+
+// Build a single typed request body from a payload, applying validation and transforms.
+export function buildConversionEvent(payload: Payload, settings: Settings): ConversionEventRequest {
+  const { a, eid, ts, ip, em, ed, gid, ua, url } = payload
+
+  if (!ip && !em) {
+    throw new PayloadValidationError('Either `ip` (IP Address) or `em` (Email) is required.')
+  }
```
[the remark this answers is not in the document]
mdkhan-tw: Fixed — `ip` is now validated with Node's `net.isIP`; a present-but-non-IPv4 value throws a `PayloadValidationError`. Test added.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return {
+    a: a as EventType,
```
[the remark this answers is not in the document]
mdkhan-tw: Fixed — added an `isEventType` type guard that checks `a` against `EVENT_TYPES` at runtime and throws a `PayloadValidationError`. No more unchecked cast.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+    const now = Date.now()
+    if (now - tsMs > MAX_EVENT_AGE_MS) {
+      throw new PayloadValidationError('`ts` (timestamp) must be within the last 7 days.')
+    }
+  }
+
+  return {
+    a: a as EventType,
+    eid,
+    aid: settings.aid,
```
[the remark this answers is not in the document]
mdkhan-tw: Fixed — `buildConversionEvent` now validates `a` against `EVENT_TYPES` via a type guard and throws `PayloadValidationError`, so the helper is safe regardless of call site.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+function serializeEventData(ed: Payload['ed']): string | undefined {
+  if (ed === undefined || ed === null) return undefined
+  if (typeof ed === 'string') return ed
+  return JSON.stringify(ed)
+}
+
+// Convert an ISO8601 / epoch timestamp to UNIX milliseconds.
+function toUnixMs(ts: Payload['ts']): number | undefined {
+  if (ts === undefined || ts === null || ts === '') return undefined
+  const ms = typeof ts === 'number' ? ts : new Date(ts).getTime()
```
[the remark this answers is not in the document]
mdkhan-tw: Added a test that passes a numeric epoch-ms `ts` and asserts it's forwarded unchanged and still runs through the 7-day validation.

packages/destination-actions/src/destinations/vibe-conversions/index.ts
```diff
@@ -0,0 +1,66 @@
...
+        description: 'The Vibe pixel ID associated with the advertiser. Found in the Vibe dashboard.',
+        required: true
+      }
+    },
+    testAuthentication: (_request, { settings }) => {
+      if (!settings.aid) {
+        throw new Error('Pixel ID is required.')
+      }
+      return true
+    }
```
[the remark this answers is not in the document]
mdkhan-tw: Fixed in c11a1f2 — `testAuthentication` now throws `InvalidAuthenticationError` instead of a generic `Error`.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+        const status = (error as { status?: number })?.status ?? 500
+        multiStatusResponse.setErrorResponseAtIndex(index, {
+          status,
+          errortype: getErrorCodeFromHttpStatus(status),
+          errormessage: (error as Error)?.message ?? 'Delivery failed',
+          sent: body as unknown as Record<string, unknown>
+        })
+      }
+    })
+  )
```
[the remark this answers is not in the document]
mdkhan-tw: Resolved by removing `performBatch` entirely — there's no more fan-out. Vibe has no batch endpoint, so the destination sends single events only.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
```diff
@@ -0,0 +1,139 @@
...
+  },
+  // Vibe's conversion API has no batch endpoint. When batching is enabled,
+  // events are sent as individual requests (one per event) with per-event
+  // error isolation via a MultiStatusResponse.
+  performBatch: (request, { payload, settings }) => {
+    return sendBatch(request, settings, payload)
+  }
```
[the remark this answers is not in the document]
mdkhan-tw: Resolved — removed `performBatch` and the `enable_batching`/`batch_size` fields, so the implementation now matches 'batching intentionally omitted'.

mdkhan-tw:
Thanks for the review @joe-ayoub-segment 🙏 — pushed changes addressing all feedback:

**Batching (removed)**
- Removed `performBatch`, `sendBatch`, and the `enable_batching`/`batch_size` fields. Vibe has no batch endpoint, so the destination sends single events via `perform` only — no `Promise.all` fan-out.

**`ed` field**
- Nested `price_usd` and `purchase_id` as properties inside the `ed` object (with `additionalProperties: true`), per your suggestion. Simplified the description.

**Validation hardening** (from Copilot)
- `a` is validated against `EVENT_TYPES` at runtime via a type guard (no more unchecked cast).
- `ip` is validated as IPv4 via `net.isIP`; non-IPv4 values throw `PayloadValidationError`.
- Future timestamps are now rejected.
- `testAuthentication` throws `InvalidAuthenticationError` instead of a generic `Error`.

**Misc**
- Removed the explanatory comments in `index.ts`.
- Added tests for numeric epoch-ms `ts` passthrough, IPv4 rejection, and future-timestamp rejection.

All 16 unit/snapshot tests pass; lint and typecheck are clean.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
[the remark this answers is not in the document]
joe-ayoub-segment: It might be worth relaxing this - as a client machine's clock might be slightly ahead of Segment's servers. 
We've never applied a guard like this in any destination AFAIK.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
mdkhan-tw: Update: after discussing with the destination owner, we've decided to keep `performBatch` so Segment can group events and so a single bad event doesn't fail the whole batch (per-event isolation via MultiStatusResponse). I've added an explicit comment at both the action and `sendBatch` explaining that Vibe has no batch endpoint and each event is still sent as its own POST. Open to further discussion if you feel strongly it should be single-event only.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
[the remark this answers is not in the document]
mdkhan-tw: Update: `performBatch` is being kept (owner decision) for per-event error isolation and Segment-side grouping. Added comments clarifying Vibe has no batch endpoint; each event is still an individual POST. `enable_batching` defaults to true.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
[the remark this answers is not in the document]
mdkhan-tw: Update: `performBatch` is being retained. Noting your concurrency point — batch sizes are bounded by `batch_size` (default 1000) and Segment's platform controls batch formation. If burst concurrency becomes an issue in practice we can add a limiter, but keeping it simple for now.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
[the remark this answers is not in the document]
mdkhan-tw: Update: reverted this — rejecting future timestamps isn't part of the requirements, so the guard was removed. Only the 7-day-age check remains.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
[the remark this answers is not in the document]
mdkhan-tw: Update: the future-timestamp guard was removed as it's not a stated requirement. Only the 7-day-age validation remains.

packages/destination-actions/src/destinations/vibe-conversions/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        required: true
+      }
+    },
+    testAuthentication: (_request, { settings }) => {
+      if (!settings.aid) {
+        throw new InvalidAuthenticationError('Pixel ID is required.')
+      }
+      return true
+    }
+  },
```
joe-ayoub-segment: This isn't really a valid auth test - the framework won't let the customer turn on the destination without providing aid (which is a required field). So for now I'd remove this function, or find a more valid way to do the test. 

For analytics destinations it's often not possible to check if the settings are valid - so this isn't a disaster.

packages/destination-actions/src/destinations/vibe-conversions/types.ts
```diff
@@ -0,0 +1,32 @@
+import type { EVENT_TYPES } from './constants'
```
joe-ayoub-segment: This file is only referenced within the trackConversion action. Let's move it in there.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,94 @@
+import { isIP } from 'net'
```
joe-ayoub-segment, line 1: This file is only referenced within the trackConversion action. Let's move it in there.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+    em: em || undefined,
+    ed: serializeEventData(ed),
+    gid: gid || undefined,
+    ua: ua || undefined,
+    url: url || undefined
+  }
+}
+
+// Post a single typed request body to the Vibe conversion endpoint.
+function postEvent(request: RequestClient, body: ConversionEventRequest) {
```
joe-ayoub-segment: let's rename to json: 


```suggestion
function postEvent(request: RequestClient, json: ConversionEventRequest) {
```

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  return request<ConversionEventResponse>(CONVERSION_ENDPOINT, {
+    method: 'POST',
+    headers: {
+      accept: 'application/json',
+      'content-type': 'application/json'
+    },
+    json: body
```
joe-ayoub-segment:
```suggestion
    json
```

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+import type { RequestClient } from '@segment/actions-core'
+import { PayloadValidationError } from '@segment/actions-core'
+import type { Settings } from './generated-types'
+import type { Payload } from './trackConversion/generated-types'
+import type { ConversionEventRequest, ConversionEventResponse, EventType } from './types'
+import { CONVERSION_ENDPOINT, EVENT_TYPES, MAX_EVENT_AGE_MS } from './constants'
+
+function isEventType(value: string): value is EventType {
+  return (EVENT_TYPES as readonly string[]).includes(value)
+}
```
joe-ayoub-segment, line 11: not really needed as the field will do this validation automagically. no real harm leaving this in place though.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!isEventType(a)) {
+    throw new PayloadValidationError(`\`a\` (Event Type) must be one of: ${EVENT_TYPES.join(', ')}.`)
+  }
+
+  if (!ip && !em) {
+    throw new PayloadValidationError('Either `ip` (IP Address) or `em` (Email) is required.')
+  }
+
+  if (ip && isIP(ip) !== 4) {
```
joe-ayoub-segment, line 55: nice! I never knew about that library.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+function toUnixMs(ts: Payload['ts']): number | undefined {
+  if (ts === undefined || ts === null || ts === '') {
+    return undefined
+  }
+  const ms = typeof ts === 'number' ? ts : new Date(ts).getTime()
+  if (Number.isNaN(ms)) {
+    throw new PayloadValidationError('`ts` (timestamp) is not a valid date.')
+  }
+  return ms
+}
```
joe-ayoub-segment, line 41: I think we should ask Vibe about which unix timestamp format they need. We could then tighten up this function.

joe-ayoub-segment (changes requested):

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  if (ip && isIP(ip) !== 4) {
+    throw new PayloadValidationError('`ip` (IP Address) must be a valid IPv4 address.')
+  }
+
+  const tsMs = toUnixMs(ts)
+  if (tsMs !== undefined && Date.now() - tsMs > MAX_EVENT_AGE_MS) {
+    throw new PayloadValidationError('`ts` (timestamp) must be within the last 7 days.')
+  }
```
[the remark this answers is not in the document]
mdkhan-tw: This is not required as per spec

packages/destination-actions/src/destinations/vibe-conversions/types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export type EventType = typeof EVENT_TYPES[number]
```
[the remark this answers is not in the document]
mdkhan-tw: Not a biggest concern imo

packages/destination-actions/src/destinations/vibe-conversions/index.ts
mdkhan-tw: Removed `testAuthentication`. Since `aid` is a required field the framework already blocks enabling the destination without it, so the check added nothing. Left it out rather than inventing a synthetic call, as you suggested.

packages/destination-actions/src/destinations/vibe-conversions/types.ts
mdkhan-tw: Moved `types.ts` into `trackConversion/` (`git mv`) and fixed the relative imports.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
mdkhan-tw: Moved `utils.ts` into `trackConversion/` (`git mv`) and fixed the relative imports.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
mdkhan-tw: Done — renamed the param to `json`.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
mdkhan-tw: Done — using the `json` shorthand.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
mdkhan-tw: Applied the suggested wording.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
mdkhan-tw: Agreed — removed `performBatch`. There's no dedicated batch endpoint, so the destination now does single-event POSTs only rather than fanning out with `Promise.all`.

packages/destination-actions/src/destinations/vibe-conversions/utils.ts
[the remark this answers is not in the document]
mdkhan-tw: Relaxed — dropped the future-timestamp guard. The only remaining check is the 7-day lower bound.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
mdkhan-tw: Done — `purchase_id` and `price_usd` are now typed sub-properties of `ed` with `@path` defaults, and get merged into the serialized Event Data.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/index.ts
mdkhan-tw: Left as a single-event path for now. Since we've dropped batching entirely (no batch endpoint on Vibe's side), there's no `sendBatch`/multistatus branch to share into — happy to revisit if a batch endpoint appears.

packages/destination-actions/src/destinations/vibe-conversions/index.ts
mdkhan-tw: Removed.

packages/destination-actions/src/destinations/vibe-conversions/index.ts
mdkhan-tw: Removed.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
mdkhan-tw: Kept it — as you noted it's harmless, and it also narrows the type so `a` is typed as `EventType` (not `string`) inside `buildConversionEvent`.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
mdkhan-tw: Good call — flagged this with Vibe. Will tighten `toUnixMs` once they confirm the exact unix format (seconds vs milliseconds) they expect.

packages/destination-actions/src/destinations/vibe-conversions/trackConversion/utils.ts
mdkhan-tw: 🙌 `net.isIP` is a handy built-in — no dependency needed.

joe-ayoub-segment (approved):

PR #3859, opened 2026-07-06
fix(ms-bing-capi): item price should be number not integer

scottlepich-lz:
## Summary

The Microsoft Bing CAPI destination's `items` InputField defines the per-item `price` property with `type: 'integer'`. This is too strict: Microsoft's own UET Conversions API accepts **decimal** item prices in whole currency units.

Because of this, Segment's mapping-kit rejects perfectly valid decimal prices (e.g. `9.99`) with:

```
400: Item Price must be an integer but it was a number
```

Since mapping validation fails, the **entire event is dropped** — not just the price field. Any storefront selling at non-round prices (which is essentially all of them) silently loses conversion events.

## Evidence (Microsoft docs)

Microsoft's official UET Conversions API integration guide documents `price` as a decimal:

- Docs: https://learn.microsoft.com/en-us/advertising/guides/uet-conversion-api-integration
- The parameter table gives the `price` example value as `25.1` (a decimal).
- Every JSON sample in that guide uses decimal prices, e.g. `"price": 25.1` and `"price": 27.3`.
- The related top-level `ecommTotalValue` is typed `number($double)`.

So `integer` is incorrect; the correct type is `number`.

## Fix

Change the `items[].price` field from `type: 'integer'` to `type: 'number'`. `quantity` is left as `integer` (correct for quantity). Nothing else is changed.

`generated-types.ts` is unchanged, because both `integer` and `number` map to the TypeScript type `number`. Codegen (`generate:types`) was run for the destination and produced no diff.

## Testing

- [x] Added a unit test in `sendEvent/__tests__/sendEvent.test.ts` proving a decimal item price (`9.99`) is now accepted and forwarded to the request payload without a validation error.
- Verified the new test fails against the old `type: 'integer'` with the exact error `Item Price must be an integer but it was a number`, and passes with `type: 'number'`.
- Ran the scoped test suite: **16 passed, 16 total** (`jest src/destinations/ms-bing-capi`).
- Ran typecheck and eslint for the touched destination — clean.

## Follow-up observation (out of scope, not changed here)

While in this file I noticed a separate latent bug: the `items` field's `default` mapping writes to key `item_price`, but the defined property is `price`. Since the field is `additionalProperties: false`, `item_price` is silently dropped, so the **default mapping never sends a price at all**. I've deliberately left this out of this PR to keep it focused on the type fix; flagging it as a follow-up.


itsarijitray (approved):

AnkitSegment (approved):

nk1107:
PR is deployed @scottlepich-lz

PR #3860, opened 2026-07-07
[Quora Conversions API] - new cloud-mode destination

mdkhan-tw:
Adds a new **cloud-mode** destination for the **Quora Conversions API**, allowing customers to send server-side conversion events to Quora for ad attribution and optimization.

The destination exposes a single `trackConversion` action that supports all 10 standard Quora conversion types. A shared `sendEvents(request, settings, items, isBatch)` builder backs both `perform` (single) and `performBatch` (batch), branching only where the request body and response handling genuinely differ.

**Highlights**
- **Auth:** custom Bearer token (`Authorization: Bearer <api_token>`) + numeric `account_id`. `testAuthentication` is a presence check — Quora exposes no auth-verification endpoint, and a live probe would create a test conversion.
- **Single** (`POST /ads/v0/conversion`) posts a flat `{ user, device, conversion, account_id }` body.
- **Batch** (`POST /ads/v0/conversions`, up to 1000 events) posts `{ account_id, data: [ { user, device, conversion } ] }` with `account_id` hoisted to the root. Returns an HTTP 200 **multi-status** body which is interrogated per-`index` and mapped into a `MultiStatusResponse`.
- **Presets (8):** Application Installed→AppInstall, Order Completed→Purchase, Signed Up→CompleteRegistration, Payment Info Entered→AddPaymentInfo, Product Added→AddToCart, Product Added to Wishlist→AddToWishlist, Checkout Started→InitiateCheckout, Products Searched→Search. Selecting `Generic` passes the Segment event name through as `event_name`.
- **Field details** (per Quora clarifications): API field is spelled `referer` (single r); `mobile_device_id` is the advertising ID (IDFA/AAID); `date_of_birth` is `YYYY-MM-DD`; `country` is ISO 3166-1 alpha-2; `language` is a locale string passed through. Timestamps are converted to **epoch microseconds**. PII is sent **plaintext** (no hashing), matching the API docs.

**⚠️ TODO before merge:** this destination is intentionally **not** registered in `packages/destination-actions/src/destinations/index.ts`. Registration requires the production-assigned metadata ID (a real MongoDB ObjectId created when the destination is set up in Segment's control plane) — a placeholder ID would pollute the shared registry.

**Open questions to confirm with Quora** (code uses sensible defaults, none blocking): endpoint rate limits; the `click_id`/`qclid` source path for cloud-only mode (a Hybrid Action to capture `qclid` from the URL is planned as a follow-up); the `value` source (`properties.value` → `properties.total` fallback); and confirmation that hashing is not required.

## Testing

- [x] Added [unit tests](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) for new functionality — 26 tests (action, utils/multi-status, auth, snapshot)
- [x] Tested end-to-end using the [local server](https://github.com/segmentio/action-destinations/blob/main/docs/testing.md#local-end-to-end-testing) — drove single + batch events through `./bin/run serve`; verified correct endpoint routing, body shape, `account_id` coercion, µs timestamp, and Bearer auth against the live host
- [ ] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change. — N/A, new destination
- [ ] [Segmenters] Tested in the staging environment — pending deploy
- [ ] [Segmenters] [If applicable for this change] Tested for regression with Hadron. — N/A

## Security Review

- [x] **Reviewed all field definitions** for sensitive data (API keys, tokens, passwords, client secrets) and confirmed they use `type: 'password'` — `api_token` is `type: 'password'`; `account_id` is a non-sensitive identifier

## New Destination Checklist

- [ ] Extracted all action API versions to `versioning-info.ts` file. — The API version (`v0`) is captured in endpoint path constants in `constants.ts`; can extract to `versioning-info.ts` if preferred.


joe-ayoub-segment:
Hi @mdkhan-tw - let's also renamed the folder containing this destination to simply 'quora'. currently it's actions-quora.

packages/destination-actions/src/destinations/quora/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      // otherwise create a test conversion).
+      if (!settings.api_token) {
+        throw new InvalidAuthenticationError('Please provide an API Token.')
+      }
+      if (!settings.account_id || !/^\d+$/.test(settings.account_id.trim())) {
+        throw new InvalidAuthenticationError('Please provide a valid numeric Account ID.')
+      }
+      return Promise.resolve(true)
+    }
+  },
```
joe-ayoub-segment, line 39: The settings fields, being required, already do this validation. 
We should either do some proper validation via an API request to Quora (if possible), or remove this function.

packages/destination-actions/src/destinations/quora/trackConversion/constants.ts
```diff
@@ -0,0 +1,46 @@
+/**
```
joe-ayoub-segment, line 1: this file is only referenced inside the trackConversion action. let's move it into that folder.

packages/destination-actions/src/destinations/quora/trackConversion/types.ts
```diff
@@ -0,0 +1,75 @@
+/**
```
joe-ayoub-segment, line 1: this file is only referenced inside the trackConversion action. let's move it into that folder.

packages/destination-actions/src/destinations/quora/trackConversion/utils.ts
```diff
@@ -0,0 +1,213 @@
+import {
```
joe-ayoub-segment, line 1: this file is only referenced inside the trackConversion action. let's move it into that folder.

packages/destination-actions/src/destinations/actions-quora/trackConversion/fields.ts
```diff
@@ -0,0 +1,156 @@
...
+    default: 'Generic'
+  },
+  segment_event_name: {
+    label: 'Segment Event Name',
+    description:
+      'The raw Segment event name. Only used when Event Name is set to `Generic`, in which case this value is sent as the Quora `event_name`.',
+    type: 'string',
+    required: false,
+    default: { '@path': '$.event' }
+  },
```
joe-ayoub-segment: Let's make this field depend on what the user selects for event_name. 


```suggestion
  segment_event_name: {
    label: 'Segment Event Name',
    description:
      'The raw Segment event name. Only used when Event Name is set to `Generic`, in which case this value is sent as the Quora `event_name`.',
    type: 'string',
    required: {
      conditions: [
        {
          fieldKey: 'event_name',
          operator: 'is',
          value: 'Generic'
        }
      ]
    },
    default: { '@path': '$.event' }, 
    depends_on: { 
      conditions: [
        {
          fieldKey: 'event_name',
          operator: 'is',
          value: 'Generic'
        }
      ]
    } 
  },
```

packages/destination-actions/src/destinations/actions-quora/trackConversion/fields.ts
```diff
@@ -0,0 +1,156 @@
...
+    label: 'Click ID (qclid)',
+    description:
+      'The Quora click ID (`qclid`), used to attribute the conversion to a specific ad click. Required by Quora for attribution.',
+    type: 'string',
+    required: false,
+    default: {
+      '@if': {
+        exists: { '@path': '$.properties.qclid' },
+        then: { '@path': '$.properties.qclid' },
+        else: { '@path': '$.context.qclid' }
```
joe-ayoub-segment: We're going to build a hybrid action to pass this from the front end. This is the path we will use for it. 

```suggestion
        else: { '@path': '$.integrations.Quora Conversions API.qclid' }
```

packages/destination-actions/src/destinations/actions-quora/trackConversion/fields.ts
```diff
@@ -0,0 +1,156 @@
...
+  value: {
+    label: 'Value',
+    description: 'The monetary value associated with the conversion.',
+    type: 'number',
+    required: false,
+    default: {
+      '@if': {
+        exists: { '@path': '$.properties.value' },
+        then: { '@path': '$.properties.value' },
+        else: { '@path': '$.properties.total' }
```
joe-ayoub-segment: Segment ecommerce spec uses revenue and total. I don't believe there is a value attribute.... 

```suggestion
        exists: { '@path': '$.properties.revenue' },
        then: { '@path': '$.properties.revenue' },
        else: { '@path': '$.properties.total' }
```

packages/destination-actions/src/destinations/quora/trackConversion/fields.ts
```diff
@@ -0,0 +1,156 @@
...
+      phone_number: {
+        label: 'Phone Number',
+        description: "The user's phone number. E.164 format is preferred but not enforced.",
+        type: 'string'
+      },
+      date_of_birth: {
+        label: 'Date of Birth',
+        description: "The user's date of birth in YYYY-MM-DD format.",
+        type: 'string'
+      },
```
joe-ayoub-segment, line 74: Lets change this so a regular date is accepted by Segment - and we can then transform it to the YYYY-MM-DD format in the action code. Customers are used to sending dates in ISO8601 format. And $.context.traits.birthday is usually going to be in that format. 

```suggestion
      date_of_birth: {
        label: 'Date of Birth',
        description: "The user's date of birth.",
        type: 'date'
      },
```

packages/destination-actions/src/destinations/quora/trackConversion/fields.ts
```diff
@@ -0,0 +1,156 @@
...
+      referer: {
+        label: 'Referer',
+        description: 'The referring URL. Note the API spells this field "referer" (single r).',
+        type: 'string'
+      }
+    },
+    default: {
+      mobile_device_id: { '@path': '$.context.device.advertisingId' },
+      user_agent: { '@path': '$.context.userAgent' },
+      language: { '@path': '$.context.locale' },
```
joe-ayoub-segment, line 129: we might need to split out the language part from the locale. need to check.

packages/destination-actions/src/destinations/actions-quora/trackConversion/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+}
+
+/**
+ * Resolves the Quora `event_name`. When the mapped event name is `Generic`,
+ * the raw Segment event name is passed through instead.
+ */
+export function resolveEventName(payload: Payload): string {
+  if (payload.event_name === GENERIC_EVENT_NAME) {
+    return clean(payload.segment_event_name) ?? GENERIC_EVENT_NAME
+  }
```
joe-ayoub-segment: segment_event_name will be required if payload.event_name === GENERIC_EVENT_NAME. so a fallback is not needed. 

```suggestion
  if (payload.event_name === GENERIC_EVENT_NAME) {
    return clean(payload.segment_event_name)
  }
```

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Builds a single Quora conversion item (`{ user, device, conversion }`) from a payload.
+ * Shared by both the single and batch delivery paths. Empty sub-objects are omitted.
+ * Throws a PayloadValidationError if no event name is resolvable.
+ */
+export function buildConversionItem(payload: Payload): QuoraConversionItem {
+  const eventName = resolveEventName(payload)
+  if (!clean(eventName)) {
+    throw new PayloadValidationError('Event Name is required.')
+  }
```
joe-ayoub-segment: Not needed. eventName will always have a value.

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
+import {
+  RequestClient,
+  MultiStatusResponse,
+  JSONLikeObject,
+  ModifiedResponse,
+  PayloadValidationError,
+  IntegrationError
```
joe-ayoub-segment:
```suggestion
  IntegrationError
  removeUndefined,
```

packages/destination-actions/src/destinations/quora/trackConversion/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+    name: clean(payload.user?.name),
+    phone_number: clean(payload.user?.phone_number),
+    date_of_birth: clean(payload.user?.date_of_birth),
+    ip: clean(payload.user?.ip),
+    country: clean(payload.user?.country),
+    region: clean(payload.user?.region),
+    city: clean(payload.user?.city),
+    postal_code: clean(payload.user?.postal_code)
+  }
+
```
joe-ayoub-segment, line 112:
```suggestion
  const user = removeUndefined({
    email: clean(payload.user?.email),
    name: clean(payload.user?.name),
    phone_number: clean(payload.user?.phone_number),
    date_of_birth: clean(payload.user?.date_of_birth),
    ip: clean(payload.user?.ip),
    country: clean(payload.user?.country),
    region: clean(payload.user?.region),
    city: clean(payload.user?.city),
    postal_code: clean(payload.user?.postal_code)
  })
```

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+    city: clean(payload.user?.city),
+    postal_code: clean(payload.user?.postal_code)
+  }
+
+  const device = {
+    mobile_device_id: clean(payload.device?.mobile_device_id),
+    user_agent: clean(payload.device?.user_agent),
+    language: clean(payload.device?.language),
+    referer: clean(payload.device?.referer)
+  }
```
joe-ayoub-segment:
```suggestion
  const device = removeUndefined({
    mobile_device_id: clean(payload.device?.mobile_device_id),
    user_agent: clean(payload.device?.user_agent),
    language: clean(payload.device?.language),
    referer: clean(payload.device?.referer)
  })
```

packages/destination-actions/src/destinations/quora/trackConversion/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+  const item: QuoraConversionItem = {
+    conversion: {
+      event_name: eventName,
+      timestamp: toEpochMicroseconds(payload.timestamp),
+      click_id: clean(payload.click_id),
+      value: payload.value ?? undefined,
+      event_id: clean(payload.event_id)
+    }
+  }
```
joe-ayoub-segment, line 128:
```suggestion
  const item: QuoraConversionItem = {
    conversion: removeUndefined({
      event_name: eventName,
      timestamp: toEpochMicroseconds(payload.timestamp),
      click_id: clean(payload.click_id),
      value: payload.value ?? undefined,
      event_id: clean(payload.event_id)
    })
  }
```

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+      json: body as unknown as JSONLikeObject,
+      throwHttpErrors: false
+    })
+  }
+
+  const body: QuoraSingleRequest = { account_id, ...items[0] }
+  return request<QuoraSingleResponse>(SINGLE_ENDPOINT, {
+    method: 'POST',
+    json: body as unknown as JSONLikeObject
+  })
```
joe-ayoub-segment:
```suggestion
  const json: QuoraSingleRequest = { account_id, ...items[0] }
  return request<QuoraSingleResponse>(SINGLE_ENDPOINT, {
    method: 'POST',
    json
  })
```

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+export async function sendEvents(
+  request: RequestClient,
+  settings: Settings,
+  items: QuoraConversionItem[],
+  isBatch: boolean
+): Promise<ModifiedResponse<QuoraSingleResponse | QuoraBatchResponse>> {
+  const account_id = resolveAccountId(settings)
+
+  if (isBatch) {
+    const body: QuoraBatchRequest = { account_id, data: items }
```
joe-ayoub-segment:
```suggestion
    const json: QuoraBatchRequest = { account_id, data: items }
```

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+    return request<QuoraBatchResponse>(BATCH_ENDPOINT, {
+      method: 'POST',
+      json: body as unknown as JSONLikeObject,
```
joe-ayoub-segment:
```suggestion
      json
```

packages/destination-actions/src/destinations/actions-quora/trackConversion/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  description:
+    'Send a conversion event to Quora. Supports all standard Quora conversion types, plus Generic pass-through of the Segment event name.',
+  defaultSubscription: 'type = "track"',
+  fields,
+
+  perform: async (request, { payload, settings }) => {
+    const item = buildConversionItem(payload)
+    const response = await sendEvents(request, settings, [item], false)
+    assertSingleEventSucceeded(response)
+    return response
```
joe-ayoub-segment, line 19: The framework will handle the response for a single event. no processing of single event response needed here. 

```suggestion
    return await sendEvents(request, settings, [item], false)
```

packages/destination-actions/src/destinations/quora/trackConversion/index.ts
```diff
@@ -0,0 +1,49 @@
+import { ActionDefinition, MultiStatusResponse } from '@segment/actions-core'
```
joe-ayoub-segment, line 1:
```suggestion
import { ActionDefinition, MultiStatusResponse, JSONLikeObject } from '@segment/actions-core'
```

packages/destination-actions/src/destinations/actions-quora/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+export function assertSingleEventSucceeded(response: ModifiedResponse<QuoraSingleResponse | QuoraBatchResponse>): void {
+  const data = response.data
+  if (!data) return
+  const errored = data.events?.find((event) => event.status !== 'OK')
+  if (data.events_errored || errored) {
+    const code = errored?.error_code ?? 'unknown error'
+    const message = errored?.error_message ?? `Quora rejected the event: ${code}`
+    throw new PayloadValidationError(message)
+  }
+}
```
joe-ayoub-segment: this function is not needed. The framework will handle the response for single event.

Let's delete assertSingleEventSucceeded

packages/destination-actions/src/destinations/actions-quora/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+    }
+  },
+
+  presets: [
+    {
+      name: 'App Install',
+      subscribe: 'type = "track" and event = "Application Installed"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AppInstall' },
```
joe-ayoub-segment:
```suggestion
      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AppInstall', value: '@path': '$.properties.value' },
```

packages/destination-actions/src/destinations/actions-quora/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      subscribe: 'type = "track" and event = "Order Completed"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'Purchase' },
+      type: 'automatic'
+    },
+    {
+      name: 'Complete Registration',
+      subscribe: 'type = "track" and event = "Signed Up"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'CompleteRegistration' },
```
joe-ayoub-segment:
```suggestion
      mapping: { ...defaultValues(trackConversion.fields), event_name: 'CompleteRegistration', value: '@path': '$.properties.value' },
```

packages/destination-actions/src/destinations/actions-quora/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      subscribe: 'type = "track" and event = "Payment Info Entered"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AddPaymentInfo' },
+      type: 'automatic'
+    },
+    {
+      name: 'Add to Cart',
+      subscribe: 'type = "track" and event = "Product Added"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AddToCart' },
```
joe-ayoub-segment:
```suggestion
      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AddToCart', value: '@path': '$.properties.price'  },
```

packages/destination-actions/src/destinations/actions-quora/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      type: 'automatic'
+    },
+    {
+      name: 'Add to Wishlist',
+      subscribe: 'type = "track" and event = "Product Added to Wishlist"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AddToWishlist' },
```
joe-ayoub-segment:
```suggestion
      mapping: { ...defaultValues(trackConversion.fields), event_name: 'AddToWishlist', value: '@path': '$.properties.price'  },
```

packages/destination-actions/src/destinations/actions-quora/index.ts
```diff
@@ -0,0 +1,116 @@
...
+      subscribe: 'type = "track" and event = "Checkout Started"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'InitiateCheckout' },
+      type: 'automatic'
+    },
+    {
+      name: 'Search',
+      subscribe: 'type = "track" and event = "Products Searched"',
+      partnerAction: 'trackConversion',
+      mapping: { ...defaultValues(trackConversion.fields), event_name: 'Search' },
```
joe-ayoub-segment:
```suggestion
      mapping: { ...defaultValues(trackConversion.fields), event_name: 'Search', value: '@path': '$.properties.value'  },
```

joe-ayoub-segment (changes requested):

quoraallan0:
Ran this through our review tooling and checked it against our Conversions API contract. Strong implementation overall, and it gets the tricky parts right: plaintext email (no client-side hashing), `referer` (single r) from `context.page.referrer`, `event_id` from `messageId`, timestamp to microseconds, `mobile_device_id` from `advertisingId` (not the IDFV), and the MultiStatus-by-returned-index mapping. A few things to fix:

**Must fix**

1. **Generic passthrough sends an invalid `event_name` (utils.ts:42-47, fields.ts:12).** `event_name` defaults to `Generic` and `segment_event_name` defaults to `$.event`, so out of the box the raw Segment event name (e.g. "Product Viewed") is sent as `conversion.event_name`. Our API accepts only the 10 enum values and rejects anything else (400 single / per-item error in batch), so every default-configured event fails. When `Generic` is selected, send the literal `"Generic"`; the raw-name passthrough isn't compatible with the enum.

2. **A 429 (or 5xx) marks the whole batch as terminal per-item errors instead of retrying (utils.ts:170-183).** Our batch endpoint rejects the entire batch with 429 when it would exceed the 1,000 events/min rate limit. The transport-failure branch calls `setErrorResponseAtIndex` for every item, which Segment treats as terminal, so a rate-limited batch is dropped rather than retried with backoff. For retryable statuses (429, 5xx), throw so Segment retries the batch.

3. **click_id has no URL capture and a dead fallback (fields.ts:35-41).** `click_id` reads `$.properties.qclid` then `$.context.qclid` (not a real Segment path). qclid arrives on the landing-page URL, and there's no web/hybrid action to capture and cache it. Since qclid is required for attribution, conversions won't attribute unless the advertiser manually populates `properties.qclid`. Is the web action a follow-up PR? If so, great; the `$.context.qclid` fallback should be dropped or fixed meanwhile.

**Should fix**

4. **`value` misses `revenue` (fields.ts:48-54).** It reads `properties.value` then `properties.total`, but Segment's Order Completed carries the amount in `properties.revenue`, so Purchase values usually come through empty. Suggest revenue then value then total.

5. **Single-endpoint response is modeled as batch JSON it never returns (utils.ts:146, types.ts:71-75).** The single endpoint returns HTTP 200 body `"OK"` or a 400 plain-text body, not `{events:[...]}`. So `assertSingleEventSucceeded` is a no-op on success, and single-event errors only surface because the framework throws on the 400 (the single path doesn't set `throwHttpErrors:false`). The `QuoraSingleResponse` type is fictional. Either drop the dead check and rely on the framework's HTTP error, or model the real text contract.

**Minor**

6. `toEpochMicroseconds` (utils.ts:33) assumes a numeric timestamp is epoch-ms; a numeric epoch-seconds value would be 1000x off, and the server then silently uses now(). Mitigated by the `datetime` field type coercing to ISO, but worth guarding for custom numeric mappings.
7. `date_of_birth` (fields.ts:101) forwards `context.traits.birthday` unnormalized; Segment's birthday is often a full ISO datetime, not the `YYYY-MM-DD` the field describes.
8. `email` reads only `context.traits.email`; consider a `properties.email` fallback.
9. `GenerateLead` is in the enum but has no preset.
10. Batch: items absent from `events[]` default to success (utils.ts:203). Safe with our API today (it echoes every item), but it would mask data loss if that ever changes.

**Nits (optional cleanup):** `QuoraSingleResponse` and `QuoraBatchResponse` are identical (dedupe to one type); `sendEvents`' `isBatch` boolean forks endpoint, body, and return type (two functions would read cleaner); the 8 presets are copy-paste that can drift from `QUORA_EVENT_NAMES` (a small mapping table keeps them in sync).

**On `testAuthentication`:** the presence-check is the right call for now. There's no clean way to live-validate credentials today (a valid authenticated request would create a real conversion), but we're adding a `test` flag to the CAPI that will let a testAuthentication call validate auth without recording anything. No change needed on your side now; we'll follow up when it lands.

quoraallan0:
Correction from our call on the conversion `value` currency. I said to convert values to USD, and that's wrong, apologies.

I traced it in our code: conversion value is stored and reported in the **ad account's own currency**, with no FX conversion applied anywhere. It flows through the same path as spend and budget (both account-currency): the value is rescaled to our internal `cc` unit at ingestion and rendered back in the account's currency for reporting, never converted via an exchange rate.

So the `value` field should be documented as "the value in your ad account's currency," not USD. An EUR advertiser sends `value: 100` for €100 and QAM reports €100; converting to USD first would misreport it.

Two internal artifacts of ours (a `# Convert value from USD to cc` comment and a `VALUE_OUT_OF_RANGE` error string that says "ensure the value is in USD") were the source of the bad steer, and we're correcting those separately.

quoraallan0:
Screenshot 2026-07-14 at 10 06 46

btw: you'll see qclid when you click on one of our native ads (looks like above), likely on feed to find one. It is not attached to our programmatic ads so the right-hand rail one did not have it.

packages/destination-actions/src/destinations/quora/trackConversion/index.ts
```diff
@@ -0,0 +1,49 @@
...
+      try {
+        validItems.push(buildConversionItem(payload))
+        validPayloadIndices.push(index)
+      } catch (error) {
+        multiStatusResponse.setErrorResponseAtIndex(index, {
+          status: error?.status || 400,
+          errortype: error?.code || 'PAYLOAD_VALIDATION_FAILED',
+          errormessage: error?.message || 'Validation failed'
+        })
+      }
```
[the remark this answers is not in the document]
joe-ayoub-segment: Not applicable here: this package sets useUnknownInCatchVariables: false in tsconfig.build.json, so the catch variable is typed as any and error?.status/code/message type-check fine. Resolving.

packages/destination-actions/src/destinations/quora/trackConversion/utils.ts
```diff
@@ -0,0 +1,254 @@
...
+      event_id: clean(payload.event_id)
+    })
+  }
+
+  if (Object.values(user).some((v) => v !== undefined)) {
+    item.user = user
+  }
+  if (Object.values(device).some((v) => v !== undefined)) {
+    item.device = device
+  }
```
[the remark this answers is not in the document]
joe-ayoub-segment: Style preference with no behavioral difference — leaving as-is. Resolving.

packages/destination-actions/src/destinations/quora/trackConversion/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    payloads.forEach((payload, index) => {
+      try {
+        validItems.push(buildConversionItem(payload))
+        validPayloadIndices.push(index)
+      } catch (error) {
+        multiStatusResponse.setErrorResponseAtIndex(index, {
+          status: error?.status || 400,
+          errortype: error?.code || 'PAYLOAD_VALIDATION_FAILED',
+          errormessage: error?.message || 'Validation failed'
+        })
```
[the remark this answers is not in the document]
joe-ayoub-segment: Not applicable: this package sets useUnknownInCatchVariables: false in tsconfig.build.json, so the catch variable is any and error?.status/code/message type-check fine. Resolving.

packages/destination-actions/src/destinations/quora/trackConversion/fields.ts
```diff
@@ -0,0 +1,149 @@
...
+      'The monetary value associated with the conversion, in your account currency. Sent to Quora as a number, for example `99.99`. Do not include currency symbols or codes.',
+    type: 'number',
+    required: false,
+    default: {
+      '@if': {
+        exists: { '@path': '$.properties.revenue' },
+        then: { '@path': '$.properties.revenue' },
+        else: { '@path': '$.properties.total' }
+      }
+    }
```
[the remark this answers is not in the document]
joe-ayoub-segment: The value defaults are intentionally preset-dependent: the field-level default uses properties.revenue → properties.total (the standard Segment revenue convention) for direct/unmapped use, while individual presets override the source to match their event semantics (e.g. properties.value for purchase-style events, properties.price for product events). This is deliberate. Resolving.

joe-ayoub-segment (approved):

PR #3863, opened 2026-07-07
fix(ms-bing-capi): do not fail batched events on validation warnings

scottlepich-lz:
## Summary

The Microsoft Bing CAPI `sendEvent` action sends events with `continueOnValidationError: true`, which means Microsoft's CAPI **accepts** an event (HTTP `200`, `eventsReceived: 1`) even when it has non-fatal issues, and reports those issues as entries in `error.details[]` flagged with `"isWarning": true`.

The batch response handler (`performBatch`) matched `error.details[]` by `index` **only** and treated ANY matching detail as a hard failure. As a result, a batched event whose only detail was a warning was marked `status: 400` and reported back to Segment as a **failed delivery**, even though Microsoft had already accepted it. This caused real, silent delivery failures for any batched event that triggered a Microsoft warning.

The single-event `perform` path returns the raw `200` response, so the bug only manifested in batching.

Example real response detail (Microsoft returns `200` with):

```json
{ "errorCode": "Empty", "errorMessage": "'price' must not be empty.", "index": 0, "isWarning": true, "propertyName": "data[0].customData.items[0].price" }
```

### Fix

- `index.ts`: only treat a detail as a failure when it is **not** a warning — `details.find((detail) => detail.index === index && !detail.isWarning)`. A warning-only event is now marked SUCCESS (`200`); an event with a real (non-warning) error at its index is still marked `400`. `continueOnValidationError: true` and the `perform`/`performBatch` structure are unchanged.
- `types.ts`: extended the `MSMultiStatusResponse.error.details[]` element type with the fields the API actually returns (`isWarning?: boolean`, `errorCode?: string`).

No `generated-types.ts` change — this is response handling, not fields.

## Testing

Added `performBatch` unit tests proving:
- a `200` response with an `isWarning: true` detail at an event's index -> that event is a **SUCCESS (200)**, not a `400`;
- a real error detail (`isWarning: false`) at an index -> that event is still marked **`400`**, while a warning at a different index stays `200`;
- an error detail with no `isWarning` field -> defaults to a real failure (`400`), preserving existing behavior.

Ran only the ms-bing-capi tests (`jest src/destinations/ms-bing-capi`): **18 passed, 18 total**. Typecheck and ESLint on the changed files are clean; Prettier reports no formatting issues.

- [x] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change. _(No new fields; a fix to existing response handling.)_

## Security Review

- [x] **Reviewed all field definitions** for sensitive data — no field definitions changed in this PR.

## Related

Sibling to the item-price fix in #3859 (https://github.com/segmentio/action-destinations/pull/3859), which addresses a separate issue in the same destination. This PR does not depend on or touch that change.


scottlepich-lz:
Superseded by #3859. The batch-warning fix (`do not fail batched events on validation warnings`) has been cherry-picked onto #3859, which also carries the `items.price` integer→number field fix and the regenerated `metadata.json`. Consolidating both ms-bing-capi fixes into #3859 so they share a single review and metadata regeneration.

PR #3870, opened 2026-07-09
Add VoiceOps name mapping fields

toddtarsi:
## What changed

Adds explicit optional name mapping fields to the VoiceOps Send Call Completed action:

- customer_first_name
- customer_last_name
- agent_first_name
- agent_last_name
- assign_first_last_name_by_splitting_email

When email splitting is enabled, missing primary agent and agent-leg names are derived from the email local-part before the request is sent to VoiceOps. For example, ava-agent@voiceops.com derives Ava / Agent.

## Why

VoiceOps customers need clearer mapping controls for customer and agent names, plus a fallback for deriving agent names from email addresses when names are not provided by the source event.

## Compatibility

- No new required fields.
- Scope is limited to the VoiceOps integration.
- Existing required call fields and validation remain unchanged.

## Validation

- corepack yarn build
- TZ=UTC ./node_modules/.bin/jest --testPathPattern=voiceops --runInBand -u
- git diff --check

Note: the local pre-commit hook could not run because it invokes yarn directly and yarn is not on PATH in this shell; the equivalent build/test checks above were run manually.

## Segment review note

Before marking this ready for review, attach an Action Tester video showing the new mapping fields and email-splitting behavior with a realistic Call Completed payload.


toddtarsi:
Video of test:

[bf1496384aac109e26b89d1ab0529d13.webm](https://github.com/user-attachments/assets/06997e4c-4875-4a5b-b0b4-835612df1a1d)

joe-ayoub-segment:
Hi @toddtarsi thanks for the PR. 

Is this destination in use by customers yet?
If so, then changing the name of the fields first_name -> agent_first_name, and last_name -> agent_last_name is essentially doing a delete of first_name and last_name fields, and an addition of new fields for agent_first_name and agent_last_name. 

If any customers are already using the first_name and last_name fields, the data from these fields will not be collected any more. 

I would advise leaving the names of the fields as they were, and simply change the field labels and descriptions - doing this is not a breaking change. 

Please let me know how you'd like to proceed.

toddtarsi:
@joe-ayoub-segment - That sounds good, will make that tweak instead. Thank you

joe-ayoub-segment:
> @joe-ayoub-segment - That sounds good, will make that tweak instead. Thank you

Thanks @toddtarsi . 

You also need to revert the name for the new agent_last_name field. 
Then make sure you to run `yarn types` and commit the changes.

toddtarsi:
@joe-ayoub-segment - Updated the fields to support the current customer and gives a path forward for better field naming going forward. Also, regenerated types

toddtarsi:
@joe-ayoub-segment - Thank you for the review!

joe-ayoub-segment:
Hi @toddtarsi , 

Apologies but I think you misunderstood me. 

there should be no need to mark some of the fields as legacy. instead, you could keep the fields named as they are, but give them different labels and descriptions. then when you are building the JSON to send to your platform you compose the JSON in the new shape you need it in. 

e.g. 

Keep the field named `first_name` then when building your JSON do something like this (but more defensively): 
```
{
  customer_first_name: first_name
}
```

Also, I think you should remove the default, as this field is a simple boolean toggle, so the customer will probably not map payload data to it: 
```
      default: {
        '@path': '$.properties.assign_first_last_name_by_splitting_email'
      }
```
You could set the default to true or false if you like: 

```
default: true
```

then the customer can change if if they want.

joe-ayoub-segment:
Hi @toddtarsi just following up to see if this PR is still something you want to proceed with. If not I'll close it. 
Best regards,
Joe

joe-ayoub-segment:
Hi @toddtarsi closing this PR due to inactivity. Feel free to reopen if you'd like to continue.

PR #3873, opened 2026-07-13
chore: localize yarn.lock registry URLs on checkout/merge

itsarijitray:
## Summary
- Adds the inverse of the existing pre-commit hook (`scripts/normalize-yarn-lock-registry.sh`, from #3872): after `git pull`/`git merge`/`git checkout`, rewrite `yarn.lock`'s committed `registry.yarnpkg.com` URLs to the local npm registry from `$npm_config_registry` (e.g. an internal Artifactory proxy), if one is configured.
- Why: yarn classic fetches from the exact `resolved` URL already recorded in the lockfile — it doesn't re-derive that from the currently configured registry for entries that already exist in `yarn.lock`. On a machine where the public registry isn't directly reachable (traffic only allowed through an internal Artifactory proxy), `yarn install` fails after pulling a lockfile that points at `registry.yarnpkg.com`, until it's rewritten locally.
- This is a local, uncommitted rewrite only. The pre-commit hook still normalizes `yarn.lock` back to `registry.yarnpkg.com` before every commit, so the two round-trip cleanly and the committed lockfile always stays on the public registry — only your working tree looks different, and only if `$npm_config_registry` is set.
- Wired into `.husky/post-checkout` and `.husky/post-merge` (covers `git pull`, which is a merge under the hood, and branch/file checkouts).

## Test plan
- [x] Manually ran `scripts/localize-yarn-lock-registry.sh` against a public-registry `yarn.lock` — rewrites all entries to the Artifactory URL from `$npm_config_registry`
- [x] Ran `scripts/normalize-yarn-lock-registry.sh` afterward — round-trips back to an exact byte-for-byte match of the original public-registry `yarn.lock`
- [x] Verified end-to-end via a real `git checkout` — the `post-checkout` hook fired automatically and localized `yarn.lock` without manual invocation
- [x] Confirmed the hook is a no-op (exits immediately) when `$npm_config_registry` is unset

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,8 @@
 **/*/generated-types.ts linguist-generated=true
+
+# Keep the committed yarn.lock pointed at the public registry.yarnpkg.com while
+# letting Twilio/Artifactory machines carry proxied URLs in the working tree.
+# The filter is registered per-clone by manually running
+# `sh scripts/setup-yarnlock-filter.sh` once (before the first `yarn install`)
+# and is a no-op where it isn't registered (external contributors, CI).
+yarn.lock filter=yarnlock-registry
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -1,5 +1,0 @@
-#!/usr/bin/env bash
-
-. "$(dirname "$0")/_/husky.sh"
-
-bash scripts/localize-yarn-lock-registry.sh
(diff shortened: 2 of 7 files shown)


sayan-das-in (approved):

joe-ayoub-segment:
This PR has been deployed.

PR #3912, opened 2026-07-30
fix(security): reject URL injection in Qualtrics datacenter and Piped…

AnkitSegment:

Validates that user-supplied datacenter (Qualtrics) and domain (Pipedrive) fields contain only alphanumeric characters and hyphens before interpolating them into API base URLs. Without this check, an attacker-controlled value could redirect requests — including bearer tokens and API keys — to an arbitrary host. Fixes SECOPS-25215 (Qualtrics) and SECOPS-25240 (Pipedrive).

## Testing

## Security Review

## New Destination Checklist


packages/destination-actions/src/destinations/pipedrive/pipedriveApi/pipedrive-client.ts
```diff
@@ -52,6 +53,7 @@ class PipedriveClient {
   private _request: RequestClient
 
   constructor(settings: Settings, request: RequestClient) {
+    validateDomain(settings.domain)
```
mdkhan-tw: do we need to perform this check multiple times? will it not suffice to do only in `testAuthentication`?

packages/destination-actions/src/destinations/pipedrive/utils.ts
```diff
@@ -1,5 +1,15 @@
...
+
 type PayloadWithCustomFields = { custom_fields?: { [k: string]: unknown } }
 
+export function validateDomain(domain: string): void {
+  if (!/^[a-zA-Z0-9-]+$/.test(domain)) {
+    throw new InvalidAuthenticationError(
+      'Invalid domain. Domain must contain only alphanumeric characters and hyphens.'
+    )
+  }
+}
```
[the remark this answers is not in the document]
mdkhan-tw: This is a valid comment, please check

packages/destination-actions/src/destinations/pipedrive/pipedriveApi/pipedrive-client.ts
AnkitSegment: Good catch — this was redundant. Removed the `validateDomain` call from the `PipedriveClient` constructor; validation now happens once in `testAuthentication`, which is the single point where `settings.domain` is verified. See b0a6fbcca.

packages/destination-actions/src/destinations/pipedrive/utils.ts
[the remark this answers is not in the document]
AnkitSegment: Agreed, and applied the same reasoning to Qualtrics for consistency: `validateDatacenter`/`validateDomain` now live only in `testAuthentication` rather than in the client constructors (which ran on every perform). Kept the regex as alphanumeric+hyphen for now rather than adding full DNS-label rules (leading/trailing hyphen, length <= 63) — happy to tighten that in a follow-up if you think it's worth doing now.

packages/destination-actions/src/destinations/qualtrics/utils.ts
```diff
@@ -7,11 +17,11 @@ export function parsedEmbeddedData(
...
     if (typeof data[key] === 'string') {
-      parsedData[key] = data[key] as string
+      parsedData[key] = data[key]
     } else if (typeof data[key] === 'number') {
-      parsedData[key] = data[key] as number
+      parsedData[key] = data[key]
     } else if (typeof data[key] === 'boolean') {
-      parsedData[key] = data[key] as boolean
+      parsedData[key] = data[key]
     } else {
```
mdkhan-tw, line 25: Why are we making this change?

AnkitSegment: Not an intentional change — this repo's lint-staged pre-commit hook runs `eslint --fix` over the whole file whenever any part of it is staged, and the project's own `@typescript-eslint/no-unnecessary-type-assertion` rule flags these three casts as redundant (the preceding `typeof` checks already narrow the type). Since `utils.ts` got touched for `validateDatacenter`, the hook auto-stripped them. No behavior change — tried reverting it manually but the hook just re-strips it on every commit that touches this file, so leaving it as-is.

PR #3915, opened 2026-08-03
SECOPS-25213 [Salesforce Marketing Cloud]: reject subdomain host/path injection

harsh-joshi99:
## Summary

Fixes **SECOPS-25213** (HackerOne) — the Salesforce Marketing Cloud (Actions) destination accepted a `subdomain` setting that could inject a host/path and forward SFMC credentials to an attacker-controlled server.

The customer-supplied `subdomain` setting is interpolated directly into the **host** portion of every request URL, e.g.:

```ts
`https://${settings.subdomain}.auth.marketingcloudapis.com/v2/token`
```

A value such as `attacker.example/` or `mc123@attacker.example` rewrites the effective host. Because the token-refresh request carries `client_id`/`client_secret` and every other call carries the bearer token, those secrets could be sent to a host the customer controls but Segment does not.

## Fix

- Added a central `validateSubdomain()` helper that enforces a single DNS label (`/^[a-zA-Z0-9-]+$/`) and throws a **non-retryable** `PayloadValidationError` for anything else (`/`, `@`, `:`, `.`, whitespace, empty, etc.).
- Enforced it **at settings-save time** via `testAuthentication`. An invalid or injection subdomain is rejected when the customer connects/saves settings, so it can never be persisted — the customer gets a clear, immediate error instead of a silent failure later.
- We deliberately do **not** re-validate on every event. Validation at save-time means a malicious value is never stored in the first place, and skipping the per-request check avoids any risk of breaking delivery for a pre-existing configuration.

## Backward compatibility

No previously-working configuration is affected. Validation runs only at settings-save, and existing stored settings were analysed against the regex to confirm none of the currently-working ones are rejected (see the pre-existing-domains results linked below). Because the check is not on the event path, no in-flight delivery can start failing as a result of this change.

## Testing

[Stage Testing Doc](https://docs.google.com/document/d/1_JDxW7zkB7ww0Dz8MACiPnN1Uc8Kns50N_-9V0rFQHI/edit?usp=sharing)

[Results of testing with Pre-existing Domains](https://docs.google.com/document/d/1R8cY2TIo5xAmvD8wVW0MaU_zcExh-R0gx1f1NesXb40/edit?usp=sharing)

- `_tests_/subdomainValidation.test.ts`:
  - `validateSubdomain()` unit tests — valid subdomains accepted; 10 injection vectors + non-string values rejected.
  - `testAuthentication` tests — a valid subdomain saves cleanly; each injection vector is rejected at settings-save time with the subdomain error message.
- Full SFMC suite passes (no regressions).
- ESLint clean.

### DB analysis of pre-existing subdomains

Ran the shipped `validateSubdomain()` over all stored SFMC subdomain settings. No currently-working configuration is rejected: every rejected value is empty, an already-non-functional full-URL paste, or an active host-injection/SSRF attempt. Details in the pre-existing-domains results doc above.


packages/destination-actions/src/destinations/salesforce-marketing-cloud/sfmc-operations.ts
```diff
@@ -22,6 +23,24 @@ import {
...
+const SUBDOMAIN_PATTERN = /^[a-zA-Z0-9-]+$/
+
+export function validateSubdomain(subdomain: unknown): string {
+  if (typeof subdomain !== 'string' || !SUBDOMAIN_PATTERN.test(subdomain)) {
+    throw new PayloadValidationError(
+      'Invalid Salesforce Marketing Cloud subdomain. The subdomain may only contain letters, numbers and hyphens, and must not include the ".rest.marketingcloudapis.com" part of your subdomain URL.'
+    )
+  }
+  return subdomain
+}
```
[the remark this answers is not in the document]
harsh-joshi99: Deliberately keeping the regex as-is rather than tightening to strict DNS-label rules (length ≤ 63, must start/end alphanumeric), for backward-compatibility reasons.

This is a live, high-volume destination. The security objective — blocking host/path injection so credentials can't be redirected — is fully met by rejecting `/`, `@`, `:`, `.`, and whitespace. Adding stricter structural rules would risk rejecting an unusual-but-currently-working tenant subdomain, i.e. breaking an existing customer to guard against a "confusing downstream DNS failure" that isn't a security issue (a malformed-but-safe subdomain already just fails the request). Given the guardrail here, the conservative choice is to reject only what could be an injection, not to impose full DNS-label validation.

I've updated the code comment to say the pattern restricts to DNS-label *characters* (rather than claiming it enforces a full single-label), so the doc matches the behavior.

packages/destination-actions/src/destinations/salesforce-marketing-cloud/apiEvent/index.ts
```diff
@@ -13,6 +14,7 @@ const action: ActionDefinition<Settings, Payload> = {
     data: eventData
   },
   perform: (request, { settings, payload }) => {
+    validateSubdomain(settings.subdomain)
```
mdkhan-tw: Can we validate it when we are saving the settings in `testAuthentication` so that customer receives the error early?

mdkhan-tw: Applicable for all files

packages/destination-actions/src/destinations/salesforce-marketing-cloud/sfmc-operations.ts
```diff
@@ -24,6 +25,26 @@ import {
...
+// A Salesforce Marketing Cloud subdomain is a single DNS label (a tenant-specific
+// string such as "mc563885gzs27c5t9-63k636ttgm"). It is interpolated directly into
+// the host portion of every request URL, so we restrict it to the characters valid
+// in a DNS label - letters, digits and hyphens. This intentionally does not enforce
+// full DNS-label structure (length, no leading/trailing hyphen); the goal is to block
+// host/path injection, not to reject unusual-but-working subdomains. Allowing other
+// characters (e.g. "/", "@", ":", ".") would let a malicious subdomain rewrite the
+// request host and exfiltrate the OAuth client secret / access token to an
+// attacker-controlled server. See SECOPS-25213.
+const SUBDOMAIN_PATTERN = /^[a-zA-Z0-9-]+$/
```
mdkhan-tw, line 41: can we do analysis with the settings stored in the db to ensure that this regex agrees. Otherwise, customer will complain failure while updating existing settings.

harsh-joshi99: Updated, added test results in PR.

packages/destination-actions/src/destinations/salesforce-marketing-cloud/apiEvent/index.ts
harsh-joshi99: Updated, added test in stage testing doc.

mdkhan-tw (approved):

mdkhan-tw:
Approved, could you please add the analysis that existing settings agrees with this regex?

AnkitSegment (approved):

PR #3919, opened 2026-08-04
[Amplitude Cohorts] Encode owner email + add audience name prefix setting

joe-ayoub-segment:
## Summary

Two changes to the Amplitude Cohorts destination. 

This is a new destination, not in use by any customers. And is intended to be internal only. 

### 1. URL-encode owner email in auth test
`owner_email` (from the `default_owner_email` setting) was interpolated directly into the User Search query string in `testAuthentication` without URL-encoding. Emails containing characters like `+` (common for aliases, e.g. `tools+amplitude@segment.com`) were misinterpreted in the URL — the `+` decodes to a space server-side — causing authentication tests to fail unexpectedly.

- Trim `default_owner_email` and throw a clear `MISSING_REQUIRED_FIELD` error if it is empty after trimming.
- Wrap the value in `encodeURIComponent` before interpolating it into the query string.

### 2. New `prefix_audience_names` setting
A new global boolean setting that prefixes all audience (cohort) names sent to Amplitude with `[Segment] `. Defaults to on; customers can turn it off. The prefix is applied only when the setting is explicitly `true`.

## Testing

- Unit test: plain email (`testing@test.com`) is not altered beyond the expected `@` → `%40`.
- Unit test: plus-alias email (`owner+alias@example.com`) is encoded as `user=owner%2Balias%40example.com`.
- Unit test: prefix is omitted when `prefix_audience_names` is `false`.
- Unit test: prefix is applied when `prefix_audience_names` is explicitly `true`.
- Full `amplitude-cohorts` suite passes (79 tests).
- Verified against the **live Amplitude API**: `testAuthentication` succeeds with the real `+`-alias owner email that previously failed; a real cohort was created and read back from Amplitude with the name `[Segment] ...`, confirming the prefix reaches Amplitude.

### DIFF
--- a/packages/destination-actions/src/destinations/amplitude-cohorts/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/amplitude-cohorts/__tests__/index.test.ts
@@ -26,6 +26,28 @@ describe('Amplitude Cohorts', () => {
 
       await expect(testDestination.testAuthentication(settings)).resolves.not.toThrowError()
     })
+
+    it('should not alter a plain email address in the query string', async () => {
+      const plainSettings = { ...settings, default_owner_email: 'testing@test.com' }
+
+      // Match the raw query string: '@' encodes to %40, and nothing else should change.
+      const scope = nock('https://amplitude.com').get('/api/2/usersearch?user=testing%40test.com').reply(200, {})
+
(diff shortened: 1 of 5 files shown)


sayan-das-in (approved):
Approving as this is still in Private Beta.

joe-ayoub-segment:
PR deployed

PR #3968, opened 2026-08-20
STRATCONN-6834 - [Mixpanel] - correcting a bug with token fallback

joe-ayoub-segment:
Fixes a bug with Mixpanel where lack of a apiSecret is not falling back properly to project token. 
This affects 0 customers as all customers have apiSecret values, but it will become a problem as we start to migrate customers over to using only project tokens. 

## Testing

Tested locally with Mixpanel team on a video call. 
- `settings.apiSecret ?? settings.projectToken` failed when apiSecret was not present but projectToken was present. 
- `settings.apiSecret || settings.projectToken` passed when apiSecret was not present but projectToken was present.  

## Security Review

## New Destination Checklist


itsarijitray (approved):

joe-ayoub-segment:
PR deployed last week.

PR #3974, opened 2026-08-24
Update merchant authentication to handle multiple merchants on one Segment instance

jramorris:

Previously a single merchant authentication was cached, causing the same token to be used for multiple merchants when running on the same instance. This update sets a cached token per merchant.

## Testing

## Security Review

## New Destination Checklist

leaving this unchecked as this is not a new destination

### DIFF
--- a/packages/destination-actions/src/destinations/friendbuy/__tests__/cloudUtil.test.ts
+++ b/packages/destination-actions/src/destinations/friendbuy/__tests__/cloudUtil.test.ts
@@ -1,9 +1,16 @@
 import nock from 'nock'
 import { createTestEvent, createTestIntegration } from '@segment/actions-core'
-import type { DecoratedResponse } from '@segment/actions-core'
+import type { DecoratedResponse, RequestClient } from '@segment/actions-core'
 import Destination from '../index'
 
-import { defaultMapiBaseUrl, resetAuthCache } from '../cloudUtil'
+import {
+  AUTH_CACHE_MAX_ENTRIES,
+  AUTH_CACHE_MAX_ENTRY_LENGTH,
+  authCacheSize,
+  defaultMapiBaseUrl,
+  getAuthToken,
+  resetAuthCache
+} from '../cloudUtil'
 
 const testDestination = createTestIntegration(Destination)
 
@@ -65,7 +72,7 @@ function trackPurchase(
     event: purchaseEvent(orderId),
     settings,
     useDefaultMappings: true
-  }) as Promise<DecoratedResponse[]>
+  })
 }
 
 describe('Friendbuy cloud auth token cache', () => {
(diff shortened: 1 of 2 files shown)


joe-ayoub-segment:
Hi @jramorris thanks for raising this PR to plug this gap. 

I had a quick look and noted the following: 

1. **Eviction is FIFO-by-first-insertion, not LRU** (cloudUtil.ts:65-72). Map.set() on an existing key doesn't move it in iteration order. Only brand-new keys append to the end. So once the cache is at capacity, the eviction loop always kills whichever merchant was first ever cached, even if that merchant is your busiest/most-active one, while a merchant cached more recently but never used again sits safe. 

Under sustained load with >1000 distinct active credentials per worker, this thrashes the hottest merchant's token instead of the coldest. Worth switching to true LRU (delete+re-set the key on cache hit) or just accepting this as a known limitation.

IMO - I'm OK with this as long as you are doing this deliberately. 

2. **Cache key omits authSecret** (authCacheKey, line 54-56), deliberate per the test at cloudUtil.test.ts:103-115. Rotating a merchant's secret does not invalidate their cached token; the old token keeps serving until its own TTL lapses. Fine if MAPI's issued tokens stay valid independent of secret rotation, but if rotation is meant to immediately revoke access, this is a gap. Worth confirming intent with the author rather than assuming.

Again, I'm OK with this if it's deliberate on your part. 

3. **No cache-bust on auth failure**. If a cached token gets rejected server-side before its recorded expiry (e.g. secret revoked upstream), there's no 401 handling in the action performs (checked trackPurchase/index.ts) to force re-auth; it'll just keep failing until natural expiry. Pre-existing behavior, not introduced by this PR, but still no better after it.

Because this is pre-existing it's probably OK. I'll leave it to you to decide. 

4. **No test covers the 1000-entry cap or the eviction path itself.**  Only correctness of keying, not the size-bound mechanism the caching policy exception hinges on.

Can you please add a unit test which verifies that there can only ever be a max of 1000 items in the cache please? Also, please add a check to enforce the max character length of the items which can be added to the cache, say 256 chars, or something like that. 

One important note, Segment's architecture is stateless and horizontally scales with load across thousands of nodes. Traffic is not sticky and will be routed to different nodes at random. Because of this it's unlikely that you'll get many cache hits. I just want to make sure you understand this as it may be important for you. 

Best regards,
Joe

joe-ayoub-segment (dismissed):
see comment.

jramorris:
Hi @joe-ayoub-segment, thanks the review.

1. Good callout. I'm okay keeping this as a known limitation. The auth call is low cost. Re-caching in this case should be fine.
2. You're right. That's deliberate. Thanks for mentioning. 
3. Another good callout. Like you mention, this is pre-existing. I'm in favor of leaving it in order to plug this production gap as soon as possible.
4. Tests added.

On your note, that's okay. Inevitably we'll have to make the auth call at least once per merchant per instance.

Cheers,
James

joe-ayoub-segment:
Thanks @jramorris, I forgot to ask for some proof of testing. 
Did you test this using the local testing tool? 

You will need to be signed into Segment to use it. 
Then run the following to start it up. 

`./bin/run serve`

then select the destination, add credentials, send in a test payload and add a couple of screenshots to this PR. 

I'm going on leave so will ask another member of the team to complete the PR review.

jramorris:
Hi @joe-ayoub-segment, I did test this locally but don't have screenshots. If you have a moment, is there anything specific I can be sure to include?

joe-ayoub-segment:
If you can run it again, just show the input JSON and a successful 200 response - this demonstrates the action is sending data to your platform end to end. Any other evidence would be useful.

jramorris:
Input json:
inputJSON

200 response:
screenshot

monutwilio (approved):

joe-ayoub-segment:
PR deployed last week.

PR #3975, opened 2026-08-24
LinkedIn Conversions: clarify the conversion rule hook label and unlinked-rule error

harsh-joshi99:
Improves two user-facing strings in the LinkedIn Conversions API destination. No behaviour changes.

### Background

The Stream Conversion Event action reads the conversion rule ID only from the persisted `onMappingSave` hook output. Entering a value in the "Existing Conversion Rule ID" hook input is not sufficient on its own — the hook has to run for that value to be stored where delivery reads it. A mapping can therefore be saved with the field visibly populated while no rule is actually linked, and every event then fails with `400: Conversion Rule ID is required`. That message contradicts what the user is looking at and offers no way forward.

Reproducible on `main`: enter an existing Conversion Rule ID, press Save without pressing the hook button, send a test event.

### Changes

**1. The error message.** Replaced `Conversion Rule ID is required.` with one that states no rule is linked, names the button that fixes it, and notes that supplying an existing ID links that rule rather than duplicating it. Extracted to a shared constant so the `perform` and `performBatch` sites cannot drift.

**2. The hook label and description.** The hook's `label` is rendered as the button text in the mapping editor, so `Create a Conversion Rule` was the only thing users saw. It names one of the operations the hook performs, and not the common one: supplying an existing Conversion Rule ID short-circuits to a `GET`, so the rule is looked up and linked, never created. Users understandably decline to press "Create" for a rule they already have. Renamed to `Link or Create a Conversion Rule`, consistent with the existing `Create or Select Data Extension` precedent for Salesforce Marketing Cloud.

The description opened with "When saving this mapping, we will create a conversion rule", which is not what happens and contradicted its own second sentence. Replaced with an accurate summary.

Deliberately **not** claimed: that the existing rule is left untouched. `bulkAssociateCampaignToConversion` runs on every successful path including the lookup path, so campaigns selected in the card are associated to the rule.

### Also included

Two redundant `DynamicFieldResponse` casts and the now-unused import were removed from the test file. These are auto-removed by the `eslint --fix` pass in `lint-staged`, which otherwise fails the pre-commit hook for anyone editing this file.

### Not addressed here

The underlying cause is app-side: the mapping editor saves a hook-using mapping with no hook outputs and reports success, and its hardcoded copy tells users to click Save to run the hook. Being raised separately, as it affects every destination using `onMappingSave`.

## Testing

- [x] [If destination is already live] Tested for backward compatibility of destination.

Two tests added covering the missing-hook-output case and the inputs-saved-without-outputs case. All 59 tests across the four `linkedin-conversions` suites pass.

Note: the local server (`./bin/run serve`) cannot exercise this path — the CLI dev server returns hook results over HTTP and discards them, so hook output persistence is not reproducible locally.

Backward compatibility: only string literals and one display label change. No field definitions, no request shapes, no new required fields. The hook key remains `onMappingSave`, so existing mappings are unaffected.

## Security Review

- [x] **Reviewed all field definitions** for sensitive data and confirmed they use `type: 'password'` — no field definitions were changed by this PR.


harsh-joshi99:
Closing — pushed from the wrong HEAD, so this diff is staging's divergence from main rather than the intended two-commit change. Reopening a correct PR shortly.

PR #3994, opened 2026-09-02
STRATCONN-6988: fix filename extension corruption in syncToS3 uploadS3

mdkhan-tw:
### Problem
In `syncToS3/client.ts`, when a `filename_prefix` already ended with the file extension, the code inserted the date suffix via:

```js
filename_prefix.replace(fileExtension, `_${dateSuffix}.${fileExtension}`)
```

`String.prototype.replace` with a **string** argument replaces the **first** occurrence of `fileExtension` **anywhere** in the name — not the trailing extension. So a prefix like `csv_export.csv` (extension `csv`) had its **leading** `csv` replaced, producing a corrupted key:

```
_<date>.csv_export.csv        ← actual (wrong)
csv_export_<date>.csv         ← expected
```

The old code also hard-coded `.csv`/`.txt` in the guard instead of using the configured `fileExtension`.

### Fix
Extracted the logic into a pure, exported `buildTimestampedFilename()` helper that strips the trailing `.<ext>` **by length** and re-appends `_<date>.<ext>`. Behavior for prefixes without the extension is unchanged.

### Tests
Added a `buildTimestampedFilename` unit suite, including a regression case for a base name that contains the extension string (`csv_export.csv`). Full `s3` suite passes (52 tests).

image
image

### Validation
Human-validated against a real S3 upload (stage credentials): the object key was corrupted before the fix and correct after.

### DIFF
--- a/packages/destination-actions/src/destinations/s3/__tests__/client.test.ts
+++ b/packages/destination-actions/src/destinations/s3/__tests__/client.test.ts
@@ -1,4 +1,4 @@
-import { isAWSError } from '../syncToS3/client'
+import { isAWSError, buildTimestampedFilename } from '../syncToS3/client'
 import { _Error as AWSError } from '@aws-sdk/client-s3'
 
 // Mock AWS SDK before any imports to avoid initialization issues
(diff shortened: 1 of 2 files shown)


mdkhan-tw:
### Validation evidence — real S3 upload (stage mode)

Uploaded a real event via `syncToS3` under `aws-okta exec stage-write` (two-hop assume-role succeeds), with `filename_prefix="csv_export-6988-<tag>.csv"`, `file_extension="csv"`, folder `defect2pr-validation/`. Upload returned **200**.

**S3 object listing (current `main`, before this fix) — the object key is corrupted:**

```
defect2pr-validation/
  _2026-09-02T11-23-46-283Z.csv_export-6988-2026-09-02T11-23-42-574Z.csv   ·   77.0 B   ·   csv
```

- **Expected key:** `csv_export-6988-<tag>_<date>.csv`
- **Actual key (bug):** `_<date>.csv_export-6988-<tag>.csv`

`filename_prefix.replace(fileExtension, …)` replaced the **leading** `csv` in `csv_export` (first occurrence of the string `"csv"`), so `.csv` lands mid-name and the key starts with `_<date>`. Confirmed visually in the S3 console object listing.

After this fix (`buildTimestampedFilename`) the key is `csv_export-6988-<tag>_<date>.csv` — covered by the new unit tests.

itsarijitray (approved):

PR #3995, opened 2026-09-02
STRATCONN-6989: fix CSV header desync (clean all delimiters + encode headers)

mdkhan-tw:
### Problem
Two issues in `syncToS3/functions.ts` combined to desync the CSV header row from the data rows:

1. **`clean()` removed only the first delimiter.** `str.replace(delimiter, '')` (string arg) replaces a single occurrence, so a column name with multiple delimiters (e.g. `tag,list,csv`) kept residual delimiters.
2. **Header cells skipped `encodeString`.** Data cells go through `processField → encodeString` (quoted), but the header row was written raw. So a residual delimiter in a header name split the header into extra columns, while quoted data rows kept their column count.

Net effect: header column count ≠ data column count → every column mapped wrong by downstream loaders.

```
event_name,taglist,csv,amount,audience_action,batch_size   ← 6 columns (raw, split)
"Purchase","u-1","100","","2"                              ← 5 columns (quoted)
```

### Fix
- `clean()` now removes **every** delimiter (`split(delimiter).join('')`; literal, so it's safe for `|`, `;`, `:` too).
- Header cells are now passed through `encodeString`, exactly like data cells.

### Tests
- `clean` regression: removes all delimiters, not just the first.
- `generateFile` regression: header column count == data column count when a column name contains the delimiter; header cells quoted like data; embedded quote in a column name is escaped.
- Updated existing golden/index-based assertions to the corrected (quoted-header) output. Full `s3` suite passes (50 tests).

image
image
image

### Validation
Human-validated against a real S3 upload (stage): a `tag,list,csv` column produced a 6-vs-5 desynced object before the fix; aligned after.

### DIFF
--- a/packages/destination-actions/src/destinations/s3/syncToS3/__tests__/index.test.ts
+++ b/packages/destination-actions/src/destinations/s3/syncToS3/__tests__/index.test.ts
@@ -40,6 +40,13 @@ describe('clean', () => {
     expect(clean(',', 'abcd,Efg')).toEqual('abcdEfg')
   })
 
+  it('removes EVERY delimiter, not just the first (regression: STRATCONN-6989)', () => {
+    // The old code used str.replace(delimiter, '') which removed only the first match,
+    // leaving residual delimiters that split the (unencoded) header row.
+    expect(clean(',', 'tag,list,csv')).toBe('taglistcsv')
+    expect(clean(';', 'a;b;c;d')).toBe('abcd')
+  })
+
   it('should handle undefined input', () => {
     expect(clean(',', '')).toBe('')
   })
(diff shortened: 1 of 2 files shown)


mdkhan-tw:
### Validation evidence — real S3 uploads (stage mode)

Two real `syncToS3` uploads to `defect2pr-validation/` under stage creds (both returned **200**). Contents read back from the S3 objects:

**Test 1 — plain column names (control). Bug NOT triggered → output correct:**

```
event_name,user_id,city,amount,note,audience_action,batch_size
Purchase,u-6989-1,"Austin, TX",100,"a,b,c",,2
Refund,u-6989-2,NYC,50,x,,2
```

7 headers / 7 data fields, aligned. Comma-containing **values** (`"Austin, TX"`, `"a,b,c"`) are quoted correctly — data-cell encoding was never the bug. (A CSV viewer normalizes quoting on the simple cells; the raw file quotes every data cell.)

**Test 2 — a column NAME contains the delimiter (`tag,list,csv`). Desync reproduced:**

```
event_name,taglist,csv,amount,audience_action,batch_size      <- 6 columns
Purchase,u-1,100,,2                                           <- 5 columns
Refund,u-2,50,,2                                              <- 5 columns
```

- `clean()` stripped only the **first** comma of `tag,list,csv` → `taglist,csv` (a residual comma survives).
- The header row skips `encodeString`, so that residual comma **splits the header into 6 columns** while the quoted data rows have **5** → header/data **desync**; downstream loaders map every column wrong.

After this fix (`clean()` removes all delimiters + headers run through `encodeString`) the header and data column counts match — covered by the new `clean()` / `generateFile` regression tests.

itsarijitray (approved):

PR #3996, opened 2026-09-03
feat(s3): reject S3 object keys over 1024 bytes up front

itsarijitray:
## Summary

S3 object keys have a hard AWS limit of **1024 bytes** (UTF-8). `actions-s3` built the key (`folderName` + `filename_prefix`) with no length check, so an over-long key failed opaquely at PUT time with a raw AWS error — and, because that error code isn't handled specially, it was thrown as a **`RetryableError`**, storming retries.

This adds a guard in `uploadS3()` that measures the key's UTF-8 **byte** length (`Buffer.byteLength(objectKey, 'utf8')`, not `.length`) and throws a non-retryable **`PayloadValidationError`** when it exceeds 1024 bytes. The guard runs **before** `assumeRole()`, so it fails fast without making any AWS/STS call, and the error message omits the key content to avoid leaking PII.

Keys already under the limit are untouched — no behaviour change for working configs.

https://twilio-engineering.atlassian.net/browse/STRATCONN-6990

## Testing

New unit tests in `syncToS3/__tests__/client.test.ts` (mocked STS + S3 client):

- Object key > 1024 bytes built from multi-byte (`€`) chars → `PayloadValidationError`, and asserts **neither STS nor the PUT** is attempted. Uses 400 `€` chars (400 chars / 1200 bytes) to prove the guard measures **bytes, not characters**.
- Over-limit key of secret content → error message does **not** leak the key content (PII-safe).
- Normal short key → proceeds to PUT, returns `{ statusCode: 200 }`.

Full `s3/syncToS3` suite passes (40/40). ESLint clean.

- [x] Tested end-to-end using the local server
- [x] [If destination is already live] Tested for backward compatibility of destination. **Note:** New required fields are a breaking change. — _No new fields; over-limit keys already failed today, just opaquely. No behaviour change for keys under the limit._

### e2e testing
image
image

## Security Review

- [x] **Reviewed all field definitions** for sensitive data — no field changes in this PR. The new error message deliberately excludes the object key content to avoid logging PII.

### DIFF
--- a/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
+++ b/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
@@ -3,8 +3,18 @@ import { STSClient, AssumeRoleCommand } from '@aws-sdk/client-sts'
 import { S3Client, PutObjectCommandInput, PutObjectCommand, _Error as AWSError } from '@aws-sdk/client-s3'
 import { v4 as uuidv4 } from '@lukeed/uuid'
 import * as process from 'process'
-import { ErrorCodes, IntegrationError, RetryableError, APIError, RequestTimeoutError } from '@segment/actions-core'
+import {
+  ErrorCodes,
+  IntegrationError,
+  RetryableError,
+  APIError,
+  RequestTimeoutError,
+  PayloadValidationError
+} from '@segment/actions-core'
 import { Credentials } from './types'
+
+// AWS enforces a hard limit of 1024 bytes (UTF-8) on S3 object keys.
+const MAX_S3_OBJECT_KEY_BYTES = 1024
 
 export class Client {
   roleArn: string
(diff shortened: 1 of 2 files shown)


packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -77,6 +87,21 @@ export class Client {
       : s3_aws_folder_name?.endsWith('/')
       ? s3_aws_folder_name
       : `${s3_aws_folder_name}/`
+    const contentType = fileExtension === 'csv' ? 'text/csv' : 'text/plain'
```
mdkhan-tw, line 90: I am wondering if we can add this validation during mapping saving so that we can give early feedback to customers

https://github.com/segmentio/action-destinations/blob/87ee29f4f0f9ffa15250ef3f0deec7289a66793a/packages/destination-actions/src/destinations/s3/syncToS3/fields.ts#L195-L209

itsarijitray: TL;DR: Save-time validation can't fully replace the runtime guard, so I'd keep this as-is and track save-time UX as a follow-up.

Three blockers:
1. **Cross-field + runtime suffix** — the real key is `folder + filename + _<timestamp>.<ext>`. Field-level `maximum` only caps each field alone; it can't validate the concatenation or the ~28-byte suffix we append at runtime.
2. **Bytes vs chars** — `maximum` is character length; the S3 limit (and our guard) is UTF-8 bytes. E.g. 350 `€` = 350 chars but 1050 bytes, so a char cap would miss it.
3. **Dynamic folder** — `s3_aws_folder_name` allows mapped values (resolved per-event at runtime), so its length is unknown at save time.

So the runtime guard stays the source of truth. Happy to file a follow-up to add a conservative (approximate) `maximum` on the static `filename_prefix` for earlier UI feedback if you think it's worth it.

mdkhan-tw (approved):

joe-ayoub-segment:
PR deployed

PR #4004, opened 2026-09-07
STRATCONN-6987 (pt.2): cache STS assume-role credentials in actions-s3 to cut STS throttling

itsarijitray:
## STRATCONN-6987 (pt.2) — s3 `syncToS3`: cache STS assume-role credentials

> **Stacked on #3993.** Base branch is `t2f/STRATCONN-6987`, so this PR's diff shows only the caching change. Review/merge #3993 first, then this. Once #3993 merges to `main`, GitHub will retarget this PR to `main` automatically.

### 🎫 Ticket
Completes the third scope item of **STRATCONN-6987 — "STS wrap + error classification + credential caching"**. PR #3993 delivered the wrap + error-classification; this PR adds the **credential caching**.

### 🔍 Why
A new `Client` is constructed on **every upload** (`syncToS3/functions.ts:51`), and `assumeRole()` performs a **two-hop STS chain** on each one:

```
assumeRole() -> getSTSCredentials(intermediary role)   // shared across all customers
             -> getSTSCredentials(customer role)         // per (roleArn, externalId, region)
```

So under a high-volume audience sync (many files), STS `AssumeRole` was being called **twice per file**. This is the most likely source of the `rate exceeded` / STS throttling errors observed during NYT load testing — Vara flagged STS (not S3 `PutObject`) as the probable culprit, and STS credential caching was called out as a pre-GA enhancement.

### 🛠 The fix
A **module-level** credential cache (shared across `Client` instances — a per-instance cache would never be reused given the per-upload construction), keyed by `region | roleArn | externalId`:

- Returns cached credentials until **5 minutes before** their STS-reported `Expiration` (safety buffer so nothing expires mid-upload).
- Caches **both hops** — the shared intermediary role (highest value under load, since it's identical for all customers) and the per-customer role.
- **Requires `Expiration`** — it's folded into the existing required-fields validation alongside `AccessKeyId`/`SecretAccessKey`/`SessionToken`, so a response missing it fails fast (403); otherwise credentials are always cached. STS always returns it (confirmed in DataDog).
- No public API or contract change; purely internal.

**Net effect under load:** STS call volume goes from `2 × fileCount` to roughly `2` per credential lifetime (~1 hour default), independent of file count.

### 🧪 Tests
Added a `STS credential caching` suite to `client.test.ts` (extends the mock STS harness introduced in #3993):
- reuses cached credentials across uploads instead of re-calling STS for every file (2 STS calls for 2 uploads, not 4);
- refreshes once credentials fall within the 5-min expiry buffer;
- caches per role identity while sharing the intermediary role;
- fails fast with a 403 when STS returns credentials without an `Expiration` (malformed response).

**Datadog Notebook:** https://segment.datadoghq.com/notebook/15492443/-proj-actions-s3-ga?refresh_mode=paused&tpl_var_env=stage&from_ts=1788952339507&to_ts=1788955939507
image

### ⏳ Cache expiry
Not a fixed TTL — each credential is cached until its **STS-reported `Expiration` minus a 5-minute safety buffer** (`CREDENTIALS_EXPIRY_BUFFER_MS` in `constants.ts`); a cache hit requires `cached.expiration − buffer > now`. We store the real `Expiration` STS returns, not a guessed duration, so the cache adapts automatically if the session length ever changes and can never serve a credential past its real lifetime.

The **5-minute buffer** is the one chosen value:
- **Avoids mid-upload expiry** — any credential handed out has comfortably more life than a single S3 PUT needs.
- **Matches existing precedent** — `lib/AWS/sts.ts` `getAWSCredentialsFromEKS()` caches 55 min against STS's default 1-hour (`DurationSeconds=3600`) token, i.e. the same 5-minute margin. Same as AWS SDK providers, which also refresh a few minutes early.
- **Good ratio** — against a 1-hour session it sacrifices ~8% of the window (~55 min still cached) to eliminate the edge case.

Caveat: 5 min is a conservative default, not empirically tuned. If an upload could ever exceed ~5 min, raise the buffer above the worst-case upload time — it's a single constant. Once the new metrics are live, an unexpectedly high miss rate is the signal to revisit the buffer (or session duration).

### 📊 Observability (DataDog)
Emits three counters from `Client.getSTSCredentials`, each tagged `role_type:intermediary` / `role_type:customer` (plus the caller's `statsContext.tags`):
- `sts_credential_cache_hit` — served from cache
- `sts_credential_cache_miss` — had to call STS `AssumeRole`
- `sts_credential_cache_set` — freshly minted creds written to cache

`statsContext` is threaded from the `perform`/`performBatch` bundle through `send()` into the `Client`; when absent, the `incr` calls are no-ops. Watch the hit ratio `hit / (hit + miss)` during the NYT load-test re-run — it should climb toward ~1.0 as file count grows.

### 🧩 Design note: why an in-memory cache and not `StateContext`

A reasonable question is whether this should use the framework's `StateContext` (the platform-managed, cross-request K/V store with TTL) instead of a module-level in-memory `Map`. We deliberately chose in-memory. Rationale:

- **Security (the deciding factor).** `StateContext` values are persisted by the monoservice and carried in the request/response `context` fields — they can be logged and live wherever subscription state lives. Caching there would mean persisting **live AWS `secretAccessKey` + `sessionToken`** outside process memory. Every current `StateContext` user in this repo stores a *non-secret* scalar (linkedin-audiences → `dmpSegmentId`, twilio-studio → a cooling-off marker, braze-cohorts → `cohort_name`); none stores a credential. In-memory keeps the STS secrets in process memory only and never serializes them anywhere.

- **Consistent with existing AWS precedent.** The two existing STS/credential caches in this repo are both in-memory module-level, not `StateContext`: `lib/AWS/sts.ts` `getAWSCredentialsFromEKS()` (caches assumed creds, ~55 min) and `lib/AWS/s3.ts` `getS3Client()` (caches the S3 client per monoservice instance). This PR mirrors that pattern for the customer assume-role path.

- **Fits the runtime.** The destination runs in the long-lived Integrations Monoservice (K8s pods), so a module-level cache persists across requests per pod. The throttling risk is per-prefix / per-flow, so per-pod caching is sufficient to flatten STS call volume (from `2 × fileCount` to ~2 per role per credential lifetime) and clear the `rate exceeded` errors observed in load testing.

- **`StateContext` is string-only, optional, and out of reach here.** It only exists in the `perform` bundle (our STS call is two layers down in `client.ts`), is not guaranteed present (`stateContext?`), and would require JSON-serializing the credential blob — i.e. secrets-in-a-string in shared state. Its coarse TTL also can't do the precise "STS `Expiration` minus 5-min buffer" freshness check we have here.

If cross-pod credential sharing is ever needed, that's a deliberate platform/security decision — but `StateContext` would still be the wrong home for live credentials, and there's no evidence it's needed for the current scale.

### ✅ Verification
- ✅ `tsc --noEmit` — clean for the changed files
- ✅ `eslint` — clean for the changed files
- ⚠️ **Jest not run locally** — this machine's `node_modules` has a pre-existing jest major-version skew (hoisted `jest-runtime@27.3.1` vs `jest@30`) that fails **all** test suites (including untouched ones like `slack`), and the internal registry is unreachable from here to repair it. The new tests follow the exact mock pattern already in `client.test.ts`; **CI will execute them.**

### ▶️ Follow-up
After this merges, re-run the actions-s3 load test with @Parag to confirm the STS throttling is gone before the NYT GA go-ahead.

---

### DIFF
--- a/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
+++ b/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
@@ -34,7 +34,15 @@ export class Client {
       RoleSessionName: this.roleSessionName,
       ExternalId: externalId
     })
-    const result = await stsClient.send(command)
+    let result
+    try {
+      result = await stsClient.send(command)
+    } catch (err) {
+      // STS failures used to escape uploadS3's try/catch entirely, so they reached the platform
+      // with no status/code, were classified type:internal and force-retried (even permanent auth
+      // failures). Map them to Segment error classes here so classification is correct.
+      throw mapAWSError(err, 'Failed to assume AWS role')
+    }
     if (
       !result.Credentials ||
       !result.Credentials.AccessKeyId ||
@@ -107,22 +115,49 @@ export class Client {
         throw new RequestTimeoutError()
       }
 
-      if (isAWSError(err)) {
-        // https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/_Error/
-        if (err.Code && accessDeniedCodes.has(err.Code)) {
-          throw new APIError(err.Message || err.Code, 403)
-        } else if (err.Code === 'NoSuchBucket') {
-          throw new APIError(err.Message || err.Code, 404)
-        } else if (err.Code === 'SlowDown') {
-          throw new APIError(err.Message || err.Code, 429)
-        } else {
-          throw new RetryableError(err.Message || err.Code || 'Unknown AWS Put error: ' + err)
-        }
-      } else {
-        throw new APIError('Unknown error during AWS PUT: ' + err, 500)
-      }
+      throw mapAWSError(err, 'AWS PUT failed')
     }
   }
+}
+
+/**
+ * Maps an AWS SDK error (S3 `_Error` shape or an STS/service exception) to the appropriate
+ * Segment error class. Permanent, client-side failures (access denied, invalid config, expired
+ * credentials, missing bucket) are surfaced as non-retryable errors; only transient/server-side
+ * or throttling failures are marked retryable. This prevents the platform from force-retrying
+ * errors that will never succeed.
+ */
+export function mapAWSError(err: unknown, context: string): Error {
+  const e = err as {
+    Code?: string
+    Message?: string
+    name?: string
+    message?: string
+    $fault?: 'client' | 'server'
+    $metadata?: { httpStatusCode?: number }
+  }
+  // S3 `_Error` uses Code/Message; STS/service exceptions use name/message.
+  const code = e?.Code ?? e?.name
+  const message = e?.Message ?? e?.message ?? code ?? String(err)
+  const httpStatus = e?.$metadata?.httpStatusCode
+  const detail = `${context}: ${message}`
+
+  if (code && accessDeniedCodes.has(code)) {
+    // Permanent authentication/authorization failure. Not retryable.
+    return new APIError(detail, 403)
+  }
+  if (code === 'NoSuchBucket') {
+    return new APIError(detail, 404)
+  }
+  if (code && throttlingCodes.has(code)) {
+    return new APIError(detail, 429)
+  }
+  // A client fault (4xx that is not throttling) is permanent - do not retry.
+  if (e?.$fault === 'client' || (typeof httpStatus === 'number' && httpStatus >= 400 && httpStatus < 500)) {
+    return new IntegrationError(detail, ErrorCodes.INVALID_AUTHENTICATION, httpStatus ?? 400)
+  }
+  // Transient / server-side / unclassified failures are safe to retry.
+  return new RetryableError(detail)
 }
 
 const accessDeniedCodes = new Set([
@@ -134,8 +169,14 @@ const accessDeniedCodes = new Set([
   'NotSignedUp',
   'AmbiguousGrantByEmailAddress',
   'AuthorizationHeaderMalformed',
-  'RequestExpired'
+  'RequestExpired',
+  // STS assume-role authorization/credential failures
+  'ExpiredToken',
+  'ExpiredTokenException',
+  'AccessDeniedException'
 ])
+
+const throttlingCodes = new Set(['SlowDown', 'Throttling', 'ThrottlingException', 'TooManyRequestsException'])
 
 // isAWSError validates that the error is an generic AWS error
 export function isAWSError(err: unknown): err is AWSError {
--- a/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
+++ b/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
@@ -106,6 +106,11 @@ export class Client {
     if (expiration instanceof Date) {
       credentialsCache.set(cacheKey, { credentials: creds, expiration: expiration.getTime() })
       statsClient?.incr('sts_credential_cache_set', 1, tags)
+    } else {
+      // STS should always return an Expiration; if it ever doesn't we can't safely cache (unknown
+      // lifetime), so we fall back to fetching every time. Emit a metric so this is observable in
+      // DataDog rather than surfacing only as a mysteriously high miss rate.
+      statsClient?.incr('sts_credential_no_expiration', 1, tags)
     }
 
     return creds
--- a/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
+++ b/packages/destination-actions/src/destinations/s3/syncToS3/client.ts
@@ -4,7 +4,8 @@ import { S3Client, PutObjectCommandInput, PutObjectCommand, _Error as AWSError }
 import { v4 as uuidv4 } from '@lukeed/uuid'
 import * as process from 'process'
 import { ErrorCodes, IntegrationError, RetryableError, APIError, RequestTimeoutError } from '@segment/actions-core'
-import { Credentials } from './types'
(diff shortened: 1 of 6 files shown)


packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -6,6 +6,34 @@ import * as process from 'process'
...
+ * A new Client is constructed on every upload (see syncToS3/functions.ts), so a per-instance
+ * cache would never be reused. Under high-volume audience syncs the two-hop assume-role chain
+ * (intermediary role -> customer role) re-ran STS on every file, which is the most likely source
+ * of the `rate exceeded` / STS throttling errors seen during load testing. Caching the minted
+ * credentials until just before their STS-reported expiry keeps STS call volume flat as the
+ * number of files grows.
+ *
+ * Keyed by region + role ARN + external id, the inputs that determine the returned credentials.
+ */
+interface CachedCredentials {
```
joe-ayoub-segment: can we add these interfaces and types into the types.ts file?

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -6,6 +6,34 @@ import * as process from 'process'
...
+  credentials: Credentials
+  expiration: number // epoch millis, from STS Credentials.Expiration
+}
+
+const credentialsCache = new Map<string, CachedCredentials>()
+
+// Refresh this long before the STS-reported expiration so we never hand out credentials that
+// would expire mid-upload.
+const CREDENTIALS_EXPIRY_BUFFER_MS = 5 * 60 * 1000
```
joe-ayoub-segment: can we add constants into the constants.ts file?

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
itsarijitray: Done — moved the `CachedCredentials` interface into `types.ts` in 11627ca3ad.

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
itsarijitray: Done — moved `CREDENTIALS_EXPIRY_BUFFER_MS` into `constants.ts` in 11627ca3ad.

itsarijitray:
### Design note: why an in-memory cache and not `StateContext`

A reasonable question is whether this should use the framework's `StateContext` (the platform-managed, cross-request K/V store with TTL) instead of a module-level in-memory `Map`. We deliberately chose in-memory. Rationale:

- **Security (the deciding factor).** `StateContext` values are persisted by the monoservice and carried in the request/response `context` fields — they can be logged and live wherever subscription state lives. Caching there would mean persisting **live AWS `secretAccessKey` + `sessionToken`** outside process memory. Every current `StateContext` user in this repo stores a *non-secret* scalar (linkedin-audiences → `dmpSegmentId`, twilio-studio → a cooling-off marker, braze-cohorts → `cohort_name`); none stores a credential. In-memory keeps the STS secrets in process memory only and never serializes them anywhere.

- **Consistent with existing AWS precedent.** The two existing STS/credential caches in this repo are both in-memory module-level, not `StateContext`: `lib/AWS/sts.ts` `getAWSCredentialsFromEKS()` (caches assumed creds, ~55 min) and `lib/AWS/s3.ts` `getS3Client()` (caches the S3 client per monoservice instance). This PR mirrors that pattern for the customer assume-role path.

- **Fits the runtime.** The destination runs in the long-lived Integrations Monoservice (K8s pods), so a module-level cache persists across requests per pod. The throttling risk is per-prefix / per-flow, so per-pod caching is sufficient to flatten STS call volume (from `2 × fileCount` to ~2 per role per credential lifetime) and clear the `rate exceeded` errors observed in load testing.

- **`StateContext` is string-only, optional, and out of reach here.** It only exists in the `perform` bundle (our STS call is two layers down in `client.ts`), is not guaranteed present (`stateContext?`), and would require JSON-serializing the credential blob — i.e. secrets-in-a-string in shared state. Its coarse TTL also can't do the precise "STS `Expiration` minus 5-min buffer" freshness check we have here.

If cross-pod credential sharing is ever needed, that's a deliberate platform/security decision — but `StateContext` would still be the wrong home for live credentials, and there's no evidence it's needed for the current scale.

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -4,29 +4,70 @@ import { S3Client, PutObjectCommandInput, PutObjectCommand, _Error as AWSError }
...
+ * A new Client is constructed on every upload (see syncToS3/functions.ts), so a per-instance
+ * cache would never be reused. Under high-volume audience syncs the two-hop assume-role chain
+ * (intermediary role -> customer role) re-ran STS on every file, which is the most likely source
+ * of the `rate exceeded` / STS throttling errors seen during load testing. Caching the minted
+ * credentials until just before their STS-reported expiry keeps STS call volume flat as the
+ * number of files grows.
+ *
+ * Keyed by region + role ARN + external id, the inputs that determine the returned credentials.
+ */
+const credentialsCache = new Map<string, CachedCredentials>()
```
mdkhan-tw, line 30: We could have used moize which does it well. Example https://github.com/segmentio/control-plane/blob/master/src/domains/destinations/core/services/encryption/encryption-service.ts#L200-L212

mdkhan-tw (dismissed):

mdkhan-tw (approved):

PR #4038, opened 2026-09-21
Re-ship actions-s3: STS wrap/caching, filename fix, key length guard (excludes header quoting + char validation)

itsarijitray:
## Re-ship of the safe actions-s3 fixes after the SEV3 (INC 20659) revert

Per the finalized [PIR](https://docs.google.com/document/d/1DrEpDtG-O5f78gn_OzieRKZxHAKNo__faO45b2Yn7QI/edit) and [RFO](https://docs.google.com/document/d/1ph2nA9YD_7PtlGaTjhuqZHiUyyU10NWGnaIWFTyYWQw/edit) for the actions-s3 outage, this PR re-applies the four `s3/syncToS3` changes that were reverted by #4031 but are **not implicated in the incident's root cause and don't alter delivered output**:

| Original PR | Change | Why it's safe to re-ship as-is |
| --- | --- | --- |
| #3996 | Reject S3 object keys over 1024 bytes | PIR repair item explicitly keeps this — AWS's length limit is a **hard, mandated** constraint (unlike the character-set check). Non-breaking: working keys are untouched, only genuinely-too-long keys (which would fail regardless) are rejected earlier with a clear error. |
| #3993 | Wrap STS assume-role errors, stop over-retrying permanent AWS errors | Internal error-classification only — no change to delivered output. Unrelated to the SEV. |
| #4004 | Cache STS assume-role credentials (stacked on #3993) | Internal only — no output change. Cuts STS call volume ~31x per DataDog data gathered during the original review. |
| #3994 | Fix filename extension corruption in `uploadS3` | Only changes output for the *currently-broken* case (e.g. `csv_export.csv` → stranded extension); byte-identical for already-working filenames. |

### Deliberately excluded from this re-ship

- **#3995 (CSV header quoting)** — this is the change that broke Fidelity's downstream ingestion. Per the PIR's Incident Prevention repair items, any change that alters delivered file *format* must ship behind an opt-in flag (default off) and get product approval before being enabled for any customer. That flagging work is a separate follow-up, not included here.
- **#4012 / STRATCONN-7017 (S3 key character validation)** — this is the change that caused the `PAYLOAD_VALIDATION_FAILED` failures (rejected valid keys containing spaces). Per the PIR's finalized repair item: *"S3 naming conventions guide is just a regular guideline and not strict validation... only add the length check as the strict validation item, as it is mandated in the guidelines."* This PR does **not** attempt to re-introduce character validation in any form.

### How this was built
`main` is fully reverted to the pre-cluster baseline (confirmed — no trace of any of the 6 GA-readiness changes). This branch cherry-picks the 4 original squash-merge commits for #3996, #3993, #4004, #3994 directly onto current `main`, in dependency order (#4004 is stacked on #3993). All 4 applied with **zero conflicts**, confirming there's no entanglement with the two excluded changes.

Verified after cherry-picking that #3995 and #4012's code is indeed absent (`grep` for `DISALLOWED_S3_OBJECT_KEY_CHARS` / header `encodeString` — neither present) and that all 4 target changes are present.

### Verification
- ✅ `tsc --noEmit` — clean
- ✅ `eslint` — clean
- ⚠️ Jest not run locally — this workspace's `node_modules` has a pre-existing jest major-version skew (`jest-runtime@27` hoisted vs `jest@30`) unrelated to this change, confirmed several times this session on unrelated files too. **CI will run the full suite.**

### Follow-ups (not in this PR)
1. Flag-gate CSV header quoting (#3995) behind an opt-in, default-off setting; get product sign-off before enabling for any customer.
2. Decide whether any lightweight S3-key guidance (docs-only, non-blocking) is still wanted, per the PIR's conclusion that strict character enforcement is not appropriate.


itsarijitray:
Splitting into 4 independently-mergeable/revertible PRs per the PIR's own repair item (don't bundle multiple destination behavior changes into one release unit):
- #4039 — S3 key length guard (STRATCONN-6990)
- #4040 — STS error wrap/classification (STRATCONN-6987)
- #4041 — STS credential caching (STRATCONN-6987 pt.2, stacked on #4040)
- #4042 — filename extension fix (STRATCONN-6988)

Closing this combined PR in favor of those four.

PR #4039, opened 2026-09-21
Re-ship: reject S3 object keys over 1024 bytes up front (STRATCONN-6990)

itsarijitray:
## Re-ship after SEV3 INC 20659

Split out of #4038 per the PIR's own repair item — ship independently-revertible units instead of a cluster.

Re-ships **#3996** (originally reverted by #4031) standalone. Cherry-picked cleanly onto current `main` (fully reverted baseline) with zero conflicts.

**Why this one is safe to re-ship as-is:** the [finalized PIR](https://docs.google.com/document/d/1DrEpDtG-O5f78gn_OzieRKZxHAKNo__faO45b2Yn7QI/edit) explicitly keeps this check — AWS's 1024-byte object-key limit is a **hard, mandated** constraint (unlike the character-set validation in #4012, which the PIR concluded should *not* be re-introduced). Non-breaking: working keys are untouched; only genuinely-too-long keys (which would fail at the PUT regardless) are rejected earlier with a clear, non-retryable error instead of an opaque late AWS failure.

Unrelated to the incident's root cause (that was #4012's character check and #3995's header-quoting change — neither is part of this re-ship).

### Verification
- ✅ `tsc --noEmit` clean
- ✅ `eslint` clean
- ⚠️ Jest not run locally (pre-existing workspace jest version skew, unrelated to this change). CI will run the suite.


itsarijitray:
**Update: gated behind a feature flag for safer rollout.**

Added \`S3_KEY_LENGTH_GUARD_FLAG\` (\`'actions-s3-key-length-guard'\`) in \`s3/constants.ts\`, off by default — following the exact same \`features[]\` pattern this destination already uses for \`S3_HASHING_FEATURE_FLAG\`. \`features\` is now threaded through \`send() → uploadS3()\`, and the length check only runs when the flag is enabled for a workspace.

- **Flag off (default):** identical to today's live behavior — no length check at all.
- **Flag on:** the length guard from the original #3996.

Tests updated: added an explicit "off by default, does not reject" case alongside the existing (now flag-gated) reject/PUT tests.

itsarijitray:
## Review Summary

review-toolkit ran 6 agents against this diff (bug-hunter, security-reviewer, silent-failure-hunter, intent-reviewer, doc-grounding-reviewer, test-analyzer). 1 finding: 0 critical, 1 high (test-gap). No bugs, security issues, or intent concerns found — this is a clean, well-scoped re-ship.

**Recommendation:** COMMENT

<details><summary>🟠 High Findings (1)</summary>

#### `features` wiring from `send()` to `uploadS3` is untested — `packages/destination-actions/src/destinations/s3/syncToS3/functions.ts:53`
The `uploadS3` guard itself is well unit-tested directly against `Client`, but the actual production wiring path — `index.ts` (perform/performBatch) → `functions.ts`'s `send()` → `Client.uploadS3` — has zero coverage for the new `features` parameter. `syncToS3/__tests__/index.test.ts` mocks `Client.uploadS3` entirely (`jest.fn().mockResolvedValue(...)`), so it never asserts what `send()` actually passes through. A future positional-argument regression (e.g. reordering `features`/`signal`) would silently disable the flag in production while every existing test stayed green — notable given this exact flag has prior incident history (STRATCONN-6986/INC 20659).

**Suggestion:** Add a test asserting `send()` forwards the `features` object to `uploadS3` intact and in the correct position, either via a spy on the mocked `Client.uploadS3` or a full integration test via `testDestination.testAction`.
_high confidence_

</details>

**Ran:** bug-hunter, security-reviewer, silent-failure-hunter, intent-reviewer, doc-grounding-reviewer, test-analyzer
**Complexity tier:** standard (single-chunk, 198-line diff)

All claims in the PR description were verified against the diff and `origin/main` by doc-grounding-reviewer — no contradictions found (flag-off byte-identical behavior, byte-boundary math, fail-fast ordering, PII-safe error message, and the revert history #3996/#4031/#4012/#3995 all check out).

</details>

---

packages/destination-actions/src/destinations/s3/syncToS3/functions.ts
```diff
@@ -56,6 +56,7 @@ export async function send(
     payloads[0]?.filename_prefix ?? '',
     payloads[0]?.s3_aws_folder_name ?? '',
     payloads[0]?.file_extension,
+    features,
```
itsarijitray: **[high] `features` wiring from `send()` to `uploadS3` is untested**

The guard itself is well unit-tested directly against `Client`, but this production wiring path has zero coverage for the new `features` parameter -- `index.test.ts` mocks `Client.uploadS3` entirely. A future positional-argument regression here would silently disable the flag while all tests stayed green.

**Suggestion:** Add a test asserting `send()` forwards `features` to `uploadS3` intact and in the correct position.

itsarijitray (commented):
Inline comments from review-toolkit agents.

itsarijitray:
Closing per team discussion (https://twilio.slack.com/archives/C0BUCQL3YJW/p1790259171982109?thread_ts=1790254357.545009&cid=C0BUCQL3YJW): pre-emptive key-length validation only renames the error *after* the customer has already misconfigured filename_prefix/folder, and adds a feature flag + new code path for it. AWS's own `KeyTooLongError` is already descriptive enough — we'll classify it properly in #4040's `mapAWSError` instead, which is simpler and lower-risk (no flag needed).

PR #4040, opened 2026-09-21
Re-ship: wrap STS assume-role errors, stop over-retrying permanent AWS errors (STRATCONN-6987)

itsarijitray:
## Re-ship after SEV3 INC 20659

Split out of #4038 per the PIR's repair item to ship independently-revertible units.

Re-ships **#3993** (originally reverted by #4031) standalone. Cherry-picked cleanly onto current `main` with zero conflicts.

**Why this is safe:** purely internal error classification (STS/AWS error → correct Segment error class + status + retryability). **No change to delivered output** and **not implicated in the incident** — the SEV was caused by #4012 (character validation) and #3995 (header quoting), neither of which is part of this re-ship. This was the original GA blocker (opaque `type:internal` 500s force-retrying permanent failures).

### Verification
- ✅ `tsc --noEmit` clean
- ✅ `eslint` clean
- ⚠️ Jest not run locally (pre-existing workspace jest version skew, unrelated). CI will run the suite.


itsarijitray:
**Update: gated behind a feature flag for safer rollout.**

Added \`S3_STS_ERROR_CLASSIFICATION_FLAG\` (\`'actions-s3-sts-error-classification'\`) in \`s3/constants.ts\`, off by default. \`features\` is now threaded through the \`Client\` constructor (\`send() → new Client(...)\`), gating **both**:
- the STS assume-role try/catch wrapping, and
- the PUT-catch error classification (\`mapAWSError\`, including the wrong-region → 401 fix).

**Flag off (default):** byte-identical to today's live behavior — STS errors escape unwrapped, and the PUT catch block uses the original inline classification (kept as \`legacyAccessDeniedCodes\` / the original if-else chain), so disabling the flag fully restores current production behavior.
**Flag on:** the full classification fix from the original #3993.

Tests updated: added an explicit "off by default, STS error escapes unclassified" case alongside the existing (now flag-gated) classification tests.

itsarijitray:
## Review Summary

review-toolkit ran 6 agents against this diff (bug-hunter, security-reviewer, silent-failure-hunter, intent-reviewer, doc-grounding-reviewer, test-analyzer). 6 findings: 0 critical, 6 high (2 bug, 1 error-handling-cluster of 3 merged into 2 distinct issues, 2 test-gap).

**Recommendation:** REQUEST_CHANGES

<details><summary>🟠 High Findings (5)</summary>

#### Region-mismatch redirect mislabeled as `INVALID_AUTHENTICATION` — `packages/destination-actions/src/destinations/s3/syncToS3/client.ts:310`
[Found by 2 agents: bug-hunter, silent-failure-hunter] `mapAWSError` maps `PermanentRedirect`/`TemporaryRedirect` (a bucket-region misconfiguration) to `IntegrationError(detail, ErrorCodes.INVALID_AUTHENTICATION, 401)` — the same code/status a genuine bad-credentials failure gets. Any downstream logic or customer messaging keyed off `INVALID_AUTHENTICATION` will tell the customer to fix their credentials when the actual fix is the `s3_aws_region` setting. Notably, the PR's own code comment on the *generic* 4xx fallback explicitly warns against this exact mislabeling pattern — the redirect branch does precisely what that comment cautions against.

**Suggestion:** Use a neutral/config error code (e.g. `ErrorCodes.UNKNOWN_ERROR` or a dedicated configuration code) instead of `INVALID_AUTHENTICATION` for the redirect case.
_high confidence_

#### Segment's own intermediary-role STS failures now classified identically to customer-role failures — `packages/destination-actions/src/destinations/s3/syncToS3/client.ts:22`
`assumeRole()` calls the shared `getSTSCredentials()` for both Segment's internal bridging role and the customer's role, and the new error-wrapping applies the same non-retryable classification to both with no way to distinguish which failed. A transient AWS-side issue on Segment's *own* internal role assumption that happens to surface with an access-denied-shaped code will now be permanently classified as non-retryable `APIError(403)` instead of retried — a regression risk that could drop customer events that previously would have succeeded on retry.

**Suggestion:** Pass context to `getSTSCredentials` distinguishing intermediary vs. customer role assumption, and only apply strict non-retryable classification to the customer-role case.
_high confidence_

#### Unrecognized/non-AWS errors silently folded into the same "retryable" bucket as real transient AWS failures — `packages/destination-actions/src/destinations/s3/syncToS3/client.ts:319`
`mapAWSError`'s fallback has no shape check equivalent to the old `isAWSError()` guard, so a genuine programming defect (e.g. a `TypeError` from a bug elsewhere in the call path) gets the identical `RetryableError` treatment as an expected AWS transient condition, with no distinguishing signal in logs. Given the PR's whole premise is "don't force-retry things that will never succeed," this reintroduces that exact failure mode for anything the classifier doesn't recognize.

**Suggestion:** Keep a shape check before the retryable fallback; label genuinely unclassified/non-AWS errors distinctly (e.g. `ErrorCodes.UNKNOWN_ERROR`) rather than lumping them in with known-transient AWS conditions.
_high confidence_

#### `mapAWSError` discards the original error — no stack trace, no AWS request id
The new classifier builds a fresh error from just a string `detail`, with no `cause`/stack/`$metadata.requestId` preserved. For a PR whose stated goal is improving incident diagnosability after INC 20659, this makes future incidents harder to correlate back to the actual AWS request.

**Suggestion:** Attach `$metadata?.requestId` to `detail`, or preserve the original error via a `cause`-style field.
_medium confidence_

#### New STS access-denied codes (`ExpiredToken`, `ExpiredTokenException`, `AccessDeniedException`) untested on the flag-off (legacy) path — `packages/destination-actions/src/destinations/s3/syncToS3/client.ts:347`
No test verifies flag-off behavior is unchanged for these specific new codes, which is exactly the guarantee the PR claims to preserve.

**Suggestion:** Add a test asserting flag-off does NOT classify these codes as 403 via the legacy path.
_high confidence_

</details>

<details><summary>🟠 Additional test-gap finding</summary>

#### `features` param wiring from `send()` into `Client` is untested — `packages/destination-actions/src/destinations/s3/syncToS3/functions.ts:51`
No test exercises `send()` to confirm `features` actually propagates end-to-end into `Client` and changes classification behavior.
_high confidence_

</details>

**Ran:** bug-hunter, security-reviewer, silent-failure-hunter, intent-reviewer, doc-grounding-reviewer, test-analyzer
**Complexity tier:** standard (single-chunk, 386-line diff)

Flag-off byte-identical claim, the STS-escapes-try/catch bug being fixed, and both named classification-bug fixes were all verified grounded by doc-grounding-reviewer. security-reviewer found no critical/high issues (no new injection surface, no secret exposure, no auth bypass). intent-reviewer raised one non-blocking process nudge: confirm a follow-up plan exists to delete the flag-off legacy path once rollout completes.

</details>

---

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -125,7 +149,68 @@ export class Client {
...
+    return new APIError(detail, 429)
+  }
+  if (code && redirectCodes.has(code)) {
+    // S3 returns a redirect (e.g. PermanentRedirect, HTTP 301) when the bucket lives in a
+    // different region than configured. It's a permanent client misconfiguration, so surface a
+    // non-retryable 401 rather than leaking the raw 3xx redirect status.
+    // Note: TemporaryRedirect (bucket mid-migration) and PermanentRedirect (fixed misconfiguration)
+    // are intentionally collapsed into the same non-retryable bucket here — a normal retry window
+    // won't resolve either, even though their root causes differ.
+    return new IntegrationError(detail, ErrorCodes.INVALID_AUTHENTICATION, 401)
```
itsarijitray: **[high] Region-mismatch redirect mislabeled as INVALID_AUTHENTICATION**

[Found by 2 agents: bug-hunter, silent-failure-hunter] `PermanentRedirect`/`TemporaryRedirect` (a bucket-region misconfiguration) is mapped to `IntegrationError(detail, ErrorCodes.INVALID_AUTHENTICATION, 401)` -- the same code/status a real bad-credentials failure gets. Downstream logic/customer messaging keyed off `INVALID_AUTHENTICATION` will point at credentials when the real fix is the `s3_aws_region` setting. Notably this is exactly the mislabeling pattern the PR's own comment on the generic 4xx fallback warns against.

**Suggestion:** Use a neutral/config error code (e.g. `ErrorCodes.UNKNOWN_ERROR`) instead of `INVALID_AUTHENTICATION` for the redirect case.

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -34,7 +38,21 @@ export class Client {
...
     })
-    const result = await stsClient.send(command)
+    let result
+    if (this.features?.[S3_STS_ERROR_CLASSIFICATION_FLAG]) {
+      try {
+        result = await stsClient.send(command)
+      } catch (err) {
+        // STS failures used to escape uploadS3's try/catch entirely, so they reached the platform
+        // with no status/code, were classified type:internal and force-retried (even permanent auth
+        // failures). Map them to Segment error classes here so classification is correct.
```
itsarijitray: **[high] Segment's own intermediary-role STS failures now classified identically to customer-role failures**

`assumeRole()` calls this same wrapped path for both Segment's internal bridging role and the customer's role, with no way to distinguish which failed. A transient AWS-side issue on Segment's own internal role assumption that surfaces with an access-denied-shaped code will now be permanently classified as non-retryable `APIError(403)` instead of retried -- a regression risk that could drop customer events that previously succeeded on retry.

**Suggestion:** Pass context distinguishing intermediary vs. customer role assumption, and only apply strict non-retryable classification to the customer-role case.

packages/destination-actions/src/destinations/s3/syncToS3/client.ts
```diff
@@ -125,7 +149,68 @@ export class Client {
...
+  // A client fault (4xx that is not throttling/transient) is permanent - do not retry.
+  if (e?.$fault === 'client' || (typeof httpStatus === 'number' && httpStatus >= 400 && httpStatus < 500)) {
+    // Only ever surface a genuine 4xx as the status; never leak a non-4xx (e.g. a 3xx redirect).
+    const status = typeof httpStatus === 'number' && httpStatus >= 400 && httpStatus < 500 ? httpStatus : 400
+    // Not necessarily an auth problem — just an unclassified client-side failure — so use a
+    // neutral error code rather than mislabeling it as an authentication issue.
+    return new IntegrationError(detail, ErrorCodes.UNKNOWN_ERROR, status)
+  }
+  // Transient / server-side / unclassified failures are safe to retry.
+  return new RetryableError(detail)
```
itsarijitray: **[high] Unrecognized/non-AWS errors silently folded into the same "retryable" bucket as real transient AWS failures**

`mapAWSError`'s fallback has no shape check equivalent to the old `isAWSError()` guard, so a genuine programming defect (e.g. a `TypeError` from a bug elsewhere) gets the identical `RetryableError` treatment as an expected AWS transient condition, with no distinguishing signal in logs. This also discards the original error's stack trace/AWS request id, making future incidents harder to correlate back to the actual AWS request.

**Suggestion:** Keep a shape check before the retryable fallback; label genuinely unclassified/non-AWS errors distinctly, and consider preserving `$metadata?.requestId`/a `cause` reference.

packages/destination-actions/src/destinations/s3/syncToS3/functions.ts
```diff
@@ -48,7 +48,7 @@ export async function send(
 
   const fileContent = generateFile(payloads, headers, delimiter, actionColName, batchColName, columnTransforms)
 
-  const s3Client = new Client(settings.s3_aws_region, settings.iam_role_arn, settings.iam_external_id)
+  const s3Client = new Client(settings.s3_aws_region, settings.iam_role_arn, settings.iam_external_id, features)
```
itsarijitray: **[high] `features` param wiring from `send()` into `Client` is untested**

No test exercises `send()` to confirm `features` actually propagates end-to-end into `Client` and changes classification behavior -- existing tests mock `Client`/`uploadS3` entirely.

**Suggestion:** Add a test asserting the mocked `Client` constructor receives the `features` object passed to `send()`.

itsarijitray (commented):
Inline comments from review-toolkit agents.

PR #4043, opened 2026-09-21
STRATCONN-7039 - [First Party DV360] - Add phone number normalization and validation to Sync Audience

joe-ayoub-segment:
Display & Video 360 only matches a phone number hashed from its E.164 form, so a number sent in any other format is silently unmatched — no error, just a member who never matches. The `Sync Audience` Action can now convert numbers before hashing them. It is off by default, so every existing mapping is byte identical.

JIRA: [STRATCONN-7039](https://twilio-engineering.atlassian.net/browse/STRATCONN-7039)

### What's new

A `Phone Number Normalization` object field with three settings:

| setting | |
|---|---|
| **Normalization** | `Do not normalize` (default), `Normalize`, or `Normalize and validate` |
| **Default Country** | the country the numbers belong to, from a list of 245 |
| **Use Country Code From Contact Info Details** | take the country per user instead, from the mapped Country Code |

`Contact Info Details > Country Code` becomes a dropdown of the same 245 countries rather than free text, so a mapped value has to resolve to an ISO 3166-1 alpha-2 code.

### How a number is read

Three shapes, and only three:

- starts with `+` — states its own country, converted with no country set
- starts with `00` — same
- anything else — a local number, converted only against the country it belongs to

`Default Country` means the country a number *belongs to*, never where a call was placed from, so other countries' dialling prefixes (`0011`, `810`, `011`) are not treated as international.

Anything that cannot be converted is sent exactly as mapped under `Normalize`, or dropped under `Normalize and validate`. An event left with no identifier then fails with the existing `No usable contact info identifiers found` error, per event in the MultiStatusResponse.

### Testing

34 new unit tests; 233 pass for the destination, the full cloud suite is green at 7130. `syncAudience/functions.ts` is at 100% statements, lines and functions. Typecheck, lint, prettier and `yarn validate` are clean, and `metadata.json` was regenerated with `yarn generate:metadata-payload`.

The 199 pre-existing tests pass unchanged, which is the evidence that the default path is untouched.

[STRATCONN-7039]: https://twilio-engineering.atlassian.net/browse/STRATCONN-7039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Prithviraj-rathore-segment (dismissed):

packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts
```diff
@@ -263,31 +268,102 @@ export function normaliseEmail(value: string): string | undefined {
...
+
+  return country ? tryParse(phone, country) : undefined
+}
+
+function tryParse(phone: string, country?: CountryCode) {
+  try {
+    const parsed = phoneUtil.parse(phone, country)
+
+    return phoneUtil.isValidNumber(parsed) ? parsed : undefined
+  } catch (error) {
```
charlottestjohn, line 332: :yellow_circle: **MEDIUM: No visibility when a number is dropped or sent unconverted**
<details>
<summary>see reasoning</summary>

`tryParse` swallows every parse/validation failure and returns `undefined` with no telemetry. Downstream, `buildContactInfo` drops those numbers via `.filter(isPresent)` (validate mode) or `normalisePhone` returns the raw string (normalize mode). A per-event error only surfaces when ALL identifiers are dropped; partial drops and pass-through-unmatched numbers are invisible. This is the exact 'silently unmatched' problem the PR exists to solve, now moved one layer up. This is intentional per the design and does not block merge — but operators get no signal that a mapping's country config is wrong and match rates are quietly falling.

**Suggestion:** Emit a stat/metric (via the available StatsContext) counting numbers dropped in validate mode and numbers passed through unconverted in normalize mode, so misconfiguration is observable without echoing the PII itself.

</details>

charlottestjohn (commented):
> ℹ️ Auto-reviewed because you requested me/my group as reviewer — draft the PR to skip this.
### Reviewed :speech_balloon:

No confirmed critical/high defects survived verification and the PII-hashing path is sound, but the free-text-to-enum change on Country Code and the lack of visibility into silently-dropped/unmatched numbers warrant a human's confirmation before merge.

Findings: 1 Medium — expand below.

---

<details>
<summary><strong>:memo: Conceptual Review — 1 item for human confirmation</strong></summary>

**Should this change exist / is it the right approach?** Yes. DV360 only matches phone numbers hashed from their E.164 form, so normalizing before hashing is the correct fix, and delegating parsing/validation to `google-libphonenumber` (the standard library) rather than hand-rolling is the right call. The feature is opt-in (`normalization` defaults to `PHONE_NORMALIZATION_NONE`), and the default output path was verified byte-identical to the prior behavior.

**One coherence question worth a human glance (non-blocking):** the `Contact Info Details > Country Code` field changes from free-text `type: 'string'` to the same string type plus `choices: COUNTRY_CHOICES` (fields.ts). This is presentational in the actions framework — the emitted `countryCode` in the ContactInfo payload is still `countryCode?.trim().toUpperCase()`, identical to before, and the country is only consumed for phone parsing when the opt-in `Use Country Code From Contact Info Details` toggle is on (off by default). So no runtime break is evident in the diff. The residual uncertainty is whether the mapping/validation layer rejects previously-saved non-ISO free-text values; the intent-reviewer raised this and it is worth confirming, but the source does not support it being a hard break. Downgraded from a blocking concern to a confirmation item on that basis.

</details>

<details>
<summary><strong>:memo: Security (overlay) — 0 critical, 0 high</strong></summary>

PII (phone numbers) is untrusted external input, normalized then hashed before leaving for DV360. Verified against source:

- **Hashing correctness:** numbers are canonicalized to E.164 (`phoneUtil.format(parsed, PhoneNumberFormat.E164)`) *before* `hash(phoneNumber, trimOnly)` → `processHashing(value, 'sha256', 'hex', ...)`, so the same person hashes consistently; already-hashed values are detected via `isHashed` (`isAlreadyHashed(value, 'sha256', 'hex')`) and passed through unchanged, avoiding double-hashing.
- **No PII leakage:** the parser's `catch (error)` block returns `undefined` without logging; no `console.log`/thrown error echoes the number, and the per-event error remains the generic `No usable contact info identifiers found`.
- **Parsing safety:** parsing is fully delegated to `google-libphonenumber` (legitimate, widely-used Google port) inside try/catch — no hand-rolled regex ReDoS surface.
- **AuthZ/tokens:** the change does not touch auth headers, OAuth, or token/refresh logic.
- **Secret/supply-chain scans:** CI Wiz Secret Scanner, Snyk (security + license), and Semgrep all passed clean.

The security passes raised only speculative findings (PII leaking *if* an upstream handler logged the exception; info disclosure *if* future code added specific messages; audience exclusion from self-controlled country codes). None is supported by the current source; see the re-verified & dismissed rows. No security finding floors the verdict.

</details>

<details>
<summary><strong>:memo: Code Review Results (review-toolkit) — 1 medium observation :yellow_circle:</strong></summary>

The normalization/validation logic is well-structured and thoroughly tested (34 new unit tests; `functions.ts` reported at 100% statements/lines/functions). The behavior for non-convertible numbers (send unchanged under Normalize, drop under Normalize-and-validate) is the documented, intentional contract, not a defect. One observability gap is worth surfacing inline.

**Findings**

- :yellow_circle: **MEDIUM: No visibility when a number is dropped or sent unconverted** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:332`)
  `tryParse` swallows every parse/validation failure and returns `undefined` with no telemetry. Downstream, `buildContactInfo` drops those numbers via `.filter(isPresent)` (validate mode) or `normalisePhone` returns the raw string (normalize mode). A per-event error only surfaces when ALL identifiers are dropped; partial drops and pass-through-unmatched numbers are invisible. This is the exact 'silently unmatched' problem the PR exists to solve, now moved one layer up. This is intentional per the design and does not block merge — but operators get no signal that a mapping's country config is wrong and match rates are quietly falling.
  **Suggestion:** Emit a stat/metric (via the available StatsContext) counting numbers dropped in validate mode and numbers passed through unconverted in normalize mode, so misconfiguration is observable without echoing the PII itself.

</details>

<details>
<summary><strong>:test_tube: CI Checks — All checks passed</strong></summary>

`gh pr checks` snapshot (head 976747672e0db8566b2c16a73a9dbe3b0f99feea):

| Check | Result |
|---|---|
| Lint (22.x) | pass |
| Unit tests (22.x) | pass |
| Validate (22.x) | pass |
| install-and-test (22.x) | pass |
| required-field-check (22.x) | pass |
| secret-fields-check (22.x) | pass |
| Browser Destination Bundle QA (22.x) | pass |
| Browser tests: actions-core (22.x) | pass |
| Wiz Secret / IaC / Data / SwMgmt / Vuln Scanners | pass |
| Snyk security + license (segmentio) | pass |
| Semgrep cloud-platform scan | pass |
| opa policy validation | pass |

All 22 checks green. Green lint + typecheck + 34 passing unit tests that call `normalisePhone` were the deciding evidence that dropped the false 'undefined isHashed' critical.

</details>

<details>
<summary><strong>:bulb: Suggestions (1)</strong></summary>

1. Confirm existing free-text `Country Code` mappings still resolve after adding `choices: COUNTRY_CHOICES` to the field. The runtime output is byte-identical (the value is still trimmed/uppercased and sent as-is) and the enum is UI-level, but verify the Segment mapping layer does not reject stored non-ISO values (e.g. 'United States', 'USA') for audiences already syncing Contact Info.

</details>

<details>
<summary><strong>:recycle: Re-verified &amp; dismissed (6)</strong></summary>

- **Undefined function isHashed causes runtime crash** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:281`)
  False. isHashed is defined at functions.ts:257 (local wrapper: return isAlreadyHashed(value, 'sha256', 'hex')), outside the diff hunk the agents saw. Green lint/typecheck and 34 passing tests that call normalisePhone confirm it resolves.
- **Unsafe CountryCode cast allows invalid country codes** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:350`)
  The `as CountryCode` is a compile-time-only assertion with no runtime effect; an invalid region makes phoneUtil.parse throw, is caught, and the number is sent as-is (normalize) or dropped (validate) per the documented contract. The emitted countryCode output is byte-identical to prior behavior. No crash or data-corruption path.
- **Module-level PhoneNumberUtil.getInstance() can crash on load** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:271`)
  google-libphonenumber is pure JavaScript with no native dependency; getInstance() is a standard top-level pattern used across the codebase. Speculative crash scenario not supported by the library or the diff.
- **Phone parsing exceptions may leak PII in error logs** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:332`)
  Speculative ('if any upstream handler logs the exception'). The catch returns undefined without logging and no code in the diff logs the phone value. No actual leak path in the change.
- **buildMember error messages may expose phone validation details** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:301`)
  About hypothetical future code adding specific messages. The current error is the generic, pre-existing 'No usable contact info identifiers found'. No defect in the change under review.
- **Country code manipulation can cause silent audience exclusion** (`packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:301`)
  The 'attacker' controls only their own event's country code, affecting only their own number's match. This is the documented normalize-mode behavior (send unconverted), gated behind an opt-in toggle that is off by default. Not a security vulnerability.

</details>


<sub>Reproduce locally: `/batch-pr:review https://github.com/segmentio/action-destinations/pull/4043`</sub>
<sub>Generated by batch-pr v0.7.15 | claude-opus-4-8 | review-toolkit dispatch (bug-hunter x2, security-reviewer x2, silent-failure-hunter, intent-reviewer, security-scanner), reverify-findings gate, gh pr checks | 2026-09-21</sub>

packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts
joe-ayoub-segment: I disagree. the metric isn't useful for customers as it's internal to Segment. 
If a customer is only sending a single identifier then an error message will be returned. But if there is a valid identifier in the payload then the payload will be sent. 
The customer can always decide if they want to send the invalid identifier anyway (they can turn off validation for phone field).

nk1107:
## Review Summary

Opt-in phone-number normalization for the DV360 Sync Audience action is well-built — the parsing model, hashing, PII handling, performance (module-level `phoneUtil` singleton, opt-in per-event cost), and the generated `metadata.json` all check out, and the phone default path (`normalization: 'none'`) is genuinely byte-identical. **However, one change in the same PR is not opt-in and is a breaking change: converting the existing free-text `countryCode` field to a `choices` dropdown enforces a JSON-schema `enum` at runtime, so mappings that previously sent lowercase/whitespace/free-text country values now hard-fail the whole event.** 3 findings: 0 critical, 3 high (+1 conceptual).

**Recommendation:** COMMENT — but the `countryCode` finding is REQUEST_CHANGES-class. This repo ships no `.claude/review-toolkit.yml`, so the tool defaults to comment-only and does not post a formal change request; a human should treat finding #1 as blocking.

### 🧭 Conceptual Review

#### 245-country table is hand-maintained and duplicates data `google-libphonenumber` already exposes
The feature itself is well-scoped, opt-in, non-breaking (on the phone path), and solves a real matching problem the right way — no necessity or coherence objection. The one nudge: the ~245-row `COUNTRIES` table (name + ISO code + dialing code) hand-transcribes a dataset the imported library already owns — `phoneUtil.getSupportedRegions()` + `getCountryCodeForRegion(code)`, with display names from `Intl.DisplayNames`. On a library upgrade that adds/renames a region the hardcoded list silently drifts, and the invariant is not enforced anywhere (see finding #2). Legitimate reasons exist to keep a curated list (stable spellings like "Türkiye", stable ordering, ICU-version independence), so this is a conscious-tradeoff call for the author, not a defect.
**Simpler path:** derive `COUNTRIES`/`PHONE_REGION_CHOICES` from `getSupportedRegions()`, or — if the curated list is intentional — at minimum enforce it against the library (finding #2).
_medium confidence_

<details><summary>🟠 High Findings (3)</summary>

#### Adding `choices` to the existing free-text `countryCode` field is a breaking change — `packages/destination-actions/src/destinations/first-party-dv360/syncAudience/fields.ts:90`
[Found by 2 agents: bug-hunter, code-reviewer]

`contact_info.countryCode` changed from free-text `type: 'string'` to `type: 'string'` + `choices: COUNTRY_CHOICES`. In this framework `choices` is **not** a UI hint — `fieldsToJsonSchema` compiles it to a JSON-schema `enum` (`packages/core/src/destination-kit/fields-to-jsonschema.ts:332`) and recurses into object `properties`, so the nested `countryCode` subfield gets `enum: ['AF',…,'US',…]` (uppercase, AC/TA/XK removed). At runtime `Action.execute` calls `validateSchema(payload, this.schema, …)` (`packages/core/src/destination-kit/action.ts:639`) with the default `throwIfInvalid: true` (`packages/core/src/schema-validation.ts:54`), and AJV is configured with only `coerceTypes: 'array'` — **no case coercion** — so an enum mismatch throws `AggregateAjvError` for the *whole* payload.

**Verified against source.** Concrete inputs that synced before and now hard-fail with a non-retryable `PAYLOAD_VALIDATION_FAILED`:
- lowercase/mixed-case codes (`'us'`, `'gb'`) — the perform code's own `countryCode?.trim().toUpperCase()` proves lowercase was expected/valid before; but validation runs on the raw resolved value *before* perform uppercases it;
- whitespace-padded values (`' US '`);
- full names or any free text (`'United States'`, `'USA'`, numeric);
- `'XK'` (Kosovo), `'AC'`, `'TA'` — deliberately excluded from `COUNTRY_CHOICES` but still valid regions.

Because an enum violation throws for the entire payload, a customer who supplied email + a non-enum countryCode is now dropped entirely instead of syncing on email. This is **not** gated behind the phone-normalization opt-in, so the PR's "default path is byte-identical" claim does not hold for anyone who maps `countryCode`. Directly violates CLAUDE.md: "Don't change field types in ways that could break existing integrations." Note the new unit test at `functions.test.ts` feeds lowercase `countryCode: 'gb'` straight into `buildContactInfo`, bypassing schema validation — so the test passes while real traffic would be rejected upstream.

**Suggestion:** Do not enum-constrain the pre-existing `countryCode` field — drop `choices: COUNTRY_CHOICES` from `contact_info.countryCode` and keep it free-text (the code already uppercases it before use); validate/normalize the country *inside* `buildContactInfo` so bad values degrade gracefully instead of failing the event. If a UI picker is genuinely wanted, note `choices` still enforces the enum at runtime and cannot accept the case-insensitive / XK-AC-TA / free-text values existing mappings already send. The new `phone_options.*` choices are fine — brand-new optional fields with valid default enum members.
_high confidence_

#### `PHONE_REGION_CHOICES` region test is tautological — never checks codes against libphonenumber — `packages/destination-actions/src/destinations/first-party-dv360/syncAudience/__tests__/functions.test.ts:299`
The test named "offers every libphonenumber region as a phone region" asserts `values(PHONE_REGION_CHOICES)` equals `COUNTRIES.map(({ code }) => code)`. But `PHONE_REGION_CHOICES` is *defined* as `COUNTRIES.map(toChoice)` with `value: code`, so it compares `COUNTRIES.map(code)` to itself — it can never fail, and never touches libphonenumber despite its name. The 245-entry hand-maintained `COUNTRIES` list is the correctness promise of the whole feature: any region a customer picks as Default Country must actually be a region `google-libphonenumber` can parse against. A typo'd or unsupported ISO code would silently drop/skip every national-format number for that region, uncaught. (Ties to the conceptual drift note above.)

**Suggestion:** Intersect against the real library, e.g. `const supported = new Set(PhoneNumberUtil.getInstance().getSupportedRegions()); COUNTRIES.forEach(({ code }) => expect(supported.has(code)).toBe(true))`, so an unsupported/typo'd region — or an upstream upgrade that renames one — fails CI instead of silently drifting.
_high confidence_

#### No test for per-user country enabled + user has no country + no default (silent drop under VALIDATE) — `packages/destination-actions/src/destinations/first-party-dv360/syncAudience/functions.ts:759`
In `phoneCountry`, when `useContactInfoCountryCode: true` and the per-user `contactInfoCountryCode` is absent, resolution falls through to `defaultCountryCode`; if that is also unset the result is `undefined`, so any national-format number is unresolvable and — under `PHONE_NORMALIZATION_VALIDATE` — is dropped. Tests cover the fall-back-to-default case (with a default supplied) and the no-country case with the per-user toggle off, but not the realistic misconfiguration: per-user country ON, a user missing their country, and no Default Country configured. That composition causes silent per-user data loss (national numbers of every country-less user vanish) and is only covered by transitivity, not an explicit assertion.

**Suggestion:** Add `expect(normalisePhone('(212) 565-0000', { normalization: PHONE_NORMALIZATION_VALIDATE, useContactInfoCountryCode: true })).toBeUndefined()` and a NORMALIZE counterpart asserting pass-through unchanged, both with no `defaultCountryCode` and no `contactInfoCountryCode`.
_medium confidence_

</details>

<details><summary>📝 Sub-threshold notes (not blocking)</summary>

These were surfaced below the critical/high bar and are recorded for awareness only:
- **`phoneCountry` does not fall back to default when the per-user country is populated-but-invalid** (silent-failure-hunter, medium): with `useContactInfoCountryCode` on, a truthy-but-unrecognized per-user `countryCode` is used as-is and does *not* fall back to `defaultCountryCode`; `parse` then throws → number dropped (VALIDATE) or passed through (NORMALIZE). Matches the field's documented "whenever it is not populated" contract, but the populated-but-garbage case may surprise.
- **Wrong-per-user-country cross-matching** (security-reviewer, low/medium): a national-format number parsed against a wrong-but-valid per-user country can format to a valid E.164 belonging to a different real subscriber and hash-match the wrong individual. Requires the customer to explicitly enable the option and trust their own event data; it is the documented purpose of the field.
- **Legacy RETL path is unaffected** (bug-hunter): `first-party-dv360/functions.ts` `buildContactInfoList` still hashes phone numbers without normalization. Pre-existing and untouched by this PR — noted only so the coverage boundary is explicit.

</details>

**Ran:** bug-hunter (ensemble a/b), security-reviewer (ensemble a/b), silent-failure-hunter, code-reviewer, performance-reviewer, test-analyzer, intent-reviewer, metadata-consistency
**Skipped:** architecture-reviewer (localized single-action change), type-design-reviewer (only two narrow casts), api-standards-reviewer (no HTTP surface), dependency-scanner (`google-libphonenumber` already a declared dep, package.json untouched), doc-grounding / docs-validator / language-pattern / simplification (low signal for this diff)
**Complexity tier:** standard — diff reviewed in 4 chunks (3 are the generated `metadata.json`; all substantive code in one chunk)

</details>

---

packages/destination-actions/src/destinations/first-party-dv360/syncAudience/fields.ts
```diff
@@ -75,8 +85,9 @@ export const contact_info: InputField = {
     },
     countryCode: {
       label: 'Country Code',
-      description: `The user's country code. ${ADDRESS_GROUP}`,
-      type: 'string'
+      description: `The user's country, as an ISO 3166-1 alpha-2 code. ${ADDRESS_GROUP} It is also used as the country for this user's phone numbers when 'Use Country Code From Contact Info Details' is enabled.`,
+      type: 'string',
+      choices: COUNTRY_CHOICES
```
nk1107, line 90: **[high] Adding `choices` to the existing free-text `countryCode` field is a breaking change**

Verified against source: `choices` compiles to a JSON-schema `enum` (fields-to-jsonschema.ts:332, recurses into object properties) and `validateSchema` at action.ts:639 defaults to `throwIfInvalid: true` with only `coerceTypes: 'array'` (no case coercion). So a resolved `contact_info.countryCode` that is lowercase (`us`/`gb`), whitespace-padded (` US `), free text (`United States`), or `XK`/`AC`/`TA` now throws for the *whole* payload — a non-retryable PAYLOAD_VALIDATION_FAILED — where it previously synced. NOT gated by the phone-normalization opt-in, so the "default path is byte-identical" claim does not hold for anyone who maps countryCode. [Found by 2 agents: bug-hunter, code-reviewer]

**Suggestion:** Drop `choices: COUNTRY_CHOICES` from `contact_info.countryCode` (keep it free-text; perform already uppercases it) and validate/normalize the country inside `buildContactInfo` so bad values degrade gracefully. `choices` cannot accept the case-insensitive / XK-AC-TA / free-text values existing mappings send. The new `phone_options.*` choices are fine (new optional fields). (Findings #2 and #3 are in the summary comment.)

nk1107 (commented):
Inline comment from review-toolkit (breaking-change finding).

packages/destination-actions/src/destinations/first-party-dv360/syncAudience/fields.ts
joe-ayoub-segment: It's a new Action and is not in use yet.

nk1107 (approved):

joe-ayoub-segment:
PR deployed

